PR #10, opened 2021-08-09
Add user types

sgrimm:
In preparation for revamping the way API client authentication works, add a
concept of "user type." Most users are of type "Individual." Super admins have
a different type, as will API clients.

In this change, the user type just replaces the old `super_admin` flag. A later
change will update API authentication to use API client users instead of a
dedicated `api_keys` table.

CU-185t84y


terraformation-inc:
Task linked: [CU-185t84y Make API keys per-user rather than having a single one in the server config](https://app.clickup.com/t/185t84y)

PR #12, opened 2021-08-12
Remove old API client authentication scheme

sgrimm:
Previously, to support API endpoints for reporting hardware metrics from the
device manager, we had a config setting that defined a single API key for each
server instance, with the assumption the server would be running on site and
the key would only be used by other processes on the same physical host.
That approach is no good with a cloud-hosted server.

We'll now be treating each API client as a special kind of user. A previous
change added the ability to differentiate between regular human users and API
clients. This one removes the original authentication scheme:

* The server-wide API key configuration option
* The `api_keys` table
* The `ClientIdentity` class and its supporting code
* The server-wide facility ID, which was only used in the context of the old
  API authentication scheme

A subsequent change will add a way for administrators to create API client
users with the appropriate settings for use by the device manager; for now,
this can be done manually by signing up as an ordinary user and then updating
the user type in the database.

CU-185t84y


sgrimm:
@kallen07 I know the migration number here conflicts with your PR; I'll renumber mine once yours is merged.

terraformation-inc:
Task linked: [CU-185t84y Make API keys per-user rather than having a single one in the server config](https://app.clickup.com/t/185t84y)

PR #22, opened 2021-08-23
Support PostGIS GEOMETRY columns

sgrimm:
Add a jOOQ data type binding class so that columns of type `GEOMETRY` can be
accessed from application code in a clean way. Currently, the only column of
that type in the database is `features.geom`.

Previously, `features.geom` was unconstrained: it could hold a mix of 2D and 3D
shapes in a mix of coordinate systems. It was up to the application code to ensure
that the shapes were consistent.

Update the database to require that all shapes be in 3D and in the spherical
Mercator (aka "pseudo-Mercator") coordinate system. Attempts to insert 2D shapes
or shapes in a different coordinate system will be rejected by the database.

As part of that, remove the `altitude` column, copying its value into the Z
coordinate of any existing 2D shapes in the database. A single `altitude` column
meant we were unable to express things like polygons on hillsides where not all 
the points had the same altitude.

It's annoying to have to explicitly convert client input to the right coordinate
system, so make the jOOQ bindings automatically do the transformation when
geometry values are passed as arguments to database queries.

Update `FeatureModel` to include a `geom` field and remove `altitude`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
@@ -90,7 +90,6 @@ class FeatureController(private val featureStore: FeatureStore) {
 data class CreateFeatureRequestPayload(
     val layerId: LayerId,
     val shapeType: ShapeType,
-    val altitude: Double? = null,
     val gpsHorizAccuracy: Double? = null,
     val gpsVertAccuracy: Double? = null,
     val attrib: String? = null,
@@ -113,7 +112,6 @@ data class CreateFeatureRequestPayload(
 data class UpdateFeatureRequestPayload(
     val layerId: LayerId,
     val shapeType: ShapeType,
-    val altitude: Double? = null,
     val gpsHorizAccuracy: Double? = null,
     val gpsVertAccuracy: Double? = null,
     val attrib: String? = null,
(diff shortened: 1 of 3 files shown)


sgrimm:
The jOOQ stuff here is probably pretty hard to review without a bunch of extra research, and the binding API documentation isn't stellar. Probably not worth spending tons of time on that part of this and instead focusing on the application code and test coverage. But if you're curious: https://www.jooq.org/doc/3.0/manual/code-generation/custom-data-type-bindings/

The class documentation for the jOOQ `Binding` interface is somewhat more informative, though still assumes a bunch of knowledge: https://github.com/jOOQ/jOOQ/blob/main/jOOQ/src/main/java/org/jooq/Binding.java

sgrimm:
@kallen07 Don't merge this, please. The general gist of it is okay but I realized it isn't dealing adequately with coordinate systems.

I'm kind of out of steam for today but will revise this tomorrow.

While I'm at it I'll throw in some helper code to easily convert to and from GeoJSON. There's a clean way to do it, just need a couple extension methods.

sgrimm:
This turned into a bigger change than I expected, but now it should be much more solid: the database is locked down to a consistent coordinate space. I also implemented GeoJSON support which was pretty cleanly separable from this code; expect another PR for that.

sgrimm:
Oh, also, I updated the initial comment on this PR to cover the more recent changes. If you squash merge this, you'll want to copy-paste that initial comment (click "Edit" on it to get the Markdown version) to use as the commit message body.

PR #23, opened 2021-08-23
Make PhotoRepositoryTest talk to the database

sgrimm:
An upcoming change will change `PhotoRepository` to use PostGIS to store the
GPS coordinates of photos, and it won't make sense to stub out the database at
that point since we'll need to test the conversion from PostGIS to latitude
and longitude.

Make the exising test talk to the database.

This change also uncovered a minor bug in the previous test's stubbing; if the
user doesn't have permission to read an accession, it should look to them like
the accession doesn't exist, not like there's a permission problem.


sgrimm:
Failing check here was some kind of GitHub glitch, I think. It never actually failed, just timed out after sitting there for ~5 hours. I reran the CI job and it worked.

PR #24, opened 2021-08-24
Implemented Plants and Plant Observations API

kallen07:
Implemented create, read, and update endpoints. Included two
batch read endpoints, one of which provides a summary of which
plant species exist in a layer. Did not include a delete endpoint.
If a caller wishes to delete a plant, then they should delete the
entire feature through the Features API.

Overloaded a couple of the layer data permissions checking
functions so that the caller can pass either a layer id or a
feature id.

Changed the Feature and Layer APIs to use plurals - "features" and
"layers". Made the delete feature endpoint throw an error if the
feature does not exist.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/api/PlantObservationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/api/PlantObservationsController.kt
@@ -81,7 +81,6 @@ class PlantObservationsController(private val observStore: PlantObservationsStor
   // through the Features API
 }
 
-@Schema(description = "height and diameterAtBreastHeight must be in meters")
 data class CreateObservRequestPayload(
     val featureId: FeatureId,
     val timestamp: Instant,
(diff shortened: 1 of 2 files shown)


kallen07:
@sgrimm Please hold off on reviewing this. I found a few things that need to be changed when I was implementing the Plant Observations API. I will push those changes tomorrow.

kallen07:
@sgrimm Here is the new commit message. If you decide to squash and merge this PR, then use this instead of any of the old commit messages. I'd focus your review on the tests.

--- Main API Changes ---
Implemented create, read, and update endpoints for Plants and
Plant Observations. Plants API includes two somewhat more
involved list endpoints (`/list` and `/list/summary`). Neither API
includes a delete endpoint. If a client wishes to delete a plant 
or a plant observation, then they should delete the entire feature
through the Features API. Especially for the Plant Observations
API, we may want to include delete functionality in the future. 

There are a lot of tests associated with both APIs. Since I'm using
the autogenerated PlantsRow and PlantObservationsRow, there
are more edge cases to handle.

--- Permissions Changes ---
Overloaded a couple of the "layer data" permissions check
functions so that the caller can pass either a layer id or a
feature id.

--- Other API Changes ---
Changed the Feature and Layer APIs to use plurals ("features"
and "layers"). Made the `delete` feature endpoint throw an error
if the feature does not exist. Currently, the permissions check will
also check if the feature exists, but I didn't want to depend on
that in the future.

PR #25, opened 2021-08-25
Support GeoJSON serialization of Geometry objects

sgrimm:
Add support for serializing and deserializing PostGIS geometry classes in GeoJSON
format.

Controllers don't have to do anything special to use this; they can just include a
`Geometry` field in a request or response payload and the serializer will
automatically do the JSON conversion.

A test suite verifies that the JSON is identical to the output of the PostGIS
`ST_AsGeoJSON` function.

With this change, the application code should largely not have to know that
GeoJSON exists at all; it uses `Geometry` objects to talk to both the database and
the client.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
@@ -90,7 +90,6 @@ class FeatureController(private val featureStore: FeatureStore) {
 data class CreateFeatureRequestPayload(
     val layerId: LayerId,
     val shapeType: ShapeType,
-    val altitude: Double? = null,
     val gpsHorizAccuracy: Double? = null,
     val gpsVertAccuracy: Double? = null,
     val attrib: String? = null,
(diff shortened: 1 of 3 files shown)


sgrimm:
The bulk of the code here is the custom serialization and deserialization logic. Terraware-server uses a library called Jackson for JSON serialization; it's by far the most widely-used JVM library for JSON. It's the thing that turns JSON HTTP payloads into Kotlin objects and vice versa.

The test suite is probably the most important thing to review here.

If you want to dig into the serializer implementation, a good resource is the official Jackson docs hub: https://github.com/FasterXML/jackson-docs

And a decent quick intro to writing custom serializers: https://www.baeldung.com/jackson-custom-serialization

sgrimm:
For posterity: I initially looked at a third-party library https://github.com/mayconbordin/postgis-geojson for this, but it didn't have any support for the `crs` field which is critical for us. At some point perhaps it will make sense to split this off into an open-source library of our own since it's a pretty standalone thing.

sgrimm:
I'll add the new fields to the API payloads in a separate PR.

PR #27, opened 2021-08-26
Include GIS endpoints in OpenAPI documentation

sgrimm:
There is an explicit list of packages that get scanned by the documentation
generator so it doesn't end up including a bunch of internal classes in the
schema document; needed to add the GSS API package to that list.


sgrimm:
@kallen07 I forgot about this bit of configuration when you were asking me about the doc generation earlier. Sorry you had to waste time figuring this out!

PR #33, opened 2021-09-02
Allow ID wrappers as path variables

sgrimm:
Previously, if we wanted to include an ID value in the URL path, we had to declare
the path variable as `Long` in the controller and then explicitly create an ID
wrapper object.

This was needed because Spring MVC didn't know how to convert a string to an
instance of an ID wrapper class. So the simple fix is to add a constructor to the
wrapper classes that takes a string argument. If the client uses a non-numeric
value, it will fail with HTTP 400 the same way it fails with `Long` parameters.

Update all the controller methods that used to take `Long` arguments to instead
use the appropriate wrapper classes.

This caused us to hit https://github.com/swagger-api/swagger-core/issues/3904 so
upgrade the OpenAPI library to get the fix for that.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -68,7 +68,7 @@ dependencies {
   implementation("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
   implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jacksonVersion")
   implementation("com.opencsv:opencsv:5.3")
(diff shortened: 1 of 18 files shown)


sgrimm:
In addition to running the test suite, I also tested this by running the seedbank-app Cypress test suite to confirm this doesn't introduce any behavior changes that break the front end.

PR #40, opened 2021-09-09
HW-15 Revamp device and timeseries endpoints

sgrimm:
Replace the old rhizo-server-style timeseries and device configuration API with
a new one that has the functionality the device manager is going to need when
it starts talking directly to terraware-server.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
@@ -22,22 +22,14 @@ class FacilityStore(private val facilitiesDao: FacilitiesDao) {
     }
   }
 
-  fun fetchById(ids: Collection<FacilityId>): List<FacilityModel> {
-    val user = currentUser()
-    val availableIds = ids.filter { user.canReadFacility(it) }.toTypedArray()
-
+  /** Returns a list of all the facilities the current user can access. */
+  fun fetchAll(): List<FacilityModel> {
+    val availableIds = currentUser().facilityRoles.keys.toTypedArray()
     return if (availableIds.isEmpty()) {
       emptyList()
     } else {
       facilitiesDao.fetchById(*availableIds).map { it.toModel() }
     }
-  }
-
-  /** Returns a list of all the facilities the current user can access. */
-  fun fetchAll(): List<FacilityModel> {
-    val user = currentUser()
-    val facilityIds = user.facilityRoles.keys
-    return fetchById(facilityIds)
   }
 
   /** Returns all the facilities the current user can access at a site. */


sgrimm:
@kallen07 Sorry for another PR you don't have much context on! These changes are all related to the device manager being able to record metrics (timeseries values) from various pieces of hardware.

Given that this is still kind of an initial version that may need to be tweaked once the device manager is actually using it, a reasonable review strategy for this would be just to see if anything sticks out to you as obviously wrong or questionable.

src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
```diff
@@ -14,8 +14,23 @@ import org.springframework.security.access.AccessDeniedException
...
+    return if (!currentUser().canReadFacility(facilityId)) {
+      null
+    } else {
+      facilitiesDao.fetchOneById(facilityId)?.toModel()
+    }
+  }
+
   fun fetchById(ids: Collection<FacilityId>): List<FacilityModel> {
-    return facilitiesDao.fetchById(*ids.toTypedArray()).map { it.toModel() }
+    val user = currentUser()
```
kallen07: Do we maybe want to rename this second function to make it clearer that it takes/returns a list? How come you decided to overload the function above?

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
```diff
@@ -1,86 +1,158 @@
...
+          typeId = type,
+          units = units,
+      )
+}
+
+data class TimeseriesValuesPayload(
+    @Schema(description = "ID of device that produced this value.") val deviceId: DeviceId,
+    @Schema(
+        description =
+            "Name of timeseries. This must be the name of a timeseries that has already been created for the device.")
```
kallen07, line 121: This is non urgent and also not super relevant to this diff, but it's a bit challenging to read the schema descriptions because of the way the code is formatted. It seems unnecessary to put all of this on three lines when it could fit into two, at least... this bothered me when I was writing schema descriptions as well

src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
```diff
@@ -1,156 +1,18 @@
...
+import com.terraformation.backend.db.tables.daos.DevicesDao
+import com.terraformation.backend.db.tables.pojos.DevicesRow
 import javax.annotation.ManagedBean
-import org.jooq.DSLContext
-import org.jooq.Record1
-import org.jooq.SelectConditionStep
-import org.jooq.impl.DSL
 
 @ManagedBean
-class DeviceStore(private val dslContext: DSLContext) {
```
kallen07, line 19: W o W !

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
sgrimm: Maybe worth filing a bug report about this with ktfmt. To me, the inconsistency is what makes it bad, not the fact that it's three lines instead of two.

But I just realized there's a way to at least get it to consistently put the `val`s on their own lines: put a trailing comma after the description if it doesn't already wrap to multiple lines. So:

```kotlin
    @Schema(
        description = "ID of device that produced this value.",
    )
    val deviceId: DeviceId,
    @Schema(
        description =
            "Name of timeseries. This must be the name of a timeseries that has already been " +
                "created for the device.")
    val timeseriesName: String,
```

What do you think? Feels a little nicer to me than "this `val` is at the beginning of the line, that one is off on the right side somewhere." Could also put a trailing comma after the longer descriptions so the closing parenthesis is more consistent, but for me the `val`s lining up is the key thing.

src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
sgrimm: In fact, we don't even need this method at all; it's only called by `fetchAll` so I'll just inline it there which will simplify the code anyway.

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
kallen07: Yes, I think that's much better. I don't have a strong opinion on making the parens more consistent for the longer descriptions.

sgrimm: OK, we can do that as a separate PR that covers all the payload classes.

PR #42, opened 2021-09-13
Type-safe wrapper for URL columns in database

sgrimm:
Use Java standard library `URI` objects to represent URLs in the database.
The application code was mostly already doing that, but was converting URIs to
strings and vice versa explicitly at each call site.


sgrimm:
Force push was to rebase on main.

PR #43, opened 2021-09-14
SW-194 Generate and cache photo thumbnails

sgrimm:
Add optional `maxWidth` and `maxHeight` query string parameters to the endpoints
    that download photos. If one or both are specified, the server will return a
    scaled-down copy of the photo.
    
    Currently, the scaling happens on demand the first time a user requests a version
    with a particular size. The scaled-down images are cached in the file store and
    the cached versions are returned on subsequent requests.


sgrimm:
Depends on #42

PR #44, opened 2021-09-15
SW-178 Simplify file creation

sgrimm:
Previously, code that wanted to create new files in a file store had to first
generate a unique path using a `PathGenerator`, then turn the path into a URL
using a `FileStore`. The return value of `PathGenerator.generatePath()` is
always immediately passed to `FileStore.getUrl()`, so there's no point making
callers do the extra work.


sgrimm:
I also tried another version of this where I got rid of `PathGenerator` completely, and put the path generation logic directly in `FileStore`, but that ended up being uglier than this because I had to turn `FileStore` from an interface into an abstract class and then give tests a way to stub out the random number generator.

PR #47, opened 2021-09-21
SW-260 Remove dependency on oauth2-proxy

sgrimm:
Move request authentication into terraware-server so we don't need to use
oauth2-proxy any more. This also gives us a lot more flexibility to change how
authentication works in our system. Out of the box, it should eliminate the CORS
problems we used to run into when doing front-end development where the front end
code lived on a different host+port than the terraware-server API.

Some of the configuration keys have changed, reflecting the fact that we're now
using a Keycloak client library. These are just 1:1 mappings from the old
Terraware config keys, but may require people to update their environment
variables or local application profiles.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
@@ -72,10 +72,29 @@ class SecurityConfig(private val userStore: UserStore) : KeycloakWebSecurityConf
 
       httpBasic { disable() }
 
-      // Don't create a new session (and set a session cookie) on each request, but use the session
-      // if one exists. The Keycloak adapter creates a session when the user logs in interactively
-      // (so we'll get a session cookie for web sessions) but not for requests that use bearer
-      // tokens (so we won't get useless cookies for non-web sessions).
+      // By default, Spring Security will create a new session for each request that doesn't already
+      // have one. It will persist the session to the session store and generate a cookie, and then
+      // when a new request comes in, it will see the cookie and look up the existing session to
+      // get the user's identity.
+      //
+      // For noninteractive clients, we don't want that; they'll include an authentication token
+      // with each request and with the default behavior, we'd be constantly creating a bunch of
+      // sessions the client would just ignore because it's not looking at the Cookie header in the
+      // response.
+      //
+      // But for interactive clients, we do want a cookie so they don't have to authenticate each
+      // request separately.
+      //
+      // Setting the session creation policy to NEVER tells Spring Security to never automatically
+      // create a session. But the application code is still allowed to explicitly create one, and
+      // Spring Security will honor it. In this case, the "application code" is the Keycloak
+      // adapter's interactive login endpoint (/sso/login, which the user gets redirected to after
+      // entering their password): the adapter explicitly creates a session so it can attach the
+      // user's Keycloak identity to the session.
+      //
+      // This gives us the behavior we want: a session cookie for interactive users, and no cookie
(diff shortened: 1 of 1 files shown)


sgrimm:
@kallen07 Obviously no need to go become an expert on Spring Security just to review this. Obviously I want to know if you spot anything wrong, but aside from that, one helpful thing to look at would be whether the docs I've written here seem like a good enough starting point for you to get your bearings if you wanted to dig into this stuff more deeply for whatever reason. I didn't want to write reams of text that just duplicated the Spring docs, but hopefully I gave enough context that someone new to this code would at least know what it's for.

sgrimm:
https://github.com/terraware/seedbank-app/pull/171 is what I was using on the front end to test logging in from a web browser.

sgrimm:
Just for posterity so it's recorded somewhere, though I don't think it's worth putting in the commit message, this was the end result of a series of attempts.

First, I tried using the "cookie" token store type that's built into the Keycloak adapter, such that there's no session and thus no need to persist session data anywhere or look it up at the start of each request. I got this working but the cookie was huge, like 3KB, which seemed excessive and likely to noticeably impact performance on slow network connections.

Continuing that line of thought, I tried adding code to compress the cookie data on the way to the user and uncompress it on the way back, such that the Keycloak code wouldn't know anything had changed. This turned out to be a pain because Keycloak hooks into the web server at somewhat of a low level, but I got it working. It chopped the cookie from 3KB to about 2.3KB. Digging further, the bad compression was thanks to the fact that OAuth tokens include digital signatures, which need to be cryptographically-secure random-ish numbers that are impossible to compress.

I thought about trying to remove the signatures and just encrypt the cookie with a secret key to get data integrity, but it didn't look like that was going to be doable without forking the Keycloak adapter library which wasn't my preference. So I switched to using the default behavior which is to persist the authentication data in a session store.

Up until that point, I'd been associating the `UserModel` with the `KeycloakAuthenticationToken` by subclassing that class to add a `UserModel` field, which `currentUser()` could then grab from Spring's `SecurityContextHolder` the same way it does in the current production version of the code. But that turned out to be a headache when those objects needed to be serialized and stored in the database: `UserModel` has references to `FeatureStore` and other service objects which obviously aren't serializable. It's easy enough to exclude them from serialization by just annotating them with `@Transient` but then how do they get populated when the session is loaded from the database again? I went down the road of adding a custom serializer that would wire them up again, but it was getting very fiddly and complicated and felt like the wrong solution.

So I abandoned the idea of making the `UserModel` reachable from `SecurityContextHolder` and switched to the implementation in this PR, where it's just a separate thread-local variable that works the same way `SecurityContextHolder` does. That turned out to be much cleaner, and I was happy enough with it to send it out for review.

src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
```diff
@@ -24,81 +33,105 @@ import org.springframework.web.cors.UrlBasedCorsConfigurationSource
...
         authorize("/api/**", fullyAuthenticated)
         authorize("/admin/**", fullyAuthenticated)
       }
-      httpBasic {}
-      sessionManagement { sessionCreationPolicy = SessionCreationPolicy.STATELESS }
+
+      httpBasic { disable() }
+
+      // Don't create a new session (and set a session cookie) on each request, but use the session
+      // if one exists. The Keycloak adapter creates a session when the user logs in interactively
```
kallen07: Hmm, I'm a little confused by the phrase "but use the session if one exists". Do you mean that we create a new session and cookie on each login request only, and otherwise there should be an existing session?

PR #51, opened 2021-09-24
SW-290 Add device configuration endpoints

sgrimm:
Add endpoints for creating, updating, and fetching device configuration data.
(There was already one for listing a facility's devices.)

These are intended to be called by API clients for the moment, though at some
point we'll build an admin UI that'll also use them.

The permissions are a little weird here because of the fact that API clients
are created with the "contributor" role. So for now, any operations that the
device manager performs will need to be accessible to contributors. We will
want to revisit this (and there is a task for it already, SW-292).

This also includes a change to make API clients implicit members of all the
projects in their organizations, following the same model we already use for
organization admins, so that they're allowed to access the facilities where
the devices live.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
@@ -175,19 +175,19 @@ data class CreateDeviceRequestPayload(
 ) {
   fun toRow(objectMapper: ObjectMapper): DevicesRow {
     return DevicesRow(
-        null,
-        facilityId,
-        name,
-        type,
-        make,
-        model,
-        protocol,
-        address,
-        port,
-        pollingInterval,
-        true,
-        settings?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
-        parentId,
+        id = null,
+        facilityId = facilityId,
+        name = name,
+        deviceType = type,
+        make = make,
+        model = model,
+        protocol = protocol,
+        address = address,
+        port = port,
+        pollingInterval = pollingInterval,
+        enabled = true,
+        settings = settings?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
+        parentId = parentId,
     )
   }
 }
@@ -232,19 +232,19 @@ data class UpdateDeviceRequestPayload(
 ) {
   fun toRow(deviceId: DeviceId, objectMapper: ObjectMapper): DevicesRow {
     return DevicesRow(
-        deviceId,
-        null,
-        name,
-        type,
-        make,
-        model,
-        protocol,
-        address,
-        port,
-        pollingInterval,
-        true,
-        settings?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
-        parentId,
+        id = deviceId,
+        facilityId = null,
+        name = name,
+        deviceType = type,
+        make = make,
+        model = model,
+        protocol = protocol,
+        address = address,
+        port = port,
+        pollingInterval = pollingInterval,
+        enabled = true,
+        settings = settings?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
+        parentId = parentId,
     )
   }
 }
--- a/src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
@@ -61,6 +61,9 @@ class DeviceStore(private val devicesDao: DevicesDao) {
 
     val existing = devicesDao.fetchOneById(deviceId) ?: throw DeviceNotFoundException(deviceId)
 
+    // A null facility ID is fine here because we don't require clients to pass in facility IDs
+    // when they update devices. But if they _do_ pass one in, it needs to be the same as the
+    // existing one; they can't attempt to move a device to a different facility.
     if (devicesRow.facilityId != null && devicesRow.facilityId != existing.facilityId) {
       throw IllegalArgumentException(
           "Devices may not be moved between facilities; create a new device instead.")


sgrimm:
CC @petersand so you can see what the new API endpoints look like in `openapi.yaml` while this is being reviewed -- they are pretty much what you'd expect, I hope.

src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
```diff
@@ -1,18 +1,72 @@
...
+    val deviceId = devicesRow.id ?: throw IllegalArgumentException("No device ID specified")
+
+    if (!currentUser().canUpdateDevice(deviceId)) {
+      throw AccessDeniedException("No permission to update device")
+    }
+
+    val parentId = devicesRow.parentId
+    if (parentId != null && !currentUser().canUpdateDevice(parentId)) {
+      throw AccessDeniedException("No permission to update parent device")
+    }
```
kallen07, line 60: I'm a little confused about the permission model here. I would have thought that the user would need update permissions on the parent device in order to have update permissions on the current device? I guess, why are there two permission checks here? 

Put another way, on creation, we check that the user has permissions on the parent device. Therefore, current device and parent device permissions should be in sync? 

I'm guessing it's this way because of some future changes we might make, but I still wonder if we can collapse both checks into one somehow.

src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
```diff
@@ -30,19 +37,52 @@ class DeviceController(
...
-        devices.map { DeviceConfig(it, it.settings?.let { objectMapper.readValue(it.data()) }) })
+    return ListDeviceConfigsResponse(devices.map { DeviceConfig(it, objectMapper) })
+  }
+
+  @ApiResponseSimpleSuccess
+  @PostMapping("/api/v1/devices")
+  fun createDevice(@RequestBody payload: CreateDeviceRequestPayload): CreateDeviceResponsePayload {
+    val devicesRow = payload.toRow(objectMapper)
+    val deviceId = deviceStore.create(devicesRow)
+    return CreateDeviceResponsePayload(deviceId)
```
kallen07, line 48: How come we don't need to wrap this in a try/catch? Same question for updateDevice. Do those exceptions filter all the way up to Spring?

src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
```diff
@@ -1,18 +1,72 @@
...
+
+    val existing = devicesDao.fetchOneById(deviceId) ?: throw DeviceNotFoundException(deviceId)
+
+    if (devicesRow.facilityId != null && devicesRow.facilityId != existing.facilityId) {
+      throw IllegalArgumentException(
+          "Devices may not be moved between facilities; create a new device instead.")
+    }
+
+    // Devices can't be moved between facilities.
+    devicesDao.update(devicesRow.copy(facilityId = existing.facilityId))
```
kallen07, line 73: Maybe we can make a note that the deviceId passed in by the user may be null because we don't request it in the API, not because the user wants to set it to null

src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
```diff
@@ -85,9 +127,130 @@ data class DeviceConfig(
...
+    @Schema(
+        description = "How often the device manager should poll for status updates, in seconds.")
+    val pollingInterval: Int? = null,
+    @Schema(
+        description =
+            "ID of parent device such as a hub or gateway, if any. The parent device must exist.")
+    val parentId: DeviceId? = null,
+) {
+  fun toRow(objectMapper: ObjectMapper): DevicesRow {
+    return DevicesRow(
```
kallen07, line 177: kinda nitty and I'm not sure if I did this in my diffs, but what do you think about specifying which params are being set? I can imagine a case where the DevicesRow changes, but this is not updated and it becomes a hard to detect bug

e.g. `id = null` instead of just `null`

src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
```diff
@@ -85,9 +127,130 @@ data class DeviceConfig(
...
+    @Schema(
+        description = "How often the device manager should poll for status updates, in seconds.")
+    val pollingInterval: Int? = null,
+    @Schema(
+        description =
+            "ID of parent device such as a hub or gateway, if any. The parent device must exist.")
+    val parentId: DeviceId? = null,
+) {
+  fun toRow(deviceId: DeviceId, objectMapper: ObjectMapper): DevicesRow {
+    return DevicesRow(
```
kallen07, line 234: same nit as above about param names

src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
```diff
@@ -54,7 +56,8 @@ class PermissionStore(private val dslContext: DSLContext) {
             ORGANIZATION_USERS
                 .ROLE_ID
                 .`in`(Role.OWNER.id, Role.ADMIN.id)
-                .or(PROJECT_USERS.USER_ID.isNotNull))
+                .or(PROJECT_USERS.USER_ID.isNotNull)
+                .or(USERS.USER_TYPE_ID.eq(UserType.APIClient)))
```
kallen07, line 60: honestly this SQL query is hard for me to validate without drawing out all the tables so I'm just going to trust you here. same with below.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -247,6 +247,7 @@ internal class PermissionTest : DatabaseTest() {
         FacilityId(1110),
         FacilityId(1111),
         createAccession = true,
+        createDevice = true,
```
kallen07, line 250: This file was hard to read/validate initially but as it gets bigger, it's not getting easier. Not a blocker, but I wonder if we can do anything about that in the future. The part I find hard is associating all the numbers with the permissions. They don't mean anything unless I scroll up and reorient myself with all the data that's been inserted. 

On that note, I've taken a cursory look at this but I didn't check all the numbers.

src/main/kotlin/com/terraformation/backend/device/db/DeviceStore.kt
sgrimm: The honest answer is we have no idea what the permissions need to look like here. Permissions are kind of a mess right now on device stuff in general because it's unclear what kind of user an API client needs to be. The product design around roles and permissions was 100% focused on the seed bank and GIS apps, and has nothing to say about device management. So this is mostly just a plausible placeholder, really.

src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
sgrimm: Yes, the exceptions will bubble up and get rendered as a server error. Right now it is kind of an ugly server error but I have a task open to improve that. The idea will be that we can write a single unified "handle all our application exceptions" bit of code that applies across all the controllers, rather than having to do individual try/catch statements everywhere.

PR #49, opened 2021-09-22
SW-134 Deploy to staging environment

sgrimm:
When a CI build on the main branch finishes, deploy it to the staging environment.

PR #52, opened 2021-09-27
SW-74: Production deployment setup

cormano:
- Trigger update script with service name and commit
- Set environment based on ref being branch or tag
- Run pipeline for version tags too. Version tags being formatted like v[0-9].[0-9]+.[0-9]+

### DIFF
--- a/.github/scripts/set-environment.sh
+++ b/.github/scripts/set-environment.sh
@@ -16,10 +16,6 @@ else
   exit
 fi
 
-# docker_tags="${docker_image}:${GITHUB_REF:0:12}"
-
-# if [[ -n "$TIER" ]]; then
-# fi
 docker_image='terraware/terraware-server'
 docker_tags="${docker_image}:${GITHUB_REF:0:12}\n${docker_image}:${TIER}"
 


sgrimm:
Hmm, did GitHub not send out notifications about the PR I did for this last week? https://github.com/terraware/terraware-server/pull/49

sgrimm:
Ah, never mind, that PR is out of date now anyway; will close it.

.github/workflows/main.yml
```diff
@@ -86,20 +91,38 @@ jobs:
...
       uses: docker/build-push-action@v2
       with:
         context: build/docker
-        tags: terraware/terraware-server:latest
+        tags: |
+          terraware/terraware-server:${{ env.TIER }}
+          terraware/terraware-server:${{ env.COMMIT_SHA }}
         platforms: linux/amd64,linux/arm64
-        push: ${{ github.ref == 'refs/heads/main' }}
+        push: env.IS_CD == true
```
sgrimm: This needs to be enclosed in `${{ }}` so it'll be evaluated as an expression.

.github/workflows/main.yml
```diff
@@ -30,7 +35,7 @@ jobs:
         path: |
           ~/.gradle
           ~/.m2
-        key: 2-${{ matrix.java-version }}-${{ hashFiles('*.gradle.kts', 'gradle.properties') }}
+        key: 2-${{ matrix.java-version }}-${{ hashFiles('*.gradle.kts', 'gradle.properties') }}-${{ env.TIER }}
```
sgrimm: I don't think we want a tier suffix here. This is for the dependency cache; if the dependency list is the same between a CI build and a release, we shouldn't be pulling a brand-new copy of everything (if for no other reason than that a download might fail).

.github/workflows/main.yml
```diff
@@ -86,20 +91,38 @@ jobs:
...
-    - name: Build Docker image and push if on main branch
+    - name: Build Docker image and push if on main branch or tag
       if: matrix.java-version == '15'
       uses: docker/build-push-action@v2
       with:
         context: build/docker
-        tags: terraware/terraware-server:latest
+        tags: |
+          terraware/terraware-server:${{ env.TIER }}
+          terraware/terraware-server:${{ env.COMMIT_SHA }}
```
sgrimm: If there is already an image with the commit SHA from a staging build, and we tag it for a release, what will Docker Hub do here? Will it keep the staging tag on the old image, remove the commit SHA tag from that image, and upload a new image with the prod and commit SHA tags? Hopefully so; that seems like what we want if we're going to build a fresh image for the prod deploy rather than reuse the staging image.

.github/workflows/main.yml
cormano: Got it, makes sense. Will fix.

.github/workflows/main.yml
cormano: Oh I was sure it worked both ways, e.g.: 

*https://github.com/terraware/terraware-server/blob/main/.github/workflows/main.yml#L92*

```
if: matrix.java-version == '15'
```

But will change it for clarity

.github/workflows/main.yml
cormano: If I understand your question correctly, then yes. So the build flow should be something like this:

- Building merged changes into main, tagging as `staging` and `commitX`

This creates the images tagged as:
  - `server:staging`
  - `server:commitX`
  
- Then after tagging a release with a vX.Y.Z tag, the build process will push images with tags `prod` and `commitX`. Since tagging doesn't create a new commit the "commit sha" will be the same.

Then we'll have

  - `server:staging`
  - `server:commit`
  - `server:prod`
  
The commit tag will be overwritten, same behaviour as with the "latest" tag overridden by the new "latest" being pushed

.github/workflows/main.yml
sgrimm: The difference (and I agree this is kind of annoying!) is that `if` is a built-in GitHub Actions property and it knows that the right-hand side always needs to be evaluated as an expression. But `push` is an argument to a plugin, and the workflow engine doesn't know what the value means, so it just passes the string in verbatim.

Ansible has this same annoyance where sometimes you need `foobar: "{{ expression }}"` and sometimes it's just `when: expression`. Hooray for dynamic typing!

.github/workflows/main.yml
```diff
@@ -69,19 +69,19 @@ jobs:
       run: ./gradlew test
 
     - name: Extract Docker image layers
-      if: matrix.java-version == '15'
+      if: ${{ matrix.java-version == '15' }}
```
sgrimm: You don't need this on `if`, just on the arguments that get passed to plugins in a `with` block.

cormano:
So there it is 😄 

Proper syntax on the expressions

.github/scripts/set-environment.sh
```diff
@@ -0,0 +1,30 @@
...
+  export \
+    TIER=prod \
+    IS_CD=true
+elif [[ "$GITHUB_REF" == refs/heads/main ]]; then
+  export \
+    TIER=staging \
+    IS_CD=true
+else
+  export IS_CD=false
+  exit
```
sgrimm, line 16: This seems to be making the CI build fail on this PR -- it gets to the "try building the Docker image" step and fails because `TIER` isn't set in the environment.

cormano: Oh true, checking

sgrimm: Is the `exit` still supposed to be here? The build worked, but I think it's because the list of tags ends up being empty. (Which I guess is maybe the right result for builds on PR branches...?)

cormano: You know, I made a commit before the last one that left the tags empty on purpose, but cancelled the build early because I thought "empty vars are the same as non existing ones" (in a shell sense). But thinking about how github handles the environment, it must be loaded in a map, and a non existing key will be an error. So I forgot about that `exit` for sure, but since it works with an empty set of tags will modify the script to work as intended

.github/scripts/set-environment.sh
```diff
@@ -0,0 +1,33 @@
...
+if [[ "$GITHUB_REF" =~ refs/tags/v[0-9].[0-9]+.[0-9]+ ]]; then
+  export \
+    TIER=prod \
+    IS_CD=true
+elif [[ "$GITHUB_REF" == refs/heads/main ]]; then
+  export \
+    TIER=staging \
+    IS_CD=true
+else
+  export IS_CD=false
```
sgrimm, line 15: The `export` won't do anything since we're exiting right afterwards.

.github/workflows/main.yml
```diff
@@ -76,7 +81,7 @@ jobs:
       uses: docker/setup-buildx-action@v1
 
     - name: Log into Docker Hub
-      if: matrix.java-version == '15' && github.ref == 'refs/heads/main'
+      if: matrix.java-version == '15' && env.IS_CD == true
```
sgrimm, line 84: I have a suspicion that `env.IS_CD == true` might not work because it will think `true` is a boolean but the environment variable value is a string. But maybe it will convert automatically in which case the way you have it represents the intent a little more clearly.

PR #53, opened 2021-09-27
SW-280 Support per-facility seed bank API requests

sgrimm:
In the data model, seed data is tied to seed bank facilities. Each accession, and
some other values such as storage locations, has a facility ID so we can track
seeds across multiple seed banks.

Since the UI didn't provide a way to switch between multiple seed banks the user
has access to, as a temporary measure, terraware-server used to assume they only
had access to one: it would look up the list of facilities they had access to and,
if there was just one seed bank entry in that list, use it as the facility ID for
the user's seed bank API calls. If the user had access to more than one seed bank,
it was treated as an error condition.

Modify the API to expose facility IDs to the client.

Since we want accession numbers to only be unique per-facility, this would have
meant making the client pass in a facility ID everywhere it currently uses an
accession number. Instead, modify the API so the client passes in the globally
unique accession ID, from which the server can look up the facility ID.

This requires corresponding changes on the client side.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1727,6 +1727,7 @@ components:
   schemas:
     AccessionPayload:
       required:
+      - accessionNumber
       - active
       - facilityId
       - id


sgrimm:
https://github.com/terraware/tree-location-web/pull/57 is the web front end side of this.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
```diff
@@ -262,7 +260,11 @@ data class UpdateAccessionRequestPayload(
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Schema
 data class AccessionPayload(
-    @Schema(description = "Server-generated unique identifier for the accession.")
+    @Schema(
+        description =
+            "Server-generated human-readable identifier for the accession. This is unique " +
+                "within a single seed bank, but different seed banks may have accessions with " +
+                "the same number.")
```
kallen07, line 267: Do we want to add a note about the accession numbers generally starting from 0 and increasing?

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
```diff
@@ -262,7 +260,11 @@ data class UpdateAccessionRequestPayload(
...
     val accessionNumber: String?,
```
kallen07: Why is the accessionNumber optional here?

src/main/kotlin/com/terraformation/backend/seedbank/api/PhotoController.kt
```diff
@@ -146,15 +139,9 @@ class PhotoController(
       maxHeight: Int? = null,
   ): ResponseEntity<InputStreamResource> {
     val headers = HttpHeaders()
-    val facilityId = currentUser().defaultFacilityId()
-
-    if (accessionStore.getIdByNumber(facilityId, accessionNumber) == null) {
-      throw NotFoundException("Accession $accessionNumber does not exist.")
```
kallen07, line 152: Do we want to check for an AccessionNotFound error in the try catch now that this logic is gone? Or do we want that error to bubble all the way up?

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -64,37 +64,7 @@ class AccessionStore(
...
-  }
-
-  private fun fetchByNumber(
-      facilityId: FacilityId,
-      accessionNumber: String,
-      skipPermissionCheck: Boolean
-  ): AccessionModel? {
+  fun fetchById(accessionId: AccessionId, skipPermissionCheck: Boolean = false): AccessionModel? {
     // First, fetch all the values that are either directly on the accession table or are in other
     // tables such that there is at most one value for a given accession (N:1 relation).
```
kallen07: I'm not sure what the comment means anymore.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -64,37 +64,7 @@ class AccessionStore(
...
     val parentRow =
```
kallen07, line 76: why is this called parentRow?

src/main/kotlin/com/terraformation/backend/seedbank/db/NotificationStore.kt
```diff
@@ -19,24 +19,31 @@ import org.jooq.impl.DSL
...
             NOTIFICATIONS.MESSAGE,
+            NOTIFICATIONS.accessions().ID,
             NOTIFICATIONS.accessions().NUMBER,
             NOTIFICATIONS.ACCESSION_STATE_ID)
         .from(NOTIFICATIONS)
-        .apply { if (since != null) where(NOTIFICATIONS.CREATED_TIME.ge(since)) }
+        .join(FACILITIES)
+        .on(NOTIFICATIONS.SITE_ID.eq(FACILITIES.SITE_ID))
+        .where(FACILITIES.ID.eq(facilityId))
+        .apply { if (since != null) and(NOTIFICATIONS.CREATED_TIME.ge(since)) }
```
kallen07, line 46: what does `.ge()` do?

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -214,26 +213,28 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
 
   @Test
   fun `create gives up if it can't generate an unused accession number`() {
-    repeat(10) { store.create(facilityId, AccessionModel()) }
+    repeat(10) { store.create(AccessionModel(facilityId = facilityId)) }
 
     dslContext.alterSequence(ACCESSION_NUMBER_SEQ).restartWith(197001010000000000).execute()
 
-    assertThrows(DuplicateKeyException::class.java) { store.create(facilityId, AccessionModel()) }
+    assertThrows(DuplicateKeyException::class.java) {
```
kallen07: nit: we started using the other version of assertThrows that allows you to just pass the exception class name

src/main/kotlin/com/terraformation/backend/seedbank/db/NotificationStore.kt
sgrimm: It's the method name for `>=` (greater or equal).

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
sgrimm: I think we don't want the numbering scheme to be part of the API contract; "it's a human-readable string the server assigns" is all we want the front-end code to assume about it. If we document the exact format, it kind of implicitly gives people the go-ahead to rely on, say, the fact that we include the current date as part of it.

src/main/kotlin/com/terraformation/backend/seedbank/api/PhotoController.kt
sgrimm: For now I'm going to let this bubble up. We need to fix exception handling in general so we don't need try/catch in all our controller methods, which is the topic of task SW-288.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
sgrimm: Good catch -- it shouldn't be nullable.

PR #55, opened 2021-09-27
SW-285 Add endpoints for projects and sites

sgrimm:
The front end needs to be able to walk the hierarchy of organizations, projects,
sites, and facilities, but we were missing project-related endpoints.

Add create, list, read, and update endpoints for projects. (Delete isn't supported
yet.)

Revamp the existing old-style endpoints for getting site information so they are
more consistent with the other endpoints the front end uses; previously they were
intended for use by the device manager but we never actually changed it to use
them, so they can be safely reworked without breaking anything.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/SiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/SiteStore.kt
@@ -10,6 +10,7 @@ import com.terraformation.backend.db.tables.daos.SitesDao
 import com.terraformation.backend.db.tables.pojos.SitesRow
 import com.terraformation.backend.db.tables.references.SITES
 import com.terraformation.backend.db.transformSrid
+import java.time.Clock
 import javax.annotation.ManagedBean
 import net.postgis.jdbc.geometry.Point
 import org.jooq.Condition
@@ -17,7 +18,11 @@ import org.jooq.DSLContext
 import org.springframework.security.access.AccessDeniedException
 
 @ManagedBean
-class SiteStore(private val dslContext: DSLContext, private val sitesDao: SitesDao) {
+class SiteStore(
+    private val clock: Clock,
+    private val dslContext: DSLContext,
+    private val sitesDao: SitesDao
+) {
   fun fetchById(siteId: SiteId, srid: Int = SRID.LONG_LAT): SiteModel? {
     return if (currentUser().canReadSite(siteId)) {
       return fetchModelsWhere(srid, SITES.ID.eq(siteId)).firstOrNull()
@@ -51,7 +56,13 @@ class SiteStore(private val dslContext: DSLContext, private val sitesDao: SitesD
       throw AccessDeniedException("No permission to create sites in this project")
     }
 
-    val row = SitesRow(location = location, name = name, projectId = projectId)
+    val row =
+        SitesRow(
+            createdTime = clock.instant(),
+            location = location,
+            modifiedTime = clock.instant(),
+            name = name,
+            projectId = projectId)
 
     sitesDao.insert(row)
 
--- a/src/main/resources/templates/admin/site.html
+++ b/src/main/resources/templates/admin/site.html
@@ -13,9 +13,9 @@
 <h2 th:text="|Site ${site.name} (${site.id})|">Site (id)</h2>
 
 <p>
-    Latitude: <span th:text="${site.latitude.toPlainString()}">123.45678</span>
+    Latitude: <span th:text="${site.location.y}">123.45678</span>
     <br>
-    Longitude: <span th:text="${site.longitude.toPlainString()}">-44.99997</span>
+    Longitude: <span th:text="${site.location.x}">-44.99997</span>
 </p>
 
 <h3>Facilities</h3>


openapi.yaml
```diff
@@ -768,6 +768,141 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ListOrganizationsResponse'
+  /api/v1/organizations/{organizationId}/projects:
```
kallen07, line 846: I don't normally review openAPI.yaml changes since they are auto generated. Let me know if that doesn't seem right.

src/main/kotlin/com/terraformation/backend/customer/api/SiteController.kt
```diff
@@ -1,92 +1,81 @@
...
+}
 
-@Schema(requiredProperties = ["id"])
-data class GetSiteResponse(
+data class SiteElement(
     val id: SiteId,
     val name: String,
+    val projectId: ProjectId,
     val latitude: String,
     val longitude: String,
```
kallen07: Not blocking, but in the future do we want to switch this to use a geom like the features endpoints? or is there no use for an altitude and ability to change between coordinate systems here?

src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
```diff
@@ -316,19 +325,6 @@ class UserModel(
     }
   }
 
-  /**
-   * Returns the user's organization ID if they belong to exactly one organization.
-   *
-   * TODO: Remove this once the client passes in organization IDs.
-   */
-  fun defaultOrganizationId(): OrganizationId? {
```
kallen07, line 340: yay!

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
```diff
@@ -0,0 +1,106 @@
...
+          "The user is not a member of the organization, or the organization does not exist.")
+  @GetMapping
+  @Operation(
+      summary = "Gets a list of the projects in an organization.",
+      description = "Only projects that are accessible by the current user are included.")
+  fun listOrganizationProjects(
+      @PathVariable organizationId: OrganizationId
+  ): ListProjectsResponsePayload {
+    if (!currentUser().canListProjects(organizationId)) {
+      throw NotFoundException()
```
kallen07, line 86: Why do you do this check here, rather that seeing if the result from the ProjectStore is an empty list?

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -65,6 +70,18 @@ class ProjectStore(
     return projectsRow.toModel()
   }
 
+  fun update(row: ProjectsRow) {
+    val projectId = row.id ?: throw IllegalArgumentException("Missing project ID")
+    if (!currentUser().canUpdateProject(projectId)) {
+      throw AccessDeniedException("No permission to update project")
```
kallen07: Should this be a not found error?

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -65,6 +70,18 @@ class ProjectStore(
...
+    }
+
+    val existing = projectsDao.fetchOneById(projectId) ?: throw ProjectNotFoundException(projectId)
+
+    projectsDao.update(
+        row.copy(organizationId = existing.organizationId, modifiedTime = clock.instant()))
```
kallen07: Should this be `existing.copy` that brings over the fields the user is able to change? The user can only update `name` through the API at the moment, but we may want to bring over all fields that are not `undefined` since, in theory this could be called by some other internal code. And there's a risk of adding new params that the user can update while forgetting to change this. I believe that as it's written now, database fields that are defined will be overwritten as empty.

It seems like the business of checking which fields have changed may be challenging. Are unset fields undefined? If not, then how can we differentiate an empty field that the user didn't want to update from one that the user wants to set as empty? One work around could be to alter the function signature to receive a projectId and name so that you don't need to deal with a partially empty projectsRow.

Also, do we want to do a check that the `row.organizationId` (if defined) must match the existing organizationId and throw an error if it doesn't?

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
sgrimm: To differentiate between, "The organization doesn't exist or isn't visible to you" (in which case it's HTTP 404) and, "You're in the organization but no projects exist or none are visible to you" (in which case it's 200 with an empty list of projects). That distinction is probably better off living in the store class, though; thanks for pointing it out.

openapi.yaml
sgrimm: Yeah, that seems right.

src/main/kotlin/com/terraformation/backend/customer/api/SiteController.kt
sgrimm: I was trying to minimize churn here but thinking about it, this probably falls under the "better to change it now when nothing is using it yet" umbrella like switching to accession IDs. I don't know if we'll even ever use those values for anything, but if we do, having them consistent with other location data seems like a win.

PR #56, opened 2021-09-28
Use query string for GET request parameters

sgrimm:
A couple endpoints were declared as GET requests but expected the requests to
have payloads, which GET requests generally aren't supposed to have.

Modify the endpoints to take optional arguments as query string parameters.


sgrimm:
Merging this to unblock the front end work, but happy to revisit in a followup.

PR #57, opened 2021-09-29
Various GIS endpoint changes

kallen07:
Add a tag to the Geometry type description so that we can override that type on the client side. This is necessary because the Geometry type is not displayed correctly in the Swagger UI. Therefore, `openapi-typescript` generator is not able to generate the client side type correctly. 

Update `api/v1/gis/plants/list/{layerId}` to return ALL the data from both the Features and Plants tables. Before this change, the endpoint returned some fields from each table.

Update all GIS endpoints to return Geometry coordinates in Longitude/Latitude (EPSG:4326 WGS 84). Before this change, the endpoints accepted Long/Lat coordinates, but returned coordinates in the Spherical Mercator (EPSG:3857 WGS 84) system.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
@@ -313,12 +313,13 @@ data class FeaturePhoto(
       size = photosRow.size!!)
 }
 
+@Schema(
+    description =
+        "Describes a map feature. The coordinate reference system of the \"geom\" field will be " +
+            "longitude/latitude EPSG:4326.")
 data class FeatureResponse(
     val id: FeatureId,
     val layerId: LayerId,
-    @Schema(
-        description = "The coordinate reference system will be longitude/latitude EPSG:4326.",
-    )
     val geom: Geometry? = null,
     val gpsHorizAccuracy: Double? = null,
     val gpsVertAccuracy: Double? = null,
--- a/src/test/kotlin/com/terraformation/backend/db/Helpers.kt
+++ b/src/test/kotlin/com/terraformation/backend/db/Helpers.kt
@@ -3,6 +3,7 @@ package com.terraformation.backend.db
 import net.postgis.jdbc.geometry.Geometry
 import net.postgis.jdbc.geometry.Point
 import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
 
 /** Returns a PostGIS [Point] with a specific spatial reference ID. */
 fun newPoint(x: Double, y: Double, z: Double, srid: Int): Point {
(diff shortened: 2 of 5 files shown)


kallen07:
@sgrimm if you do squash and merge this, please use the initial PR comment for the commit description.

src/test/kotlin/com/terraformation/backend/db/Helpers.kt
```diff
@@ -15,3 +17,17 @@ fun newPoint(x: Double, y: Double, z: Double, srid: Int): Point {
...
  */
 fun mercatorPoint(x: Double, y: Double, z: Double) = newPoint(x, y, z, SRID.SPHERICAL_MERCATOR)
+
+/**
+ * Assert Point coordinate data is equal. Since the coordinates probably went through at least one
+ * coordinate system conversion in the database write/read, allow a small fuzz factor in the x and y
+ * coordinates.
+ */
+fun assertPointsEqual(expected: Geometry, actual: Geometry) {
+  assertEquals(expected.type, Geometry.POINT) // this function only compares "Point" types
```
sgrimm: Not merge-blocking, but it'd be good to put messages on these assertions -- if one of the ones in the middle fails, a message would tell you which value didn't match. Maybe take an optional message argument on the function and use it as part of the assertion messages on the individual `assertEquals` calls.

src/test/kotlin/com/terraformation/backend/db/Helpers.kt
```diff
@@ -15,3 +17,17 @@ fun newPoint(x: Double, y: Double, z: Double, srid: Int): Point {
...
+  assertEquals(expected.type, actual.type)
+  assertEquals(expected.srid, actual.srid)
+  assertEquals(expected.firstPoint.x, actual.firstPoint.x, 0.0001)
+  assertEquals(expected.firstPoint.y, actual.firstPoint.y, 0.0001)
+  assertEquals(expected.firstPoint.z, actual.firstPoint.z)
```
sgrimm: The Z axis isn't being transformed, true, but there's still the usual floating point "wrote 5.123, got back 5.122999999999" problem that could happen; this could use a fuzz factor too.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
```diff
@@ -59,13 +60,20 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
   private val layerId = LayerId(100)
   private val nonExistentLayerId = LayerId(401)
   private val nonExistentFeatureId = FeatureId(400)
+  // Spherical Mercator coordinates X=261845.71, Y=6250564.35, Z=35.0
+  // map to longitude = 2.3522, latitude = 48.8566, elevation = 35.0
+  private val sphericalMercatorPoint =
+      newPoint(261845.71, 6250564.35, 35.0, SRID.SPHERICAL_MERCATOR)
```
sgrimm: Could use `mercatorPoint()` instead of `newPoint()` here.

src/test/kotlin/com/terraformation/backend/db/Helpers.kt
```diff
@@ -15,3 +17,17 @@ fun newPoint(x: Double, y: Double, z: Double, srid: Int): Point {
  * coordinates are already in that coordinate system.
  */
 fun mercatorPoint(x: Double, y: Double, z: Double) = newPoint(x, y, z, SRID.SPHERICAL_MERCATOR)
+
+/**
+ * Assert Point coordinate data is equal. Since the coordinates probably went through at least one
+ * coordinate system conversion in the database write/read, allow a small fuzz factor in the x and y
+ * coordinates.
+ */
+fun assertPointsEqual(expected: Geometry, actual: Geometry) {
```
sgrimm: This could maybe take `Geometry?` to save you the `!!` at the call sites and also so you can separately assert the two values aren't null.

src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
```diff
@@ -316,6 +316,9 @@ data class FeaturePhoto(
 data class FeatureResponse(
     val id: FeatureId,
     val layerId: LayerId,
+    @Schema(
+        description = "The coordinate reference system will be longitude/latitude EPSG:4326.",
+    )
```
sgrimm: Would have expected to see this description show up in the `openapi.yaml` diff.

sgrimm: Oh, right, it's because this is a reference to another type in the schema and, annoyingly, the schema generator can't put descriptions on individual references.

PR #58, opened 2021-09-29
SW-305 Script to generate SQL to import shapefiles

sgrimm:
This is based on the shapefile import code from gis-server, but modified to output
a SQL script rather than inserting directly to a database. With this approach, you
can import the data in a transaction in psql and examine the data before committing.
(Or you can pipe the output of this script to psql and it will autocommit.)

Another difference between this script and the gis-server one is that by default,
this creates a new layer for each import, even if there is an existing layer of
the same type. You can give it an existing layer ID to override that behavior.

It does not attempt to automatically detect layer types from subdirectory names
like the gis-server script; the example shapefiles we've gotten from the forestry
team didn't follow the naming convention that script expected so it seemed simpler
to just have whoever runs the script explicitly say what layer type they want.

### DIFF
--- a/scripts/import_shapefile.py
+++ b/scripts/import_shapefile.py
@@ -1,0 +1,160 @@
+import argparse
+import os
+import json
+import datetime
+import shapefile
+import sys
+from typing import Optional
+from sridentify import Sridentify
+from pypika import CustomFunction, Table, Query
+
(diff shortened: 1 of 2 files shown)


sgrimm:
I used this to populate a couple of layers in the staging database.

PR #61, opened 2021-09-30
Fix: deploy

cormano:
Had to troubleshoot a bunch of things in the deployment process, that even though passed the local testing didn't seem to work when deployed.


.github/workflows/main.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       uses: docker/setup-buildx-action@v1
 
     - name: Log into Docker Hub
-      if: matrix.java-version == '15' && env.IS_CD == true
+      if: matrix.java-version == '15' && env.IS_CD == 'true'
```
sgrimm, line 84: Ah, bummer this was needed, but not too surprising.

PR #62, opened 2021-09-30
SW-288 Improve JSON error responses

sgrimm:
Handle a few more kinds of exceptions bubbling up from controller methods or
from Spring's request handling infrastructure. These are rendered using our
`ErrorResponsePayload` response schema for consistency.

`AccessDeniedException` turns into HTTP 403.

Invalid request payloads turn into HTTP 400 with a description of exactly
which field had the problem, if possible. For example (formatted for clarity):

    {
      "error": {
        "message": "Required field not present: timeseries[0].values[1].timestamp"
      },
      "status": "error"
    }

Invalid URLs (path variables or query string arguments) turn into HTTP 400
with a description of which field had the problem.

This is focused on JSON responses; a subsequent change will address the fact
that we use Spring's default HTML error page for non-JSON requests.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/ControllerExceptionHandler.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/ControllerExceptionHandler.kt
@@ -1,12 +1,23 @@
 package com.terraformation.backend.api
 
+import com.fasterxml.jackson.databind.JsonMappingException
+import com.fasterxml.jackson.databind.exc.InvalidFormatException
+import com.fasterxml.jackson.databind.exc.InvalidNullException
+import com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException
+import javax.ws.rs.QueryParam
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.http.HttpHeaders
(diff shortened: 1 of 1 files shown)


sgrimm:
It is sad that the invalid payload error pretty-printing isn't just built into Spring or the JSON parsing library. I had to build pretty much the same rendering logic at my last company too. I'll look forward to our open-source release so the world can stop reinventing this particular wheel.

src/main/kotlin/com/terraformation/backend/api/ControllerExceptionHandler.kt
```diff
@@ -15,15 +26,15 @@ import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExcep
  */
 @ControllerAdvice
 class ControllerExceptionHandler : ResponseEntityExceptionHandler() {
-  @ExceptionHandler(ClientFacingException::class)
+  @ExceptionHandler
```
sgrimm, line 29: Not sure why I had the arguments on these annotations; by default Spring looks at the types of the method arguments to figure out what exceptions the handlers cover, so it was completely pointless to specify the exact same exception class in the annotation.

PR #63, opened 2021-09-30
Don't treat layer tile set name as required

sgrimm:
The corresponding column in the database is nullable, and we don't populate
it when we import data from shapefiles.

Previously, if you tried to get layer data for a layer imported from a
shapefile, the server would return HTTP 500.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3296,7 +3296,6 @@ components:
       - layerType
       - proposed
       - siteId
-      - tileSetName
       type: object
       properties:
         id:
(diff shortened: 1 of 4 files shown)


sgrimm:
Ran into this after I loaded the Future Forests shapefiles into my database and tried to view it in the web app's map UI.

sgrimm:
The front-end team is hitting this in staging today, so I'm going to go ahead and merge it to unblock them; happy to address any comments in a followup PR.

PR #65, opened 2021-09-30
Fix: SSH reading while loop's stdin

cormano:
Didn't happen before with the webhook, but now when ssh-ing to the host through the bastion it seems ssh is reading the while loop's items list. So setting the ssh command's stdio to /dev/null fixes that.

(But setting the whole block's stdin by grouping the while loop's inner block with braces seems clearer in my opinion)


cormano:
That sounds better, just made it work the first way I thought of, will change

sgrimm:
Merging this before tests finish because it doesn't touch any code that's covered by tests.

PR #67, opened 2021-10-04
SW-315 Create users from admin UI

sgrimm:
Enhance the "add user to organization" function of the placeholder admin UI to
be able to create users who aren't already registered on the system.

New users get an email message from Keycloak requiring them to set a password.
Existing users get added to the organization.

The "change your password" email redirects the user to the application home
page afterwards. The code attempts to figure out the correct redirect URL.
This works properly in dev environments, but in case it doesn't work in staging
or production thanks to the load balancer, there's a new server config option
to explicitly set the redirect URL.

This isn't intended to be a full-featured "invite user to organization"
implementation, but rather a stopgap to allow us to add alpha users quickly.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -155,7 +155,7 @@ class UserStore(
 
   /**
    * Creates a new user as a member of an organization. This registers them in Keycloak and also
-   * adds them to the local database.
+   * adds them to the `users` table.
    */
   fun createUser(
       organizationId: OrganizationId,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -153,6 +153,62 @@ class UserStore(
     return usersDao.fetchOneById(userId)?.let { rowToDetails(it) }
   }
 
+  /**
+   * Creates a new user as a member of an organization. This registers them in Keycloak and also
+   * adds them to the local database.
```
kallen07: what do you mean "local" database?

sgrimm: As opposed to Keycloak's database. I'll change this to say "terraware-server" instead of "local."

sgrimm: Or maybe "the users table" would be clearer.

PR #68, opened 2021-10-04
SW-299 Map internal "X not found" exceptions to HTTP 404

sgrimm:
Add a handler that causes exceptions such as `LayerNotFoundException` to be
rendered as HTTP 404 responses if they bubble up through a controller method.
Previously, they were treated as generic exceptions and were rendered as HTTP
500 responses.


sgrimm:
I updated the admin UI to use these exceptions so I could do an end-to-end test by hitting `/admin/organization/99999` and confirming the response was HTTP 404 with a well-formed error message.

PR #69, opened 2021-10-05
SW-315 Increase link validity time to 3 days; minor cleanup

sgrimm:
Feedback from the product team was that the default validity time on "reset
password" links, 12 hours, was too short when they were creating a bunch of
alpha users who might not be at their computers. Increase it to 72 hours as
requested.

Pass the "must set password" setting to Keycloak at user creation time rather
than setting it after creation, to reduce the number of places where a failure
could leave the system in an inconsistent state.

Stop assuming (incorrectly) that the Keycloak admin client `UsersResource.get`
method checks to see if the user exists; it just returns a client object that
can be used to make Keycloak admin API requests related to a specific user.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/KeycloakRequiredActions.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/KeycloakRequiredActions.kt
@@ -1,0 +1,14 @@
+package com.terraformation.backend.auth
+
+/**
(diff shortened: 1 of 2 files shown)


sgrimm:
Merging this now since it's blocking adding people to the production site for alpha testing, but will do a followup PR if there are any comments I need to address.

kallen07:
Looks good to me

PR #70, opened 2021-10-05
Add layer creation to admin UI

sgrimm:
The mobile app expects a site to already have a "Plants Planted" layer if the
user is trying to upload information about plants. It could create the layer on
demand but that introduces a race condition if two users upload plants to a new
site at the same time; they'd both end up creating separate layers since it's
legal (intentionally) for a site to have multiple layers of the same type.

Add layer creation to the admin UI, and by default, create a "Plants Planted"
layer when a new site is created. The admin can deselect that option to skip
layer creation, or they can select additional layer types. They can also add
layers to existing sites if needed.

This should suffice until we've worked out the requirements for user-facing
site administration.


sgrimm:
One possible use case is to differentiate between imported plants from ArcGIS and plants that were added via the app. I don't know if that's how we'll actually want it to work, but separating them by layer would make it pretty easy to, e.g., import a revised data set from ArcGIS without disturbing app-generated data.

PR #71, opened 2021-10-05
Redirect after POST in admin UI

sgrimm:
Make the POST endpoints in the admin UI redirect to GET endpoints rather than
directly returning an HTML page. This way if the user reloads the page after
submitting a form, the browser won't try to resubmit the form.

This is mostly just a straightforward use of Spring MVC flash attributes (which
are stored in the session data on the server side and get automatically added
to the model and removed from the session store after a redirect) except for
API keys, which we only want to show one time; there's now a GET endpoint to
show a just-created API key, and it redirects to the organization page with an
error message if you try to reload it.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -433,18 +433,25 @@ class AdminController(
         Point(longitude.toDouble(), latitude.toDouble(), 0.0).apply { srid = SRID.LONG_LAT }
     val site = siteStore.create(projectId, name, location)
 
-    layerTypes?.forEach { layerType ->
-      layerStore.createLayer(
-          LayerModel(
-              hidden = false,
-              layerType = layerType,
-              proposed = false,
-              siteId = site.id,
-              tileSetName = null,
-          ))
+    try {
+      layerTypes?.forEach { layerType ->
+        layerStore.createLayer(
+            LayerModel(
+                hidden = false,
+                layerType = layerType,
+                proposed = false,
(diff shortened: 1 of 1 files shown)


sgrimm:
The fact that I hadn't bothered to do this was bugging me every time I used the admin UI. Spring MVC makes it so easy to do POST-then-GET that there's no excuse not to.

This depends on #70 just so it doesn't have to be updated after that one is merged; if #70 looks bad but this looks good, I can easily rebase this on main so it can be merged first.

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -398,53 +426,104 @@ class AdminController(
...
+      redirectAttributes: RedirectAttributes,
   ): String {
     val location =
         Point(longitude.toDouble(), latitude.toDouble(), 0.0).apply { srid = SRID.LONG_LAT }
-    siteStore.create(projectId, name, location)
+    val site = siteStore.create(projectId, name, location)
+
+    layerTypes?.forEach { layerType ->
+      layerStore.createLayer(
+          LayerModel(
```
kallen07: Do we want to wrap this in a try catch since createLayer may throw an access denied exception?
Same with createLayer below.

sgrimm: Good point. I think I had the idea in my head that if you have permission to create a site, you have permission to create layers. But they are separate permissions. And also, a layer creation could fail for some other reason unrelated to permissions and we still want to tell the user that the site itself was created.

PR #72, opened 2021-10-05
SW-252 Add "Awaiting Check-In" accession state

sgrimm:
To support mobile clients uploading accession data before the seeds are delivered
to the seed bank, and to allow the web front end to show a list of such
accessions, add a new accession state that comes before Pending. If an accession
is created in that state, it must be advanced to Pending via an explicit API
request. The existing state transition logic still applies, e.g., if you enter
today's date as the storage date for an Awaiting Check-In accession, it will
transition directly to In Storage.

There's a new optional field in `CreateAccessionRequestPayload` to control
whether an accession starts off in Awaiting Check-In state. The default is the
existing behavior: start off in Pending state (that is, consider new accessions
already checked in at creation time.) Thus any accessions created by existing
clients should continue to behave as expected.

The date and time when an accession was checked in is tracked and is available as
a search field.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
@@ -66,8 +66,7 @@ class AccessionController(private val accessionStore: AccessionStore, private va
   @Operation(summary = "Create a new accession.")
   @PostMapping
   fun create(@RequestBody payload: CreateAccessionRequestPayload): CreateAccessionResponsePayload {
-    val updatedPayload =
-        accessionStore.create(payload.toModel(), checkedIn = payload.checkedIn ?: true)
+    val updatedPayload = accessionStore.create(payload.toModel())
     return CreateAccessionResponsePayload(AccessionPayload(updatedPayload, clock))
   }
 
@@ -127,13 +126,6 @@ class AccessionController(private val accessionStore: AccessionStore, private va
 @JsonAutoDetect(getterVisibility = JsonAutoDetect.Visibility.NONE)
 data class CreateAccessionRequestPayload(
     val bagNumbers: Set<String>? = null,
-    @Schema(
-        description =
-            "If false, the accession will be created in \"Awaiting Check-In\" state and must be " +
-                "checked in using a separate API request. If true, the accession will already be " +
-                "checked in when it is initially created.",
-        defaultValue = "true")
-    val checkedIn: Boolean? = null,
     val collectedDate: LocalDate? = null,
     val deviceInfo: DeviceInfoPayload? = null,
     val endangered: SpeciesEndangeredType? = null,
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.seedbank.db
 
 import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.customer.db.AppDeviceStore
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.AccessionId
@@ -57,6 +58,7 @@ class AccessionStore(
     private val withdrawalStore: WithdrawalStore,
     private val clock: Clock,
     private val support: StoreSupport,
+    private val config: TerrawareServerConfig,
 ) {
   companion object {
     /** Number of times to try generating a unique accession number before giving up. */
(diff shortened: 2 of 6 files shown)


sgrimm:
@kallen07 Don't merge this quite yet -- I want Constanza to have a chance to look over the API changes and confirm they make sense.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -439,6 +456,47 @@ class AccessionStore(
...
+   * field values.
+   *
+   * @throws AccessionNotFoundException The accession did not exist or wasn't accessible by the
+   * current user.
+   */
+  fun checkIn(accessionId: AccessionId): AccessionModel {
+    val accession = fetchById(accessionId) ?: throw AccessionNotFoundException(accessionId)
+
+    if (!currentUser().canUpdateAccession(accessionId, accession.facilityId)) {
+      throw AccessDeniedException("No permission to update accession")
```
kallen07: Based on how we're throwing permissions errors for other update requests, should this be a not found error?

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -439,6 +456,47 @@ class AccessionStore(
...
+      log.info("Accession $accessionId is already checked in; ignoring request to check in again")
+      return accession
+    }
+
+    // Don't record the time with sub-second precision; it is not useful and makes exact searches
+    // problematic in the face of systems with different levels of precision in their native time
+    // representations.
+    val checkedInTime = clock.instant().truncatedTo(ChronoUnit.SECONDS)
+    val checkedIn =
+        accession.copy(checkedInTime = checkedInTime).withCalculatedValues(clock, accession)
```
kallen07, line 480: I think it's a bit confusing to use the variable name `checkedIn` here since that sounds like it should be a boolean value rather than an accession.

sgrimm:
@kallen07 Constanza says the API changes look fine, so you can review at your leisure, and you can merge it if it looks good to you.

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -259,6 +259,23 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
 
+  @Test
+  fun `create sets checked-in timestamp`() {
+    every { clock.instant() } returns Instant.EPOCH.plusMillis(600)
+
+    store.create(AccessionModel(facilityId = facilityId), checkedIn = true)
+
+    assertEquals(
+        AccessionsRow(
+            checkedInTime = Instant.EPOCH,
```
kallen07, line 277: Maybe we can add a comment here about the rounding?

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -176,7 +178,7 @@ class AccessionStore(
     }
   }
 
-  fun create(accession: AccessionModel): AccessionModel {
+  fun create(accession: AccessionModel, checkedIn: Boolean = false): AccessionModel {
```
kallen07: I think it's a bit confusing that the default checkedIn value is false here, whereas in the API the default is true. Knowing the API behavior, the tests seem a bit counter intuitive since accessionStore.create() is defaulting to AwaitingCheckIn state rather than Pending.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
```diff
@@ -65,7 +66,8 @@ class AccessionController(private val accessionStore: AccessionStore, private va
   @Operation(summary = "Create a new accession.")
   @PostMapping
   fun create(@RequestBody payload: CreateAccessionRequestPayload): CreateAccessionResponsePayload {
-    val updatedPayload = accessionStore.create(payload.toModel())
+    val updatedPayload =
+        accessionStore.create(payload.toModel(), checkedIn = payload.checkedIn ?: true)
```
kallen07: Is the ?: necessary here when the default value on `CreateAccessionRequestPayload` is true?

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: This is closer to how we want it to work long-term, I think, though obviously #73 will move this logic out of `AccessionStore` entirely. But in this case, `fetchById` already does the read permission check, so you'd only get to line 470 here if you had read permission but not write permission.

sgrimm:
Switched this to a draft for now. When I wrote this, it was an open question whether we wanted accessions created via the web site to be considered checked-in by default (the existing behavior) and I figured the most likely answer would be yes, so we'd need an "is this already checked in?" flag the web client could set to true and the mobile client could set to false. But after I sent this PR out, we got an answer from the product team and it turns out we want web-created accessions to be awaiting check-in too.

So we probably don't need the `checkedIn` flag in the create payload at all any more, just a feature flag we can turn on once we deploy a front-end version that can deal with the new state.

sgrimm:
Ah, forgot you approved this already. I addressed the comments and I think the feature flag is pretty noncontroversial.

PR #73, opened 2021-10-07
SW-326 Centralize permission exceptions

sgrimm:
Add a new permission-checking function that throws exceptions if the user doesn't
have the permission in question. Application code like

    if (!user.canCreateSite(projectId)) {
      throw AccessDeniedException("No permission to create site")
    }

changes to

    requirePermissions { createSite(projectId) }

In addition to reducing boilerplate in the application code, this is a first step
toward making the permission checks more intelligent about situations where you
have read-only access to objects. Currently the system treats that situation
inconsistently; sometimes it returns a "not found" error even though you can read
the thing in question. Putting all the exception-throwing code in one place should
make it easier to keep everything consistent.

The KDoc for the new `PermissionRequirements` class lays out some rules for
exception logic which should help people add new code that follows the same
behavior patterns.

However, in this initial version, those rules aren't being followed. This change is
a straight refactor that attempts to exactly reproduce the existing behavior. There
were a couple places where the existing logic was wrong or inconsistent in such a
way that replicating it wouldn't have made sense, but aside from those cases, the
existing unit tests for various permissions continue to pass unmodified.

There is also a unit test for `PermissionRequirements` itself; currently it is not
very interesting but it will get more valuable once we start throwing different
exceptions in different situations.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
@@ -403,18 +403,8 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
   @Test
   fun `delete throws AccessDeniedException if the user doesn't have delete permission`() {
     val feature = store.createFeature(validCreateRequest)
-
-    // They still have read permission, so they can see the feature
     every { user.canDeleteLayerData(any()) } returns false
-
     assertThrows<AccessDeniedException> { store.deleteFeature(feature.id!!) }
-  }
-
-  @Test
-  fun `delete throws a FeatureNotFoundException if the user doesn't have read permission`() {
-    store.createFeature(validCreateRequest)
-    every { user.canReadLayerData(featureId = any()) } returns false
-    assertThrows<FeatureNotFoundException> { store.deleteFeature(nonExistentFeatureId) }
   }
 
   @Test


sgrimm:
I am pretty happy with the structure of this, though if you're not, I'd love to hear why.

I am less than thrilled with the verbosity of the name `requirePermissions`, so if you can think of a better one, please suggest it. Aside from just abbreviating it to `requirePerms` I couldn't think of something more concise that satisfied my (somewhat arbitrary, but hopefully not dumb) criteria:

* Convey that this is specifically about checking for permissions
* Convey the expectation that it will throw, rather than just returning `false`, if the check fails
* Not imply that it is going to _change_ the user's permissions
* Not be a super generic word that you'd find yourself colliding with later on

Some other names I considered: `assertAllowed` (seemed too close to unit test assertion method names), `require` (seemed too generic), `mustPermit` (just looked ugly to me).

Obviously it's trivial to rename it if we think of something better later, and `requirePermissions` isn't _that_ verbose so maybe I'm just obsessing over it for no good reason.

src/main/kotlin/com/terraformation/backend/db/Exceptions.kt
```diff
@@ -18,6 +18,9 @@ class AccessionNotFoundException(val accessionId: AccessionId) :
 class DeviceNotFoundException(val deviceId: DeviceId) :
     EntityNotFoundException("Device $deviceId not found")
 
+class FacilityNotFoundException(val facilityId: FacilityId) :
+    EntityNotFoundException("Facility $facilityId not found")
+
```
sgrimm: Oops, this isn't used yet -- will remove it from this PR.

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -0,0 +1,261 @@
...
+
+  fun readAccession(accessionId: AccessionId, facilityId: FacilityId? = null) {
+    if (!user.canReadAccession(accessionId, facilityId)) {
+      throw AccessionNotFoundException(accessionId)
+    }
+  }
+
+  fun updateAccession(accessionId: AccessionId, facilityId: FacilityId? = null) {
+    if (!user.canUpdateAccession(accessionId, facilityId)) {
+      throw AccessDeniedException("No permission to update accession $accessionId")
```
kallen07, line 92: I see what you mean about the permissions being inconsistent. updateLayer throws a NotFoundException for this same case...

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -172,14 +170,16 @@ class FeatureStore(
   }
 
   fun updateFeature(newModel: FeatureModel): FeatureModel {
-    val oldModel = noPermissionsCheckFetch(newModel.id!!)
+    val featureId = newModel.id ?: throw IllegalArgumentException("No feature ID specified")
```
kallen07, line 173: Thanks for adding the IllegalArgumentException. It's much clearer now.

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
```diff
@@ -140,8 +139,8 @@ class LayerStore(
     // Caller provided siteId must match the siteId in the database because allowing site moves
     // would add additional permissions complexity. Plus, the caller can achieve that behavior
     // using a combination of createLayer() and deleteLayer().
-    if (layerModel.siteId != currentRow[LAYERS.SITE_ID] ||
-        !currentUser().canUpdateLayer(currentRow[LAYERS.SITE_ID]!!)) {
+    requirePermissions { updateLayer(layerModel.id!!, layerModel.siteId) }
```
kallen07: Can we parse out `layerModel.id` above and throw an `IllegalArgumentException` if it doesn't exist? The not null assertion was making me nervous here, and then I realized that it actually could cause an uncaught exception on line 137.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
```diff
@@ -401,10 +401,20 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
...
-    assertThrows<FeatureNotFoundException> { store.deleteFeature(feature.id!!) }
+
+    assertThrows<AccessDeniedException> { store.deleteFeature(feature.id!!) }
+  }
+
+  @Test
+  fun `delete throws a FeatureNotFoundException if the user doesn't have read permission`() {
+    store.createFeature(validCreateRequest)
+    every { user.canReadLayerData(featureId = any()) } returns false
+    assertThrows<FeatureNotFoundException> { store.deleteFeature(nonExistentFeatureId) }
```
kallen07: I'm confused about how these tests are passing. Where in the code are we checking whether the user has read permissions on a deleteFeature request?

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -0,0 +1,261 @@
...
+
+  fun createLayerData(featureId: FeatureId) {
+    if (!user.canCreateLayerData(featureId)) {
+      throw AccessDeniedException("No permission to create data for feature $featureId")
+    }
+  }
+
+  fun readLayerData(photoId: PhotoId, featureId: FeatureId) {
+    if (!user.canReadLayerData(featureId)) {
+      throw PhotoNotFoundException(photoId)
```
kallen07, line 146: I'm wondering if this kind of fine grained checking will create many very similar functions later on. For example, are we going to throw unique exceptions for all feature data? Plants, photos, plant observations, plus new types of features associated with different layer types?

Right now, FeatureStore.fetchFeature() isn't using this function. The exception is being thrown in FeatureController. Were you planning on addressing those cases in a follow up diff?

kallen07:
I think the name `requirePermissions` is good. It's clear, and I don't mind the length.

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
sgrimm: Good call. We should probably just go through the code and do that everywhere, but I'll do it here for now.

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
sgrimm: It's a reasonable question. My goal here was to emulate the existing behavior so I didn't think about this kind of question.

One benefit is that we expose these exception messages in our error responses when these exceptions bubble up through the controller. So if the user says "give me photo 567" it would be odd to get back HTTP 404 with an error message of "layer 123 not found." We would have indirectly told the user that the photo does in fact exist. Obviously we could make the controllers generate their own error messages, though, so this isn't a super strong argument, just a minor convenience to offset the inconvenience of the extra methods.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
sgrimm: Good catch. This test was wrong.

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
kallen07: So it sounds like we could use this this new abstraction instead of throwing NotFound errors in the controller, but we could also continue returning null in the store.get() functions.

sgrimm: Yeah, that's my thinking at the moment. It might also make sense to add non-nullable store.get() variants that throw if the object isn't found. But my guess is that in some cases we won't want to treat missing/inaccessible data as an error condition, and having a nullable function seems nicer for those cases than having to catch and ignore an exception.

PR #76, opened 2021-10-12
SW-288 Improve timeseries endpoint error reporting

sgrimm:
`/api/v1/seedbank/timeseries/values` is a batch API endpoint where the client
can submit multiple values for multiple timeseries in one request. Previously,
it just reported that specific values failed to be recorded, but that made it
impossible for the client to tell if the problem was client-side or
server-side.

Add more detailed error messages to differentiate a couple of common problems:
the client specifying a nonexistent timeseries name and the client trying to
record multiple values with the same timestamp.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
@@ -4,15 +4,12 @@ import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.DeviceId
 import com.terraformation.backend.db.TimeseriesId
-import com.terraformation.backend.db.TimeseriesNotFoundException
 import com.terraformation.backend.db.tables.pojos.TimeseriesRow
 import com.terraformation.backend.db.tables.references.TIMESERIES
 import com.terraformation.backend.db.tables.references.TIMESERIES_VALUES
 import java.time.Instant
 import javax.annotation.ManagedBean
 import org.jooq.DSLContext
-import org.springframework.dao.DataIntegrityViolationException
-import org.springframework.dao.DuplicateKeyException
 
 @ManagedBean
 class TimeseriesStore(private val dslContext: DSLContext) {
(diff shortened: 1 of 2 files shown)


sgrimm:
@kallen07 Don't merge this yet -- I might need to coordinate with Peter to avoid breaking the device manager if it is parsing the existing error responses.

sgrimm:
This is safe to merge; it won't break the existing error handling code in the device manager.

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
```diff
@@ -65,14 +68,27 @@ class TimeseriesStore(private val dslContext: DSLContext) {
...
-          .execute()
+      try {
+        dslContext
+            .insertInto(TIMESERIES_VALUES)
+            .set(TIMESERIES_ID, timeseriesId)
+            .set(CREATED_TIME, createdTime)
+            .set(VALUE, value)
+            .execute()
+      } catch (e: DuplicateKeyException) {
+        throw e
```
kallen07: Why are you catching this error and then throwing it again right away?

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
```diff
@@ -65,14 +68,27 @@ class TimeseriesStore(private val dslContext: DSLContext) {
...
+      } catch (e: DataIntegrityViolationException) {
+        throw TimeseriesNotFoundException(deviceId)
```
kallen07: Which case would cause the DataIntegrityViolationException to be thrown?

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
sgrimm: You're right, this is not necessary. It's left over from a different approach I took initially. Thanks.

But to answer your question about the pattern here: It's because of the `catch` just under this. `DuplicateKeyException` is a subclass of `DataIntegrityViolationException`. I wanted `DuplicateKeyException` to bubble up to the caller as-is. But if (in my original cut of this) you tried to insert a value for a timeseries that didn't exist, the query would throw a generic `DataIntegrityViolationException` due to the foreign key constraint, and I wanted to turn that into a `TimeseriesNotFoundException`.

However, now it shouldn't ever be possible for the timeseries to not exist since the caller has to pass in the ID.

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
sgrimm: (Answered in the response to the other comment)

PR #78, opened 2021-10-13
SW-327 Move all parent ID lookups into permission check methods

sgrimm:
Previously, as an optimization, we had methods in `UserModel` that checked for
permission to do something with a child object but took the ID of a parent
object as a parameter, since we knew the permissions were always inherited.
But this makes it tough to throw the correct `EntityNotFoundException` because
it should refer to the thing the user was actually trying to access, which is
the child object. To work around that, in some places the methods took both
the parent and child object IDs, forcing calling code to know about the
permission boundaries. But sometimes the methods just took the child IDs and
looked the parent IDs up. The whole thing was inconsistent.

Update the `UserModel` and `PermissionRequirements` APIs so they only ever take
the ID of the object the user is trying to access. They now always look up
parent IDs as needed.

As a side benefit, this means we can now declare that the permission checking
methods will always do an existence check. Previously that was _usually_ the
case but not always: if a method took a parent ID as an argument, it could
pass the permission check even if the child didn't exist. Making that behavior
official means we can get rid of some existence checks in the calling code,
making this change performance-neutral in those places. (It still causes
extra database queries in other places, though.) There's a basic test case to
sanity-check that permission checks always fail on nonexistent objects.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
@@ -198,34 +198,18 @@ class UserModel(
     return canAccessLayer(layerId)
   }
 
-  // "Layer data" refers to all tables that directly or indirectly references layers.
-  // You can also think of it as all data that belongs "inside" a layer.
-  private fun canAccessLayerData(layerId: LayerId): Boolean {
-    // Currently, all gis data is lumped into one permission. In the future, permission
-    // to create/update layers may be separated from general access to layer data.
-    return canAccessLayer(layerId)
-  }
-
-  fun canCreateLayerData(layerId: LayerId): Boolean {
-    return canAccessLayerData(layerId)
-  }
-
-  fun canReadLayerData(layerId: LayerId): Boolean {
-    return canAccessLayerData(layerId)
-  }
-
-  fun canUpdateLayerData(layerId: LayerId): Boolean {
-    return canAccessLayerData(layerId)
+  fun canCreateFeature(layerId: LayerId): Boolean {
+    return canUpdateLayer(layerId)
   }
 
   fun canReadFeature(featureId: FeatureId): Boolean {
     val layerId = featuresDao.fetchOneById(featureId)?.layerId ?: return false
-    return canReadLayerData(layerId)
+    return canReadLayer(layerId)
   }
 
   fun canUpdateFeature(featureId: FeatureId): Boolean {
     val layerId = featuresDao.fetchOneById(featureId)?.layerId ?: return false
-    return canUpdateLayerData(layerId)
+    return canUpdateLayer(layerId)
   }
 
   fun canDeleteFeature(featureId: FeatureId): Boolean {
--- a/src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
@@ -46,7 +46,7 @@ class FeatureStore(
   private val log = perClassLogger()
 
   fun createFeature(model: FeatureModel): FeatureModel {
-    requirePermissions { createLayerData(model.layerId) }
+    requirePermissions { createFeature(model.layerId) }
 
     val currTime = clock.instant()
     val insertedRecord =
@@ -114,7 +114,7 @@ class FeatureStore(
   }
 
   fun countFeatures(id: LayerId): Int {
-    if (!currentUser().canReadLayerData(id)) {
+    if (!currentUser().canReadLayer(id)) {
       return 0
     }
 
@@ -128,7 +128,7 @@ class FeatureStore(
   }
 
   fun listFeatures(id: LayerId, skip: Int? = null, limit: Int? = null): List<FeatureModel> {
-    if (!currentUser().canReadLayerData(id)) {
+    if (!currentUser().canReadLayer(id)) {
       return emptyList()
     }
 
--- a/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
@@ -133,9 +133,9 @@ class PermissionRequirements(private val user: UserModel) {
     }
   }
 
-  fun createLayerData(layerId: LayerId) {
-    if (!user.canCreateLayerData(layerId)) {
-      throw AccessDeniedException("No permission to create data in layer $layerId")
+  fun createFeature(layerId: LayerId) {
+    if (!user.canCreateFeature(layerId)) {
+      throw AccessDeniedException("No permission to create feature in layer $layerId")
     }
   }
 
--- a/src/main/kotlin/com/terraformation/backend/gis/db/PlantStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/db/PlantStore.kt
@@ -83,7 +83,7 @@ class PlantStore(
       maxEnteredTime: Instant? = null,
       notes: String? = null
   ): List<FetchPlantListResult> {
-    if (!currentUser().canReadLayerData(layerId)) {
+    if (!currentUser().canReadLayer(layerId)) {
       return emptyList()
     }
 
@@ -148,7 +148,7 @@ class PlantStore(
       minEnteredTime: Instant? = null,
       maxEnteredTime: Instant? = null,
   ): Map<SpeciesId, Int> {
-    if (!currentUser().canReadLayerData(layerId)) {
+    if (!currentUser().canReadLayer(layerId)) {
       return emptyMap()
     }
     val records =
--- a/src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
@@ -137,11 +137,11 @@ internal class PermissionRequirementsTest : RunsAsUser {
   }
 
   @Test
-  fun createLayerData() {
-    assertThrows<AccessDeniedException> { requirements.createLayerData(layerId) }
+  fun createFeature() {
+    assertThrows<AccessDeniedException> { requirements.createFeature(layerId) }
 
-    grant { user.canCreateLayerData(layerId) }
-    requirements.createLayerData(layerId)
+    grant { user.canCreateFeature(layerId) }
+    requirements.createFeature(layerId)
   }
 
   @Test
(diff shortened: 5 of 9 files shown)


src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
```diff
@@ -170,68 +176,82 @@ class UserModel(
...
-  fun canUpdateLayerData(layerId: LayerId): Boolean {
-    return canAccessLayerData(layerId)
+  fun canUpdateFeature(featureId: FeatureId): Boolean {
+    val layerId = featuresDao.fetchOneById(featureId)?.layerId ?: return false
+    return canUpdateLayerData(layerId)
   }
 
-  fun canUpdateLayerData(featureId: FeatureId): Boolean {
+  fun canDeleteFeature(featureId: FeatureId): Boolean {
+    return canUpdateFeature(featureId)
```
kallen07, line 216: It seems like `canUpdateFeature`, `canDeleteFeature`, and `canUpdateFeatureData` all call `canUpdateLayerData`. I think it's a bit confusing that `canDeleteFeature` does this through `canUpdateFeature` but `canUpdateFeatureData` calls `canUpdateLayerData` directly. Then, all of these roll up into `canAccessLayerData`. 

What do you think calling `canAccessLayerData` directly from all layer data permissions to make it more clear what the check is actually doing (for now, when the permissions are all the same)? Right now I'm finding it hard to follow.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -172,7 +172,7 @@ class FeatureStore(
   fun updateFeature(newModel: FeatureModel): FeatureModel {
     val featureId = newModel.id ?: throw IllegalArgumentException("No feature ID specified")
 
-    requirePermissions { updateLayerData(newModel.layerId, featureId) }
+    requirePermissions { updateFeatureData(featureId) }
```
kallen07: Did you mean `updateFeature` here?

As a broader question, do we need to differentiate between features and feature data? I think I remember in a conversation with Leann then we'd most likely just need feature (and everything nested inside a feature) permission to be the most fine grained level of permission.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -284,9 +280,7 @@ class FeatureStore(
   }
 
   fun listPhotos(featureId: FeatureId): List<PhotosRow> {
-    if (!currentUser().canReadLayerData(featureId) || noPermissionsCheckFetch(featureId) == null) {
-      throw FeatureNotFoundException(featureId)
-    }
+    requirePermissions { readFeature(featureId) }
```
kallen07, line 283: The way it's written, would this be `canReadFeaturePhoto`? 

Same question as above, do we really need this permission or can it just be `readFeature`?

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -327,7 +321,7 @@ class FeatureStore(
   fun deletePhoto(featureId: FeatureId, photoId: PhotoId) {
     val photosRow = getPhotoMetadata(featureId, photoId)
 
-    requirePermissions { deleteLayerData(featureId) }
+    requirePermissions { deleteFeatureData(featureId) }
```
kallen07: It seems odd/inconsistent to have the `deletePhoto` function be checking `deleteFeatureData` while the `getPhotoMetadata` function checks `readFeaturePhoto`

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
```diff
@@ -56,18 +56,15 @@ class LayerStore(
             .where(LAYERS.ID.eq(id))
             .fetchOne()
 
-    if (layer == null || (layer[LAYERS.DELETED]!! && ignoreDeleted)) {
-      return null
-    }
-
-    val siteId = layer[LAYERS.SITE_ID]!!
-    if (!currentUser().canReadLayer(siteId)) {
+    if (layer == null ||
```
kallen07: do we want to remove this null check since canReadLayer does the existence check? Or maybe do the permission check before fetching the layer, and then do the layer deleted check afterwards?

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
```diff
@@ -141,7 +138,7 @@ class LayerStore(
     // Caller provided siteId must match the siteId in the database because allowing site moves
     // would add additional permissions complexity. Plus, the caller can achieve that behavior
     // using a combination of createLayer() and deleteLayer().
-    requirePermissions { updateLayer(layerId, layerModel.siteId) }
+    requirePermissions { updateLayer(layerId) }
```
kallen07: Maybe we can move the permission check to before we fetch the current row? Then, we could also remove the throw LayerNotFoundException in the database query, but I guess that's a bit tricky because the row could technically be deleted between when we do the permission check and when we fetch the data.

One thing that seems a little odd, but maybe it just takes some getting used to, is relying on the permission check to throw if the item doesn't exist. I see why we'd want to do that in the permission check, but it seems odd to then ignore these other cases where the database could fail to find the row. I'm not suggesting we remove the existence check from the permissions check, but I'm wondering if we want to keep it in other places where we'd normally have it?

src/main/kotlin/com/terraformation/backend/gis/db/PlantObservationsStore.kt
```diff
@@ -38,15 +38,15 @@ class PlantObservationsStore(
   fun fetch(id: PlantObservationId): PlantObservationsRow? {
     val row = observDao.fetchOneById(id)
 
-    if (row == null || !currentUser().canReadLayerData(row.featureId!!)) {
+    if (row == null || !currentUser().canReadFeature(row.featureId!!)) {
```
kallen07, line 41: Similar to my other comment, it seems inconsistent that we're using `canReadFeature` here but `createFeatureData` above.

This happens in a couple other places but I won't spam you with comments on all of them.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -95,8 +95,7 @@ class AccessionStore(
             .fetchOne()
             ?: return null
 
-    if (!skipPermissionCheck &&
-        !currentUser().canReadAccession(accessionId, parentRow[ACCESSIONS.FACILITY_ID])) {
+    if (!skipPermissionCheck && !currentUser().canReadAccession(accessionId)) {
```
kallen07, line 98: Is it worth it to continue having the controller throw not found exceptions? It would be more consistent in the store code if store.read() could also use the `requirePermissions` rather than needing to bypass that abstraction to allow the controller to throw.

I think in our last conversation it sounded like we could go either way, but apologies if I'm mis-remembering that.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
sgrimm: Yes, this makes sense. I think I was mostly just following the existing model but it's overly complex for no benefit. Will collapse these down to permissions on features.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
sgrimm: Agreed. Will add a `deleteFeaturePhoto` permission to make this consistent.

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
sgrimm: This section can be simplified even a bit more than that; I'll fold the deleted check into the query.

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
sgrimm: Will move the permission check.

My thinking about how the existence checks in the permission checks relate to the row getting deleted right after the permission check is that we can treat a row disappearing in the middle of an operation as a generic "something unexpected happened" kind of error rather than as an expected failure that has to be treated specially. One reason I arrived at this point of view is that this situation could also happen after the explicit read but before an update or the creation of a child object; the failure mode isn't actually being _introduced_ by putting the existence check in the permission code.

So basically: the code needs to not produce corrupted state or inconsistent results if something deletes a row we're in the middle of working with, but we can treat it as a "database query failed" condition rather than a "user passed us an invalid ID" condition.

Or that's how I'm thinking about it, anyway.

sgrimm: (Also, moved the permission check as you suggested.)

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: The thought definitely occurred to me as I worked on this that having our fetch functions return null for nonexistent items makes things more complicated. I think I will do that as a followup PR because it's an internal API change that isn't strictly related to the permission checks.

src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
sgrimm: Hopefully this is less messy without the `FeatureData` permissions. I kept the indirection in place in `canUpdateFeature` and friends because I think it makes it clearer what the inheritance relationship between the rules is. That is, you can update a feature if you can update its layer, independently of what the rules are for whether you can update the layer. If we introduce more detailed layer permissions, the feature -> layer permission mapping stays the same.

If that still seems unclear, though, I can definitely cut out the indirection.

kallen07: I definitely think it's clearer without the FeatureData permissions.

I think I'm a little unclear on when we would want to use canUpdateLayerData vs canUpdateFeature. When I created the layerData permission (I believe I created it), I imagined it being used for everything underneath layers. But if we're going to have features use it's own permission, then maybe we don't need the layerData permission at all.

src/main/kotlin/com/terraformation/backend/gis/db/LayerStore.kt
kallen07: That sounds reasonable. In that case would we end up returning a 500 error to the client?

sgrimm: Yes, most likely it'd be a 500. If it turns out to happen with any frequency we can revisit that but my intuition is that there is going to be essentially zero update/delete contention in our app so it's not worth maintaining code to check for it specifically.

src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
sgrimm: Yeah, you're right, I guess at this point, LayerData doesn't serve much of a purpose. I'll get rid of it before merging this.

PR #79, opened 2021-10-13
RFC: Move parent lookups into a helper class

sgrimm:
The number of DAOs `UserModel` needed was getting a bit out of hand, and they
are only ever used to look up parent IDs.

Try pulling that functionality out into a dedicated class so we don't have to
keep adding dependencies to `UserModel` as the data model grows.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
@@ -1,0 +1,60 @@
+package com.terraformation.backend.customer.db
+
+import com.terraformation.backend.customer.model.UserModel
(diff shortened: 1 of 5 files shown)


sgrimm:
This may or may not be the right approach; putting it out for feedback. If it seems like an improvement, I can clean it up a little and turn this into a real PR.

An alternative, to really minimize dependencies, would be to put these in `PermissionStore`, which `UserModel` needs no matter what. These lookups felt like a different category of operation than the other methods in that class, so I put them in their own spot, but maybe jumbling them together with the permission reading code is less bad than adding a separate dependency.

sgrimm:
Merging this so I can rebase the read-only permissions PR on it. I expanded on the class comment a little bit.

PR #80, opened 2021-10-13
SW-327 Implement read-only permission logic

sgrimm:
Previously, the permission logic was inconsistent about what exception should be
thrown if you try to write to an object that you only have permission to read.

Make all the permission checks consistent: if you have read access, you always
get `AccessDeniedException` because you can see the object exists. If you don't
have read access, you always get an appropriate `EntityNotFoundException` in
keeping with the idea that you shouldn't be able to see that objects exist if
you don't have access to them.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
@@ -381,13 +381,6 @@ internal class PermissionTest : DatabaseTest() {
         deleteFeaturePhoto = true)
 
     permissions.expect(
-        PhotoId(10010),
-        PhotoId(11000),
-        PhotoId(11001),
-        readFeaturePhoto = true,
-        deleteFeaturePhoto = true)
-
-    permissions.expect(
         FacilityId(1000),
         FacilityId(1001),
         FacilityId(1010),


sgrimm:
(This is based on #78 and will need to be rebased once that PR is merged.)

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
```diff
@@ -143,7 +143,8 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
 
   @Test
   fun `create fails with AccessDeniedException if user doesn't have create permission`() {
-    every { user.canCreateLayerData(layerId = any()) } returns false
+    every { user.canCreateLayerData(any()) } returns false
+    every { user.canReadLayer(any()) } returns true
```
kallen07: It seems a bit clunky that we have to set both permissions here in order to do this check. It's kind of breaking that abstraction that allows us to not worry how the permissions are calculated. Can we mock at a higher level (the require function)?

src/test/kotlin/com/terraformation/backend/gis/db/LayerStoreTest.kt
```diff
@@ -258,10 +268,18 @@ internal class LayerStoreTest : DatabaseTest(), RunsAsUser {
...
-  fun `delete fails with LayerNotFoundException if user does not have permission`() {
+  fun `delete fails with LayerNotFoundException if user does not have read permission`() {
     val layer = store.createLayer(validCreateRequestModel)
     every { user.canDeleteLayer(any()) } returns false
+    every { user.canReadLayer(any()) } returns false
     assertThrows<LayerNotFoundException> { store.deleteLayer(layer.id!!) }
+  }
+
+  @Test
+  fun `delete fails with AccessDeniedException if user does not have delete permission`() {
```
kallen07, line 279: distinguishing between the types of failures (NotFound vs AccessDenied) seems like it maybe doesn't belong in here anymore since we're not actually throwing these exceptions in the store code anymore

sgrimm: The eternal "unit test vs. integration test" debate! Agreed that these are covering code that's already covered by `PermissionRequirementsTest`. But I think these are still useful: if you do a bad copy-paste of the `requirePermissions` block and, say, check for read permission instead of delete permission, ideally a test should catch that.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
sgrimm: Hopefully this will feel less clunky once I update this PR to account for the fact that I removed the LayerData permissions.

I did consider mocking `requirePermissions` instead, but I realized I'd end up having to do something like

```kotlin
every { requirements.createFeature(any()) } throws AccessDeniedException("foo")
assertThrows<AccessDeniedException> { store.createFeature(model) }
```

which didn't feel like it was testing anything useful to me.

src/test/kotlin/com/terraformation/backend/gis/db/LayerStoreTest.kt
kallen07: Oh good point, I hadn't thought of that at all.

PR #81, opened 2021-10-15
Fetch accessions using single SQL queries

sgrimm:
Previously, to fetch an accession, we first did a query that loaded all the scalar
values for the accession, then did a separate query for each table where there
could be zero or more of something associated with a particular accession. That
requires a bunch of database round-trips.

Start using jOOQ 3.15's "multiset" feature to load everything in one query.

This not only makes accession loading a bit more efficient since the database
server can parallelize or reorder the child table lookups, but also acts as a way
to introduce multisets to the code base in preparation for using them in other
places such as the seed bank search API.


sgrimm:
Intro to multisets: https://blog.jooq.org/jooq-3-15s-new-multiset-operator-will-change-how-you-think-about-sql/

PR #84, opened 2021-10-22
SW-339 Support nested search results

sgrimm:
Rewrite the code that generates SQL queries to search for accessions. The new
implementation supports nested lists of values in search results. These
correspond to the field structure in the accession payload: there is a list
of germination tests each of which contains a list of germinations, etc.

The existing flat-table Cartesian join representation of lists of values is
still supported and can be freely mixed and matched with nested fields.
Users of the existing front end shouldn't notice any behavior difference;
this was confirmed using the web app test suite.

Currently, nested fields aren't supported for CSV export, simply because it's
unclear how one should format a list of lists of values in a CSV cell.

The implementation leans heavily on jOOQ multisets to bind nested values to
search results. It works around some limitations of multisets, which makes the
code kind of complicated. `NestedQueryBuilder.kt` has what is hopefully a
sufficiently thorough description of why it needs to work the way it works.

Although the code as it exists here has a few hardwired references to the
accessions table, it's designed with an eye toward making it easy to modify
to work with other kinds of data in the future if we want similar search
functionality.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
@@ -70,31 +70,49 @@ import org.jooq.impl.DSL
  *
  * ```kotlin
  * val queryBuilder = NestedQueryBuilder(dslContext)
- * val fields =
+ *
+ * queryBuilder.addSelectFields(
  *     listOf(
  *         searchFields["id"],
  *         searchFields["species"],
  *         searchFields["bags.number"],
  *         searchFields["germinationTests.startDate"],
- *         searchFields["germinationTests.germinations.seedsGerminated"])
- * queryBuilder.addSelectFields(fields)
- * queryBuilder.addSortFields(fields)
+ *         searchFields["germinationTests.germinations.seedsGerminated"]))
+ *
+ * // You can sort on fields you didn't select, and vice versa, though we won't do that here just
+ * // to keep the example easier to follow.
+ * queryBuilder.addSortFields(
+ *     listOf(
+ *         SearchSortField(searchFields["species"]),
+ *         SearchSortField(searchFields["germinationTests.startDate"]),
+ *         SearchSortField(searchFields["bags.number"]),
+ *         SearchSortField(searchFields["germinationTests.germinations.seedsGerminated"])))
+ *
  * val results = queryBuilder.toSelectQuery().fetch()
  * ```
  *
- * You'll get back a list of results that looks like this:
+ * You'll get back a list of results that looks like this, minus the comments, of course:
  *
  * ```json
  * [
+ *   # The first sort key is "species". "First Species" is alphabetically lower than "Second
+ *   # Species", so accession 3 is first in the results list.
  *   {
  *     "id": "3",
+ *     "species": "First Species",
  *     "bags": [
+ *       # The first (and only) sort field under "bags" is the bag number, so this list is
+ *       # sorted alphabetically on that value.
  *       { "number": "First bag for accession 3" },
  *       { "number": "Second bag for accession 3" }
  *     ],
  *     "germinationTests": [
+ *       # The first sort field under "germinationTests" is the start date, so this list is sorted
+ *       # in ascending start date order.
  *       {
  *         "germinations": [
+ *           # The first sort field under "germinationTests.germinations" is "seedsGerminated",
+ *           # so this list is sorted by that.
  *           { "seedsGerminated": "10" },
  *           { "seedsGerminated": "20" }
  *         ],
@@ -103,12 +121,14 @@ import org.jooq.impl.DSL
  *       {
  *         "startDate": "2021-10-15"
  *       }
- *     ],
- *     "species": "First Species"
+ *     ]
  *   },
  *   {
- *     "id": "4",
- *     "species": "Second Species"
+ *     # Empty lists, and scalar fields with no values, are omitted from the result. Each result
+ *     # is a Map<String,Any> which might be implemented as a hashtable, so callers shouldn't
+ *     # assume the fields are in any particular order.
+ *     "species": "Second Species",
+ *     "id": "4"
  *   }
  * ]
  * ```
@@ -146,7 +166,9 @@ import org.jooq.impl.DSL
  * ]
  * ```
  *
- * # Field types
+ * # Implementation details
+ *
+ * ## Field types
  *
  * Search results are returned as a list of maps of field names to field values. A field value can
  * either be a string or a list of maps of field names to field values. Kotlin doesn't have support
@@ -170,7 +192,7 @@ import org.jooq.impl.DSL
  * `germinationTests` each element of which contains a sublist field called `germinations` each
  * element of which contains a scalar field called `recordingDate`.
  *
- * # Query hierarchy
+ * ## Query hierarchy
  *
  * The query is represented as a tree of [NestedQueryBuilder]s. The root node represents the query
  * as a whole, and each child node represents a sublist field.
@@ -186,11 +208,12 @@ import org.jooq.impl.DSL
  * ```
  *
  * Field names are always evaluated relative to the current node (that is, the prefix is stripped
- * off to get a relative name). So for example, the middle node in the above hierarchy treats the
- * search field as having a name of `germinations.recordingDate`. The presence of the `.` in that
- * name tells it that there needs to be another sublist field called `germinations`.
+ * off to get a relative name). So for example, the middle node in the above hierarchy strips off
+ * its prefix and treats the search field as having a name of `germinations.recordingDate`. Because
+ * that name contains a `.` character, the middle node needs to peel off the part before the `.` and
+ * treat the field as a sublist called `germinations`.
  *
- * # Multisets
+ * ## Multisets
  *
  * This implementation is heavily dependent on implementation details of jOOQ's "multiset" operator.
  * For background on multisets, please see
@@ -211,7 +234,7 @@ import org.jooq.impl.DSL
  * two-dimensional array indexed by row number and field position. This implementation detail will
  * become important later.
  *
- * # Search criteria
+ * ## Search criteria
  *
  * Say you have an accession with two bags, "A" and "B". The user does a search and asks for results
  * containing bag "A".
@@ -220,8 +243,8 @@ import org.jooq.impl.DSL
  * on bag numbers (returning a result with a single bag) or treat it as a filter on accessions
  * (returning a result with two bags).
  *
- * Our product decision is to do the latter. When you search for bag "B", what you're really telling
- * the system is that you want all _accessions_ that have a bag called "B". And for each accession
+ * Our product decision is to do the latter. When you search for bag "A", what you're really telling
+ * the system is that you want all _accessions_ that have a bag called "A". And for each accession
  * that matches the search criteria, you get back the full list of bags.
  *
  * That's relevant here because it changes what the SQL looks like: we don't apply any search
@@ -249,7 +272,7 @@ import org.jooq.impl.DSL
  *
  * The key point is that the multiset will contain _all_ the bags for the accession.
  *
- * # Ordering
+ * ## Ordering
  *
  * Unfortunately, the happy-path usage of multisets breaks down once you need to sort the results
  * based on a field in a child table, because the ordering needs to bubble up through the rest of
@@ -309,7 +332,7 @@ import org.jooq.impl.DSL
  * ORDER BY bags_multiset[0][1]
  * ```
  *
- * # Lateral joins
+ * ## Lateral joins
  *
  * Unfortunately, an `ORDER BY` clause can't use a computed value from the `SELECT` clause of the
  * same query. So in the above example, `ORDER BY bags_multiset[0][1]` won't work because
@@ -337,7 +360,7 @@ import org.jooq.impl.DSL
  * Here the `bags_multiset` column is computed in the subquery and is an input to the top-level
  * query, so we can reference it in the `ORDER BY` clause.
  *
- * # Enums
+ * ## Enums
  *
  * Some fields are stored as integer IDs in the database, but rendered as human-readable strings in
  * the UI. For example, an accession with `state_id = 10` is shown as having a state of `Pending`
@@ -368,8 +391,8 @@ class NestedQueryBuilder(
 
   /**
    * Sublists indexed by the first element of their relative names. For example, if this node has a
-   * prefix of `a.b` and there is a field whose full name is `a.b.c.d.e`, the entry in this map
-   * would have a key of `c`.
+   * prefix of `a.b` and there are two fields whose full names are `a.b.c.d.e` and `a.b.f.g`, this
+   * map would contain keys `c` and `f`.
    */
   private val sublists = mutableMapOf<String, NestedQueryBuilder>()
 


sgrimm:
A lot of the diff on the test class here is just moving test methods into a couple new inner classes to make it easier for me to be selective about which tests to run. No changes to the actual test logic on those; the functional changes to the tests are all in the `NestedFieldsTest` inner class, and all in the form of brand-new test methods.

I was a little tempted to split it into multiple top-level test classes but during development it was pretty handy to be able to run all the search tests as a single unit. Happy to do that as a followup (or as part of this PR if you think the way it is now is no good.)

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,756 @@
...
+import org.jooq.OrderField
+import org.jooq.Record
+import org.jooq.RecordMapper
+import org.jooq.SelectJoinStep
+import org.jooq.SelectSeekStepN
+import org.jooq.SortField
+import org.jooq.impl.DSL
+
+/**
+ * Builds a database query that returns nested results from a tree of child tables.
```
sgrimm, line 16: I recommend starting here. Everything else in this PR is to support the logic this comment tries to explain.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,756 @@
...
+ *
+ * For example, if you query an accession that has multiple bags, the query constructed by this
+ * class will return a single row with a "bags" field that contains a list of the bags. If you were
+ * to just join the accessions and bags tables, you'd get back a separate query result for each bag.
```
kallen07: I'm not sure what the second sentence here is trying to convey. I think that having a non-trivial example would really help here. I like the example input that ends up in the swagger docs. I'm wondering if we can build off of that example here (or use it here)? Ideally, I could see a fairly complex example without needing to run the server and open the swagger UI.

I specifically want to see what the swagger docs example would output.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
kallen07: To answer your slack question, I think that leaving the comments here is good. I agree with you that having a separate file makes it easier to miss, and less likely to stay updated.

My thoughts so far on this comment are that it's currently trying to do two things
1. explain *what* problem this class solves and how to use it by using examples, defining the input types, and defining how a query is interpreted and represented
2. explain *how* this class solves that problem by talking about how multisets are used, how ordering is achieved, and lateral joins. 

I think that those two things should be kept very distinctly separate. I may only want to know what problem this class solves and how to use it. If I do want to know more about implementation details, I want to understand the interface first.

Perhaps you could refactor this comment to make a clear separation between 1. and 2. Or, you could consider putting those implementation details inside the class definition alongside the code that is actually doing that implementation. That way, it's easier to understand the implementation when I'm looking for it (honestly haven't gotten that deep yet).

One question I asked myself is, "how do I use this class"? I looked at the actual usage in code, but I think including that in the class comments *may* be useful.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,756 @@
...
+ * children, e.g., an accession has germination tests each of which has germinations.
+ *
+ * Under the covers, on PostgreSQL, jOOQ turns a multiset into an aggregation operation over a
+ * subquery. The aggregation function returns a JSON array where each element of the array
+ * represents a row of results from the subquery. Each of those elements is itself a JSON array of
+ * the field values in the subquery's SELECT clause. In other words, the multiset turns into a
+ * two-dimensional array indexed by row number and field position. This implementation detail will
+ * become important later.
+ *
+ * # Search criteria
```
kallen07: I think the answer to the question "how do we interpret a query"? Should be further up, probably with the example input and output. The implementation details can stay a bit further down here though.

src/main/kotlin/com/terraformation/backend/util/MemoizedValue.kt
```diff
@@ -0,0 +1,32 @@
...
+    get() = value != null
+
+  /**
+   * Returns the value, computing it if it hasn't previously been accessed.
+   *
+   * If another thread is currently computing the value, blocks until that thread finishes.
+   */
+  fun get(computeFunc: () -> T): T {
+    return value
+        ?: synchronized(this) {
```
kallen07, line 23: this synchronization is pretty cool!

sgrimm:
Converted this to a draft until I revise the docs based on the first round of comments.

sgrimm:
Added two new sections to the top of the class doc describing how to call it and giving an example. Hopefully that'll provide enough context to make things clearer.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ * Suppose you have the following data. (For clarity, this omits irrelevant columns.)
+ *
+ * ```sql
+ * INSERT INTO species (id, name) VALUES (1, 'First Species');
+ * INSERT INTO species (id, name) VALUES (2, 'Second Species');
+ *
+ * INSERT INTO accessions (id, species_id) VALUES (3, 1);
+ * INSERT INTO accessions (id, species_id) VALUES (4, 2);
+ *
+ * INSERT INTO bags (accession_id, number) VALUES (3, 'First bag for accession 3');
```
kallen07, line 46: How come number is a string value here?

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ * ```
+ * - A root node with no prefix with a sublist field `germinationTests`:
+ *    - A node with prefix `germinationTests` and a sublist field `germinations`:
+ *      - A node with prefix `germinationTests.germinations` and a scalar field `recordingDate`
+ * ```
+ *
+ * Field names are always evaluated relative to the current node (that is, the prefix is stripped
+ * off to get a relative name). So for example, the middle node in the above hierarchy treats the
+ * search field as having a name of `germinations.recordingDate`. The presence of the `.` in that
+ * name tells it that there needs to be another sublist field called `germinations`.
```
kallen07: I think the last sentence here is a bit confusing. Does this rephrase capture what you mean?

"The presence of the `.` in that name is the reason that middle node is a sublist field called `germinations`"

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ * # Search criteria
+ *
+ * Say you have an accession with two bags, "A" and "B". The user does a search and asks for results
+ * containing bag "A".
+ *
+ * There are two ways that could work, neither of them wrong: treat the search criterion as a filter
+ * on bag numbers (returning a result with a single bag) or treat it as a filter on accessions
+ * (returning a result with two bags).
+ *
+ * Our product decision is to do the latter. When you search for bag "B", what you're really telling
```
kallen07: Nit: change this to bag "A" since that's what is used above.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ *     "species": "First Species",
+ *     "bagNumber": "Second bag for accession 3"
+ *   },
+ *   {
+ *     "id": "4",
+ *     "species": "Second Species"
+ *   }
+ * ]
+ * ```
+ *
```
kallen07, line 168: Maybe add some kind of separation here to say that we're now talking about implementation details?

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+   */
+  private val conditions = mutableListOf<Condition>()
+
+  /** Scalar fields indexed by relative name. */
+  private val scalarFields = mutableMapOf<String, SearchField>()
+
+  /**
+   * Sublists indexed by the first element of their relative names. For example, if this node has a
+   * prefix of `a.b` and there is a field whose full name is `a.b.c.d.e`, the entry in this map
+   * would have a key of `c`.
```
kallen07: Do you mean? 

For example, if this node has a prefix of `a.b` and two fields with names `a.b.c.d.e` and `a.b.f.g`, this map would have keys `c` and `f`.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ *           { "seedsGerminated": "10" },
+ *           { "seedsGerminated": "20" }
+ *         ],
+ *         "startDate": "2021-10-01"
+ *       },
+ *       {
+ *         "startDate": "2021-10-15"
+ *       }
+ *     ],
+ *     "species": "First Species"
```
kallen07: How is the order of fields being determined? Asking because the sort fields list is in the following order `id, species, bags.number, germinationTests.startDate, germinationTests.germinations.seedsGerminated`

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+   * prefix of `a.b` and there is a field whose full name is `a.b.c.d.e`, the entry in this map
+   * would have a key of `c`.
+   */
+  private val sublists = mutableMapOf<String, NestedQueryBuilder>()
+
+  /**
+   * Zero-indexed position of each field in the `SELECT` clause. Both scalar and sublist fields are
+   * included. However, this does _not_ include fields that are only used as sort keys.
+   */
+  private val selectFieldPositions = mutableMapOf<String, Int>()
```
kallen07, line 403: Why do we need to store the position of each field in the select clause? In the above example, it seems like the fields are returned in alphabetical order.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ * [you have to repeat the subquery.](https://stackoverflow.com/questions/69577525/how-to-order-by-values-from-nested-multisets)
+ *
+ * Doing it that way would make queries more expensive, so instead we break portability and rely on
+ * the way multisets are implemented on PostgreSQL. Recall that they turn into two-dimensional
+ * arrays.
+ *
+ * We know that the multiset is already sorted correctly: the first row will always be the one that
+ * should determine the sort order of the parent row. In the example query, the first row of the
+ * multiset will have bag number "X" for accession 1 and "A" for accession 2 thanks to the `ORDER
+ * BY` in the subquery.
```
kallen07, line 314: Probably the most important paragraph in this file 😮

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
```diff
@@ -0,0 +1,880 @@
...
+ *
+ * We also know the order of the fields in the multiset's subquery. In the example, the field we
+ * want to sort on is the second one, `bags.number`.
```
kallen07, line 317: Is that because the order is based on the order we used for the SELECT? Do we select the fields based on the order we know we need them sorted?

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: I think "bag number" started out as an actual number but then the forestry folks decided it'd be better to use a shorter label while still having enough uniqueness to not have too many collisions among the bags that come in on a given day. So now, I believe, it's a three-character alphanumeric string that's randomly generated by the mobile app.

sgrimm: I used a long string here just to make the examples clearer; an actual bag number would be like `F7P`.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: Good call. I'll add an "Implementation details" header and then make the subheadings one level smaller.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: They're in whatever order they were passed to `addSelectFields()` by the caller. I'll update the example so it doesn't look like it's alphabetical.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: Yes to the first question (and that's why we need to remember the position of each field) but the order that the fields appear in the SELECT clause is unrelated to which ones are used for sorting.

However, this question and your other ones have made me realize that it wasn't a good idea to try to shorten my example by passing the same field list to both `addSelectFields()` and `addSortFields()` because it implies that they need to be the same. Will update the example to fix that.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: The order is deterministic at the moment but I don't want to make it part of the API contract; clients should assume the server is free to return fields in any order as long as they're all there somewhere. I'll update the example to make that explicit. (I'm also adding some explanatory comments inside that JSON block and I'll mention it there.)

sgrimm: I mean the order of the fields inside a single JSON object. The order of the objects in a list is determined by whatever sort fields the user asks for.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: Yeah, good point, having two examples would make this much clearer.

src/main/kotlin/com/terraformation/backend/seedbank/search/NestedQueryBuilder.kt
sgrimm: Rewrote this as

```
 * Field names are always evaluated relative to the current node (that is, the prefix is stripped
 * off to get a relative name). So for example, the middle node in the above hierarchy strips off
 * its prefix and treats the search field as having a name of `germinations.recordingDate`. Because
 * that name contains a `.` character, the middle node needs to peel off the part before the `.`
 * and treat the field as a sublist called `germinations`.
```

PR #85, opened 2021-10-22
SW-361 Return HTTP 404 if feature photo file is missing

sgrimm:
If a photo has a storage URL in the database but the file referred to by the
URL doesn't exist, we should return HTTP 404 instead of HTTP 500 when a client
tries to download it.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
@@ -311,10 +311,15 @@ class FeatureStore(
   ): SizedInputStream {
     val photosRow = getPhotoMetadata(featureId, photoId)
 
-    return if (maxWidth != null || maxHeight != null) {
-      thumbnailStore.getThumbnailData(photoId, maxWidth, maxHeight)
-    } else {
-      fileStore.read(photosRow.storageUrl!!)
+    return try {
+      if (maxWidth != null || maxHeight != null) {
+        thumbnailStore.getThumbnailData(photoId, maxWidth, maxHeight)
+      } else {
+        fileStore.read(photosRow.storageUrl!!)
+      }
+    } catch (e: NoSuchFileException) {
+      log.error("Feature $featureId photo $photoId file ${photosRow.storageUrl} not found")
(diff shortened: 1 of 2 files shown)


sgrimm:
Tested this end-to-end in the web UI by updating the storage URL to point to a nonexistent file and bringing up the feature details. The server returns 404.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -311,10 +311,15 @@ class FeatureStore(
...
-      fileStore.read(photosRow.storageUrl!!)
+    return try {
+      if (maxWidth != null || maxHeight != null) {
+        thumbnailStore.getThumbnailData(photoId, maxWidth, maxHeight)
+      } else {
+        fileStore.read(photosRow.storageUrl!!)
+      }
+    } catch (e: NoSuchFileException) {
+      log.error("Feature $featureId photo $photoId file ${photosRow.storageUrl} not found")
+      throw PhotoNotFoundException(photoId)
```
sgrimm, line 322: This is a subclass of `EntityNotFoundException` and thus gets mapped to HTTP 404 by `ControllerExceptionHandler` when it bubbles up through a controller method.

kallen07:
It seems like if we run into this error in prod, that would indicate a serious database consistency issue. Will we have a way to look for this error in the log and generate an oncall ticket if this happens?

I'm also wondering... do we use 404 when the resource isn't found due to a server issue? Or is it actually better to use 500 in that case? On the client side, if I get a 404 then I typically think that I must have passed an invalid resource ID.

sgrimm:
It's a legitimate question and I thought about it before I wrote this code. To me, the deciding factor was recoverability. When I get an HTTP 500, I usually expect that if I wait for the bug to be fixed on the server side and try again, it'll work. But in this case, the file is legitimately gone (we get a different exception if there's an I/O error) and trying again later won't cause it to reappear.

We can already see this in the logs in Datadog by filtering for log level. But that means remembering to search. Once we've integrated Sentry with our setup (there's a Jira task for it already), we'll be able to set up Slack alerts for ERROR-level log messages like this one.

PR #87, opened 2021-10-22
SW-278 Add API endpoints for facility automations

sgrimm:
The device manager will support various automations that can be run at a
facility. Each automation needs automation-specific configuration data. Add a
series of CRUD endpoints to manage the data.

The automations are stored in a new `automations` table. The Flyway script that
creates that table introduces something we should have been using all along:
instead of having a separate SQL file for Postgres and non-Postgres databases
where the only difference is the type of the JSONB column, instead it uses
Flyway's built-in variable substitution feature and defines a variable for
the column type that should be used for JSON data.


src/main/kotlin/com/terraformation/backend/customer/model/AutomationModel.kt
```diff
@@ -0,0 +1,30 @@
...
+    val createdTime: Instant,
+    val modifiedTime: Instant,
+    val configuration: Map<String, Any?>?
+) {
+  constructor(
+      row: AutomationsRow,
+      objectMapper: ObjectMapper
+  ) : this(
+      id = row.id ?: throw IllegalArgumentException("id is required"),
+      facilityId = row.facilityId ?: throw IllegalArgumentException("facilityId is required"),
```
kallen07, line 24: Will this end up throwing a 5xx or a 4xx error if some data is missing when we create an AutomationModel instance inside the AutomationStore?

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -406,6 +438,20 @@ internal class PermissionTest : DatabaseTest() {
         updateAccession = true,
     )
 
+    permissions.expect(
```
kallen07, line 441: What do you think about storing the list of IDs as a variable here? And then constructing the list of Accession IDs, Automation IDs, Device IDs, etc from that list?

src/main/kotlin/com/terraformation/backend/customer/model/AutomationModel.kt
sgrimm: It'll be a 5xx, which seems like what we'd want.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
sgrimm: Oh, that's a good idea. I'll do that in a separate PR.

PR #88, opened 2021-10-26
SW-279 Add device manager notification endpoint

sgrimm:
Add an API endpoint the device manager can use to send email notifications when it
detects a problem at a facility. This is a stopgap until we have a more complete
design for how alerting and monitoring should work.

The notifications are sent to a per-facility list of email addresses. The admin UI
now has a facility page (previously, facilities were just listed on the site page)
where the notification addresses can be edited. Other facility information can be
edited there too.

The server can be configured to send mail using either SMTP or the AWS SES API.
A series of config options can be set for use in dev or staging environments to
control the recipients and subject lines of email messages. By default, it just
writes messages to the server log and doesn't try to send them at all.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -24,7 +24,7 @@ spring:
     # Spring Boot Mail checks for the existence of this option; we want to enable it even if
     # we aren't using SMTP.
     host: dummy
-    # Make authentication and TLS opt-out in local configurations since we usually want it.
+    # Default to using authentication and TLS. Can opt out in local configuration.
     properties:
       mail:
         smtp:


src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -0,0 +1,161 @@
...
+
+        throw e
+      }
+    } else {
+      log.info("Email sending is disabled; did not send the generated message.")
+    }
+  }
+
+  /** Sends the message using the SES SendEmail API. */
+  private fun sendSes(message: MimeMessage, rawMessage: ByteArray): String? {
```
sgrimm, line 129: There is a Spring library that abstracts away the "are you using SMTP or SES?" logic so you don't have to do the explicit AWS API calls. But it is currently based on the AWS Java SDK version 1, and we're already using version 2. I figured it was better to write this method rather than pulling in an entire additional AWS SDK as a transitive dependency. If/when they switch to the newer SDK, maybe we can get rid of this method.

README.md
```diff
@@ -116,6 +116,56 @@ For example, you could create `src/main/resources/application-dev-foo.yaml` and
 
 You can have as many local profiles in the list as you like. If the same config setting is specified in more than one of them, the last one wins.
 
+## Configuring email
+
+By default, the server will output email messages to its log rather than trying to send them. To enable email, you need to tell it which mail server to connect to and how to authenticate. It supports both SMTP and the AWS Simple Email Service (SES) version 2.
```
kallen07, line 121: Maybe silly question, but how do we set these variables in prod and staging?

src/main/resources/application.yaml
```diff
@@ -20,6 +20,19 @@ spring:
     placeholders:
       jsonColumnType: "JSONB"
 
+  mail:
+    # Spring Boot Mail checks for the existence of this option; we want to enable it even if
+    # we aren't using SMTP.
+    host: dummy
+    # Make authentication and TLS opt-out in local configurations since we usually want it.
```
kallen07: Maybe rephrase as "Default to using authentication and TLS. Can opt out in local configuration."

README.md
sgrimm: In prod and staging, we set environment variables in the `docker-compose.yml` that gets installed using Ansible.

PR #90, opened 2021-11-06
Remove explicit ID lists from PermissionTest

sgrimm:
Previously, the assertions in `PermissionTest` checked a bunch of hardwired
object IDs for the correct permissions. This was repetitive and hard to read
because it wasn't immediately obvious what a given list of IDs was supposed to
represent.

Extract those ID lists into variables. This reduces repetition and the variable
names should help make it clearer what behavior the tests are trying to verify.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
@@ -4,6 +4,7 @@ import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.db.PermissionStore
 import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.model.PermissionTest.PermissionsTracker
 import com.terraformation.backend.db.AccessionId
 import com.terraformation.backend.db.AccessionState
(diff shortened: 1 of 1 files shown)


sgrimm:
I was changing the permissions logic for species management which meant editing this test, and it seemed like a good time to actually do the simplification we've talked about. It can probably be improved further but I think this is at least a step in the right direction.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -4,6 +4,7 @@ import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.db.PermissionStore
 import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.model.PermissionTest.PermissionsTracker
```
sgrimm, line 7: A bit counterintuitive that this file needs to import from itself, but it's to resolve the reference I added to the class doc.

sgrimm:
If the thought occurs to you that it should be possible to dynamically compute the list of child IDs for a given project or organization based on the powers-of-10 ID numbering scheme in the test, well, you're in good company. I tried it. Several different ways. The results were all more verbose and less readable than what's in the PR.

PR #94, opened 2021-11-09
TRACK-418 Decouple search fields from seed tables

sgrimm:
Continuing the groundwork for search APIs for organization and GIS data, move the
search field classes out of the seedbank package.

Since the classes are no longer inner classes of the seedbank `SearchFields` class,
they can no longer refer to that class's `fuzzySeachOperators` and instead have to
take it as an argument, which would be annoying to have to do for every text field
because there are a lot of them.

It also no longer makes sense for the `accessions` search table to be the default.
But it would still be pretty annoying to have to pass it to their constructors
a zillion times in `SearchFields`.

To solve both those problems, `SearchTable` is now an abstract class instead of an
interface, and it has a bunch of helper methods to create fields. The net effect
is that in `SearchFields`, instead of doing

    TextField(
            "fieldName",
            "displayName",
            GERMINATIONS.WHATEVER,
            SearchTables.Germination)

you now do

    searchTables.germinations.textField(
            fieldName",
            "displayName",
            GERMINATIONS.WHATEVER)

There are helper methods for all the field types we use in the existing seedbank
search API, except for a handful such as the `active` field that are very specific
to the seed data model and wouldn't be useful elsewhere. Those fields continue to
be implemented as one-off inner classes in `SearchFields`.

The search table definitions in the seedbank `SearchTables` class are now `val`s
instead of static `object`s because they need to take constructor arguments that
aren't available until dependency injection happens.

With these changes, it should be possible to define a set of search fields and
tables that have nothing to do with the seedbank data model.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
@@ -6,7 +6,7 @@ import com.terraformation.backend.seedbank.search.SearchService
 data class SearchResults(
     /**
      * List of results containing the fields specified by the caller. Each element of the list is a
-     * map of field name to non-null value. If an accession does not have a value for a particular
+     * map of field name to non-null value. If a result does not have a value for a particular
      * field, it is omitted from the map.
      *
      * Each value is either `String` or `List<Map<String, Any>>`.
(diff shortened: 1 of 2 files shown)


sgrimm:
Marking this as a draft -- there was something I didn't take into account when I was reorganizing the classes here and I'm not sure if I'll have to do something different.

PR #95, opened 2021-11-11
TRACK-419 Merge plants and features internally

sgrimm:
Change the internal modeling of plants so they are treated as optional
supplementary data about features, rather than as first-class entities. As we
add other feature types, there is now a straightforward way to model them in
the application code.

This is purely an implementation change; the client-facing API continues to behave
exactly as before.

With this structure, it no longer makes sense for plants to have their own
creation and modification timestamps, so drop those columns from the plants table.

`PlantStore` is now merged into `FeatureStore`. It still has methods to directly
operate on plants, though those will go away once we no longer need to support the
plant-centric API endpoints.

Likewise, `PlantStoreTest` is rolled into `FeatureStoreTest`, minus test cases
related to plants having timestamps.

As part of this change, I also fixed two bugs in `fetchPlantsList`: it wasn't
filtering by layer ID, and it was returning search results in an undefined order.
The latter issue caused test cases to fail if the database didn't happen to return
rows in insertion order.


src/main/kotlin/com/terraformation/backend/gis/model/FeatureModel.kt
```diff
@@ -16,4 +17,13 @@ data class FeatureModel(
     val enteredTime: Instant? = null,
     val createdTime: Instant? = null,
     val modifiedTime: Instant? = null,
-)
+    val plant: PlantsRow? = null,
```
kallen07, line 20: In the future, if we had an "irrigation line" feature, would there be a way to ensure that plant and irrigationLine fields don't both exist on the feature model?

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -182,31 +142,44 @@ class FeatureStore(
...
-              .set(MODIFIED_TIME, currTime)
-              .where(ID.eq(featureId))
-              .returningResult(GEOM.transformSrid(SRID.LONG_LAT).`as`(GEOM))
-              .fetchOne()
-              ?.getValue(GEOM)
-              ?: throw DataAccessException("Database did not return Geom")
+    return dslContext.transactionResult { _ ->
+      val currTime = clock.instant()
+
+      val longLatGeom =
```
kallen07, line 152: did you choose to grab the geom here because of rounding changes that may happen when saving to the database?

src/main/kotlin/com/terraformation/backend/gis/model/FeatureModel.kt
sgrimm: We'd validate it in the constructor (or rather, in the `init` block).

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
sgrimm: More because of the coordinate transformation: the client can pass us a geometry in any coordinate system PostGIS supports, but we want to always return geometry in the same system for consistency.

PR #96, opened 2021-11-12
TRACK-419 Add plant details to features API

sgrimm:
Update the API payload classes for features to add plant information in the
form of an optional child object. Clients can include plant information when
they create features, add it to existing features, and retrieve it as part of
the feature data.

The plants API continues to be supported for backward compatibility, but its
functionality should now be covered by the features API. We can remove it once
clients are updated.

As part of updating the features API, I also cleaned up a couple minor things:

* There's no longer a need to catch `FeatureNotFoundException` and rethrow it
  as `NotFoundException`.
* Clients no longer have to pass in a layer ID when updating a feature.
  Previously, it was required but wasn't modifiable.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
@@ -5,7 +5,7 @@ import com.terraformation.backend.customer.api.ModifyAutomationRequestPayload
 import com.terraformation.backend.device.api.CreateDeviceRequestPayload
 import com.terraformation.backend.device.api.DeviceConfig
 import com.terraformation.backend.device.api.UpdateDeviceRequestPayload
-import com.terraformation.backend.search.SearchField
+import com.terraformation.backend.search.field.SearchField
 import com.terraformation.backend.seedbank.api.AccessionPayload
 import com.terraformation.backend.seedbank.api.GerminationTestPayload
 import com.terraformation.backend.seedbank.api.SearchResponsePayload
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.search
 
+import com.terraformation.backend.search.field.SearchField
 import com.terraformation.backend.seedbank.search.SearchFields
 import com.terraformation.backend.seedbank.search.SearchService
 import com.terraformation.backend.seedbank.search.SearchTables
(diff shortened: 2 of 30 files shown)


src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
```diff
@@ -63,17 +66,22 @@ class FeatureController(private val featureStore: FeatureStore) {
...
+  @Operation(
+      summary =
+          "Lists the features associated with a layer. Can optionally filter by entered time, " +
+              "notes, and (for plants) species name.")
   @GetMapping("/list/{layerId}")
   fun list(
+      @QueryParam("minEnteredTime") minEnteredTime: Instant? = null,
+      @QueryParam("maxEnteredTime") maxEnteredTime: Instant? = null,
+      @QueryParam("notes") notes: String? = null,
+      @QueryParam("speciesName") speciesName: String? = null,
```
kallen07: For filters, currently the user is able to select the species name from a dropdown. Since we're filtering by exact values, rather than applying some kind of species name search, I'm wondering what you think about passing the species ID here instead of the name.

The reason I say that is because we're only returning the species ID on the plant object, and in the future we may have to deal with the fact that some species have the same secondary names.

src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
```diff
@@ -63,17 +66,22 @@ class FeatureController(private val featureStore: FeatureStore) {
...
 
     return GetFeatureResponsePayload(FeatureResponse(model))
   }
 
   @ApiResponse(responseCode = "200")
-  @Operation(summary = "List all features associated with a layer.")
+  @Operation(
+      summary =
+          "Lists the features associated with a layer. Can optionally filter by entered time, " +
+              "notes, and (for plants) species name.")
```
kallen07, line 78: Do we want to leave a note here about notes being used for fuzzy searching (pun intended)?

src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
```diff
@@ -333,18 +343,47 @@ data class FeatureResponse(
     val attrib: String? = null,
     val notes: String? = null,
     val enteredTime: Instant? = null,
+    val plant: PlantDetailsPayload? = null,
```
kallen07, line 351: It looks like the client needs to specify a non-null value for at least one plant field in order for the feature to be marked as a plant. If so, that changes the data model in a meaningful way because all the plants information is currently optional. Is that what we want?

sgrimm: You can do `{"layerId": 12345, "plant": {}}` and it'll create the feature as a plant with no plant fields filled in.

src/main/kotlin/com/terraformation/backend/gis/api/FeatureController.kt
sgrimm: That makes sense, yeah. Ambiguous names hopefully won't be an actual problem; as part of my change for per-org species, I'm requiring species names to be unique within an organization, just not globally unique. We'll see if people find that acceptable but the "what species do you mean by this name" problem is arguably pretty bad user experience, never mind the technical side.

sgrimm: Also just noting that it would be super easy to make species name use fuzzy search, if we want that. But for this change, I'll replace name with ID on this endpoint since there's no backward compatibility question. I'll keep name as-is on the plants API.

PR #98, opened 2021-11-16
TRACK-428 Fetch organization hierarchy

sgrimm:
Update the organization list endpoint to optionally also return a list of 
projects, sites, and facilities for each organization. The client can specify
how deep it wants the results to be.

Add an endpoint to get the information for a single organization as opposed to
listing all of the available ones.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
@@ -48,10 +48,12 @@ class OrganizationStore(
       organizationId: OrganizationId,
       depth: FetchDepth = FetchDepth.Organization
   ): OrganizationModel? {
-    return if (organizationId in currentUser().organizationRoles) {
+    val user = currentUser()
+
+    return if (organizationId in user.organizationRoles) {
       selectForDepth(depth, ORGANIZATIONS.ID.eq(organizationId)).firstOrNull()
     } else {
-      log.warn("User ${currentUser().userId} attempted to fetch organization $organizationId")
+      log.warn("User ${user.userId} attempted to fetch organization $organizationId")
       null
     }
   }
@@ -62,6 +64,10 @@ class OrganizationStore(
   ): List<OrganizationModel> {
     val user = currentUser()
     val organizationIds = user.organizationRoles.keys
+
+    if (organizationIds.isEmpty()) {
+      return emptyList()
+    }
 
     val facilitiesMultiset =
         if (depth.level >= FetchDepth.Facility.level) {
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx2400m -Dfil
 
 awsSdkVersion=2.17.22
 jacksonVersion=2.11.3
-jooqVersion=3.15.3
+jooqVersion=3.15.4
```
sgrimm, line 13: This was to pick up a minor bugfix related to JSONB values in multisets.

src/main/kotlin/com/terraformation/backend/customer/api/FacilityController.kt
```diff
@@ -67,12 +61,9 @@ class FacilityController(
   @Operation(summary = "Gets information about a single facility")
   fun getFacility(@PathVariable facilityId: FacilityId): GetFacilityResponse {
     val facility = facilityStore.fetchById(facilityId) ?: throw NotFoundException()
-    val role =
-        currentUser().facilityRoles[facilityId]
-            ?: throw IllegalStateException("BUG! No role for facility")
 
     return GetFacilityResponse(
-        FacilityPayload(facilityId, facility.siteId, facility.name, facility.type, role.name))
```
kallen07, line 75: Why are we removing the facility role here?

src/main/kotlin/com/terraformation/backend/customer/api/SiteController.kt
```diff
@@ -78,23 +79,26 @@ class ProjectSitesController(private val siteStore: SiteStore) {
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
 data class SiteElement(
     val id: SiteId,
     val name: String,
     val projectId: ProjectId,
     val location: Point,
     val locale: String?,
-    val timezone: String?
+    val timezone: String?,
+    val facilities: List<FacilityPayload>?,
```
kallen07, line 90: Facilities and map layers are both tied to sites. How come facilities are an option here but map layers are not?

src/main/kotlin/com/terraformation/backend/customer/model/SiteModel.kt
```diff
@@ -15,7 +18,23 @@ data class SiteModel(
...
+) {
+  constructor(
+      record: Record,
+      facilitiesMultiset: Field<List<FacilityModel>?>? = null,
+  ) : this(
+      record[SITES.ID] ?: throw IllegalArgumentException("ID is required"),
+      record[SITES.PROJECT_ID] ?: throw IllegalArgumentException("Project ID is required"),
+      record[SITES.NAME] ?: throw IllegalArgumentException("Name is required"),
+      record[SITES.LOCATION]?.firstPoint ?: throw IllegalArgumentException("Location is required"),
+      record[SITES.CREATED_TIME] ?: throw IllegalArgumentException("Created time is required"),
```
kallen07, line 31: How come the created and modified times are on the SiteModel but not the org or project models?

src/main/kotlin/com/terraformation/backend/customer/api/FacilityController.kt
sgrimm: Good catch. It was added before we'd really worked out what our roles/permissions setup was going to be. It no longer really makes sense as part of the facility data given that roles are per-org, not per-facility. Nothing was using it for anything anyway.

src/main/kotlin/com/terraformation/backend/customer/api/SiteController.kt
sgrimm: The idea here wasn't to allow you to traverse the whole data hierarchy, and it didn't seem like layers were needed to show the project/site/facility selector in the top nav which was kind of the target use case here.

If I'm wrong about that, I can add them, though then I maybe need to rethink the concept of "fetch depth" since facilities and layers are siblings in our data model.

My preference is to cover "list all the layers in all my projects" in the search API instead, unless not having it here is going to block front-end work.

kallen07: Ah, I see. Yeah, that makes sense. I wasn't thinking about this being used specifically to populate the dropdowns. 

Not a blocking question, but will these endpoints end up providing a subset of the eventual search functionality? As in, could we use the search APIs to get this same information? If so, does that mean we'll end up removing these later on?

sgrimm: My intention is that yes, you will eventually be able to get all this via the search API. I think it will likely still make sense to provide special-purpose GET endpoints with fixed responses, so the client doesn't have to construct a complete search request with the right parameters for fetching commonly-used data.

src/main/kotlin/com/terraformation/backend/customer/model/SiteModel.kt
sgrimm: We're pretty inconsistent about that across the API. Just filed https://terraformation.atlassian.net/browse/TRACK-471 about it.

PR #100, opened 2021-11-18
Fix incorrect field mapping in search results

sgrimm:
We were incorrectly returning the remaining quantity in grams in the search
results for withdrawals, instead of the raw remaining quantity value whose
units are specified in a separate field.


sgrimm:
Noticed this in the course of revamping the way search fields are represented internally.

I started trying to write a JUnit test to guarantee that all search fields are mapped to the correct underlying values, but it quickly became a fiddly mess since the search field values don't map 1:1 to the underlying database columns. I'll revisit that later since it'd be good to catch this kind of thing in the future, but didn't want to get sidetracked from the thing I'm actually trying to work on right now.

PR #101, opened 2021-11-22
TRACK-465 Richer data model for search field paths

sgrimm:
Rework the way the code represents search field names internally. Previously, they
were represented as strings, and the code used string manipulation to navigate
them while constructing the tree of sublists and scalar fields. This was
inflexible and would have made it awkward to support a more flexible search API
that can navigate the data model more freely.

The client-facing API still uses strings, but now those strings are parsed into
path objects where each element of the path carries information about what part of
the data model it refers to. The list of field definitions is now broken apart
into separate "namespaces" that are arranged in a hierarchy, rather than defined
as a flat list where the hierarchy of sublists was implicit in the field names.

There is no functionality change here! This is a pure refactor and the accession
search API should behave identically from the client's point of view. The existing
test suite is unmodified here aside from switching to a different method for
resolving field names.

Subsequent changes will build on this to start building a more sophisticated
client-facing API.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx2400m -Dfil
 
 awsSdkVersion=2.17.22
 jacksonVersion=2.11.3
-jooqVersion=3.15.3
+jooqVersion=3.15.4
 kotlinVersion=1.5.10
 postgresJdbcVersion=42.2.24
 springDocVersion=1.5.10
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1080,9 +1080,22 @@ paths:
     get:
       tags:
       - Customer
-      summary: List all organizations
-      description: List all organizations the user can access.
-      operationId: listAll_1
+      summary: Lists all organizations.
+      description: Lists all organizations the user can access.
+      operationId: listOrganizations
+      parameters:
+      - name: depth
+        in: query
+        required: false
+        schema:
+          type: string
+          description: Return this level of information about the organization's contents.
+          default: Organization
+          enum:
+          - Organization
+          - Project
+          - Site
+          - Facility
       responses:
         "200":
           description: OK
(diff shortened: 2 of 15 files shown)


src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
```diff
@@ -196,19 +199,50 @@ import org.jooq.impl.DSL
  * `germinationTests` each element of which contains a sublist field called `germinations` each
  * element of which contains a scalar field called `recordingDate`.
  *
- * ## Query hierarchy
- *
- * The query is represented as a tree of [NestedQueryBuilder]s. The root node represents the query
- * as a whole, and each child node represents a sublist field.
+ * A field name is represented as a [SearchFieldPath] which consists of a prefix and a scalar field.
+ * The prefix is represented as a [SearchFieldPrefix] and includes a list of path elements. Each
+ * element is a sublist sublist field, forming a path to the location of the scalar field
```
kallen07: sublist sublist?

src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
```diff
@@ -0,0 +1,154 @@
+package com.terraformation.backend.search
+
+import com.fasterxml.jackson.annotation.JsonValue
+import com.terraformation.backend.search.field.SearchField
+
+/**
+ * An element of a search field path prefix.
+ *
+ * Currently, there is only one kind of element (sublists) but later revisions of this code will
+ * introduce an additional type.
```
sgrimm, line 10: This comment turns out to be wrong, or at least only temporarily right; the next PR in the series does introduce an additional type, but then the (upcoming) one after that collapses it back to a single type and moves the single/multi value distinction somewhere else.

src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
```diff
@@ -0,0 +1,154 @@
...
+data class SublistPrefixElement(
+    override val name: String,
+    override val namespace: SearchFieldNamespace
+) : SearchFieldPrefixElement
+
+/**
+ * A partial location of a search field in the application's data hierarchy. The prefix specifies
+ * the starting point (root) of a relative path, and zero or more path elements underneath the root.
+ */
+data class SearchFieldPrefix(
```
kallen07, line 43: Is it possible to merge this with `SearchFieldPath`?

src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
```diff
@@ -0,0 +1,154 @@
...
+    val root: SearchFieldNamespace,
```
kallen07, line 44: I'm a little unsure why the comment says `The prefix specifies the starting point (root) of a relative path`, but then the data member is called `root` instead of `prefix`.

src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
sgrimm: In an early draft of this PR, it actually worked that way: the `SearchField` on a path was nullable, and a prefix was just a path with a null search field.

It's valid to do some things with a prefix that you can't do with a full field path and vice versa. For example, you can add more path elements to a prefix but it makes no sense to add elements to a field. So I had a bunch of `if (searchField != null) throw IllegalStateException()` kinds of sanity checks all over the place. It stood out as a code smell to me, a sign I wasn't using the type system to define those invalid operations out of existence.

src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
sgrimm: Good catch. This comment made more sense in an earlier version of the PR. Will reword it to reflect reality.

PR #102, opened 2021-11-22
TRACK-462 Support navigating up the data hierarchy

sgrimm:
Sublists now come in two flavors: multi-value and single-value.

Multi-value sublists are the new name for what used to just be called "sublists"
prior to this change. They represent children in the data model hierarchy, and a
given parent can have zero or more children.

Single-value sublists represent many-to-one relationships. Often these are
child-to-parent relationships, e.g., the relationship between an accession and
its facility. But there can also be other such relationships, e.g., between an
accession and its species. Thus we use the terms "multi-value" and "single-value"
rather than "child" and "parent" to make it clear that the difference is the
cardinality of the sublist, not the semantic relationship.

In search results, single-value sublists turn into single JSON objects, whereas
multi-value sublists continue to turn into lists of JSON objects.

This change adds `facility` as a single-value sublist of accessions, and tweaks
`SearchService` to use that sublist for filtering by facility ID. A future change
will allow clients to pass in multiple facility IDs or none at all.

As of this change, it is now possible to refer to a given field using an infinite
(in principle, anyway) number of different paths, simply by bouncing up and down
the hierarchy. It's possible we will want to disallow clients from doing this, but
for now, it is permitted and is verified by unit tests.

The fact that field paths can now be nested arbitrarily deep means it's impossible
to fully enumerate the complete list of valid paths as part of the OpenAPI schema.
However, the existing schema continues to describe a useful subset of the valid
inputs, so it is left as-is in this version of the code. A future change in this
series will get rid of the canned list of field names in favor of another way for
clients to find out the list of recognized fields.

This change also renames the namespace class that holds the list of accession
fields, since that field list no longer represents the "top" of the field
hierarchy. Future changes will continue to make the accessions namespace less of a
special case in the code.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
@@ -70,8 +70,6 @@ class SearchServiceTest : DatabaseTest(), RunsAsUser {
   private val checkedInTime = Instant.parse(checkedInTimeString)
 
   private val searchTables = SearchTables(PostgresFuzzySearchOperators())
-  private val searchTables = SearchTables(searchTables)
-  private val rootPrefix = SearchFieldPrefix(root = searchFields)
   private val accessionsNamespace = AccessionsNamespace(searchTables)
   private val rootPrefix = SearchFieldPrefix(root = accessionsNamespace)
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #102**  (this PR)
    * **PR #103** 
      * **PR #104** 
        * **PR #105** 
          * **PR #107** 
            * **PR #106** 

PR #103, opened 2021-11-25
TRACK-464 Rename property to reduce ambiguity

sgrimm:
An upcoming change will introduce another sublist-related property. To
reduce confusion over which sublist-related thing is which, rename
`NestedQueryBuilder.sublists` to `sublistQueryBuilders` to reflect the
fact that the values of that map are `NestedQueryBuilder` objects that
construct SQL queries for sublists.

This is purely a name change, no difference in functionality.


sgrimm:
In case it's not obvious: this depends on #102 .

sgrimm:
* main
  * **PR #103**  (this PR)
    * **PR #104** 
      * **PR #105** 
        * **PR #107** 
          * **PR #106** 
            * **PR #108** 

PR #104, opened 2021-11-25
TRACK-464 Generalize permission checking in search queries

sgrimm:
When we construct a query in the search code, we need to make sure the query
takes into account what data the user has permission to see. Concretely, that
means we need to make sure there's a join to a table that has an identifier we
can check against the list of things the user is allowed to access, e.g., a
facility ID we can check against `UserModel.facilityRoles`.

Previously, the search code was always searching the accessions table, so it
could count on being able to look there for a facility ID, and all we had to do
was make sure we joined with that table if we were doing a search that only
selected fields from a child or grandchild table. In practice, that was only
an issue with the `germinations` table where we had to make sure to join with
`germination_tests` so we could join to `accessions` from there.

But once clients are able to specify the root of a search query, we can't count
on there being a single hardwired place to look for permission information,
and we can't count on the presence of a facility ID. They might be searching
map layers or projects.

Permissions are inherited: your ability to see a germination test is inherited
from your ability to see its accession, which is inherited from your ability
to see the accession's facility, etc.

Add a property to `SearchTable` to represent the permission inheritance
hierarchy, and change `SearchService` to walk that hierarchy (recursively, if
necessary) to construct the proper `JOIN` statements for permission checking.

There is no change in behavior here; for accession queries, this ends up
constructing the same joins as the existing code. Existing test cases continue
to pass without changes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
@@ -40,24 +40,32 @@ abstract class SearchTable(private val fuzzySearchOperators: FuzzySearchOperator
   abstract fun <T : Record> leftJoinWithMain(query: SelectJoinStep<T>): SelectJoinStep<T>
 
   /**
-   * Adds a LEFT JOIN clause to a query to connect this table to any other tables required to filter
-   * out values the user doesn't have permission to see.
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #104**  (this PR)
    * **PR #105** 
      * **PR #107** 
        * **PR #106** 
          * **PR #108** 

PR #105, opened 2021-11-29
TRACK-464 Move accession search to a separate service

sgrimm:
Introduce `AccessionSearchService` so there's a separate entrypoint for searching
accessions that mimics the existing search API. This is groundwork for changing
`SearchService` to make it more general; at that point the logic for adding the
default set of accession fields and the facility ID filter will need to live
somewhere else.

This change is a mostly-mechanical one with no functionality difference, to reduce
clutter in later changes that change how the code works. For now,
`AccessionSearchService.search` is a pass-through for `SearchService.search`.

Specifically:

* Move `SearchService` out of the `com.terraformation.backend.seedbank.search`
  package and into `com.terraformation.backend.search`.
* Add `AccessionSearchService` with a single method that calls through to
  `SearchService`.
* Update calling code to call `AccessionSearchService.search`.

A subsequent change will turn the new wrapper method into more than just a no-op,
but with this change in place, callers will remain unaffected.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -2,7 +2,6 @@ package com.terraformation.backend.search
 
 import com.terraformation.backend.search.field.SearchField
 import com.terraformation.backend.seedbank.search.AccessionsNamespace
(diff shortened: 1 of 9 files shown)


sgrimm:
* main
  * **PR #105**  (this PR)
    * **PR #107** 
      * **PR #106** 
        * **PR #108** 

PR #106, opened 2021-11-29
TRACK-464 Search arbitrary root namespaces

sgrimm:
Update `SearchService` so the caller can pass in an arbitrary namespace as the
starting point for a search. The existing accession search functionality is now
just a canned set of parameters for this more general search API.

`SearchService` no longer has any accession-specific code.

This change doesn't define any new search tables, but it does include new test
cases to run accession-related searches using tables other than `accessions` as
the root.


sgrimm:
* main
  * **PR #106**  (this PR)
    * **PR #108** 
      * **PR #113** 

PR #107, opened 2021-11-29
TRACK-464 Model sublist fields explicitly

sgrimm:
This change introduces the `SublistField` class. As its name suggests, it
represents a sublist field, in the same way the existing `SearchField` class
represents a scalar field. This is a simpler and hopefully easier to understand
model of the search field hierarchy.

The new class, to be more precise, represents the _relationship_ between two
namespaces. Knowledge of whether a relationship is one-to-many or many-to-one now
lives in `SublistField`, which means there's no longer a need for separate
`SearchFieldNamespace` properties for single-value and multi-value sublists.

The SQL snippet required to connect a multiset query to its enclosing query also
lives in `SublistField` now. This replaces the `conditionForMultiset()` method on
`SearchTable`, which we need to do in order to fully support navigating
to tables such as `photos` that are referenced from multiple places in the data
model.

There are no behavior changes here; existing tests continue to pass as-is.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -246,12 +246,16 @@ import org.jooq.impl.DSL
  * sublistQueryBuilders:
  *   germinationTests:
  *     prefix:
+ *       # This prefix refers to the germinationTests sublist under accessions; the root
+ *       # is the same as the parent's root.
  *       root: accessions
  *       sublists: [germinationTests]
  *     scalarFields: []
  *     sublistQueryBuilders:
  *       germinations:
  *         prefix:
+ *           # This prefix refers to the germinations sublist under the germinationTests
+ *           # sublist under accessions. As before, the root is the same as the parent's.
  *           root: accessions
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #107**  (this PR)
    * **PR #106** 
      * **PR #108** 
        * **PR #113** 

src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
```diff
@@ -235,27 +235,26 @@ import org.jooq.impl.DSL
...
  *     scalarFields: []
  *     sublistQueryBuilders:
  *       germinations:
  *         prefix:
- *           namespace: accessions
- *           path: germinationTests.germinations
- *         scalarFields:
- *           - recordingDate
+ *           root: accessions
+ *           sublists: [germinationTests, germinations]
```
kallen07, line 260: Why is `germinations` included in the sublist for `germinations`?

sgrimm: Oh, yeah, that's confusing. Will add a comment to explain why it's that way.

PR #109, opened 2021-11-30
TRACK-488 Add current user's role to organization info

sgrimm:
Make the organization list and fetch endpoints return which role the user has
in each of their organizations.


sgrimm:
* main
  * **PR #109**  (this PR)
    * **PR #111** 
      * **PR #110** 

PR #110, opened 2021-11-30
Add plural aliases for organization endpoints

sgrimm:
In most of the API, we've standardized on using plural nouns in endpoint
names, but `OrganizationController` predated that decision. Add
`/api/v1/organizations` as an alternate name for `/api/v1/organization`.
Once clients have moved over to the new name, we can get rid of the old one.

No functional change here. If clients are using the operation ID from the
OpenAPI schema to look up endpoint paths, they will switch over to the new
names automatically; otherwise the old names continue to work for now.


sgrimm:
* main
  * **PR #110**  (this PR)

kallen07:
Will you create a jira ticket so that we can keep track of this on the frontend?

sgrimm:
Filed TRACK-503 to track updating the client code.

PR #111, opened 2021-11-30
Use consistent naming scheme for organization payloads

sgrimm:
Bring the payload names for `/api/v1/organization` in line with the naming
convention used in most of the code.

No functional change here, though clients that are using the OpenAPI schema for
code generation will need to be updated to use the new names.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1102,7 +1102,7 @@ paths:
           content:
             application/json:
               schema:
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #111**  (this PR)
    * **PR #110** 

kallen07:
Not going to merge because it looks like the Graphite integration didn't rebase after I landed the first diff.

PR #112, opened 2021-12-01
Reduce noise in database tests

sgrimm:
The Spring test framework logs a few very verbose INFO messages for each test
method as it sets up and tears down the database connection. This makes it more
of a pain to see the test results when running tests interactively.

Change the logging configuration to only show Spring log messages at WARN and
ERROR levels.


sgrimm:
* main
  * **PR #112**  (this PR)

PR #113, opened 2021-12-01
TRACK-485 Introduce field aliases

sgrimm:
The existing accession search API has fields at the accession level that are
actually sublist fields, e.g., `germinationTestType` is another name for
`germinationTests.type` but without the nesting.

There will shortly be a more general mechanism to ask for search results in
tabular form, but it will require specifying the field names in a different
format and we don't want to break existing client code.

Introduce a concept of "alias fields" which, as the name suggests, are a way
to refer to a field by a different name. This is done transparently to the
client: search results use the alias name rather than the canonical one.

Subsequent changes will replace the flattened-view-of-sublist fields in
`AccessionsNamespace` with aliases; this change just adds the capability to
define aliases at all.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
@@ -649,7 +649,7 @@ class SearchServiceTest : DatabaseTest(), RunsAsUser {
 
   @Test
   fun `can sort by aliased field that is not in select list`() {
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #113**  (this PR)
    * **PR #114** 
      * **PR #115** 
        * **PR #116** 
          * **PR #117** 
            * **PR #119** 
              * **PR #120** 
                * **PR #121** 
                  * **PR #123** 

PR #114, opened 2021-12-02
TRACK-485 Add namespaces for accession "parent" tables

sgrimm:
The accessions table has many-to-one relationships with some secondary tables.
These can now be represented as separate namespaces and accessed using
single-value sublist fields.

This change doesn't update any of the existing fields that point to values in
those secondary tables, just adds the sublists.  A subsequent change will turn
those fields into aliases.

Also, some of the sublists have less-than-ideal names because the ideal names
are already taken. For example, there is already a `species` field for the
species name, so the species sublist is called `speciesInfo` here. A subsequent
change will introduce some aliases for the colliding names; we'll update the
client to use those new aliases, then rename the sublists.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4454,6 +4454,7 @@ components:
       - estimatedSeedsIncoming
       - family
       - geolocation
+      - geolocations.coordinates
       - germinationEndDate
       - germinationPercentGerminated
       - germinationSeedType
@@ -4505,6 +4506,7 @@ components:
       - totalViabilityPercent
       - treesCollectedFrom
       - viabilityTestType
+      - viabilityTestTypes.type
       - withdrawalDate
       - withdrawalDestination
       - withdrawalGrams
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #114**  (this PR)
    * **PR #115** 
      * **PR #116** 
        * **PR #117** 
          * **PR #119** 
            * **PR #120** 
              * **PR #121** 
                * **PR #123** 
                  * **PR #124** 

src/main/kotlin/com/terraformation/backend/search/namespace/AccessionGerminationTestTypesNamespace.kt
```diff
@@ -0,0 +1,21 @@
+package com.terraformation.backend.search.namespace
+
+import com.terraformation.backend.db.tables.references.ACCESSION_GERMINATION_TEST_TYPES
+import com.terraformation.backend.search.SearchFieldNamespace
+import com.terraformation.backend.search.SublistField
+import com.terraformation.backend.search.field.SearchField
+import com.terraformation.backend.seedbank.search.SearchTables
+
+class AccessionGerminationTestTypesNamespace(searchTables: SearchTables) : SearchFieldNamespace() {
```
sgrimm, line 9: This name is gross, and it doesn't even match the client-facing field name. But renaming it seemed out of scope for this PR.

PR #116, opened 2021-12-02
TRACK-485 Add flattened sublists

sgrimm:
Before sublists were introduced, you had to query accession data using a flat
set of fields, some of which could have multiple values for a single accession.
Those fields can now be queried using multi-value sublists, but for CSV export,
we still need to be able to return tabular search results with no nested
values. Also, the seed bank web app hasn’t been updated to know about sublists
yet, and we can’t break it as we evolve the search code.

Even after we update the web app, we’ll still need CSV export, and we’ll want
to be able to export tabular search results for queries against the full data
model, not just accessions.

This change introduces the ability to “flatten” a sublist. Flattening a sublist
essentially pulls its fields up into the parent so they’re not nested any more.
The docs in `NestedQueryBuilder` describe flattening in more detail with some
examples. Flattened sublists are specified by using `_` instead of `.` as the
delimiter. (I chose `_` both because it makes the flattened paths resemble
valid identifiers in most major programming languages, and because visually it
looks like the result of flattening something.)

All the accession search fields that used to refer to secondary tables are now
aliases for fields in flattened sublists.

Flattened sublists replace the special-case “join with the accessions table”
logic that used to be needed to support querying accession data that lived in
secondary tables. This simplifies `SearchTable` and also makes it possible to
turn `SearchService.fetchValues` into a small wrapper around the regular
`search` method.

There are a few new test cases for flattened sublists, but many of the existing
tests also exercise the new functionality by virtue of the fact that they refer
to fields that have turned into aliases.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
@@ -76,16 +76,38 @@ class SearchService(private val dslContext: DSLContext) {
       limit: Int = Int.MAX_VALUE,
       distinct: Boolean = false,
   ): SearchResults {
+    // TODO: Better cursor support. Should remember the most recent values of the sort fields
+    //       and pass them to skip(). For now, just treat the cursor as an offset.
+    val offset = cursor?.toIntOrNull() ?: 0
+
+    val results =
+        runQuery(rootPrefix, fields, criteria, sortOrder, limit, offset, distinct).filterNotNull()
+
+    val newCursor =
+        if (results.size > limit) {
+          "${offset + limit}"
+        } else {
+          null
+        }
+
+    return SearchResults(results.take(limit), newCursor)
+  }
+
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #116**  (this PR)
    * **PR #117** 
      * **PR #119** 
        * **PR #120** 
          * **PR #121** 
            * **PR #123** 
              * **PR #124** 
                * **PR #125** 
                  * **PR #126** 
                    * **PR #127** 
                      * **PR #130** 
                        * **PR #128** 
                          * **PR #129** 
                            * **PR #132** 
                              * **PR #133** 

sgrimm:
Sorry this one is so big! Other than a couple trivial bits like switching to the more concise form of `assertThrows` in the test code, I didn't see a way to break this down into smaller chunks. Hopefully the new wall of text in `NestedQueryBuilder` will adequately explain what it's all about.

PR #117, opened 2021-12-02
TRACK-466 Simplify SearchTable declarations

sgrimm:
In preparation for adding a bunch of new search tables, make it less
cumbersome to declare them by providing more default values and adding
a helper base class for tables with per-facility permissions.

Move SearchTables out of the seedbank package since it is going to
have tables other than seed-bank-related ones.

PR #115, opened 2021-12-02
Add named constant for sublist field delimiter

sgrimm:
Using a hardwired constant was okay-ish when there was only one delimiter,
but there is shortly going to be a second one. Give the delimiter a name so
it will be clearer what the meaning is.

PR #108, opened 2021-11-30
TRACK-466 Split AccessionsNamespace

sgrimm:
Introduce a new class `SearchFieldNamespaces` that constructs the namespaces and
exposes them to the application as needed. For now, it only constructs the
existing accessions namespace.

Split `AccessionsNamespace` into a series of plain (non-`@ManagedBean`) classes.

No change in functionality here. A subsequent change will start introducing more
namespaces to the hierarchy.

PR #118, opened 2021-12-03
Don't include empty maps in search results

sgrimm:
This fixes a bug that was masked by the fact that accession searches always
include the accession ID and accession number in the field list. If you
searched for a sublist field, you would get back an entry in the results list
for every root-level entity that matched the filter criteria, even if there
were no values for the sublist you were asking for. That violates the contract
of the search API, which says you will never get back a completely empty map.

Fix the bug and add a regression test to verify the fix.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -579,8 +579,10 @@ class NestedQueryBuilder(
    * The result of that conversion becomes the value of the multiset field in the parent query. See
    * the jOOQ docs for a more detailed explanation of how ad-hoc converters and nested queries
    * interact with each other.
+   *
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #118**  (this PR)
    * **PR #108** 
      * **PR #113** 
        * **PR #114** 
          * **PR #115** 
            * **PR #116** 
              * **PR #117** 
                * **PR #119** 
                  * **PR #120** 
                    * **PR #121** 

sgrimm:
This is a standalone bugfix, not part of the ongoing stack.

sgrimm:
Turns out this fix doesn't play well with some of the changes I made in the big stack. I'm going to change the big stack so it is based on this change, which will cause some of the PRs in the stack to start failing CI checks; then I'll go fix the failures.

All of which is to say: review this PR first.

PR #120, opened 2021-12-03
TRACK-466 Handle circular references among namespaces

sgrimm:
Previously, namespaces were constructed using an implicit tree structure, where
each one was created as part of its the initialization of its parent in the
data hierarchy.

That works when the namespaces genuinely form a tree structure, but it won't
work once they form a more complicated graph with cycles and cross-linking. For
example, `species` will appear as a sublist of `plants` as well as
`accessions`.

Fix it by switching to a two-phase initialization scheme where we create all
the namespaces and then link them together (by initializing their `sublists`
fields) only after they all exist.


sgrimm:
* main
  * **PR #120**  (this PR)
    * **PR #121** 
      * **PR #123** 
        * **PR #124** 
          * **PR #125** 
            * **PR #126** 
              * **PR #127** 
                * **PR #130** 
                  * **PR #128** 
                    * **PR #129** 
                      * **PR #132** 
                        * **PR #133** 

PR #121, opened 2021-12-03
TRACK-466 Add organizations, projects, sites namespaces

sgrimm:
Allow searches to ask for information about the upper layers of the data model.

The list of fields in these namespaces is minimal for now, because we're still
finalizing requirements like how a project's location should be modeled. This is
mostly just to get the top-level structure in place so we can start expanding
the search API into other parts of the data model.

PR #123, opened 2021-12-03
TRACK-513 Add aliases for fields that ought to be sublists

sgrimm:
There are a few accession fields that refer to values in secondary tables and
whose field names are the names that we'd ideally want to use for the
corresponding sublists. To avoid name collisions, the sublists have `Info`
suffixes for now.

Add alternate aliases the client can use temporarily. The migration sequence
will look like this:

1. Add alternate names (this change)
2. Update the client to use the alternate names
3. Remove the offending fields and rename the sublists
4. Update the client to use the flattened sublist names instead of the aliases
5. Remove the aliases

For example, `species` is the existing field for the species name, which is
pulled from a separate table. Ideally, we'd want the separate table to
correspond to a sublist called `species`, at which point you could ask for
`species_name` to get the name. But thanks to the name collision, you
have to do `speciesInfo_name`. This diff adds an alias `speciesName`.
Once the client is using `speciesName` instead of `species`, we can
change the sublist's name to `species` which is what we want long term.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
@@ -2029,6 +2029,7 @@ class SearchServiceTest : DatabaseTest(), RunsAsUser {
                   "active" to "Active",
                   "bags" to listOf(mapOf("number" to "2"), mapOf("number" to "6")),
                   "species" to "Other Dogwood",
+                  "speciesName" to "Other Dogwood",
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #121** 
    * **PR #123**  (this PR)
      * **PR #124** 
        * **PR #125** 
          * **PR #126** 
            * **PR #127** 
              * **PR #130** 
                * **PR #128** 
                  * **PR #129** 
                    * **PR #132** 
                      * **PR #133** 

PR #124, opened 2021-12-04
TRACK-463 Remove search field list from OpenAPI schema

sgrimm:
Having the list of search fields in the OpenAPI schema made sense when searches
could only be done on accessions and the list of fields was strictly fixed.

When we added sublists, we extended the schema to include them, and just ignored
the fact that you can use the "navigate up to the parent" functionality to
construct arbitrarily long field names. The list of fields in the schema was a
subset of the set of valid field names, but it was still a useful subset.

Once we add the ability to specify the root prefix for a search, though, it will
become impossible to construct a static list of valid field names because the
field names are all relative to the prefix and the prefix can point to an
arbitrary place in our data model. At that point, the schema will be flat-out
wrong, not just incomplete.

Nothing in the existing front-end code is actually using the schema's field
list anyway, so rather than trying to figure out some clever way to represent
it as a mutually-recursive series of enums or somesuch, just remove it. In the
future, we can add an endpoint to dynamically return the list of fields under a
particular prefix, and also add some automation to generate a human-readable
field list as documentation.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3299,7 +3299,7 @@ components:
         fields:
           type: array
           items:
-            $ref: '#/components/schemas/SearchField'
+            type: string
         sortOrder:
           type: array
           items:
@@ -3422,7 +3422,7 @@ components:
       - type: object
         properties:
           field:
-            $ref: '#/components/schemas/SearchField'
+            type: string
           values:
             minItems: 1
             type: array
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #124**  (this PR)
    * **PR #125** 
      * **PR #126** 
        * **PR #127** 
          * **PR #130** 
            * **PR #128** 
              * **PR #129** 
                * **PR #132** 
                  * **PR #133** 

PR #125, opened 2021-12-06
TRACK-463 Separate accession search payloads from generic ones

sgrimm:
In preparation for adding generic search endpoints, shuffle around some payload
classes that will be the same for both accession search and generic search, and
rename a couple of endpoints in the OpenAPI schema so they are clearly marked as
being accession-specific.

No functional change here, but if clients are using the payload or operation
names, they'll need to be updated to use the new names.

Specific changes, since git won't be able to easily figure out the renames here:

* Operations renamed: `search`->`searchAccessions`, `export`->`exportAccessions`
* Payload renamed: `SearchRequestPayload`->`SearchAccessionsRequestPayload`
* Controller class renamed: `SearchController`->`AccessionSearchController`
* All other search payload classes moved to `com.terraformation.backend.search.api`


sgrimm:
* main
  * **PR #125**  (this PR)
    * **PR #126** 
      * **PR #127** 
        * **PR #130** 
          * **PR #128** 
            * **PR #129** 
              * **PR #132** 
                * **PR #133** 

PR #126, opened 2021-12-06
TRACK-463 Resolve field names in controllers

sgrimm:
Previously, we had a custom deserializer that knew how to turn strings into
`SearchFieldPath`s for accession search. That could be done generically because
paths always had the same prefix. But once clients can specify the prefix, there's
no way to parse strings into `SearchFieldPath`s without knowing what the prefix
is, which the JSON deserializer can't know.

Remove the deserializer and make the controller classes parse the paths.

No client-visible behavior change here, though the human-readable error message
that's included with the HTTP 400 response is different if you pass in an invalid
field name.


sgrimm:
* main
  * **PR #126**  (this PR)
    * **PR #127** 
      * **PR #130** 
        * **PR #128** 
          * **PR #129** 
            * **PR #132** 
              * **PR #133** 

PR #128, opened 2021-12-07
Allow accession state and active status in same sublist

sgrimm:
Work around https://github.com/jOOQ/jOOQ/issues/12704 which causes multiset
queries to return the wrong result if the same column is included twice in the
SELECT clause. We can generate a query like that if the user searches for both
the `state` and `active` fields in the accessions namespace and puts it in a
nested sublist; `active` is a computed value based on `state` so when it gets
queried, we actually ask the database for `state`.

Fix it, as jOOQ's maintainer suggests, by using a column alias.


sgrimm:
* main
  * **PR #128**  (this PR)
    * **PR #129** 
      * **PR #132** 
        * **PR #133** 

sgrimm:
The underlying bug is already fixed, so we can revert this once the next jOOQ release comes out and we upgrade to it. But for now, this works.

PR #129, opened 2021-12-07
Fix sorting on field in grandchild sublist

sgrimm:
If you sorted by a field two or more levels deep in the nested sublist tree, and
you weren't querying any values from the sublist the field was in, the query
builder would throw an exception.

Fix the bug and add a test case to confirm the fix.


sgrimm:
* main
  * **PR #128** 
    * **PR #129**  (this PR)
      * **PR #132** 
        * **PR #133** 

PR #130, opened 2021-12-07
Fix malformed SQL when mixing aliases and nested sublists

sgrimm:
If you referenced an alias in a nested sublist, and the alias pointed to
a flattened sublist, and the query didn't include any fields from the
nested sublist itself, we would generate SQL that was missing the JOIN
clause to connect the flattened sublist to the parent query.

Fix the bug and add test cases that failed without the fix.


sgrimm:
* main
  * **PR #130**  (this PR)
    * **PR #128** 
      * **PR #129** 
        * **PR #132** 
          * **PR #133** 

PR #131, opened 2021-12-07
TRACK-483 Improve GeoJSON parsing and rendering

sgrimm:
Add more input validation to the GeoJSON parser so we can return helpful error
messages to the client, e.g., when they send us a polygon that doesn't have
enough points. Previously, we were accepting values that were syntactically
well-formed, even if they didn't make sense; PostGIS would catch the problems
when we tried to store the values in the database, but we wouldn't always get
back an error message that was suitable to return to the client.

When returning GeoJSON values to the client, round coordinates to 8 digits
after the decimal point so our rendered values don't have bogus-looking floating
point inaccuracies due to the fact that we do coordinate system conversion in
the database. This was especially glaring if you tried to create a feature
with longitude/latitude coordinates; when you read it back from the server, you
would almost always get ugly-looking values like 121.0000000000001.


sgrimm:
* main
  * **PR #131**  (this PR)

PR #132, opened 2021-12-07
Test searching all fields in hierarchy

sgrimm:
Update the "search all the fields" test so it starts at the organizations
namespace rather than the accessions one. As we add more sublists to the field
hierarchy, this will make sure they're covered by the test.

Unfortunately, this change exposed a bug in PostgreSQL: sorting on large numbers
of JSON expressions causes the query planner to slow down dramatically. We still
want to make sure `NestedQueryBuilder` can use any search field for sorting, so
move that to a separate test that tries them one at a time and makes sure the
query generation doesn't bomb out.


sgrimm:
* main
  * **PR #132**  (this PR)
    * **PR #133** 

PR #133, opened 2021-12-07
TRACK-467 Add layers, features, plants to search

sgrimm:
Add the GIS-related parts of our data model to the hierarchy of search fields.

Geometry values are returned as strings whose contents are GeoJSON objects.
For now, they can only be returned as search results and used as sort keys
(though sorting by a geometric shape isn't too useful). They can't be used as
filter criteria because none of the existing filter operations really apply to
them -- exact-match might be reasonable, but coordinate system conversion makes
its semantics a bit muddled.

Plant fields are currently represented as a sublist of the features namespace.
This is somewhat at odds with our desire to move toward an API where a plant is
just a type of feature, but flattened sublists can be used to pull the plant
fields up a level so they look like they're feature fields. This will probably
need some refinement once we start using it for real.


sgrimm:
* main
  * **PR #133**  (this PR)

src/main/kotlin/com/terraformation/backend/search/namespace/LayersNamespace.kt
```diff
@@ -0,0 +1,32 @@
...
+          features.asMultiValueSublist("features", LAYERS.ID.eq(FEATURES.LAYER_ID)),
+          sites.asSingleValueSublist("site", LAYERS.SITE_ID.eq(SITES.ID)),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      with(namespaces.searchTables.layers) {
+        listOf(
+            timestampField("createdTime", "Layer created time", LAYERS.CREATED_TIME),
```
kallen07, line 24: Does this mean that `createdTime` can be returned with Layers if the client requests that field? Right now, we aren't returning created time in `LayerController.LayersResponse`. Not suggesting that we need to remove it, but just checking that I understand the change correctly.

Same question for `createdTime` in the `FeaturesNamespace` field list.

sgrimm: Yes, that's correct. I noticed that a bunch of the UI mockups (for account management, to be clear) were showing creation times, so I figured there was a decent chance we'd want them here too. Should probably add them to the regular API payloads as well.

PR #134, opened 2021-12-10
TRACK-421 Move family name to accessions table

sgrimm:
Currently, there are two related but distinct tables `species` and `families` in
the data model. In theory, a species should be a member of a family. In practice,
family is only used in the seed bank app, and it is treated as a free-form text
field that's unrelated to the species name.

We're adding per-organization species management to our system, but there is no
plan to also add per-organization family management, and indeed, we'll likely
do something completely different when we start improving how we model taxonomic
details of species.

Rather than trying to shoehorn families into that transition, update the data
model so it reflects the way families actually work: they are free-form text
fields on the accessions table.

This does introduce a client-visible behavior change: `/api/v1/seedbank/summary`
reports the number of families as of a particular date, and it now does so by
looking at accession creation timestamps since there's no longer an explicit
family creation timestamp. The difference is mostly that the summary only
counts family names that are currently used by accessions, rather than names
that were used at any time in the past.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
@@ -51,7 +51,8 @@ class SummaryController(
                 speciesStore.countSpecies(now), speciesStore.countSpecies(startOfWeek)),
         families =
             SummaryStatistic(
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #134**  (this PR)
    * **PR #135** 

PR #135, opened 2021-12-10
TRACK-421 Remove unused species endpoints

sgrimm:
The front end code has been using `/api/v1/species` for all its species
management needs since the seed bank and tracking UIs were merged together. We
no longer need the older seed-bank-specific endpoints.

Getting rid of these will mean we can rename some payload classes that have
awkward names for the sake of avoiding name collisions with the endpoints that are
going away here, but that'll be a client-breaking change so it isn't included.


sgrimm:
* main
  * **PR #135**  (this PR)

PR #136, opened 2021-12-10
Remove backward-compatible organizations endpoints

sgrimm:
The front end code is using `/api/v1/organizations`, so there is no longer any
need for the older name `/api/v1/organization`.


sgrimm:
* main
  * **PR #136**  (this PR)
    * **PR #138** 

PR #137, opened 2021-12-13
Fix compiler warning; treat warnings as errors

sgrimm:
Fix a shadowed variable name that's been in the code for a month or so. To stop
that kind of thing from creeping into the code again, turn on the compiler
option that treats warnings as errors. We can revisit it if it turns out to cause
too many false positives.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -212,6 +212,7 @@ sourceSets.main {
 
 tasks.withType<KotlinCompile> {
   dependsOn(generateVersionFile)
+  kotlinOptions.allWarningsAsErrors = true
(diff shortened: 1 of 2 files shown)


sgrimm:
@kallen07 Curious for your thoughts on whether warnings-as-errors is a good idea. What makes it less than crystal-clear is that while there are warnings like the shadowed variable name that are really useful for pointing out potential bugs, there are also warnings like "variable xyz is unused" that are less likely to be actual problems.

In general I think I'd rather be forced to clean up harmless sloppy stuff than to miss out on potentially-harmful stuff.

PR #139, opened 2021-12-13
Add countries and subdivisions to search

sgrimm:
Allow clients to include organizations' country and US state in their search
requests when searching organization details. Since the client is likely to use
the ISO-3166 codes internally, I added the codes as search fields in the
organizations namespace as well as adding them as sublists that can be used to
retrieve the country/state names.

This also means the search API serves double duty as the API you can use to
dump out the list of known countries and subdivisions; pass in a prefix of
`country` and a field list of `code` and `name` with no search criteria to get
the full list of countries, and if you want a structured list of each country's
subdivisions, search for `subdivisions.code` and `subdivisions.name`.


sgrimm:
* main
  * **PR #139**  (this PR)
    * **PR #140** 

PR #141, opened 2021-12-15
Upgrade to Spring Boot 2.4.4

sgrimm:
While developing the search code, it was annoying to keep running into
https://github.com/spring-projects/spring-boot/issues/25717 which caused database
error messages to not show up in the logs. The bug was fixed in the patch version
of Spring Boot just after the one we were using.


sgrimm:
* main
  * **PR #141**  (this PR)

PR #142, opened 2021-12-16
TRACK-520 Support preregistered users

sgrimm:
Allow users to be created without Keycloak user identifiers. When they log in for
the first time, associate their Keycloak identities (including their first and
last names, which they will have supplied as part of registration) with the existing
user ID by matching the email address.

There is currently no externally-accessible way to create users in that state, but
future changes will use it to represent pending invitations.


sgrimm:
* main
  * **PR #142**  (this PR)

PR #143, opened 2021-12-16
Upgrade JDK and dependencies

sgrimm:
Upgrade to the latest versions of all of terraware-server's dependencies, including moving to Java 17 for both the build process and the execution environment.

A few of the dependency upgrades required minor code changes, but nothing that affects functionality.

Each dependency upgrade is in a separate commit in this PR so that if one of them introduces a problem, we can `git bisect` to figure out which one.

**_Please merge this using "Rebase and merge" rather than squash_**, or we'll lose the ability to bisect to a specific dependency upgrade.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,20 +16,16 @@ jobs:
   build:
     runs-on: ubuntu-20.04
 
-    strategy:
-      matrix:
-        java-version: ["15"]
-
     steps:
     - uses: actions/checkout@v2
 
     - name: Set environment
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #143**  (this PR)

openapi.yaml
```diff
@@ -1,18 +1,19 @@
 openapi: 3.0.1
 info:
-  title: Terraware Seed Bank
-  description: Local server API for seed banks
+  title: Terraware Server
```
kallen07, line 3: 😁

PR #145, opened 2021-12-17
TRACK-497 Add endpoints to query organization users

sgrimm:
Add `/api/v1/organizations/{orgId}/users` to fetch a list of all the users in an
organization, and `/api/v1/organizations/{orgId}/users/{userId}` to fetch
information about a single user in an organization.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/OrganizationUserModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/OrganizationUserModel.kt
@@ -8,7 +8,6 @@ import java.time.Instant
 
 data class OrganizationUserModel(
     val userId: UserId,
-    val authId: String?,
     val email: String,
     val firstName: String?,
     val lastName: String?,
--- a/src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
@@ -314,7 +314,6 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
         listOf(
             OrganizationUserModel(
                 UserId(100),
-                "auth1",
                 "user1@x.com",
                 "First1",
                 "Last1",
@@ -326,7 +325,6 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
                 emptyList()),
             OrganizationUserModel(
                 UserId(101),
-                "auth2",
                 "user2@x.com",
                 "First2",
                 "Last2",
@@ -351,7 +349,6 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
     val model =
         OrganizationUserModel(
             UserId(100),
-            "auth",
             "x@y.com",
             "First",
             "Last",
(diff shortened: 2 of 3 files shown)


sgrimm:
* main
  * **PR #145**  (this PR)
    * **PR #146** 

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -154,8 +186,49 @@ data class OrganizationPayload(
   )
 }
 
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
kallen07, line 189: Does this mean that the `firstName` and `lastName` fields will not be included in the JSON response if their value is null?

src/main/kotlin/com/terraformation/backend/customer/model/OrganizationUserModel.kt
```diff
@@ -8,13 +8,14 @@ import java.time.Instant
 
 data class OrganizationUserModel(
     val userId: UserId,
-    val authId: String,
+    val authId: String?,
```
kallen07: (maybe this is answered in a future diff) Why are we adding the authId here?

src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
```diff
@@ -302,4 +306,89 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
...
+  fun `fetchUsers throws exception if member has no permission to list users`() {
+    every { user.canListOrganizationUsers(organizationId) } returns false
+
+    assertThrows<AccessDeniedException> { store.fetchUsers(organizationId) }
+  }
+
+  @Test
+  fun `fetchUsers throws exception if member has no permission to see organization`() {
+    every { user.canReadOrganization(organizationId) } returns false
+    every { user.canListOrganizationUsers(organizationId) } returns false
```
kallen07, line 379: should this test pass without `every { user.canListOrganizationUsers(organizationId) } returns false`? Not sure if that's even a case you'd want to test though

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -78,6 +84,32 @@ class OrganizationsController(private val organizationStore: OrganizationStore)
...
+  @GetMapping("/{organizationId}/users/{userId}")
+  @Operation(summary = "Gets information about a user's membership in an organization.")
+  fun getOrganizationUser(
+      @PathVariable("organizationId") organizationId: OrganizationId,
+      @PathVariable("userId") userId: UserId,
+  ): GetOrganizationUserResponsePayload {
+    val model =
+        try {
+          organizationStore.fetchUser(organizationId, userId)
+        } catch (e: UserNotFoundException) {
```
kallen07, line 105: This won't include the access denied exception right? As in, `AccessDeniedException` will bubble all the way up as a separate exception?

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
sgrimm: That's right.

src/main/kotlin/com/terraformation/backend/customer/model/OrganizationUserModel.kt
sgrimm: It was already there; this is just making it nullable. But good catch: nothing actually uses it outside of a unit test. I'll get rid of it.

src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
sgrimm: Agreed that it's kind of a nonsensical configuration. But the test should (and does) fail in that case because no exception is thrown. Which I _think_ is the right behavior for that nonsense configuration: you are trying to list users and you have permission to list users. Maybe analogous to having execute permission but not read permission on a UNIX executable?

I'll add a test case for it to make the behavior explicit, even though it's probably not a configuration we'd allow a user to ever specify.

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
sgrimm: Yeah, that's right. And it'll be turned into HTTP 403 by `ControllerExceptionHandler`.

PR #148, opened 2021-12-27
TRACK-515 Use correct names for accession sublists

sgrimm:
The front end code has been updated to use alternate fields for species, storage
location, and primary collector names. Rename the sublists for those values to
remove the temporary-hack `Info` suffix, and remove the alias fields that used
to have name collisions.

Specifically:

* Remove the `primaryCollector` field and rename the `primaryCollectorInfo`
  sublist to `primaryCollector`
* Remove the `species` field and rename the `speciesInfo` sublist to `species`
* Remove the `storageLocation` field and rename the `storageLocationInfo`
  sublist to `storageLocation`

The next step in this sequence of changes is to modify the front end to directly
use flattened sublists instead of relying on alias fields, e.g., to start using
`species_name` instead of `speciesName`. Once that's done, we can remove those
alias fields entirely from the back end.

PR #149, opened 2021-12-28
TRACK-551 Return CSV error responses

sgrimm:
If the client sends a request with an `Accept: text/csv` header and there's an
error, we need to be able to send a CSV response since that's the only kind of
response that it's valid to return.

Update our exception-handling code to generate CSV responses if needed, but to
prefer JSON if possible.


sgrimm:
* main
  * **PR #149**  (this PR)
    * **PR #150** 
      * **PR #151** 
      * **PR #148** 
        * **PR #152** 

PR #150, opened 2021-12-28
Remove old-style filters field from new search endpoint

sgrimm:
The accession search API payload has a `filters` field that takes a list of search
filters. This has been superseded by the more powerful `search` field that takes
an arbitrary Boolean expression. The old field is still supported for backward
compatibility, but we only need to care about that on the accession search
endpoints; there is no old code that calls the non-accession-specific search
endpoint.


sgrimm:
* main
  * **PR #150**  (this PR)
    * **PR #151** 
    * **PR #148** 
      * **PR #152** 

PR #151, opened 2021-12-28
TRACK-551 Support CSV export of search results

sgrimm:
Make `/api/v1/search` return a CSV file if the incoming request specifies
`text/csv` in its `Accept` header.

Allow the client to pass a maximum result count of 0 to indicate no limit, since
pagination isn't as useful for CSV downloads as it is for interactive search in
the web UI.

The existing `/api/v1/seedbank/search/export` endpoint continues to work as
before.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
@@ -1,13 +1,21 @@
 package com.terraformation.backend.search.api
 
 import com.terraformation.backend.api.SearchEndpoint
+import com.terraformation.backend.api.csvResponse
+import com.terraformation.backend.api.writeNext
 import com.terraformation.backend.search.SearchFieldPrefix
 import com.terraformation.backend.search.SearchService
 import com.terraformation.backend.search.namespace.SearchFieldNamespaces
 import io.swagger.v3.oas.annotations.media.Content
 import io.swagger.v3.oas.annotations.media.ExampleObject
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #151**  (this PR)

src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
```diff
@@ -69,7 +72,54 @@ class SearchController(
...
+            payload.cursor,
+            count)
+
+    val dateAndTime =
+        DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss").format(LocalDateTime.now(clock))
+    val filename = "terraware-$dateAndTime.csv"
+
+    val columnNames = payload.getSearchFieldPaths(rootPrefix).map { it.searchField.displayName }
+
+    return csvResponse(filename, columnNames) { csvWriter ->
```
kallen07, line 107: Is this syntactic sugar? I'm surprised that the callback function is being specified after the closing paren.

sgrimm: If a function has a single argument with a function type like `() -> Unit`, and you're using a lambda function at the call site, the function can come outside the parentheses, and you can omit the parentheses if there are no other arguments.

That's actually how a bunch of built-in Kotlin things work. Stuff like `let` and `with` and `use` are all just functions that take functions as arguments. It's also how various Kotlin DSLs work, e.g., all the blocks like `dependencies` in `build.gradle.kts` are doing the same thing.

PR #153, opened 2021-12-28
Reject invalid enum values in request payloads

sgrimm:
If a request payload includes a field whose type is one of the enum classes we
generate from reference tables, we should only accept strings that match the
display names of the enum's values. But we had a bug: we were treating unknown
values as null.


sgrimm:
* main
  * **PR #153**  (this PR)
    * **PR #154** 
      * **PR #155** 

PR #154, opened 2021-12-28
TRACK-561,TRACK-544 Add more fields to projects

sgrimm:
Add new per-project fields to support the project management UI:

* Description
* Start date
* Status
* Types

That last one requires a data model change; previously a project had a single
type rather than a list of them.


sgrimm:
* main
  * **PR #154**  (this PR)
    * **PR #155** 

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
```diff
@@ -57,7 +67,13 @@ class ProjectsController(private val projectStore: ProjectStore) {
       @PathVariable("id") projectId: ProjectId,
```
kallen07, line 72: Could we put a note in here that this endpoint overwrites all the fields (and will "erase" the value of a field if it's left undefined)?

PR #155, opened 2021-12-28
TRACK-560 Add description to site API payload

sgrimm:
Read the description from the database and return it with the other site data.
This is only the read path; there isn't currently an update path, so it'll be
added as a separate change.


sgrimm:
* main
  * **PR #155**  (this PR)

PR #156, opened 2022-01-05
TRACK-420 Summary API endpoint for plants+species

sgrimm:
Add `/api/v1/gis/plants/summary` to fetch current and historical counts of species
and plants by organization, project, or site.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/api/PlantController.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/api/PlantController.kt
@@ -153,8 +153,8 @@ class PlantController(private val featureStore: FeatureStore) {
       @RequestParam("previousTime")
       @Schema(
           description =
-              "If present, the response will include data for plants entered at or before this " +
-                  "time.",
+              "If present, the response will also include data for plants entered at or before " +
+                  "this time, in addition to the most recent data.",
           example = "2021-01-05T11:22:33Z")
       previousTime: Instant? = null,
   ): PlantsAndSpeciesSummaryResponsePayload {
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1036,8 +1036,8 @@ paths:
         required: false
         schema:
           type: string
-          description: "If present, the response will include data for plants entered\
-            \ at or before this time."
+          description: "If present, the response will also include data for plants\
+            \ entered at or before this time, in addition to the most recent data."
           format: date-time
           example: 2021-01-05T11:22:33Z
       responses:


sgrimm:
* main
  * **PR #156**  (this PR)

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -383,13 +390,87 @@ class FeatureStore(
...
 
+  fun fetchPlantSummary(
+      layerId: LayerId,
+      minEnteredTime: Instant? = null,
+      maxEnteredTime: Instant? = null,
+  ): Map<SpeciesId, Int> {
+    return fetchPlantSummary(listOf(layerId), minEnteredTime, maxEnteredTime)
+  }
+
+  fun fetchPlantSummary(
```
sgrimm, line 408: I tried collapsing these into a single method that took all the IDs as optional parameters and constructed the query based on which of them were non-null, but it ended up being pretty messy.

src/main/kotlin/com/terraformation/backend/gis/api/PlantController.kt
```diff
@@ -112,6 +122,82 @@ class PlantController(private val featureStore: FeatureStore) {
...
+      @Schema(
+          description =
+              "ID of site to summarize. If not present, either organizationId or projectId " +
+                  "must be present.")
+      siteId: SiteId? = null,
+      @RequestParam("previousTime")
+      @Schema(
+          description =
+              "If present, the response will include data for plants entered at or before this " +
+                  "time.",
```
kallen07: Maybe clarify that the current time stats will still be included. E.g., I took a stab at a modification...

"In addition to the current data, if this parameter is present then the response will also include "previous" data. Aka, data for plants entered at or before this time."

But on second thought, this might not be needed because of the operation description. Whatever you think is best.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
```diff
@@ -383,13 +390,87 @@ class FeatureStore(
...
         .groupBy(PLANTS.SPECIES_ID)
         .fetchMap({ it.value1() }, { it.value2() })
   }
 
+  fun fetchPlantSummary(
+      layerId: LayerId,
+      minEnteredTime: Instant? = null,
+      maxEnteredTime: Instant? = null,
+  ): Map<SpeciesId, Int> {
+    return fetchPlantSummary(listOf(layerId), minEnteredTime, maxEnteredTime)
```
kallen07, line 405: Do you need this function because sometime the SELECT statements below return only one layer id and you get a type error?

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
kallen07: I like how it is. Very clear and easy to follow.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
```diff
@@ -729,7 +737,7 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
           featureIds.add(randomFeatureId)
 
           // All features are associated with the class variable layerId
```
kallen07: Nit: remove this comment since it's no longer correct.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
```diff
@@ -1030,6 +1038,68 @@ internal class FeatureStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      every { user.canReadLayer(inaccessibleLayerId) } returns false
+
+      val expected = mapOf(SpeciesId(1) to 6, SpeciesId(2) to 4, SpeciesId(3) to 2)
+      val actual = store.fetchPlantSummary(siteId)
+
+      assertEquals(expected, actual)
+    }
+
+    @Test
```
kallen07, line 1062: Is it possible to check that the permissions code was called since we're not testing for inaccessible projects? That could be an cheaper way to do the permissions check for the overloaded `project id ` and `organization id` `fetchPlantSummary` functions.

src/main/kotlin/com/terraformation/backend/gis/db/FeatureStore.kt
sgrimm: No, this is called by `PlantController` so I left it in. I could instead have made the private version of this public and changed `PlantController` to pass `listOf(layerId)`, but exposing a function that takes a single ID seemed nice for consistency with the new functions I'm adding here.

sgrimm: It looks like a completely new function in the PR, but this is actually (from callers' point of view) a refactoring of the function above this, which used to take a single layer ID and now takes a list. The list version of this is actually the new one.

src/test/kotlin/com/terraformation/backend/gis/db/FeatureStoreTest.kt
sgrimm: Not 100% sure I understand the suggestion, so apologies if this makes no sense, but the reason I did the per-layer permission check even if you're fetching by project ID and have project permission is just to avoid baking in assumptions about the permissions structure. If we add some kind of "only a subset of users are allowed to see the irrigation map" thing down the road, we'll already have the right permission checks and we'll just have to change `UserModel`.

That said, I admit I've been quite inconsistent about following that rule.

On the expense side, though, I'm purposely ignoring the fact that the permission checks require separate database queries. Until we're operating at higher scale, it's not going to be noticeable. And if/when it becomes a performance bottleneck, we can introduce an efficient caching layer for permissions rather than trying to optimize away the permission checks.

Again, though, sorry if I'm totally misunderstanding what you're suggesting here.

kallen07: The thought behind my comment is that you did permission checking in `fetchPlantSummary by site ID aggregates results across accessible layers`, but it seems like you left it out of the later functions since the fetch by site/project/org id functions all call the same function under the hood. 

So i was wondering if in the other test functions, you want to test the permissions to avoid depending on this specific implementation where every function ultimately depends on the one function that does this
```
val readableLayerIds = layerIds.filter { currentUser().canReadLayer(it) }
    if (readableLayerIds.isEmpty()) {
      return emptyMap()
    }
 ```
 
I was suggesting possibly adding some kind of cheaper (less lines of code) permissions test such as checking that the correct overloaded version of fetchPlantSummary (that does the permission check) is called. But maybe that's more confusing and worse practice than not having the permissions tested in each function.

Hopefully that makes a bit more sense than my first comment.

sgrimm: Ah, okay, gotcha. Good point; I should do the same "leaves out the inaccessible data" check in each of those tests so as not to depend on the implementation details. Thanks for pointing it out.

sgrimm: (Which I realize is literally the opposite of what you're actually suggesting, but the key problem here is I'm not testing the functionality I should be, I think.)

kallen07: Yes, I guess I could have made that a whole lot simpler by asking if you intended to not test the permissions in the later functions 😂

PR #157, opened 2022-01-05
Remove ClientFacingException

sgrimm:
Our controller exception handling code has support for turning our internal
`EntityNotFoundException` classes into HTTP 404 responses, and for translating
standard JAX-RS exceptions into Spring exceptions.

We thus don't need `ClientFacingException` or most of its subclasses.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/Exceptions.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/Exceptions.kt
@@ -1,37 +1,15 @@
 package com.terraformation.backend.api
 
-import org.springframework.http.HttpStatus
+import javax.ws.rs.ClientErrorException
+import javax.ws.rs.core.Response
 
-// Controller methods can throw these exceptions to return HTTP error statuses with a consistent
-// set of error messages.
-
-abstract class ClientFacingException(val status: HttpStatus, message: String) :
-    RuntimeException(message) {
-  override val message
-    get() = super.message!!
-}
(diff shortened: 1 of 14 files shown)


sgrimm:
* main
  * **PR #157**  (this PR)
    * **PR #160** 
      * **PR #158** 
        * **PR #159** 
          * **PR #161** 
            * **PR #162** 

src/main/kotlin/com/terraformation/backend/api/Exceptions.kt
```diff
@@ -1,37 +1,15 @@
...
 class ResourceInUseException(message: String = "The resource is currently in use") :
-    ClientFacingException(HttpStatus.CONFLICT, message)
-
-class UnsupportedPhotoFormatException(message: String = "Photos must be of type image/jpeg") :
-    ClientFacingException(HttpStatus.UNSUPPORTED_MEDIA_TYPE, message)
-
-class WrongOrganizationException(
-    message: String = "Resource is owned by a different organization"
-) : ClientFacingException(HttpStatus.FORBIDDEN, message)
+    ClientErrorException(message, Response.Status.CONFLICT)
```
kallen07, line 15: Since this is the same as the `DuplicateNameException`, I'm guessing these map internal exceptions to standard HTTP error codes. So two internal exceptions can have the same HTTP error code.

sgrimm: Yep, that's exactly right.

PR #161, opened 2022-01-06
TRACK-523 Add endpoints to add/remove project users

sgrimm:
Also add a test for `ProjectStore`, which didn't have one before. The test
covers the methods that are used by the client-visible change here.


sgrimm:
* main
  * **PR #161**  (this PR)
    * **PR #162** 

PR #164, opened 2022-01-08
TRACK-585 Add organization ID to species endpoints

sgrimm:
Add an `organizationId` parameter to all the species-related endpoints that will
need to include it once we change the species data model to add per-organization
species lists. Depending on the endpoint, it is either a query string parameter
or a payload field.

The parameter currently does absolutely nothing. But with this change, it is now
included in the OpenAPI schema and will thus show up in autogenerated client code.

Once this is merged, the client can be updated to start including the organization
ID in all its species-related requests. After that, the parameter will become
required rather than optional.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2086,6 +2086,14 @@ paths:
       - GISApp
       summary: Lists all known species.
       operationId: listSpecies
+      parameters:
+      - name: organizationId
+        in: query
+        required: false
+        schema:
+          type: integer
+          description: Organization whose species should be listed. (Currently ignored.)
+          format: int64
       responses:
         "200":
           description: OK
@@ -2123,6 +2131,15 @@ paths:
       - GISApp
       summary: Lists all species names.
       operationId: listAllSpeciesNames
+      parameters:
+      - name: organizationId
+        in: query
+        required: false
+        schema:
+          type: integer
+          description: Organization whose species names should be listed. (Currently
+            ignored.)
+          format: int64
       responses:
         "200":
           description: OK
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #164**  (this PR)

sgrimm:
(Posted this in Slack too, but putting it here for posterity.)

I did toy with putting it in the path, but ultimately I decided on a query string parameter for a few reasons.

1. Backward compatibility with existing clients. The first part of the sequence of changes is to add org ID as an optional parameter so it shows up in the OpenAPI schema and the frontend code generation will pick it up, but the existing frontend code will keep working. That’s hard to do with a URL path element because the structure of the path wouldn’t be the same with and without it.
2. Tying species to things other than organizations could be a thing. The version of the species data model in gis-server, following the lead of the forestry team’s data model, also had site and project IDs in the species_options table. I took those out as part of the per-org species changes but I think there’s a nonzero chance we’ll want them later on. At that point, having the org ID in the query string makes it seamless to change the API so you can pass a project ID instead if you want.
3. Thinking ahead to shared species metadata. Right now the implementation doesn’t let two organizations share an underlying species, but the data model does. Down the road, it may make sense to say, “Give me the generic, non-org-specific info for species X.” And the organization ID is just a different view of that underlying data. This point is pretty speculative since we don’t know what the desired behavior is going to be, admittedly.
4. The URL structures I was thinking up felt pretty awkward to me. If you want the species to be under the organization, you end up with URLs like `/api/v1/organizations/123/species/456/names` which isn’t impossibly awful, but feels clunky to me. It also means the previous two points get a little more awkward to deal with (are the per-site and non-org-specific views under a completely different path?)

All those points except the first one are speculative or subjective, but putting them together, the query string felt like the better approach to me.

PR #165, opened 2022-01-10
TRACK-508 Fold SearchFieldNamespace into SearchTable

sgrimm:
The search code had two metadata classes for each searchable database table.
This was because the way the application's data model was represented in the
search API didn't quite correspond 1:1 to the way the underlying database
tables were organized: some fields in child tables were represented as values
in the parent tables in order to present a flat view of the
data to the front end.

We now use flattened sublists and alias fields to bridge that gap, so there's no
longer any need to distinguish between a "search table" and a "search field
namespace."

Remove `SearchFieldNamespace` and fold its data into `SearchTable`.


sgrimm:
* main
  * **PR #165**  (this PR)
    * **PR #166** 

sgrimm:
This one is going to be annoying to review, I'm afraid. Couldn't really avoid touching a bunch of files all at once given the nature of the change. A lot of the changes are pretty repetitive "rename 'namespace' to 'table'" updates.

Hopefully the end result is that the search code is easier to follow, though. It was always kind of awkward to bounce back and forth between namespaces and tables.

src/main/kotlin/com/terraformation/backend/search/table/FacilitiesTable.kt
```diff
@@ -1,19 +1,27 @@
-package com.terraformation.backend.search.namespace
+package com.terraformation.backend.search.table
```
kallen07, line 1: I'm curious about why github is tracking some of these renames correctly, but not others.

sgrimm: Git's rename detection is based on file similarity. I don't remember the exact algorithm but it's along the lines of, "If more than X% of a deleted file's content is the same as a newly-created file's content, treat it as a rename+edit rather than a delete+create." Git's underlying data model doesn't have any direct representation of renames.

Unfortunately, the similarity algorithm is just a dumb fast binary diff, so things like indentation changes (which there are some of here) count toward the detection threshold.

The thinking was that if you required people to tell git about renames so they could be tracked explicitly, it would end up being even more inconsistent because you can't force all the world's software to stop calling the OS's "rename a file" functions. So you'd end up not tracking renames done by non-git-aware tools at all, even if the file contents were 100% identical. The git devs also pointed out that doing it the way they do it means that rename detection improvements over time will cause git to be able to better detect renames in existing history, whereas if you explicitly track renames, you can't fix them after the fact without rewriting history.

But yeah, it is annoying when a single commit has renamed files just above and just below the threshold and the results just look broken.

Side note: The git guys are right. In a past life I worked on a (now defunct) commercial version control system that had explicit renames. Our system used a custom filesystem, which meant we could intercept the OS-level rename operations and track them. So we caught most of the intentional renames automatically. But then we got tripped up by stuff like editors that do the equivalent of `mv foo.txt foo.txt.backup` before saving.

kallen07: I've used a command that changes the "similarity" threshold before, but I think was searching through old code. I'm guessing we can't modify the threshold for diff review.

PR #166, opened 2022-01-10
Distinguish optional vs. mandatory to-one relationships

sgrimm:
Keep track of which single-value sublists represent relationships where there
will always be a value. This will be used to determine whether or not a given
search field is nullable.


sgrimm:
* main
  * **PR #166**  (this PR)

PR #167, opened 2022-01-11
Support partial schema docs generation

sgrimm:
During development, it's handy to be able to look at the data model
diagram, but you usually only care about one or two of the subsets of
the tables that we generate diagrams for. Add support for generating
the docs for specific slices of the data model rather than always
generating all of them.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/db/SchemaDocsGenerator.kt
+++ b/src/test/kotlin/com/terraformation/backend/db/SchemaDocsGenerator.kt
@@ -16,6 +16,7 @@ import kotlin.io.path.Path
 import kotlin.io.path.exists
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Assumptions.assumeTrue
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.EnumSource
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #167**  (this PR)

sgrimm:
@kallen07 This should be a trivial one. I was on the fence about whether to even ask for a review, but better to err on the side of more eyes since it's not an urgent one.

src/test/kotlin/com/terraformation/backend/db/SchemaDocsGenerator.kt
```diff
@@ -191,6 +202,11 @@ class SchemaDocsGenerator : DatabaseTest() {
 
     assertTrue(docsDir.exists(), "$docsDir doesn't exist")
 
+    val slicesList = System.getenv(slicesEnvVar)
+    assumeTrue(
+        slicesList == null || slice.subdirectory in slicesList.split(','),
+        "Skipping slice $slice because it is not in $slicesEnvVar")
+
```
kallen07, line 209: At risk of sounding extremely dumb, where is `slicesList` or `slicesEnvVar` being used such that it's actually causing partial doc generation? It kinda seem like this diff just does a sanity check that the `slicesList` is valid and then doesn't use it...

Again, I'm sorry if this a dumb question.

sgrimm: It's not used anywhere by default, but when you're making data model changes locally, it can be handy to look at the diagrams and make sure the generated table docs look good. I did it when I was messing with species, for example. And in that case I don't want to sit there twiddling my thumbs while the `customer` and `device` slices get rebuilt when I know they have nothing to do with what I'm working on.

In CI, we always generate all slices whenever there's a schema change.

kallen07: Ah, so when developing you typically make a code change that uses the `slicesList`.

PR #168, opened 2022-01-15
TRACK-613 Add endpoints to create and update sites

sgrimm:


sgrimm:
* main
  * **PR #168**  (this PR)

src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
```diff
@@ -269,7 +269,12 @@ data class UserModel(
   }
 
   fun canReadSite(siteId: SiteId): Boolean {
-    return siteRoles[siteId] != null
+    return siteId in siteRoles
+  }
+
+  fun canUpdateSite(siteId: SiteId): Boolean {
+    val role = siteRoles[siteId]
+    return role == Role.ADMIN || role == Role.OWNER
```
kallen07: If this is up to date then I think we should include managers as well. https://terraformation.atlassian.net/wiki/spaces/PM/pages/279969793/Roles+Based+Permissions

I think that will also involve updating the manager permission tests.

src/main/kotlin/com/terraformation/backend/customer/db/SiteStore.kt
```diff
@@ -51,20 +50,50 @@ class SiteStore(
     }
   }
 
-  fun create(projectId: ProjectId, name: String, location: Point): SiteModel {
+  /**
+   * Creates a new site.
+   *
+   * @param row Initial site information. Project ID is required; ID and timestamps are ignored.
```
kallen07, line 56: Maybe I should have said something sooner when you started using this style, but I'm wondering what you think about adding some kind of delimiter (e.g. colon or dash) between the param name and the description? I find it hard to parse at a glance as written.

sgrimm: This is KDoc format: https://kotlinlang.org/docs/kotlin-doc.html

There is a syntax variant where you can put the parameter name in square brackets. How does this strike you?
```kotlin
  /**
   * Does something.
   *
   * @param[name] The name of the thing to which something needs to be done.
   */
```

sgrimm: You can also put a space before the left bracket, like
```kotlin
  /**
   * Does something.
   *
   * @param [name] The name of the thing to which something needs to be done.
   */
```

kallen07: The first option you provided is the most readable to me :)

src/main/kotlin/com/terraformation/backend/customer/model/UserModel.kt
sgrimm: Good catch! Thanks.

src/main/kotlin/com/terraformation/backend/customer/db/SiteStore.kt
sgrimm: Added a Jira task for this since we'll want to do it everywhere at once.

PR #169, opened 2022-01-18
TRACK-495 Include added time in organization user data

sgrimm:
The UI needs to show when users were added to organizations. We were tracking
that on the server side but not returning it as part of the organization user
payload. Add it.


sgrimm:
* main
  * **PR #169**  (this PR)

PR #172, opened 2022-01-20
Use consistent Postgres Docker image for build and test

sgrimm:
We configured the Docker image that gets used for code generation separately
from the one that gets used for running database tests, but we want the two
to always match. Move the image settings to Gradle properties and write them to
a generated source file so they're used verbatim by tests.

This also allows you to use a different image locally by overriding the
properties on the command line or in `$HOME/.gradle/gradle.properties`.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -159,6 +159,29 @@ val generateVersionFile by tasks.registering {
   }
 }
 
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #172**  (this PR)

PR #174, opened 2022-01-21
TRACK-657 Include createdTime in org-related payloads

sgrimm:
The web UI needs to show the creation dates/times of organizations, projects, etc.
Add `createdTime` fields to the relevant payload classes.

We weren't tracking creation time of facilities at all, so also add timestamps to
that table and set them when facilities are created or modified.


sgrimm:
* main
  * **PR #174**  (this PR)
    * **PR #175** 

PR #175, opened 2022-01-21
TRACK-649 Add organization-wide flag to projects

sgrimm:
To support seed banks being accessible organization-wide, add a project-level
flag to control whether or not the project is accessible to the whole organization.

There is currently no way for a client to create an organization-wide project; a later
change will use the new flag when it automatically creates seed bank projects in
new organizations.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
@@ -63,9 +63,9 @@ class ProjectsController(private val projectStore: ProjectStore) {
   fun createProject(@RequestBody payload: CreateProjectRequestPayload): GetProjectResponsePayload {
     val project =
         projectStore.create(
-            payload.organizationId,
-            payload.name,
-            payload.description,
+            description = payload.description,
+            name = payload.name,
+            organizationId = payload.organizationId,
             startDate = payload.startDate,
             status = payload.status,
             types = payload.types ?: emptyList())
@@ -84,12 +84,12 @@ class ProjectsController(private val projectStore: ProjectStore) {
       @RequestBody payload: UpdateProjectRequestPayload
   ): SimpleSuccessResponsePayload {
     projectStore.update(
-        projectId,
-        payload.description,
-        payload.name,
-        payload.startDate,
-        payload.status,
-        payload.types ?: emptyList())
+        description = payload.description,
+        name = payload.name,
+        projectId = projectId,
+        startDate = payload.startDate,
+        status = payload.status,
+        types = payload.types ?: emptyList())
     return SimpleSuccessResponsePayload()
   }
 


sgrimm:
* main
  * **PR #175**  (this PR)
    * **PR #176** 
      * **PR #177** 

sgrimm:
This doesn't have any functional dependency on #174; I just stacked it on top of that one to avoid DB migration numbering conflicts.

I am not in love with the term "per-user" here. I tried to think of a better word but I couldn't come up with a concise one that didn't imply behavior that isn't implemented here. For example (flipping the sense of the flag), "public" suggests the project is visible outside the org, and given we're likely to want to support actual public projects at some point, it seemed likely to be confusing. "Organization-wide" (again, flipping the sense of the flag) is accurate but kind of a mouthful. Suggestions welcome.

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
```diff
@@ -63,9 +66,9 @@ class ProjectsController(private val projectStore: ProjectStore) {
             payload.organizationId,
             payload.name,
             payload.description,
-            payload.startDate,
-            payload.status,
-            payload.types ?: emptyList())
+            startDate = payload.startDate,
+            status = payload.status,
+            types = payload.types ?: emptyList())
```
kallen07, line 71: Why are the parameter names added here but not on the lines above?

src/test/kotlin/com/terraformation/backend/customer/db/PermissionStoreTest.kt
```diff
@@ -49,13 +49,26 @@ internal class PermissionStoreTest : DatabaseTest() {
   }
 
   @Test
-  fun `fetchFacilityRoles only includes projects the user is in`() {
+  fun `fetchFacilityRoles only includes per-user projects the user is in`() {
```
kallen07: What do you think about this?
`fetchFacilityRoles only includes per-user projects when the user has project access()`

If you decide to make this change and the one below, then don't forget about the `fetchProjectRoles`  and `fetchSiteRoles` tests

src/test/kotlin/com/terraformation/backend/customer/db/PermissionStoreTest.kt
```diff
@@ -49,13 +49,26 @@ internal class PermissionStoreTest : DatabaseTest() {
...
     insertTestData()
     assertEquals(
         mapOf(FacilityId(1000) to Role.MANAGER, FacilityId(1001) to Role.MANAGER),
         permissionStore.fetchFacilityRoles(UserId(5)))
   }
 
+  @Test
+  fun `fetchFacilityRoles includes non-per-user projects`() {
```
kallen07: What about this? 
`fetchFacilityRoles includes organization-wide projects (where per-user = false)`

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
sgrimm: I did a refactoring in IntelliJ. Not sure why it decided to do that!

PR #176, opened 2022-01-21
TRACK-648 Add hidden flag to projects

sgrimm:
We'll be automatically creating a seed bank facility when a new organization is
created, which requires a project and a site. But we don't want the project and
its children to appear in the UI. Add a project-level flag to indicate to the
front end that it should hide the project and its children.

The flag doesn't do anything on the server side; it's purely for client use.

Currently the flag is read-only; we'll set it when we auto-create the seed bank
during organization creation, but clients aren't yet able to set it.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4987,6 +4987,7 @@ components:
     ProjectPayload:
       required:
       - createdTime
+      - hidden
       - id
       - name
       - organizationId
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #176**  (this PR)
    * **PR #177** 

sgrimm:
Note about the relationship between this and #175: in the initial release, where we're only using both flags for auto-created seed banks, it'll always be true that `perUser != hidden` but down the road it seems likely we'll want to expose `perUser` as a project setting people can choose, at which point they'll be independent variables.

PR #177, opened 2022-01-25
TRACK-648 Automatically create seed bank

sgrimm:
By default, creating a new organization now automatically creates a project
(hidden, organization-wide), site, and facility, all called "Seed Bank." The
client can pass an optional parameter to disable creating the seed bank.

Since organizations don't have location data, there isn't a sensible value to
use as the location for the site. We weren't really doing anything with the
site location anyway, so it is no longer a required field.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
@@ -67,7 +67,9 @@ class OrganizationService(
       val orgModel = organizationStore.createWithAdmin(row)
       val name = messages.seedBankDefaultName()
 
-      if (createSeedBank) {
+      if (!createSeedBank) {
+        orgModel
+      } else {
         val projectModel =
             projectStore.create(orgModel.id, name, hidden = true, organizationWide = true)
         val siteModel = siteStore.create(SitesRow(projectId = projectModel.id, name = name))
@@ -78,8 +80,6 @@ class OrganizationService(
                 listOf(
                     projectModel.copy(
                         sites = listOf(siteModel.copy(facilities = listOf(facilityModel))))))
-      } else {
-        orgModel
       }
     }
   }


sgrimm:
* main
  * **PR #177** Graphite  (this PR)
    * **PR #180** Graphite 

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
```diff
@@ -51,4 +61,26 @@ class OrganizationService(
...
+        val facilityModel = facilityStore.create(siteModel.id, name, FacilityType.SeedBank)
+
+        orgModel.copy(
+            projects =
+                listOf(
+                    projectModel.copy(
+                        sites = listOf(siteModel.copy(facilities = listOf(facilityModel))))))
+      } else {
+        orgModel
+      }
```
kallen07, line 83: minor style suggestion: when there's a short return I sometimes like to put it first, so
```
if (!createSeedBank) {
    return orgModel
}

... other logic no longer needs to be nested in an if
```

This might not be the standard style for Kotlin though.
Also, is the return statement implied here because this is inside `dslContext.transactionResult`?

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -192,6 +193,41 @@ data class UpdateOrganizationUserRequestPayload(
     val role: Role,
 )
 
+data class CreateOrganizationRequestPayload(
```
kallen07, line 196: I wish it were possible to make this a subclass of the `UpdateOrganizationRequestPayload` because then I could quickly scan and see that this only has one additional field. 

Are you planning on going into the frontend code to update the call, or just letting the create seedbank behavior happen by default? One tiny change that would be nice to go along with this would be to change the name of the following variable to `CreateOrganizationRequestPayload` (on the frontend).

```
type UpdateOrganizationRequestPayload =
  paths[typeof ORGANIZATIONS]['post']['requestBody']['content']['application/json'];
```

src/test/kotlin/com/terraformation/backend/customer/OrganizationServiceTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            name = "Test Organization",
+            projects =
+                listOf(
+                    ProjectModel(
+                        createdTime = clock.instant(),
+                        hidden = true,
+                        id = ProjectId(1),
+                        organizationId = OrganizationId(1),
+                        organizationWide = true,
+                        name = "Seed Bank",
```
kallen07: Is it possible to link this to `src/main/kotlin/com/terraformation/backend/i18n/Messages.kt` instead of hard coding the name?

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
sgrimm: I hadn't been planning to change the frontend code since that new argument defaults to the value we want anyway, but good point about the name change. I'll do that after landing this.

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
sgrimm: Yeah, it's because the code block here is an expression, not a statement. It's the same structure as something like `foo?.let { it.bar }` but with a larger body.

src/test/kotlin/com/terraformation/backend/customer/OrganizationServiceTest.kt
sgrimm: Yeah, that's fair. My hesitation about doing that is not really relevant yet: namely, that the i18n code will, at some point, stop returning hardcoded string constants and start doing localization logic that we might not want to force this test to know about. But on balance that's probably no worse than the string here falling out of sync with the string there.

sgrimm: And also, the better solution there is to stub `Messages` out anyway.

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
sgrimm: Flipped the order of that `if` statement to make the simple case come first.

PR #178, opened 2022-01-26
TRACK-671 Support moving sites between projects

sgrimm:
Admins and owners need to be able to select a different project when they are editing a
site's details. Add project ID as an optional input to the PUT endpoint for sites.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/db/SiteStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/db/SiteStoreTest.kt
@@ -101,7 +101,7 @@ internal class SiteStoreTest : DatabaseTest(), RunsAsUser {
 
   @Test
   fun `update throws exception if user has no permission to update site`() {
-    every { user.canUpdateSite(any()) } returns false
+    every { user.canUpdateSite(siteId) } returns false
 
     val rowWithNewName = sitesDao.fetchOneById(siteId)!!.copy(name = "New Name")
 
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #178** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/178" target="_blank">Graphite</a>  (this PR)

src/test/kotlin/com/terraformation/backend/customer/db/SiteStoreTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    assertThrows<AccessDeniedException> { store.update(rowWithNewProject) }
+  }
+
+  @Test
+  fun `update throws exception if user has no permission to create site in target project`() {
+    val newProjectId = ProjectId(3)
+    insertProject(newProjectId, organizationId)
+
+    every { user.canCreateSite(any()) } returns false
```
kallen07: Is there a reason that this function doesn't use the exact site ID when mocking the permissions?

kallen07:
FYI i made a super nitty change from "new project" to "different project" in the PR description. The word "new" made me think of a flow where the "user can create a brand new project at the same time while updating a site".

src/test/kotlin/com/terraformation/backend/customer/db/SiteStoreTest.kt
sgrimm: No particular reason, nope. Good catch; it makes the test more thorough to use a specific ID.

PR #179, opened 2022-01-26
TRACK-670 Return per-project user counts

sgrimm:
Add the ability to return the number of users in each project as part of the
project data. The client must explicitly request it, since we don't want to have
to run the additional database queries if the results aren't needed.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
@@ -246,13 +246,16 @@ class ProjectStore(
     val projectIdField = DSL.field(DSL.name("project_id"), ProjectId::class.java)
     val countField = DSL.count()
 
-    // For each project ID, a list of the organization's admin users.
-    val adminUsersQuery =
+    // For each project ID, a list of the users who have access to the project without needing to be
+    // explicitly added, either because they have a privileged role or because the project is
+    // organization-wide.
+    val implicitUsersQuery =
         DSL.select(PROJECTS.ID.`as`(projectIdField), ORGANIZATION_USERS.USER_ID)
             .from(PROJECTS)
             .join(ORGANIZATION_USERS)
             .on(PROJECTS.ORGANIZATION_ID.eq(ORGANIZATION_USERS.ORGANIZATION_ID))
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #179** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/179" target="_blank">Graphite</a>  (this PR)

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -230,4 +231,51 @@ class ProjectStore(
...
+    requirePermissions { projectIds.forEach { readProject(it) } }
+
+    if (projectIds.isEmpty()) {
+      return emptyMap()
+    }
+
+    val projectIdField = DSL.field(DSL.name("project_id"), ProjectId::class.java)
+    val countField = DSL.count()
+
+    // For each project ID, a list of the organization's admin users.
```
kallen07: nit: the comment here and on 263 is a little misleading because "admin" seems to be used to mean administrator users (owners and admins), and also "admin" is a role.

maybe we can explicitly write "admin and owner"

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -230,4 +231,51 @@ class ProjectStore(
       throw UserNotFoundException(userId)
     }
   }
+
+  /**
+   * Returns the number of users in a list of projects. Administrators and owners are considered
+   * users of all projects in the organization.
+   */
+  fun countUsers(projectIds: Collection<ProjectId>): Map<ProjectId, Int> {
```
kallen07, line 239: Do we want this function to handle organization-wide projects as a special case? It looks like, right now, this function would only return the number of admins + owners for organization-wide projects.

This doesn't impact what is displayed to the user, since all organization wide projects are currently hidden, but it still seems like we'd want this endpoint to handle the case gracefully.

sgrimm: Good catch. That's what I get for doing all this stuff as an unordered bag of independent changes.

PR #180, opened 2022-01-26
TRACK-593 Record timestamp of most recent user activity

sgrimm:
Add a column to the users table to record when the system most recently received a
request from the user.

This makes use of an application event that gets automatically published by
Spring's request handling code when a request is finished processing.


sgrimm:
* main
  * **PR #177** Graphite 
    * **PR #180** Graphite  (this PR)

sgrimm:
Stacked this one on top of #177 to avoid migration numbering conflicts.

Originally I'd planned to define a new event class and explicitly publish the event from our authentication filter, but then I realized there's already an event that's guaranteed to be published and it already includes the user's identity and the request timestamp. No point adding more moving parts.

PR #182, opened 2022-01-28
TRACK-421 Support per-organization species lists

sgrimm:
Update the data model so that each organization can have its own list of species.

From the API client's perspective, the only change here is that the response to
`/api/v1/species/name/{id}` now includes an organization ID. We already made it
mandatory for the client to pass in an organization ID where needed.

Internally, the code now makes use of the `species_options` table, which
previously existed but didn't do anything. The presence of a row in that table
is what determines whether or not a species is included in the organization's
list of species.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
@@ -149,6 +149,10 @@ class SpeciesStore(
   fun findAllSpeciesNames(organizationId: OrganizationId): List<SpeciesNamesRow> {
     requirePermissions { readOrganization(organizationId) }
 
+    // We want all the species names associated with an organization ID for which there is also
+    // a species option for the same species and the same organization. Checking the option is
+    // necessary to filter out species that have been deleted from the organization; deletion only
+    // removes the species_options row, not the names.
     return dslContext
         .selectFrom(SPECIES_NAMES)
         .where(SPECIES_NAMES.ORGANIZATION_ID.eq(organizationId))
@@ -188,19 +192,20 @@ class SpeciesStore(
           } else {
             val newRow = row.copy(createdTime = clock.instant(), modifiedTime = clock.instant())
             speciesDao.insert(newRow)
-            newRow.id ?: throw RuntimeException("BUG! Species ID not returned after insertion")
-          }
+            val newId =
+                newRow.id ?: throw RuntimeException("BUG! Species ID not returned after insertion")
 
-      if (existingSpeciesId == null) {
-        speciesNamesDao.insert(
-            SpeciesNamesRow(
-                createdTime = clock.instant(),
-                isScientific = row.isScientific,
-                name = row.name,
-                modifiedTime = clock.instant(),
-                organizationId = organizationId,
-                speciesId = speciesId))
-      }
+            speciesNamesDao.insert(
+                SpeciesNamesRow(
+                    createdTime = clock.instant(),
+                    isScientific = row.isScientific,
+                    name = row.name,
+                    modifiedTime = clock.instant(),
+                    organizationId = organizationId,
+                    speciesId = newId))
+
+            newId
+          }
 
       speciesOptionsDao.insert(
           SpeciesOptionsRow(
@@ -216,6 +221,11 @@ class SpeciesStore(
   /**
    * Updates the data for an existing species.
    *
+   * If the species name in [row] differs from the existing primary name, the new name also replaces
+   * the existing one in the species' list of names, to maintain the invariant that the primary name
+   * (the one on the `species` table) is always also included in the full list of names (the rows in
+   * the `species_names` table).
+   *
    * @throws DuplicateKeyException The requested name was already in use.
    * @throws SpeciesNotFoundException No species with the requested ID exists.
    */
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #182** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/182" target="_blank">Graphite</a>  (this PR)

src/main/kotlin/com/terraformation/backend/species/api/SpeciesController.kt
```diff
@@ -123,32 +116,33 @@ class SpeciesController(
           description = "Cannot delete the species because it is currently in use."))
   @ApiResponse404
   @DeleteMapping("/{speciesId}")
-  @Operation(summary = "Deletes an existing species.")
+  @Operation(
+      summary = "Deletes an existing species.",
+      description =
+          "The species will no longer appear in the organization's list of species, but existing " +
+              "data (plants, seeds, etc.) that refer to the species will still refer to it.")
```
kallen07, line 123: I think the product decision here may be confusing to users, and it was confusing to me before I saw how this is implemented. But I don't really have a better suggestion that isn't significantly more work. E.g. not letting the user delete a species that's being referenced by other data, but then we'd probably have to surface that data somewhere.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -27,44 +36,181 @@ class SpeciesStore(
...
-            name = row.name,
-            modifiedTime = clock.instant(),
-            speciesId = speciesId))
+  fun findAllSpeciesNames(organizationId: OrganizationId): List<SpeciesNamesRow> {
+    requirePermissions { readOrganization(organizationId) }
 
-    return speciesId
+    return dslContext
+        .selectFrom(SPECIES_NAMES)
+        .where(SPECIES_NAMES.ORGANIZATION_ID.eq(organizationId))
```
kallen07, line 158: Since we're checking the organization id here, why do we also need to check it on the `SPECIES_OPTIONS` table? I think I'm just misunderstanding the data model slightly.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -27,44 +36,181 @@ class SpeciesStore(
...
+      if (existingSpeciesId == null) {
+        speciesNamesDao.insert(
+            SpeciesNamesRow(
+                createdTime = clock.instant(),
+                isScientific = row.isScientific,
+                name = row.name,
+                modifiedTime = clock.instant(),
+                organizationId = organizationId,
+                speciesId = speciesId))
+      }
```
kallen07: Do we maybe want to nest lines 195 -203 inside the `else` on line 188?

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -86,15 +232,19 @@ class SpeciesStore(
...
 
       if (existingNamesRowForNewName != null) {
         // We are updating the primary name to a name that was previously added as a secondary.
         // Use the existing secondary name so its creation time is preserved.
         dslContext
             .deleteFrom(SPECIES_NAMES)
-            .where(SPECIES_NAMES.SPECIES_ID.eq(speciesId).and(SPECIES_NAMES.NAME.eq(existing.name)))
+            .where(SPECIES_NAMES.SPECIES_ID.eq(speciesId))
+            .and(SPECIES_NAMES.NAME.eq(existing.name))
+            .and(SPECIES_NAMES.ORGANIZATION_ID.eq(organizationId))
```
kallen07: Why are we deleting the `existingNamesRowForNewName` here and then attempting to modify it on line 250? Am I misunderstanding?

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -123,22 +274,22 @@ class SpeciesStore(
   }
 
   /**
-   * Deletes a species. This will fail if there are any entities referencing the species.
+   * Deletes a species from an organization. This doesn't remove any existing references to the
+   * species, just prevents it from being used in the future.
```
kallen07, line 307: Strictly speaking, we can still use the "species" in the future, although it seems like a new SPECIES_OPTIONS row would need to be created. Right? So, from the user perspective, nothing is blocking the user from using the species again, but it won't auto populate in lists, etc.

src/main/kotlin/com/terraformation/backend/species/api/SpeciesController.kt
sgrimm: Happily, there isn't actually a way to delete  species in the frontend yet, so we get to kick that discussion down the road a bit, but I agree 100% that it's unclear what the right behavior here is.

"Don't let people delete species that are referenced" isn't clearly (to me) better than this from a product point of view, for what it's worth. It will lead to situations like, "You can't stop this species from showing up in the species menu in the mobile app until you go edit the 1,653 existing accessions that refer to it. And also, you will no longer have a record of what species all those old accessions were."

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
sgrimm: I'll try to clarify this in comments. Thanks.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
sgrimm: Will add a comment to hopefully clarify this. It's a corner case that might never happen in reality so is kind of hard to intuit.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
sgrimm: Correct. The species ID will still be accessible (and it needs to be, so you can display species info for existing seeds/plants) but it won't be returned by the "list all the species" endpoints that populate the dropdowns and so forth.

We could maybe start rejecting deleted species in write operations on accessions and plants, but that would have some product implications we'd have to think through (e.g., if I delete a species right before someone submits seeds of that species from the mobile app, do we reject the submission, and if so, what does that look like to the mobile app user?)

kallen07: Sounds like a product question that we probably don't need to address before the release.

PR #183, opened 2022-02-03
Insert mock user into users table in tests

sgrimm:
Tracking which users created/modified things will mean adding user ID columns to a
number of tables. Those user IDs will need to refer to IDs that exist in the users
table. That has an impact on tests, where previously we could get away with just
using an in-memory mock object. We'll still do that, but the mock needs to have a
user ID, and the user will need to exist in the users table.

Address this by introducing a helper function to create a mock user with a
reasonable set of default behaviors, and modify `insertSiteData()` to insert a
user before it inserts the organization and so forth.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/MockUser.kt
+++ b/src/test/kotlin/com/terraformation/backend/MockUser.kt
@@ -1,0 +1,28 @@
+package com.terraformation.backend
+
+import com.terraformation.backend.auth.CurrentUserHolder
+import com.terraformation.backend.customer.model.IndividualUser
+import com.terraformation.backend.db.UserId
+import com.terraformation.backend.db.UserType
+import io.mockk.CapturingSlot
+import io.mockk.every
+import io.mockk.mockk
+
+fun mockUser(userId: UserId = UserId(2)): IndividualUser {
+  val user: IndividualUser = mockk(relaxed = true)
+
+  every { user.userId } returns userId
+  every { user.email } returns "$userId@terraformation.com"
+  every { user.authId } returns "$userId"
+  every { user.firstName } returns "First"
+  every { user.lastName } returns "Last"
+  every { user.userType } returns UserType.Individual
+
+  val funcSlot = CapturingSlot<() -> Any>()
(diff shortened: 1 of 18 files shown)


sgrimm:
* main
    * **PR #183** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/183" target="_blank">Graphite</a>  (this PR)

src/test/kotlin/com/terraformation/backend/MockUser.kt
```diff
@@ -0,0 +1,28 @@
...
+
+  every { user.userId } returns userId
+  every { user.email } returns "$userId@terraformation.com"
+  every { user.authId } returns "$userId"
+  every { user.firstName } returns "First"
+  every { user.lastName } returns "Last"
+  every { user.userType } returns UserType.Individual
+
+  val funcSlot = CapturingSlot<() -> Any>()
+  every { user.run(capture(funcSlot)) } answers
```
kallen07, line 22: Kotlin language question - what is the purpose lines of 21-25?

sgrimm: That's actually more of a MockK API question. Basically, if we're using the mock user and we encounter code (either in the test or in the application) like
```kotlin
someUser.run { someFunction() }
```
then it doesn't work to just return a fixed dummy value from the mock user's `run` method because `someFunction` wouldn't be called, and presumably it does work that we want the test to be able to verify.

So this code defines a "capturing slot" which is a way to tell MockK to capture the value of a method call on the mock object. When the block of code in the `answers` gets run by MockK, the callback function has been saved in `funcSlot`, and we can pass it to `CurrentUserHolder.runAs` which will execute it just like it would execute any other callback that gets passed to it.

Does that explanation make sense?

sgrimm: Docs on MockK capturing: https://mockk.io/#capturing

kallen07: Ohhhhh. That's pretty clever.
Your explanation was very helpful. Thanks!

PR #184, opened 2022-02-03
TRACK-678 Add system user

sgrimm:
We're going to add "created by" and "modified by" fields to a bunch of entities to
keep track of which users are making changes. But some changes are made
automatically by the system, e.g., advancing accessions to new states based on the
passage of time. In those cases, the changes need to be attributed somehow, but
there's no individual user to attribute them to.

Add a special "system" user that we can use for automated changes that aren't
initiated by a specific individual user.

The system user needs to be able to modify data across all organizations and
projects. Rather than explicitly adding it as a member of every project, instead
we hardwire its permission checks to always return `true`.

There's no clean way to hardwire permission checks in `UserModel` without
introducing a bunch of special-case code that could potentially be run by mistake
for regular users; it's simpler and less bug-prone to introduce a separate model
class that can only ever represent the system user, and allow it to be used in
place of `UserModel`.

To that end, this change:

* Adds a `TerrawareUser` interface that declares all the methods and properties
  that aren't specific to individual (human or API client) users. `currentUser()`
  is defined as returning a `TerrawareUser` implementation, and most code doesn't
  need to care whether that's the system user or a regular user.
* Adds a `SystemUser` class that implements `TerrawareUser` and has hardwired
  permission checks. This is a singleton: there is exactly one system user, and
  it is impossible to log in as the system user.
* Renames `UserModel` to `IndividualUser` so the class names are consistent.
* Updates `DailyTaskRunner` so that scheduled tasks are run as the system user.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -62,15 +62,10 @@ data class IndividualUser(
     private val parentStore: ParentStore,
     private val permissionStore: PermissionStore,
 ) : TerrawareUser, UserDetails {
-  /** The user's role in each organization they belong to. */
   override val organizationRoles: Map<OrganizationId, Role> by lazy {
     permissionStore.fetchOrganizationRoles(userId)
   }
 
-  /**
-   * The user's role in each project they have access to. Currently, roles are assigned
-   * per-organization, so this is really the user's role in the organization that owns each project.
-   */
   override val projectRoles: Map<ProjectId, Role> by lazy {
     permissionStore.fetchProjectRoles(userId)
   }
@@ -82,11 +77,6 @@ data class IndividualUser(
    */
   private val siteRoles: Map<SiteId, Role> by lazy { permissionStore.fetchSiteRoles(userId) }
 
-  /**
-   * The user's role in each facility they have access to. Currently, roles are assigned
-   * per-organization, so this is really the user's role in the organization that owns the project
-   * and site of each facility.
-   */
   override val facilityRoles: Map<FacilityId, Role> by lazy {
     permissionStore.fetchFacilityRoles(userId)
   }
@@ -99,17 +89,6 @@ data class IndividualUser(
   val fullName: String?
     get() = if (firstName != null && lastName != null) "$firstName $lastName" else null
 
-  /**
-   * Runs some code as this user.
-   *
-   * This is useful in two scenarios. First, if the code isn't running on a request handler thread
-   * (e.g., in a unit test or on a thread pool), any calls to [currentUser] will fail because there
-   * won't be a current user.
-   *
-   * Second, less common, is masquerading: if there is already a current user, it will be replaced
-   * with this user for the duration of the function, and then the current user will be restored
-   * afterwards.
-   */
   override fun <T> run(func: () -> T): T {
     return CurrentUserHolder.runAs(this, func, authorities)
   }


sgrimm:
* main
  * **PR #184** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/184" target="_blank">Graphite</a>  (this PR)

src/main/kotlin/com/terraformation/backend/customer/model/SystemUser.kt
```diff
@@ -0,0 +1,149 @@
...
+    return CurrentUserHolder.runAs(this, func)
+  }
+
+  override fun hasAnyAdminRole(): Boolean = true
+  override fun getName(): String = USERNAME
+
+  /*
+   * All permission checks always succeed.
+   */
+
```
kallen07, line 89: This is most likely not possible, but is there a way for us to enumerate over the list of all the functions that need to be overridden and define them to all return true?

Presumably Kotlin's type safety ensures that all each of the `TerrawareUser` interface functions are overridden, so my suggestion is more for ease of reading than anything.

src/main/kotlin/com/terraformation/backend/customer/model/TerrawareUser.kt
```diff
@@ -0,0 +1,124 @@
...
+  val userId: UserId
+  val userType: UserType
+
+  /** The user's role in each organization they belong to. */
+  val organizationRoles: Map<OrganizationId, Role>
+
+  /**
+   * The user's role in each project they have access to. Currently, roles are assigned
+   * per-organization, so this is really the user's role in the organization that owns each project.
+   */
```
kallen07, line 36: A few comments like this one are duplicated in the `IndividualUser` definition. What do you think about putting them only in here (or there)? My concern is that the comments will get out of sync.

src/main/resources/db/migration/common/V73__SystemUser.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO user_types (id, name)
+VALUES (4, 'System');
+
+-- Auth ID of 'DISABLED' has no special meaning but will never match a Keycloak UUID, meaning it is
+-- impossible to authenticate as this user.
+INSERT INTO users (auth_id, email, first_name, last_name, created_time, modified_time, user_type_id)
+VALUES ('DISABLED', 'system', 'Terraware', 'System', NOW(), NOW(), 4);
```
kallen07, line 7: Maybe a dumb question, but how do you know that Keycloak will never return a UUID of `DISABLED`? What are their guidelines for generating UUIDs?

src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
```diff
@@ -122,7 +122,7 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
 
     every { user.organizationRoles } returns mapOf(organizationId to Role.OWNER)
     every { user.projectRoles } returns mapOf(projectId to Role.OWNER)
-    every { user.userId } returns UserId(1)
+    every { user.userId } returns UserId(2)
```
kallen07, line 125: are you changing this to match the definition of `mockkUser` from the last diff?

src/test/kotlin/com/terraformation/backend/daily/DailyTaskRunnerTest.kt
```diff
@@ -110,4 +115,16 @@ internal class DailyTaskRunnerTest : DatabaseTest() {
...
+  @Test
+  fun `runs tasks as system user`() {
+    every { task.processPeriod(any(), any()) } answers
+        {
+          assertEquals(systemUser.userId, currentUser().userId)
+        }
+
+    dailyTaskRunner.runTask(task)
+
+    verify(exactly = 1) { task.processPeriod(any(), any()) }
```
kallen07, line 139: What does this line do?

src/main/kotlin/com/terraformation/backend/customer/model/SystemUser.kt
sgrimm: Yes, there is a way to do that, and I considered it. You'd use [dynamic proxies](https://docs.oracle.com/javase/8/docs/technotes/guides/reflection/proxy.html) (docs are for Java but it's a JVM feature that works in Kotlin code too). With a dynamic proxy, every method call on an instance of an interface gets routed to a generic handler function that takes the method name as an argument. In this case, that method could do something like `if (methodName.startsWith("can")) return true`.

The reason I didn't go with that approach is just that it's not very IDE-friendly. If you ask IntelliJ to show you all the implementations of a given interface, it has no way of knowing about dynamic proxies because they're defined completely at runtime.

As you say, the compiler will prevent us from forgetting to implement a new `canWhatever` method. If that turns out to be a maintenance headache, we can consider switching to dynamic proxies. I figured we add new permission types infrequently enough that it won't be a pain.

src/main/resources/db/migration/common/V73__SystemUser.sql
sgrimm: It's because I'm not using "UUID" as a generic term to mean "unique identifier" here. Keycloak user IDs are true [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier) in string form -- 32-digit-long hexadecimal numbers with some hyphens.

src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
sgrimm: Yeah, pretty much, though technically it's the other way around. This PR adds a migration that inserts the system user, which, on a fresh database, will always be assigned user ID 1. So any additional users tests need to create can't use user ID 1 because it'll always collide now (even if it didn't before this PR).

src/test/kotlin/com/terraformation/backend/daily/DailyTaskRunnerTest.kt
sgrimm: Will add a clarifying comment.

src/main/kotlin/com/terraformation/backend/customer/model/TerrawareUser.kt
sgrimm: Yeah, that's a valid concern. In a few cases, they are similar but not identical (the `IndividualUser` one has implementation notes, etc.) but I'll get rid of the duplication.

sgrimm: I was wrong! There were no similar-but-not-identical comments after all. I think I was messing with that during development and forgot where I ended up. I've removed the redundant comments entirely.

PR #185, opened 2022-02-03
TRACK-594 Add created-by and modified-by columns to tables

sgrimm:
Add `created_by` and `modified_by` columns to all the tables that represent
things that can be directly edited via the API. That includes most tables, but
excludes things like withdrawals which can only be edited by editing the
accession.

Add `created_time` and or `modified_time` columns to tables that didn't already
have them.

These new columns are all non-nullable. The database migration attempts to
populate them with reasonable defaults, which in most cases means the owner of
the organization they're associated with.

This change touches a lot of files, but most of the changes are pretty similar:
populate the new columns in the create/update methods, and modify tests to
expect values in the new columns.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/db/WithdrawalStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/db/WithdrawalStoreTest.kt
@@ -59,13 +59,18 @@ internal class WithdrawalStoreTest : DatabaseTest(), RunsAsUser {
     insertSiteData()
 
     // Insert a minimal accession in a state that allows withdrawals.
-    dslContext
-        .insertInto(ACCESSIONS)
-        .set(ACCESSIONS.ID, accessionId)
-        .set(ACCESSIONS.CREATED_TIME, Instant.now())
-        .set(ACCESSIONS.FACILITY_ID, facilityId)
-        .set(ACCESSIONS.STATE_ID, AccessionState.InStorage)
-        .execute()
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
      * **PR #185** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/185" target="_blank">Graphite</a>  (this PR)

sgrimm:
@kallen07 Obviously this will get squashed to one commit when it's merged, but I kept the PR split up into smaller commits in case that makes it easier to review.

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
```diff
@@ -36,19 +37,27 @@ class TimeseriesStore(private val dslContext: DSLContext) {
...
+    val userId = currentUser().userId
+
     return with(TIMESERIES) {
       dslContext
           .insertInto(TIMESERIES)
-          .set(NAME, row.name)
-          .set(DEVICE_ID, row.deviceId)
-          .set(UNITS, row.units)
+          .set(CREATED_BY, userId)
+          .set(CREATED_TIME, clock.instant())
```
kallen07, line 46: on conflict, will the existing `CREATED_BY` and `CREATED_TIME` values be preserved?

sgrimm: Yes, it'll only update the columns that come after the `.doUpdate()` in that case.

PR #186, opened 2022-02-07
TRACK-617 Use Terraware-themed HTML email

sgrimm:
Update the HTML version of the "You've been added to an organization" email 
message to use the Terraware email look and feel.

Terraware-themed emails are authored in MJML rather than in handcrafted HTML. 
The utility that turns MJML into HTML is a Node.js app. This change therefore
adds `package.json` and `yarn.lock` files to install the MJML tool, and adds a
Gradle module that lets us execute Yarn commands as part of a build. This is
purely a build-time thing; terraware-server itself knows nothing about MJML
and we package the generated HTML files with the distribution.

### DIFF
--- a/src/main/resources/templates/email/README.md
+++ b/src/main/resources/templates/email/README.md
@@ -4,7 +4,7 @@ This directory contains templates for server-generated email messages. Each temp
 
 There are two versions of each email message: a plaintext version and an HTML version.
 
-`body.txt` is the plaintext version. It may be edited directly.
+`body.txt` is the plaintext version.
 
 `body.mjml` is the source for the HTML version. It isn't actual HTML; the HTML is generated from [MJML](https://mjml.io/).
 


sgrimm:
* main
  * **PR #186** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/186" target="_blank">Graphite</a>  (this PR)

sgrimm:
@kallen07 This PR is a bit esoteric; I'm not expecting you to learn Gradle task authoring to review it. Maybe focus on whether the comments would be enough to tell you what's going on rather than on the nitty-gritty details of building dynamic lists of build tasks. One specific thing I'd like feedback on, since it's not my usual stomping ground, is my `package.json` file; am I doing anything dumb in there?

sgrimm:
Updated this to load images from the frontend's assets directory. Corresponding PR that actually creates the files: https://github.com/terraware/terraware-web/pull/142

src/main/resources/templates/email/README.md
```diff
@@ -0,0 +1,17 @@
+# Email templates
+
+This directory contains templates for server-generated email messages. Each template is in its own subdirectory.
+
+There are two versions of each email message: a plaintext version and an HTML version.
+
+`body.txt` is the plaintext version. It may be edited directly.
+
+`body.mjml` is the source for the HTML version. It isn't actual HTML; the HTML is generated from [MJML](https://mjml.io/).
```
kallen07, line 9: Can we also edit the mjml directly as well? I think the answer is yes but I'm checking since that's specifically called out for the plaintext file.

sgrimm: Ah, good point. The "edited directly" comment was left over from a point in my development when I was writing the generated HTML into the src directory so it could be committed to version control, and I needed to say people shouldn't hand-edit it. I've removed that sentence.

PR #190, opened 2022-02-21
TRACK-717 Fix race condition in new user creation

sgrimm:
When the user first logs in, the web app sends several API requests at once to
populate various parts of the UI. That's normally fine, but in the case where the
user just registered with Keycloak and is logging in for the first time, it
means there can be two or more request handler threads that fail to find the
user in terraware-server's database. All the threads will then query Keycloak
to get the user's information, and they'll all try to insert the user into the
database.

Previously, this caused all but one of the requests to return HTTP 500 because the
insertion only succeeded on one thread thanks to the unique constraint on the
users table, and the other request handlers aborted with `DuplicateKeyException`.

Now the system will search for the user again when the insert fails; in the case of
this race condition, the user will exist and the request handler can proceed as
normal.

If the user still doesn't exist, it means something else is going on; in that case
the exception is thrown as before.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -93,7 +93,7 @@ class UserStore(
    * @throws KeycloakUserNotFoundException There is no user with that ID in the Keycloak database.
    */
   fun fetchByAuthId(authId: String): IndividualUser {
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #190** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/190" target="_blank">Graphite</a>  (this PR)

PR #191, opened 2022-02-22
TRACK-718 Use standard error message for invalid email

sgrimm:
When there's a structural problem with a request payload, e.g., the client passing
in an integer for a field that should be a boolean, the server returns an error
message, "Field value has incorrect format: X".

Use the same message if the client makes a POST request to
`/api/v1/organizations/{id}/users` with an invalid email address.

Eventually we'll want to return machine-readable information about validation
failures, but for now, this will at least make the error consistent with
validation failures in other fields.


sgrimm:
* main
  * **PR #191** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/191" target="_blank">Graphite</a>  (this PR)

PR #193, opened 2022-02-22
TRACK-745 Require names to be nonempty

sgrimm:
Previously, the server accepted requests to create or update organizations,
projects, and sites with zero-length names. From the user's point of view, we were
effectively not treating names as required.

Add validation annotations to the appropriate request payloads, and add a handler
to render the validation failure exceptions in human-readable form. Now if you
pass in an empty name, you'll get back an HTTP 400 with a standard JSON error payload:

    {
        "error": {
            "message": "Field value invalid: name must not be empty"
        },
        "status": "error"
    }

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
@@ -203,7 +203,7 @@ data class ProjectPayload(
                 "displayed to end users.")
     val hidden: Boolean,
     val id: ProjectId,
-    @field:NotEmpty val name: String,
+    val name: String,
     val organizationId: OrganizationId,
     @Schema(
         description =


sgrimm:
* main
  * **PR #193** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/193" target="_blank">Graphite</a>  (this PR)

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
```diff
@@ -199,7 +203,7 @@ data class ProjectPayload(
                 "displayed to end users.")
     val hidden: Boolean,
     val id: ProjectId,
-    val name: String,
+    @field:NotEmpty val name: String,
```
kallen07: How does this validation help us when it's done on our response payloads?

sgrimm: Good catch. Apparently I got overzealous; you're correct that it has no effect here. I'll remove it.

PR #196, opened 2022-02-24
TRACK-774 Create default storage locations

sgrimm:
We create a default seed bank facility for new organizations, but if users try to
use it, the list of storage locations is empty. Eventually we'll probably need to
give users more control over this, but for now, create three refrigerators and
three freezers as part of the default setup of a seed bank, which is the correct
setup for our containerized seed banks.


sgrimm:
* main
  * **PR #196** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/196" target="_blank">Graphite</a>  (this PR)

sgrimm:
This was simple so I decided to just do it while I was messing with storage location code, but we're still awaiting confirmation on the Jira task that this is the behavior we actually want  for the initial release. So it's a draft for now.

sgrimm:
Confirmed that we want this behavior for initial launch.

I considered building this by adding a config option that would let you specify the default storage location names/types in `application.yaml` but it seemed like needless extra complexity given the likelihood that this hack will go away once we figure out what we want to do about facility administration in the product as a whole.

PR #197, opened 2022-03-01
Rename notifications to accession_notifications

sgrimm:
In anticipation of starting work on a more general-purpose notifications system,
rename the tables and classes that support the existing accession-centric one.
It will make more sense to use the general term "notifications" to refer to the
newer system (though this commit doesn't introduce any new code or schema
elements, just renames the old ones).

No client-visible change here other than the names of some payloads in the
OpenAPI schema; the existing notifications API was already under
`/api/v1/seedbank` and continues to behave exactly as before.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2030,7 +2030,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/NotificationListResponse'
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #197** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/197" target="_blank">Graphite</a>  (this PR)

PR #200, opened 2022-03-04
TRACK-659 Add users to search API

sgrimm:
There is now a `members` sublist under both organizations and projects that can 
be used to search for user information.

Each of these sublists has the creation date of the membership, the user's role 
(in the case of organizations), and a `user` single-value sublist that contains
the actual user information.

Search results will only include users that are members of at least one of the
current user's organizations.

Similarly, the user information has `organizationMemberships` and
`projectMemberships` sublists, which will only include organizations and
projects the current user has permission to see.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/OrganizationUsersTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/OrganizationUsersTable.kt
@@ -22,7 +22,7 @@ class OrganizationUsersTable(tables: SearchTables, fuzzySearchOperators: FuzzySe
           organizations.asSingleValueSublist(
               "organization", ORGANIZATION_USERS.ORGANIZATION_ID.eq(ORGANIZATIONS.ID)),
           projectUsers.asMultiValueSublist(
-              "projects",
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #200** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/200" target="_blank">Graphite</a>  (this PR)

sgrimm:
One thing I'm not entirely happy with here and would love feedback on is the overloading of names. The `users` sublist under organizations is actually a list of the organization's user details (the `organization_users` table) and doesn't include things like email address; you have to jump down to another sublist for that. So you get weird field paths like `users.user.email`. Similarly, if you're navigating from a user to their projects or organizations, you have a level of indirection the other way: `projects.project.name`. The middle layers need to exist, though, because they hold fields such as the user's role in the particular organization.

Would it be better if I used a totally different word to refer to those middle layers? Say, "member" and "membership"? So you'd have `members.user.email` under organizations, and `projectMemberships.project.name` under users. I thought about doing that but I'm not sure I like it any better.

PR #201, opened 2022-03-05
TRACK-775 Reject negative seed quantities

sgrimm:
Previously, there was no sanity checking on the quantities of withdrawals or
germination tests and it was possible to withdraw 1000 seeds from an accession
with 50 seeds remaining. Add validation checks to reject updates that cause the
remaining quantity to drop below zero.

It was also possible to specify a zero or negative total accession size; catch
that as well.


sgrimm:
* main
  * **PR #201** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/201" target="_blank">Graphite</a>  (this PR)

PR #204, opened 2022-03-07
Upgrade to jOOQ 3.16

sgrimm:
The latest version of jOOQ has started adding support for spatial data types
such as the PostgreSQL `GEOMETRY` type. For now, we want to continue using the
PostGIS library since it is more feature-rich, but the binding code that
converts raw database values to classes from the PostGIS library needs to be
updated to recognize that the "raw database values" are now jOOQ `Geometry`
objects rather than simple strings.

This change doesn't affect the application code aside from `GeometryBinding`,
but it's required if we ever want to upgrade to subsequent jOOQ versions to get
additional features.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Duser.country=US -Dkotlin.daemon.jvm.options=-Xmx2400m -Dfil
 
 awsSdkVersion=2.17.100
 jacksonVersion=2.13.0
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #204** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/204" target="_blank">Graphite</a>  (this PR)

PR #205, opened 2022-03-09
Remove workaround for old jOOQ bug

sgrimm:
Commit a23eae83 added a workaround for https://github.com/jOOQ/jOOQ/issues/12704
which was causing queries to fail when the same column was included twice in the
SELECT clause. The bug is fixed in jOOQ 3.16, which we're now using. Remove the
workaround, but keep the test case that was added as part of that same commit.


sgrimm:
* main
  * **PR #205** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/205" target="_blank">Graphite</a>  (this PR)

PR #206, opened 2022-03-11
TRACK-818 Replace whitelabel error page with redirect

sgrimm:
Previously, if the user navigated their browser to a URL that got routed to the
server (for example, an API endpoint) and the request failed, the server returned
the generic Spring Boot "Whitelabel Error Page."

Replace this with a redirect to the web app. For most errors, the server will
redirect to `/error` on the web app, and will include an error message in the
query string; the frontend code can consume that and show the message to the user.

As a special case, if the OAuth2 redirect endpoint `/sso/login` returns an HTTP
400 (Bad Request) response, the server will redirect to the web app home page.
That response from that endpoint usually means the user tried to navigate to a
login link that included an expired OAuth2 session identifier, in which case we
want them to see the web app (if they are already logged in) or get redirected to
the login page.


sgrimm:
* main
  * **PR #206** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/206" target="_blank">Graphite</a>  (this PR)

PR #207, opened 2022-03-14
Be more specific about timestamp format in JSON payloads

sgrimm:
If the client searches on a timestamp field and passes a malformed value,
return an error message that's more specific about the correct format: we want
RFC 3339 timestamps, not arbitrary ISO-8601 timestamps. ISO-8601 allows a huge
range of time formats most of which don't make sense as search terms.


sgrimm:
* main
  * **PR #207** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/207" target="_blank">Graphite</a>  (this PR)

PR #208, opened 2022-03-15
TRACK-838 Use cluster-aware job scheduler

sgrimm:
Currently, in our AWS environments, the daily tasks (accession state transitions,
etc.) run separately on each server instance. There is locking in place to prevent
the same task from running concurrently on two instances, but it still causes the
system to do unnecessary work.

Replace the daily task runner's background thread with a third-party job scheduler
library, JobRunr. The library automatically ensures that only one instance runs
each scheduled occurrence of a job.

The more sophisticated scheduler will also be useful for other periodic tasks such
as scanning for idle device managers.

This change doesn't modify any of the logic of the daily tasks themselves,
including the logic that figures out which time ranges to scan; it just replaces
the mechanism by which the tasks are launched.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -93,6 +93,7 @@ dependencies {
   implementation("org.apache.tika:tika-core:2.1.0")
   implementation("org.codehaus.janino:janino:3.1.6")
   implementation("org.flywaydb:flyway-core:8.2.2")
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #208** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/208" target="_blank">Graphite</a>  (this PR)

PR #209, opened 2022-03-16
Add test helper method to insert into devices table

sgrimm:

PR #210, opened 2022-03-16
TRACK-838 Send email when device manager goes idle

sgrimm:
Add logic to send an email alert to the facility's alert recipients when too much
time has passed since the most recent batch of metrics for the facility's devices.

This works by keeping track of the most recent data from each facility. There
is a facility-level setting (available via the admin UI) to control how much
time is allowed to pass between updates from the device manager before a
facility is considered idle. When more than that amount of time has passed, the
facility is marked as idle and email is sent.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
@@ -4,6 +4,7 @@ import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.model.FacilityModel
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.customer.model.toModel
+import com.terraformation.backend.db.DeviceId
 import com.terraformation.backend.db.FacilityId
 import com.terraformation.backend.db.FacilityNotFoundException
 import com.terraformation.backend.db.FacilityType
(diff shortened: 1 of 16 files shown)


sgrimm:
* main
  * **PR #209** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/209" target="_blank">Graphite</a> 
    * **PR #210** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/210" target="_blank">Graphite</a>  (this PR)

src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
```diff
@@ -200,4 +218,87 @@ class FacilityStore(
...
+        log.info("Found newly idle facilities: $facilityIds")
+
+        dslContext
+            .update(FACILITIES)
+            .setNull(FACILITIES.IDLE_AFTER_TIME)
+            .set(FACILITIES.IDLE_SINCE_TIME, clock.instant())
+            .where(FACILITIES.ID.`in`(facilityIds))
+            .execute()
+
+        func(facilityIds)
```
kallen07, line 300: I learned something reviewing this code, thanks!

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
```diff
@@ -99,6 +103,8 @@ class TimeseriesController(private val timeSeriesStore: TimeseriesStore) {
       }
     }
 
+    facilityStore.updateLastTimeseriesTimes(payload.timeseries.map { it.deviceId })
```
kallen07, line 106: As the code is written, will we update the last timeseries time even if there were errors when adding data to a specific device?

sgrimm: Yes, that's the idea -- in that case, the device manager is still awake and reporting in, so flagging it as idle wouldn't be accurate.

PR #211, opened 2022-03-17
Reduce log spew from daily accession tasks

sgrimm:
The daily tasks that move accessions to new states and generate notifications
about due dates were implemented when we were planning to have the server run
locally in each seed bank. When we moved to a centralized model, we updated the
logic to treat each seed bank facility as its own little universe so there
wouldn't be any potential behavior changes.

Unfortunately, that means we started generating a separate set of log messages
for each facility, whether or not there was actually any work to do. It's now at a
point where these messages account for a significant percentage of the server's
log volume.

Improve the situation a few ways:

* Make some of the daily tasks operate on the entire set of accessions rather than
  looping through each facility one by one. This will have the added bonus of
  making the jobs run faster since there'll be far fewer database queries needed.
  The state summary task remains per-facility since it needs to generate
  statistics on a per-facility basis anyway.
* Remove some uninteresting log messages like "No notifications generated."
* Fold some query timing messages together.


sgrimm:
* main
  * **PR #211** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/211" target="_blank">Graphite</a>  (this PR)

PR #212, opened 2022-03-17
TRACK-654 Remove obsolete filters field from search API

sgrimm:
The initial version of the search API had a less sophisticated way to specify
search criteria. We kept it around for backward compatibility since it was
still being used in the web app. The last use of it has now been removed from
the frontend code: it was only being used by the accession number search bar,
which was removed from the UI. Get rid of it on the server side.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3872,10 +3872,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/SearchSortOrderElement'
-        filters:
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #212** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/212" target="_blank">Graphite</a>  (this PR)

PR #215, opened 2022-03-23
Fix missing update of germination test totals

sgrimm:
Editing the list of germinations on an existing germination test wasn't updating
the germination test's total seeds germinated or total percent germinated. (The
accession-level percentage was being updated correctly.)

In addition, though this wasn't client-visible, the updated `GerminationTestModel`
returned after inserting a new test didn't include those calculated totals, though
they were correctly inserted into the database.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
@@ -610,11 +610,12 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     val initial = createAndUpdate {
       it.copy(
           processingMethod = ProcessingMethod.Count,
-          initialQuantity = seeds(100),
+          initialQuantity = seeds(2000),
           germinationTests =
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #215** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/215" target="_blank">Graphite</a>  (this PR)

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -610,11 +610,12 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     val initial = createAndUpdate {
       it.copy(
           processingMethod = ProcessingMethod.Count,
-          initialQuantity = seeds(100),
+          initialQuantity = seeds(2000),
```
sgrimm, line 613: Changed the quantities here and in `seedsSown` so the "total seeds" and "total percentage" numbers wouldn't be identical (this test wouldn't have caught a bug where the code was swapping the two values).

PR #257, opened 2022-03-30
Don't retry failing scans for idle facilities

sgrimm:
By default, JobRunr retries failed jobs up to 10 times with an exponential
backoff. While awaiting a retry, the job is considered in progress. That's fine
for most jobs but for the once-a-minute scan for idle facilities, the exponential
backoff quickly grows to more than a minute, meaning the job potentially won't
get retried for a significant amount of time even after the underlying problem is
fixed.

Instead, set the job to never retry on failure; it will run again a minute later.


sgrimm:
* main
  * **PR #257** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/257" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/257?utm_source=stack-comment).

PR #258, opened 2022-03-30
Render email templates using FreeMarker

sgrimm:
Replace the not-really-a-template-system string replacement in `EmailService` with
Apache FreeMarker. This will give us more templating flexibility as we introduce
additional kinds of email notifications.

The templates for the plaintext email bodies now have `.txt.ftl` extensions so
that editors will detect them as plaintext FreeMarker templates and give us syntax
highlighting and so forth. The templates for HTML email bodies, which are still
generated from MJML, now have `.ftlh` extensions which causes FreeMarker to treat
them as HTML and automatically enable escaping of HTML special characters.


sgrimm:
* main
  * **PR #258** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/258" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/258?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -133,26 +128,22 @@ class EmailService(
...
+    val model =
         mapOf(
-            "\${admin.email}" to admin.email,
-            "\${admin.fullName}" to (admin.fullName ?: ""),
-            "\${organization.name}" to organization.name,
-            "\${organizationHomeUrl}" to organizationHomeUrl,
+            "admin" to admin,
+            "organization" to organization,
+            "organizationHomeUrl" to organizationHomeUrl,
+            "webAppUrl" to webAppUrl,
```
kallen07, line 136: Why are we passing the webAppUrl now?

sgrimm: It is used to construct the URLs of the image files (company logo, social media icons).

PR #268, opened 2022-04-05
TRACK-881 /api/v1/users/me should return user ID

sgrimm:
The client needs to be able to call APIs that require a user ID as input,
but currently, if the user isn't an admin, there's no way to find out their
ID. Add an `id` field to the response payload of `/api/v1/users/me`.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -6137,8 +6137,14 @@ components:
     UserProfilePayload:
       required:
       - email
+      - id
       type: object
       properties:
+        id:
+          type: integer
+          description: "User's unique ID. This should not be shown to the user, but\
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #268** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/268" target="_blank">Graphite</a>  (this PR)
    * **PR #273** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/273" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/268?utm_source=stack-comment).

PR #273, opened 2022-04-06
SW-885 Allow users to leave organizations and projects

sgrimm:
Currently, only admins can remove users from organizations, which means there's no
way for a non-admin user to leave an organization.

Update the permission checks so that people always have permission to remove
themselves from organizations and projects they're in, but only admins can remove
people other than themselves.


sgrimm:
* main
  * **PR #273** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/273" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/273?utm_source=stack-comment).

PR #277, opened 2022-04-14
SW-882 Import GBIF backbone data

sgrimm:
To support autocompletion and validation of species names, we need to have an
authoritative source of taxonomic data. We've chosen the GBIF backbone data.

GBIF distributes its data in the form of a zipfile that contains a number of
tab-delimited text files:

https://hosted-datasets.gbif.org/datasets/backbone/current/backbone.zip

Add an admin endpoint to read the relevant files and populate a set of new
database tables with the relevant information. The files can be on the local
host (in which case we'd pass a `file:` URL to the endpoint) or the S3 bucket
where we keep other data such as photos (in which case we'd pass an `s3:` URL).

The new endpoint is restricted to users who are marked as "super admins" in the
user database, since this is a systemwide operation.

It's likely we'll need to iterate on the exact structure here as we build out
the user interface for species management, but this initial version supports
some key operations we know we'll need:

* Looking up family names
* Listing common (vernacular) names for a species
* Searching for species by word prefixes on either common or scientific names
  (for typeaheads)
* Looking up conservation status, though the GBIF data only includes this
  information for about 10% of listed species
* Mapping outdated names to the currently accepted names


sgrimm:
* main
  * **PR #277** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/277" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/277?utm_source=stack-comment).

PR #280, opened 2022-04-14
SW-884 Remove manager role from the system

sgrimm:
We removed the manager role from the user interface, but it was still
supported by the backend code. Upgrade all existing manager users to
admins (there aren't any managers other than Terraformation employees)
and remove the role from the code.

PR #279, opened 2022-04-14
SW-893 Remove tracking features

sgrimm:
We aren't going to ship our original concept of the plant tracking features.
Remove its implementation from the code base.

Nearly all of this is just straightforward deletion of code. One minor
exception is that this _adds_ a previously-missing field to `SitesTable` so we
can continue to test that the search code returns geometry values in the
correct format.

PR #278, opened 2022-04-14
SW-897 Add user count to organization API payload

sgrimm:
When fetching information about organizations, include the total number of users
since the web app needs to display it as part of the user's list of organizations.

PR #281, opened 2022-04-14
SW-899 Add user setting to enable email notifications

sgrimm:
Add a per-user setting to control whether or not the system should send
notifications via email. It defaults to disabled (that is, email notifications
are opt-in).

Nothing makes use of this setting yet; this change just makes the setting
available to the client.


sgrimm:
* main
  * **PR #277** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/277" target="_blank">Graphite</a> 
    * **PR #278** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/278" target="_blank">Graphite</a> 
      * **PR #279** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/279" target="_blank">Graphite</a> 
        * **PR #280** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/280" target="_blank">Graphite</a> 
          * **PR #281** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/281" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/281?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
```diff
@@ -53,13 +63,26 @@ data class UserProfilePayload(
...
 data class UpdateUserRequestPayload(
+    @Schema(
+        description =
+            "If true, the user wants to receive all the notifications for their organization and " +
+                "projects via email. This does not apply to certain kinds of notifications such " +
+                "as \"You've been added to a new organization.\" If null, leave the existing " +
+                "value as-is.")
+    val emailNotificationsEnabled: Boolean? = null,
     val firstName: String,
     val lastName: String,
```
kallen07: Do we want to make first and last name nullable as well to match the pattern being used with `emailNotificationsEnabled`? That would allow a client to set the emailNotificationsEnabled without needing to update the user's name.

sgrimm: That's a good question -- the reason I did it here wasn't so much to give the client the option to not set the flag, but for backward compatibility since existing clients won't supply a value for this at all.

I'll add a note about that, but we should maybe have a broader discussion about whether we want to allow a "modify specific fields of an object" API style rather than just the current "completely replace the object" one. (PUT vs. PATCH, to put it in HTTP verb terms.)

PR #285, opened 2022-04-18
SW-626 Remove FacilityIdConverter from codebase

karthikbtf:
- No longer used to convert config option to facility id. The config option is gone since use-case of running server within a facility (and hence the config option to id the facility) is no longer valid.

- Ran server after change, was able to run Web-app against server and load page views, create data, etc.

- Ran tests `gradlew check` successfully


karthikbtf:
> Looks good! Hope it wasn't so trivial as to be obnoxious.

Absolutely not. This was a good starter task :)

PR #286, opened 2022-04-18
SW-781 Refactor PermissionRequirementsTest (WIP)

karthikbtf:
- WIP, wanted to get feedback on direction/approach
- refactored 'createXYZ' tests with a helper function
- not sure if the usage is friendly enough at this point


karthikbtf:
Definitely cleaner with lambdas. Thanks for the example.
This was a nice learning experience for me, especially Kotlin inline capability.

PR #290, opened 2022-04-19
SW-856 Use application events to trigger notifications

sgrimm:
Replace direct calls to `EmailService` with event listeners. The application
code that detects situations that should result in notifications can just
publish an event and doesn't have to know whether that event will result in an
in-app notification, an email notification, or whatever.


sgrimm:
* main
  * **PR #290** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/290" target="_blank">Graphite</a>  (this PR)
    * **PR #292** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/292" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/290?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/event/FacilityAlertRequestedEvent.kt
```diff
@@ -0,0 +1,11 @@
+package com.terraformation.backend.customer.event
+
+import com.terraformation.backend.db.FacilityId
+import com.terraformation.backend.db.UserId
+
+data class FacilityAlertRequestedEvent(
```
kallen07, line 11: Why do we call this `FacilityAlertRequestedEvent` instead of `FacilityAlertEvent`?

sgrimm: Added a comment to clarify what "requested" refers to.

PR #291, opened 2022-04-20
SW-663 Refactor how we identify sequences to clear when running tests

karthikbtf:
Tests currently reset sequences in order to have deterministic ids (or whatever else field that is incremented by a sequence) - this helps write test that check for exact values of ids, etc.
When resetting sequences before test runs, replace hardcoded sequence names with table objects instead.
Use these table objects to figure out relevant sequences to clear.
Current approach is to find sequences based off the table's primary key columns.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
@@ -147,8 +147,8 @@ abstract class DatabaseTest {
                         SQLDataType.VARCHAR,
                         DSL.value(table.name),
                         DSL.value(field.name)))
-                .fetchOne(PG_GET_SERIAL_SEQUENCE)!!
-            .toString()
+                .fetchOne()!!
+            .value1()
       }
 
   /**


src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -106,13 +117,49 @@ abstract class DatabaseTest {
   protected val sequencesToReset: List<String>
     get() = emptyList()
```
karthikbtf: I need to remove this once I figure out how to handle "accession_number_seq"

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -90,16 +98,13 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
   override val sequencesToReset
     get() =
         listOf(
-            "accession_id_seq",
             "accession_number_seq",
```
karthikbtf: need to identify if this is used... or which table/column uses this.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: perhaps leave this as-is for now, for those sequences that can't be inferred from the table

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
karthikbtf: sequence is used directly in code base

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
sgrimm: Not sure why I didn't do it this way to start with, but I wonder if it would make sense to change this to `List<Sequence>` and use the jOOQ sequence objects like `ACCESSION_NUMBER_SEQ`. That'll at least protect against typos. (Beware that there is a standard Kotlin library type called `Sequence`, but I don't think it's referenced in `DatabaseTest` anywhere, so just importing the jOOQ one should be fine.)

karthikbtf: Good idea, I will make that change.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
+        null
+      }
     }
   }
 
+  @BeforeEach
+  fun resetSequencesForTables() {
+    tablesToResetSequences
+        .map { table -> getSerialSequenceNames(table) }
+        .flatten()
```
sgrimm: Could use `flatMap` here.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
+
+  fun getSerialSequenceNames(table: Table<out Record>): List<String> {
+    return table.primaryKey!!.fields.mapNotNull { field ->
+      try {
+        dslContext
+                .select(
+                    DSL.function(
+                        PG_GET_SERIAL_SEQUENCE,
+                        SQLDataType.VARCHAR,
+                        DSL.field("{0}, {1}", table.name, field.name)))
```
sgrimm: Does it work to split this into two scalar arguments? That is, instead of `DSL.field("{0}, {1}", table.name, field.name)`, use `DSL.value(table.name), DSL.value(field.name)`? If that works, it might make it clearer that the intent is to pass two arguments to the function. Right now it sort of looks like it's passing a single argument that has a comma in the middle, though it makes sense when you think about the fact that this is representing raw SQL.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
+                .fetchOne(PG_GET_SERIAL_SEQUENCE)!!
+            .toString()
+      } catch (e: Exception) {
+        println(e)
```
sgrimm: This is probably just for debugging as you work on the change, but I think it's actually fine to not catch the exception here at all; it'll bubble up and make the test fail, which I think is what we'd want. (Or does that break existing tests?)

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
   }
 
+  @BeforeEach
+  fun resetSequencesForTables() {
+    tablesToResetSequences
+        .map { table -> getSerialSequenceNames(table) }
+        .flatten()
+        .toSet()
+        .forEach { sequenceName ->
+          dslContext.alterSequence(DSL.unquotedName(sequenceName)).restart().execute()
```
sgrimm: The `DSL.unquotedName()` is because the function returns fully-qualified names, I suppose?

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
+    return table.primaryKey!!.fields.mapNotNull { field ->
+      try {
+        dslContext
+                .select(
+                    DSL.function(
+                        PG_GET_SERIAL_SEQUENCE,
+                        SQLDataType.VARCHAR,
+                        DSL.field("{0}, {1}", table.name, field.name)))
+                .fetchOne(PG_GET_SERIAL_SEQUENCE)!!
+            .toString()
```
sgrimm: If you wanted to avoid the `toString()` here, this could be `.fetchOne()!!.value1()` -- the type signature at this point in the query includes the fact that the first column of each result row is a string thanks to the `SQLDataType.VARCHAR`, so `value1()` will be known to be of type `String?`.

The way you have it is fine too; consider this an FYI more than a suggestion.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -103,16 +115,50 @@ abstract class DatabaseTest {
...
     get() = emptyList()
 
   @BeforeEach
   fun resetSequences() {
-    sequencesToReset.forEach { sequenceName ->
-      dslContext.alterSequence(sequenceName).restart().execute()
+    sequencesToReset.forEach { sequence -> dslContext.alterSequence(sequence).restart().execute() }
+  }
+
+  fun getSerialSequenceNames(table: Table<out Record>): List<String> {
```
sgrimm: This function can be private, I think. Maybe put it under both of the `@BeforeEach` methods just so those are all in one place at the top of the file for easier skimming.

karthikbtf:
Thanks for the review, Steve. I will follow up.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: Correct, prefixed with schema. For example: public.accession_id_seq
The alter sequence command is executed as
`alter "public.accession_id_seq"... ` which errors out as not found. I was going to strip the schema prefix, but that seemed hacky to me. Guessing unquoting is ok here since it is a test and not production code?

karthikbtf: `alter public.accession_id_seq..` for instance works, so does `alter "accession_id_seq"...`

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: Does not break any tests, just old habits. I will remove this and let the test fail if there's an error, that is cleaner.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: Just tried it, works great. Will make that change.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: I couldn't get this to work, let me dig into this more and understand the type signature

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -130,35 +130,29 @@ abstract class DatabaseTest {
...
-    }
-  }
-
   @BeforeEach
   fun resetSequencesForTables() {
     tablesToResetSequences
         .map { table -> getSerialSequenceNames(table) }
-        .flatten()
-        .toSet()
+        .flatMap { it.toSet() }
```
sgrimm: Oops, sorry, I meant you could replace `.map().flatten().toSet()` with `.flatMap().toSet()`. My comment was ambiguous!

Not that there's likely to be any overlap no matter what, but doing `.flatMap { it.toSet() }` would potentially leave you with duplicate elements, if the same sequence somehow happened to be used for two tables and you put both of them in the table list; each table's list of sequences would be unique thanks to the `toSet()`, but the sets would be concatenated together into a list by `flatMap()`.

karthikbtf: Makes sense, completely missed the flatMap opportunity.

karthikbtf:
> This is looking good now. I'm happy to merge this as is, but you mentioned wanting to look at the type signature of the `pg_get_serial_sequence` query -- if you still want to do that, I'll hold off.

I was able to get it to work after switching from `fetchOne(<field name>)` which returns `Object`.. to just `fetchOne()` which returns `Record` with support for `value1()`.
In our use-case, there is no need to specify the field name since the response has just one field anyways.
Will submit this change shortly.

PR #292, opened 2022-04-21
SW-856 Move email header/footer to include files

sgrimm:
To avoid having to include the boilerplate header and footer content in every
email template, move them to separate MJML files that get included by the
MJML files for individual notification types.

Since the header and footer need to know the base URL of the web app (because
they use it to construct URLs of the company logo and other images), move it
to `EmailTemplateModel` and turn that from a marker interface to a class.


sgrimm:
* main
  * **PR #292** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/292" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/292?utm_source=stack-comment).

PR #289, opened 2022-04-19
SW-856 Unify notification template rendering

sgrimm:
Refactor `EmailService` to use a set of generic methods to send notifications to
a particular target audience using the contents of a particular template directory.
This will be less repetitive and error-prone than constructing each notification
message completely from scratch.

As part of this, subject lines are now rendered from templates rather than treated
as localizable messages.

PR #287, opened 2022-04-18
SW-856 Replace facility alert recipients with opted-in project members

sgrimm:
The `facility_alert_recipients` table was a temporary placeholder so we could
alert about device problems at our existing sites while we figured out how we
wanted the notifications system to work.

Now that we have a firmer grasp of how our initial notifications setup will
figure out which users should receive which messages, we can start broadcasting
facility alerts using the same setup we'll be using for other kinds of
notifications.

Specifically, we will send out facility alerts to all members of the facility's
project and to all organization admins, but only if they have email
notifications enabled.

To transition from the placeholder implementation, all users with email
addresses that were previously listed as alert recipients are automatically
opted into email notifications. Facility alerts were only enabled for a couple
of Terraformation sites, so this won't impact external customers at all.

PR #293, opened 2022-04-22
SW-916 Support search filters on compound primary keys

sgrimm:
Previously, the search table definitions for the `organization_users` and
`project_users` tables treated user ID as the primary key. But that caused a
problem: since search criteria are implemented using an `IN` clause that
selects the top-level objects by primary key, the results weren't being filtered
by organization, just by user.

So for example, if you searched for organization members with a role of "Admin"
in a particular organization ID, you would only get search results for users who
had the admin role in that organization, but you would get the membership details
for _all_ their organizations, not just the one you specified.

The fix is to not ignore the fact that organization users (and project users) have
compound primary keys.

Luckily, jOOQ has a feature called "embeddable types" that lets us solve this
without having to revamp our search implementation: you can define synthetic data
types that combine multiple columns into a single field that acts, from the
application's point of view, like a single atomic value. Configuring the search
tables to use these embeddable types as the primary key, rather than the user ID,
causes the filtering to be applied to both the organization and user at the same
time, and the queries return the expected results.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -253,12 +253,14 @@ tasks {
         withIncludes(".*")
         withExcludes(generator.excludes())
         withForcedTypes(generator.forcedTypes(basePackageName))
+        withEmbeddables(generator.embeddables())
         // Fix compiler warnings for PostGIS functions; see https://github.com/jOOQ/jOOQ/issues/8587
         withTableValuedFunctions(false)
       }
 
       generate.apply {
         isDaos = true
+        isEmbeddables = true
         isJavaTimeTypes = true
         isPojos = true
         isPojosAsKotlinDataClasses = true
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #287** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/287" target="_blank">Graphite</a> 
    * **PR #289** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/289" target="_blank">Graphite</a> 
      * **PR #290** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/290" target="_blank">Graphite</a> 
        * **PR #292** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/292" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/293?utm_source=stack-comment).

PR #294, opened 2022-04-22
SW-911 Add endpoint to list organization roles

sgrimm:
Add an endpoint to list all the available roles in an organization and how many
users have each role.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
@@ -109,7 +109,7 @@ class OrganizationsController(
   fun listOrganizationRoles(
       @PathVariable("organizationId") organizationId: OrganizationId
   ): ListOrganizationRolesResponsePayload {
-    val roleCounts = mapOf(Role.ADMIN to 0)
+    val roleCounts = organizationStore.countRoleUsers(organizationId)
     return ListOrganizationRolesResponsePayload(
         roleCounts.map { (role, count) -> OrganizationRolePayload(role, count) })
   }


sgrimm:
* main
  * **PR #294** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/294" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/294?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -104,6 +104,16 @@ class OrganizationsController(
     return SimpleSuccessResponsePayload()
   }
 
+  @Operation(summary = "Lists the roles in an organization.")
+  @GetMapping("/{organizationId}/roles")
+  fun listOrganizationRoles(
+      @PathVariable("organizationId") organizationId: OrganizationId
+  ): ListOrganizationRolesResponsePayload {
+    val roleCounts = mapOf(Role.ADMIN to 0)
```
karthikbtf: to be replaced with `organizationStore.countRoleUsers`?

sgrimm: Haha, wow, that's embarrassing. You can tell I was testing this with the unit test, not with actual requests.

karthikbtf: All good, the api delegation layer is very clear and will obviously work, the unit test coverage is great.

openapi.yaml
```diff
@@ -496,6 +496,26 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/SimpleErrorResponsePayload'
+  /api/v1/organizations/{organizationId}/roles:
```
karthikbtf, line 499: Just for my understanding.. are these manual updates?

sgrimm: We generate `openapi.yaml` from annotations in the code, but we need to run the build command manually to do it. (It's `./gradlew generateOpenApiDocs`.) There is a CI check to make sure the file that's checked into the repo is up to date.

Arguably this shouldn't be considered a source file at all since it gets generated on demand if you hit the OpenAPI schema endpoint on the server; the code is the real source of truth. I decided to check it into the repo just because it turned out to be handy for the frontend team to take a peek at a pending PR and see the API changes without having to pull the PR branch and build the server locally.

PR #301, opened 2022-04-26
SW-929 Add endpoint to delete organization

sgrimm:
Allow the owner to delete an organization if there aren't any other users.

Currently, this just removes the owner from the organization, making it
inaccessible. A subsequent change will add a background job to purge the
organization's data from the database.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
@@ -6,11 +6,13 @@ import com.terraformation.backend.customer.db.OrganizationStore
 import com.terraformation.backend.customer.db.ProjectStore
 import com.terraformation.backend.customer.db.SiteStore
 import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.event.OrganizationDeletedEvent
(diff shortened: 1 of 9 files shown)


sgrimm:
* main
  * **PR #301** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/301" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/301?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
```diff
@@ -95,4 +100,22 @@ class OrganizationService(
...
+      if (users.size != 1 || users[0].userId != currentUser().userId) {
+        throw OrganizationHasOtherUsersException(organizationId)
+      }
+
+      organizationStore.removeUser(
+          organizationId, currentUser().userId, allowRemovingLastOwner = true)
+
+      log.info("Deleted last owner from organization $organizationId")
+
+      publisher.publishEvent(OrganizationDeletedEvent(organizationId))
```
sgrimm, line 118: The idea, in case it's not self-evident, is that this event will trigger the logic to purge the organization's data from the system.

PR #304, opened 2022-04-27
support additional columns in enum table generator

karthikbtf:
Support an optional list of column names for the enum table generation. Assumption is the column values are of type string.

Got this working on my dev branch for notifications.
For a table like:
```
CREATE TYPE notification_criticality AS ENUM ('Info', 'Warning', 'Error', 'Success');

CREATE TABLE notification_types (
  id INTEGER PRIMARY KEY,
  name TEXT NOT NULL,
  criticality notification_criticality DEFAULT 'Info'
);
```
and some dummy values inserted, the enum generation was

```
enum class NotificationType(
    override val id: Int,
    @get:JsonValue override val displayName: String,
    val criticality: String
) : EnumFromReferenceTable<NotificationType> {
    UserAddedtoOrganization(1, "User Added to Organization", "Info"),
    FacilityIdle(2, "Facility Idle", "Warning"),
    FacilityAlertRequested(3, "Facility Alert Requested", "Error");
    
    override val tableName get() = "notification_types"

    companion object {
        private val byDisplayName = values().associateBy { it.displayName }
        private val byId = values().associateBy { it.id }
        
        @JsonCreator
        @JvmStatic
        fun forDisplayName(name: String) = byDisplayName[name]
            ?: throw IllegalArgumentException("Unrecognized value: $name")
        
        fun forId(id: Int) = byId[id]
    }
}

```
I was able to use this class successfully for jooq notification reads and writes, worked very clean/smooth.

Note: the additional properties are not overridden from the base EnumFromReferenceTable<>, they are data members specific to the derived Enum table.

### DIFF
--- a/buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
+++ b/buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
@@ -38,8 +38,7 @@ class TerrawareGenerator : KotlinGenerator() {
         val tableName: String
       }
 
-    """.trimIndent()
-    )
+    """.trimIndent())
 
     ENUM_TABLES.forEach { printEnum(out, it, schema.database.connection) }
     ID_WRAPPERS.forEach { it.render(out) }
@@ -122,15 +121,13 @@ class TerrawareGenerator : KotlinGenerator() {
           override fun to(enumValue: $enumName?) = enumValue?.id
       }
 
-    """.trimIndent(),
-    )
+    """.trimIndent())
   }
 
   override fun generateTable(schema: SchemaDefinition, table: TableDefinition) {
     if (ENUM_TABLES.any { it.toString() == table.name }) {
       throw IllegalArgumentException(
-          "${table.name} is generated as an enum and must be excluded from the table list"
-      )
+          "${table.name} is generated as an enum and must be excluded from the table list")
     }
 
     super.generateTable(schema, table)
@@ -149,8 +146,7 @@ class TerrawareGenerator : KotlinGenerator() {
             ForcedType()
                 .withIncludeExpression("(?i:.*_ur[li])")
                 .withConverter("com.terraformation.backend.db.UriConverter")
-                .withUserType("java.net.URI")
-        )
+                .withUserType("java.net.URI"))
 
     ENUM_TABLES.forEach { types.add(it.forcedType(targetPackage)) }
     ID_WRAPPERS.forEach { types.add(it.forcedType(targetPackage)) }


buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -74,11 +82,17 @@ class TerrawareGenerator : KotlinGenerator() {
     // https://youtrack.jetbrains.com/issue/KT-2425
     val dollarSign = '$'
 
+    val additionalProperties = table.additionalColumns.map { p -> "val $p: String" }
```
karthikbtf: this should be converted to camelCase perhaps

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/EnumTable.kt
```diff
@@ -10,13 +10,20 @@ class EnumTable(
      * This allows the generated code for those tables to replace their raw ID value with our enum.
      */
     includeExpressions: List<String>,
-    val enumName: String = tableName.trimEnd('s').toPascalCase()
+    val enumName: String = tableName.trimEnd('s').toPascalCase(),
+    val additionalColumns: List<String> = emptyList()
```
sgrimm: May as well put a trailing comma on line 14 too. (Earlier versions of Kotlin didn't allow trailing commas and I still haven't trained myself to add them on multi-line argument lists like this.)

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -51,14 +51,22 @@ class TerrawareGenerator : KotlinGenerator() {
 
     log.info("Generating enum for reference table $table")
 
-    connection.prepareStatement("SELECT id, name FROM $table ORDER BY id").use { ps ->
+    val columns = listOf("id", "name", *(table.additionalColumns.toTypedArray())).joinToString()
```
sgrimm: A more idiomatic way to concatenate lists is to use the `+` operator, like `(listOf("id", "name") + table.additionalColumns).joinToString()`.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -51,14 +51,22 @@ class TerrawareGenerator : KotlinGenerator() {
...
             val capitalizedName = name.replace(Regex("[-/ ]"), "").capitalize()
-            values.add("$capitalizedName($id, \"$name\")")
+            if (table.additionalColumns.isEmpty()) {
+              values.add("$capitalizedName($id, \"$name\")")
+            } else {
+              val properties = table.additionalColumns.mapIndexed { i, _ ->
+                "\"${rs.getString(2 + i + 1)}\""
+              }.joinToString()
+              values.add("$capitalizedName($id, \"$name\", $properties)")
+            }
```
sgrimm: This feels like it could be simplified by making a combined column list that starts with `name`.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
sgrimm: Yeah, agreed.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -74,11 +82,17 @@ class TerrawareGenerator : KotlinGenerator() {
     // https://youtrack.jetbrains.com/issue/KT-2425
     val dollarSign = '$'
 
+    val additionalProperties = table.additionalColumns.map { p -> "val $p: String" }
+    val properties = listOf(
+        "override val id: Int",
+        "@get:JsonValue override val displayName: String",
+        *(additionalProperties.toTypedArray()),
+    ).filter { it.trim().isNotEmpty() }.joinToString(separator = ",\n          ")
```
sgrimm: I'd be inclined to leave the `filter` out here; unless I'm misreading, the only way there could be an empty string in the list is if someone put an empty additional column name in `Config.kt` and it'd be better if that caused a compile failure.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
karthikbtf: name has some special semantics, i can still simplify this without an if/else

karthikbtf:
> Some nitpicks inline, but this seems like the right high-level approach.
> 
> Do we maybe want to have a way of specifying non-string properties? In the initial use case, it seems like it'd be preferable to have the extra property be of type `NotificationCriticality` (which itself could be a generated enum) than `String`. We probably wouldn't need to try to automatically figure the types out; maybe make `additionalColumns` a `List<Pair<String, String>>` and then in the config, you'd do `listOf("criticality" to "NotificationCriticality")` or something.
> 
> Also, though I know it's not part of the actual code change here, in the example you're using `CREATE TYPE`. We've avoided using Postgres enum types in general, both because it isn't portable to other DBs and, more importantly, because enum types are kind of a pain to modify after the fact compared to using foreign key references to lookup tables.

Thanks for the review, Steve. Let me look into adding support for non-string properties.

karthikbtf:
Removed the postgres type.
For testing, I have 2 tables now.

```
CREATE TABLE notification_criticality (
  id INTEGER PRIMARY KEY,
  name TEXT NOT NULL
);

CREATE TABLE notification_types (
  id INTEGER PRIMARY KEY,
  name TEXT NOT NULL,
  notification_criticality_id INTEGER NOT NULL REFERENCES notification_criticality
);

INSERT INTO notification_criticality (id, name)
VALUES (1, 'Info'),
       (2, 'Warning'),
       (3, 'Error'),
       (4, 'Success');

INSERT INTO notification_types (id, name, notification_criticality_id)
VALUES (1, 'User Added to Organization', 1),
       (2, 'Facility Idle', 2),
       (3, 'Facility Alert Requested', 3)
ON CONFLICT (id) DO UPDATE SET name = excluded.name, notification_criticality_id = excluded.notification_criticality_id;

```

Added both notification_criticality and notification_types to the enum table generation list.

```
        EnumTable("notification_criticality", listOf(".*\\.notification_criticality_id")),
        EnumTable(
            "notification_types",
            listOf(".*\\.notification_type_id"),
            additionalColumns = listOf("notification_criticality_id" to "NotificationCriticality"),
``` 

Code generated looks like this:

```enum class NotificationType(
    override val id: Int,
    @get:JsonValue override val displayName: String,
    val notificationCriticalityId: NotificationCriticality
) : EnumFromReferenceTable<NotificationType> {
    UserAddedtoOrganization(1, "User Added to Organization", 1),
    FacilityIdle(2, "Facility Idle", 2),
    FacilityAlertRequested(3, "Facility Alert Requested", 3);
```

which throws a compile error `The integer literal does not conform to the expected type NotificationCriticality`
I need to write this as NotificationCriticality.forId(integer value) instead,,, wondering if I should pass this context (that it is an enum type) in the additionalColumns list.

karthikbtf:
My latest change was tested as follows:

```
CREATE TABLE notification_criticality (
  id INTEGER PRIMARY KEY,
  name TEXT NOT NULL
);

CREATE TABLE notification_types (
  id INTEGER PRIMARY KEY,
  name TEXT NOT NULL,
  testid INTEGER NOT NULL,
  notification_criticality_id INTEGER NOT NULL REFERENCES notification_criticality
);

INSERT INTO notification_criticality (id, name)
VALUES (1, 'Info'),
       (2, 'Warning'),
       (3, 'Error'),
       (4, 'Success');

INSERT INTO notification_types (id, name, testid, notification_criticality_id)
VALUES (1, 'User Added to Organization', 5, 1),
       (2, 'Facility Idle', 4, 2),
       (3, 'Facility Alert Requested', 3, 3)
ON CONFLICT (id) DO UPDATE SET name = excluded.name, notification_criticality_id = excluded.notification_criticality_id;
```

generated code 

```
enum class NotificationType(
    override val id: Int,
    @get:JsonValue override val displayName: String,
    val notificationCriticalityId: NotificationCriticality,
    val testid: Int
) : EnumFromReferenceTable<NotificationType> {
    UserAddedtoOrganization(1, "User Added to Organization", NotificationCriticality.forId(1)!!, 5),
    FacilityIdle(2, "Facility Idle", NotificationCriticality.forId(2)!!, 4),
    FacilityAlertRequested(3, "Facility Alert Requested", NotificationCriticality.forId(3)!!, 3);
```

Works good, but wondering if it is a bit hacky.
One option is to stick with standard data types, have client code reinterpret nested enum values (which may not be that often).

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -51,22 +52,22 @@ class TerrawareGenerator : KotlinGenerator() {
...
             val capitalizedName = name.replace(Regex("[-/ ]"), "").capitalize()
-            if (table.additionalColumns.isEmpty()) {
-              values.add("$capitalizedName($id, \"$name\")")
-            } else {
-              val properties = table.additionalColumns.mapIndexed { i, _ ->
-                "\"${rs.getString(2 + i + 1)}\""
-              }.joinToString()
-              values.add("$capitalizedName($id, \"$name\", $properties)")
-            }
+            val properties = (listOf("\"$name\"") + table.additionalColumns.mapIndexed { i, it ->
```
sgrimm: It'd be better to replace `it` with a more descriptive name; typically you only want to use `it` for one-line lambdas.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -51,22 +52,22 @@ class TerrawareGenerator : KotlinGenerator() {
...
+              val obj = rs.getObject(2 + i + 1)
+              when (obj) {
+                is String -> "\"$obj\""
+                else -> if (it.isTableEnum) "${it.columnDataType}.forId($obj)!!" else "$obj"
+              }
+            }).joinToString()
```
sgrimm: This might be clearer as three branches in the `when` rather than a nested `when`/`if`:

```kotlin
when {
  obj is String -> "\"$obj\""
  variableFormerlyKnownAsIt.isTableEnum -> ...
  else -> "$obj"
}
```

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -116,13 +120,15 @@ class TerrawareGenerator : KotlinGenerator() {
...
-    """.trimIndent())
+    """.trimIndent(),
+    )
   }
 
   override fun generateTable(schema: SchemaDefinition, table: TableDefinition) {
     if (ENUM_TABLES.any { it.toString() == table.name }) {
       throw IllegalArgumentException(
-          "${table.name} is generated as an enum and must be excluded from the table list")
+          "${table.name} is generated as an enum and must be excluded from the table list",
```
sgrimm: SUPER DUPER NITPICKY and I know this is an annoying kind of review comment, but I'd leave off the trailing comma here and on line 150 (and I guess on line 41 too). The main reason I like to add it in places like argument lists is just so that if I need to add another argument, or reorder the arguments, I don't have to edit the last item differently from the other ones. But in this case, there's never going to be a second argument.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
karthikbtf: Will do, I agree.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
karthikbtf: This is definitely clearer, thanks much.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
karthikbtf: This change was auto generated, not sure at what point this happened. will remove this change....
I tend to 'reformat code' in IntelliJ.. perhaps that did it

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -127,7 +147,8 @@ class TerrawareGenerator : KotlinGenerator() {
             ForcedType()
                 .withIncludeExpression("(?i:.*_ur[li])")
                 .withConverter("com.terraformation.backend.db.UriConverter")
-                .withUserType("java.net.URI"))
+                .withUserType("java.net.URI"),
```
karthikbtf: same here

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
sgrimm: Ah, that makes sense. You'll want to install the "ktfmt" plugin and restart IntelliJ. It will override IntelliJ's built-in formatter with the same one we use in the CI builds. (Annoyingly, installing it doesn't force you to restart the IDE, but until you restart, it doesn't override the default formatter.)

sgrimm: How it relates to this, I should add, is that ktfmt will preserve trailing commas (and use them as a hint that arguments should be split onto separate lines) but never add them.

karthikbtf: Great, thanks. I just installed it and restarted...

PR #305, opened 2022-04-27
Return error payload for missing query parameters

sgrimm:
If an endpoint requires certain parameters in the query string, Spring knows to
respond with HTTP 400 if they're missing from a request. But the default response
doesn't have a body and doesn't tell the client which parameter was missing.

Add a handler for the exception that gets thrown when a parameter is missing, such
that it returns an error payload in our normal error format with a message that
identifies the missing parameter.


sgrimm:
* main
  * **PR #305** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/305" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/305?utm_source=stack-comment).

PR #307, opened 2022-04-28
SW-857 Add API to support new functionality in notifications

karthikbtf:
Defined new schema for notifications.
Added API to fetch a user's notifications by organization or globally scoped notifications.
Added API to fetch a user's unread notifications count.
Added API to mark a  user's single notification as read or unread.
Added API to mark all notifications for a user (scoped by organization or globally) as read or unread.
Added an AppNotificationService whose purpose is to subscribe to relevant events that require publishing notifications. Currently subscribes to an event for a user being added to an organization, and creates an appropriate notification for that user.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
@@ -48,7 +48,7 @@ class AppNotificationService(
         organizationStore.fetchById(event.organizationId)
             ?: throw OrganizationNotFoundException(event.organizationId)
 
-    val organizationHomeUrl = webAppUrls.inAppOrganizationHome(event.organizationId).toString()
+    val organizationHomeUrl = webAppUrls.organizationHome(event.organizationId).toString()
     val message = messages.userAddedToOrganizationNotification(organization.name)
 
     log.info(
--- a/src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
+++ b/src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
@@ -12,7 +12,7 @@ import javax.ws.rs.core.UriBuilder
  */
 @ManagedBean
 class WebAppUrls(private val config: TerrawareServerConfig) {
-  fun organizationHome(organizationId: OrganizationId): URI {
+  fun fullOrganizationHome(organizationId: OrganizationId): URI {
     return UriBuilder.fromUri(config.webAppUrl)
         .path("/home")
         .queryParam("organizationId", organizationId)
@@ -20,7 +20,7 @@ class WebAppUrls(private val config: TerrawareServerConfig) {
   }
 
   /** Generates a relative path of organization home within the web app */
-  fun inAppOrganizationHome(organizationId: OrganizationId): URI {
+  fun organizationHome(organizationId: OrganizationId): URI {
     return UriBuilder.fromPath("/home").queryParam("organizationId", organizationId).build()
   }
 }
--- a/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
+++ b/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
@@ -65,5 +65,7 @@ class Messages {
 
   /** Title and body to use for "user added to organization" app notification */
   fun userAddedToOrganizationNotification(orgName: String): NotificationMessage =
-      NotificationMessage(title = "Lorem ipsum dolor sit amet", body = "You have joined $orgName.")
+      NotificationMessage(
+          title = "You've been added to a new organization!",
+          body = "You are now a member of $orgName. Welcome!")
 }
--- a/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
@@ -78,7 +78,7 @@ class EmailNotificationService(
         organizationStore.fetchById(event.organizationId)
             ?: throw OrganizationNotFoundException(event.organizationId)
 
-    val organizationHomeUrl = webAppUrls.organizationHome(event.organizationId).toString()
+    val organizationHomeUrl = webAppUrls.fullOrganizationHome(event.organizationId).toString()
 
     emailService.sendUserNotification(
         user,


buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -31,6 +31,10 @@ val ENUM_TABLES =
                 "germination_tests\\.test_type",
                 "accession_germination_test_types\\.germination_test_type_id")),
         EnumTable("germination_treatments", "germination_tests\\.treatment_id"),
+        EnumTable("notification_criticality", listOf(".*\\.notification_criticality_id")),
```
sgrimm: Minor naming convention thing: We use plural table names, so this should be `notification_criticalities` (and then you'll need to pass in the enum name to depluralize it).

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -0,0 +1,66 @@
...
+  fun on(event: UserAddedToOrganizationEvent) {
+    val admin = userStore.fetchById(event.addedBy) ?: throw UserNotFoundException(event.addedBy)
+    val user = userStore.fetchById(event.userId) ?: throw UserNotFoundException(event.userId)
+    val organization =
+        organizationStore.fetchById(event.organizationId)
+            ?: throw OrganizationNotFoundException(event.organizationId)
+
+    val organizationHomeUrl = webAppUrls.organizationHome(event.organizationId).toString()
+    val title = "You have joined \"${organization.name}\""
+    val body = "You were added to organization \"${organization.name}\" by \"${admin.email}\"."
```
sgrimm: The text for these should be pulled out into functions in the `Messages` class so they'll be easier to localize down the road.

Also, not a comment on your code per se, but we should check with the product team about what content they want in the title vs. the body. Looks like the requirements doc with the list of notifications doesn't break it down to that level of detail.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    log.info("Creating app notification for user being added to an organization.")
```
sgrimm: I'd maybe log the user and organization IDs here.

src/main/kotlin/com/terraformation/backend/customer/model/Notification.kt
```diff
@@ -0,0 +1,51 @@
...
+data class NotificationCountModel(val organizationId: OrganizationId?, val unread: Int)
+
+data class CreateNotificationModel(
+    val notificationType: NotificationType,
+    val userId: UserId,
+    val organizationId: OrganizationId?,
+    val title: String,
+    val body: String,
+    val localUrl: URI
+)
```
sgrimm: Ah, the design conundrum of null-safe ORM in Kotlin. To be honest I still haven't figured out what approach I like best to solve the "fields are required on read, not on write" problem while keeping duplication to a minimum, as you've no doubt noticed while reading through the code. This is a decent approach.

karthikbtf:
Some testing data:

```
MacBook-Pro:terraware-server kb$ curl -X GET -H "Cookie: $COOKIE" "http://localhost:8080/api/v1/notifications" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   337    0   337    0     0    631      0 --:--:-- --:--:-- --:--:--   648
{
  "notifications": [
    {
      "id": 1,
      "notificationType": "User Added to Organization",
      "notificationCriticality": "Info",
      "userId": 6,
      "metadata": {
        "adminEmail": "karthik@terraformation.com",
        "organizationId": 1,
        "organizationName": "my org"
      },
      "localUrl": "http://localhost:3000/home?organizationId=1",
      "createdTime": "2022-04-28T17:18:51.002043Z"
    }
  ],
  "status": "ok"
}

MacBook-Pro:terraware-server kb$ curl -X GET -H "Cookie: $COOKIE" "http://localhost:8080/api/v1/notifications/count" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    68    0    68    0     0    728      0 --:--:-- --:--:-- --:--:--   764
{
  "notifications": [
    {
      "organizationId": null,
      "unread": 1
    }
  ],
  "status": "ok"
}
```

src/main/kotlin/com/terraformation/backend/customer/model/TerrawareUser.kt
```diff
@@ -114,4 +114,7 @@ interface TerrawareUser : Principal {
   fun canUpdateSpeciesName(speciesNameId: SpeciesNameId): Boolean
   fun canUpdateStorageLocation(storageLocationId: StorageLocationId): Boolean
   fun canUpdateTimeseries(deviceId: DeviceId): Boolean
+  fun canReadNotification(): Boolean
+  fun canCreateNotification(): Boolean
+  fun canUpdateNotification(): Boolean
```
sgrimm: These permissions should be scoped, if only because it should be a permission failure to read someone else's notifications. The question then becomes what the scope should be: user, org, or notification?

A couple bits of design context on this stuff:

The goal is to be able to eventually roll out the ability for users to create custom roles for their organizations and pick and choose fairly fine-grained permissions for those roles. That's why there's this big long list of permission checks even though at the moment we only have three roles (and two of them, owner and admin, are nearly identical). When we add custom roles, I don't want to have to scour the code base adding a zillion fine-grained permission checks all at once because it'd be both annoying and mistake-prone.

I made the deliberate design decision to not worry too much about the performance cost of these permission checks. That is, it's fine if each check ends up having to do a separate `SELECT something FROM table WHERE id = X` and we do a couple dozen of them in a single request. My thinking is it's much better to have performance issues than security holes, and we can address performance by caching permission data in memory, which we already do for a few things as you've seen. If we end up doing hundreds of them in a single request, then it's time to optimize (the search code is an example of that).

With that in mind, I think the read and update methods should probably take a notification ID. In theory they could take a user ID but you won't always know the notification's user ID at the point where you need to do the checks. Semantically, "Does the current user have the right to update this particular notification?" seems like a reasonable thing to be able to ask the permission system.

You could add a separate `canListNotifications` permission that takes a user ID.

The right scope for the create permission is a little fuzzier: should it take an organization ID or a user ID or both? I'd be inclined to go with a user ID for now.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
karthikbtf: I was going to follow up with product as a second round of changes. 
Was wondering if maybe these messages should be formulated in the UI based on metadata in the notification. Gives UI flexibility on how to style the messages, etc. My most recent change has the 'metadata' version.
I can roll it back to support title/body, in which case yes I will move the content into `Messages`

src/main/resources/db/migration/common/V84__NotificationsTable.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE notifications (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    notification_type_id INTEGER NOT NULL REFERENCES notification_types,
+    user_id BIGINT NOT NULL REFERENCES users,
+    organization_id BIGINT REFERENCES organizations,
+    local_url TEXT NOT NULL,
+    metadata JSONB NOT NULL,
```
sgrimm: Is the idea that the notification messages will be rendered client-side?

If not, skip to the last paragraph of this comment.

Otherwise, I totally get that there are good reasons to want to do that, but I would argue against that approach, and in favor of what you had in the previous revision of the PR (title/body fields) for a few reasons:

* Adding a new notification type, or changing what information is included in the rendered notifications, now requires a coordinated frontend and backend release.
* We're going to want to consume these notifications on mobile as well, I assume, and we'd either need to have a second rendering implementation in the mobile code or set up a way to share code between the web and mobile apps. Either way, it's an even more complicated release coordination problem, especially given that users don't always upgrade their mobile apps immediately.
* There's no great way to guarantee that the client and server agree on what metadata fields should be present for a particular notification type.

If we _do_ take this approach, we'll probably want to version the metadata so we can deal sanely with existing notifications when we add/remove fields. We could either version it explicitly or make it implicit in the notification type (that is, if we add a field, we add a new notification type rather than modifying the existing one).

karthikbtf: Good points. I will revert this to the title/body approach as they satisfy the design requirements, and avoid unnecessary complexity.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -0,0 +1,124 @@
...
+        .select(NOTIFICATIONS.ORGANIZATION_ID, DSL.count(NOTIFICATIONS.ID))
+        .from(NOTIFICATIONS)
+        .where(NOTIFICATIONS.READ_TIME.isNull)
+        .groupBy(NOTIFICATIONS.ORGANIZATION_ID)
+        .fetch { row -> NotificationCountModel(row.value1(), row.value2()) }
+  }
+
+  fun markRead(
+      notificationId: NotificationId,
+      readTime: Instant?,
```
sgrimm: Is the idea that passing null for `readTime` marks a message as unread?

If so, I think it'd make the intent clearer to either have separate "mark read" and "mark unread" functions in this class's public API (internally, they can call a method that uses a nullable read time) or to turn this into a boolean argument and pull the time from `clock.instant()` if it's marking as read.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -0,0 +1,124 @@
...
+      organizationId: OrganizationId?
```
sgrimm: Is the organization ID needed if you have a notification ID?

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -0,0 +1,124 @@
...
+            .insertInto(
+                NOTIFICATIONS,
+                NOTIFICATION_TYPE_ID,
+                USER_ID,
+                ORGANIZATION_ID,
+                METADATA,
+                LOCAL_URL,
+                CREATED_TIME,
+            )
+            .values(
```
sgrimm: This is totally fine, but FYI, there's an alternate syntax you can use for this if you want:

```kotlin
dslContext
    .insertInto(NOTIFICATIONS)
    .set(USER_ID, userId)
    .set(LOCAL_URL, localUrl)
    // ...
    .execute()
```
jOOQ will turn that into a correct INSERT statement. Like I said, up to you and the way you have it here is fine, just pointing out the other style exists in case you decide you like it better.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -0,0 +1,124 @@
...
+          .where(
+              NOTIFICATIONS
+                  .ID
+                  .eq(notificationId)
+                  .and(NOTIFICATIONS.USER_ID.eq(currentUser().userId))
+                  .and(getOrganizationIdClause(organizationId)))
+          .execute()
+      return true
+    } else {
+      return false
```
sgrimm: Calling this method with a nonexistent notification ID isn't an expected usage pattern, right? It'd almost always be a sign of a bug or data corruption, I'd think, but maybe I'm wrong. If so, though, this could throw `NotificationNotFoundException` instead of returning a success value. Then the controller doesn't have to check the return value.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -0,0 +1,124 @@
...
+                .and(NOTIFICATIONS.USER_ID.eq(currentUser().userId)))
+        .fetch { row -> NotificationModel(row, objectMapper) }
+  }
+
+  fun count(): List<NotificationCountModel> {
+    requirePermissions { canReadNotification() }
+    return dslContext
+        .select(NOTIFICATIONS.ORGANIZATION_ID, DSL.count(NOTIFICATIONS.ID))
+        .from(NOTIFICATIONS)
+        .where(NOTIFICATIONS.READ_TIME.isNull)
```
sgrimm: This should filter by user ID too, right?

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Schema
+data class NotificationPayload(
+    val id: NotificationId,
+    val notificationType: NotificationType,
+    val notificationCriticality: NotificationCriticality,
+    val userId: UserId,
```
sgrimm: This will always be the current user's ID, right? Maybe we don't need to return it.

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
```diff
@@ -0,0 +1,145 @@
...
+    val organizationId: OrganizationId?,
+    val metadata: Map<String, Any?>,
+    val localUrl: URI,
+    val createdTime: Instant,
+    val readTime: Instant?,
```
sgrimm: Does the client display the read timestamp? Could make this a boolean if not, just to make the payloads a little more compact.

sgrimm:
Since you specifically asked in your PR description: this change seems directionally good to me so far.

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
karthikbtf: I can change this to a boolean, client does not show the timestamp.

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
karthikbtf: correct, current user's id. i will remove it from the payload

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
karthikbtf: Yes need to filter by user ID, very important, thanks for the catch!!

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
karthikbtf: Correct, a non-existent notification will not be used in the real world use-case.. unless someone is playing with curl. I will clean this up

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
karthikbtf: organization id is not needed, was wondering whether API should assume notification ids are unique across organizations. over thought, i will remove the org id here.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
karthikbtf: will clean this up, switch to a bool instead of instant in the schema

karthikbtf:
> A few more comments. Apologies if some of these are things you were planning to change already anyway -- I know this is a work in progress.

Thanks for the review, these are very helpful!

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -1,124 +1,101 @@
...
     private val dslContext: DSLContext,
+    private val notificationsDao: NotificationsDao,
     private val clock: Clock,
-    private val objectMapper: ObjectMapper
 ) {
 
-  fun fetchById(notificationId: NotificationId): NotificationModel? {
-    requirePermissions { canReadNotification() }
+  fun isOwner(userId: UserId, notificationId: NotificationId): Boolean =
+      notificationsDao.fetchOneById(notificationId)?.userId?.equals(userId)
```
sgrimm: Could use `notificationsDao.fetchOneById(notificationId)?.userId == userId` (the `==` operator is basically equivalent to `?.equals()`).

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -1,124 +1,101 @@
...
+    dslContext
+        .update(NOTIFICATIONS)
+        .set(NOTIFICATIONS.IS_READ, read)
+        .where(NOTIFICATIONS.ID.eq(notificationId).and(isCurrentUserOwnerClause()))
+        .execute()
   }
 
-  fun markAllRead(readTime: Instant?, organizationId: OrganizationId?): Unit {
-    requirePermissions { canUpdateNotification() }
+  fun markAllRead(read: Boolean, organizationId: OrganizationId?): Unit {
```
sgrimm: Minor consistency nit: the ID comes before the read flag in `markRead()` but the other way around here.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -1,124 +1,101 @@
...
       }
     }
   }
 
-  private fun getOrganizationIdClause(organizationId: OrganizationId?) =
+  private fun isOrganizationIdClause(organizationId: OrganizationId?) =
       if (organizationId == null) NOTIFICATIONS.ORGANIZATION_ID.isNull
       else NOTIFICATIONS.ORGANIZATION_ID.eq(organizationId)
+
+  private fun isCurrentUserOwnerClause() = NOTIFICATIONS.USER_ID.eq(currentUser().userId)
```
sgrimm: Putting this clause directly in the queries rather than in a helper method would make the queries easier to read, I think, since there's no conditional logic or anything.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -1,124 +1,101 @@
...
     dslContext
         .update(NOTIFICATIONS)
-        .set(NOTIFICATIONS.READ_TIME, readTime)
-        .where(
-            NOTIFICATIONS
-                .USER_ID
-                .eq(currentUser().userId)
-                .and(getOrganizationIdClause(organizationId)))
+        .set(NOTIFICATIONS.IS_READ, read)
+        .where(isOrganizationIdClause(organizationId).and(isCurrentUserOwnerClause()))
```
sgrimm: Could move the `and` up one level, that is, do `.where(x).and(y)` instead of `.where(x.and(y))`. No functional difference but it'd make the clauses line up better in the autoformatted code.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -361,9 +364,30 @@ data class IndividualUser(
     return userType == UserType.SuperAdmin
   }
 
-  override fun canReadNotification(): Boolean = true
-  override fun canCreateNotification(): Boolean = true
-  override fun canUpdateNotification(): Boolean = true
+  override fun canReadNotification(notificationId: NotificationId): Boolean {
+    return notificationStore.isOwner(userId, notificationId)
```
sgrimm: Looking up the owner of a notification should go in `ParentStore` rather than `NotificationStore`. The idea of consolidating all the "look up the parent of this thing as part of a permission check" methods in one place is to avoid this class gradually accumulating massive numbers of dependencies as we add permission checks on more kinds of data over time.

The other thing I'd suggest would be that the method in `ParentStore` return the owner's user ID, rather than taking the desired value as an argument and returning a boolean. That is, the code here would look like

```kotlin
  return parentStore.getUserId(notificationId) == userId
```

That way the actual business logic behind the permission is in here, and the external class is just doing a simple data fetch.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -361,9 +364,30 @@ data class IndividualUser(
...
+      canReadNotification(notificationId)
+
+  override fun canUpdateNotifications(organizationId: OrganizationId?): Boolean =
+      canListNotifications(organizationId)
+
+  override fun canCreateNotification(userId: UserId, organizationId: OrganizationId): Boolean =
+      // for now, ensure user making the request and the target user for notification,
+      // are both members of the organization in context
+      permissionStore.fetchOrganizationRoles(this.userId)[organizationId] != null &&
+          permissionStore.fetchOrganizationRoles(userId)[organizationId] != null
```
sgrimm: A few things here:

Maybe make this a block body rather than an expression body since it's multiple lines of code.

The explicit fetch of the current user's organizations isn't necessary; it's already available in the `organizationRoles` property. But this does still need to fetch the target user's organizations like you're doing.

Could use `in` here too, like above.

Maybe rename the `userId` argument to `targetUserId` or something, both to avoid shadowing the `userId` property and to make it clearer to callers that they shouldn't pass the current user's ID.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -361,9 +364,30 @@ data class IndividualUser(
...
-  override fun canCreateNotification(): Boolean = true
-  override fun canUpdateNotification(): Boolean = true
+  override fun canReadNotification(notificationId: NotificationId): Boolean {
+    return notificationStore.isOwner(userId, notificationId)
+  }
+  override fun canListNotifications(organizationId: OrganizationId?): Boolean {
+    // user can list global notifications relevant to user
+    if (organizationId == null) return true
+    // user should belong to the organization otherwise
+    else return organizationRoles[organizationId] != null
```
sgrimm: A few things here.

The idiomatic style is to hoist the `return` out of the `if`, since `if` is an expression in Kotlin. That is, `return if (x) y else z` rather than `if (x) return y else return z`. You can put curly braces around the values so there's somewhere to put the comments. The way you have it is perfectly valid, just not what the Kotlin style guides recommend.

You could use the `in` operator for the second part of this, that is, `organizationId in organizationRoles` rather than `organizationRoles[organizationId] != null`.

I was initially going to suggest that this could be shortened to `return organizationId == null || organizationId in organizationRoles` but I think the comments describing the intent of the two cases make it worth splitting out like you're doing.

src/main/kotlin/com/terraformation/backend/customer/model/Notification.kt
```diff
@@ -16,35 +14,36 @@ data class NotificationCountModel(val organizationId: OrganizationId?, val unrea
 data class CreateNotificationModel(
     val notificationType: NotificationType,
     val userId: UserId,
-    val organizationId: OrganizationId?,
+    val organizationId: OrganizationId,
```
sgrimm: Doesn't this need to be nullable to support non-organization-specific notifications? (In which case I guess the `isGlobalNotification` flag wouldn't be needed since it'd be implied by this being null.)

src/main/kotlin/com/terraformation/backend/customer/model/Notification.kt
```diff
@@ -16,35 +14,36 @@ data class NotificationCountModel(val organizationId: OrganizationId?, val unrea
...
     val localUrl: URI,
-    val metadata: Map<String, Any?>,
     val createdTime: Instant,
-    val readTime: Instant?,
+    val isRead: Boolean,
 ) {
   constructor(
-      row: Record,
-      objectMapper: ObjectMapper
+      record: Record
```
sgrimm, line 34: Good catch. I totally missed this the first time through.

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -428,21 +430,40 @@ class PermissionRequirements(private val user: TerrawareUser) {
...
+      throw NotificationNotFoundException(notificationId)
     }
   }
 
-  fun canCreateNotification() {
-    if (!user.canCreateNotification()) {
-      throw AccessDeniedException("No permission to create notifications")
+  fun listNotifications(organizationId: OrganizationId?) {
+    if (!user.canListNotifications(organizationId)) {
+      throw AccessDeniedException("No permission to list notifications")
```
sgrimm: Before throwing the exception, this should call `readOrganization(organizationId)` if the ID isn't null. It will throw `OrganizationNotFoundException` if the user isn't a member. The idea is that if you don't have access to an entity (organization, project, etc.) the system should act as if it doesn't exist at all rather than telling you that the thing exists but is inaccessible.

Same idea with a couple of the other methods here, but I won't comment on them specifically.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -6,6 +6,9 @@ import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import javax.annotation.ManagedBean
 
+/** Helper class to encapsulate notification message semantics */
+data class NotificationMessage(val title: String, val body: String)
```
sgrimm, line 10: This is a good idea. Much nicer than having separate methods for title and body.

src/main/kotlin/com/terraformation/backend/customer/model/Notification.kt
karthikbtf: Correct, I needed a way to convey the org id, for permission checks. I should be exposing those semantics in the create api instead of in the data model.

karthikbtf:
Thanks for the review Steve, appreciate your time and patience esp re: kotlin. Liking kotlin very much.
Will follow up on the feedback.. was wondering if I should intro a config flag (disabled by default) to govern persistence of notifications. I suspect this effort will be rolling for a bit.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -52,17 +47,19 @@ class NotificationStore(
...
         .from(NOTIFICATIONS)
-        .where(NOTIFICATIONS.IS_READ.isFalse.and(isCurrentUserOwnerClause()))
+        .where(NOTIFICATIONS.IS_READ.isFalse)
+        .and(NOTIFICATIONS.USER_ID.eq(currentUser().userId))
         .groupBy(NOTIFICATIONS.ORGANIZATION_ID)
         .fetch { row -> NotificationCountModel(row.value1(), row.value2()) }
   }
 
-  fun markRead(notificationId: NotificationId, read: Boolean): Unit {
+  fun markRead(read: Boolean, notificationId: NotificationId): Unit {
```
sgrimm: Missed this before, but `Unit` is the default for functions, so you can do `fun xyz(args: Whatever) {` if the function doesn't return anything.

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -71,19 +68,20 @@ class NotificationStore(
...
         .set(NOTIFICATIONS.IS_READ, read)
-        .where(isOrganizationIdClause(organizationId).and(isCurrentUserOwnerClause()))
+        .where(isOrganizationIdClause(organizationId))
+        .and(NOTIFICATIONS.USER_ID.eq(currentUser().userId))
         .execute()
   }
 
-  fun create(notification: CreateNotificationModel) {
-    requirePermissions { createNotification(notification.userId, notification.organizationId) }
+  fun create(notification: CreateNotificationModel, targetOrganizationId: OrganizationId) {
```
sgrimm, line 106: I had to read through this a bit to see what the difference between `targetOrganizationId` and `notification.organizationId` was; might be worth a KDoc comment for this method (which is good practice anyway) that includes a `@param` line saying what this parameter is for. (I totally get this is a work in progress and you might be planning to go add doc comments later, of course.)

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -437,6 +437,9 @@ class PermissionRequirements(private val user: TerrawareUser) {
   }
 
   fun listNotifications(organizationId: OrganizationId?) {
+    if (organizationId != null) {
+      readOrganization(organizationId)
+    }
```
sgrimm: It'll throw the correct exception either way given that we're unlikely to ever have a role with write-only permissions on things, but I've been putting these "can the user read the object they're trying to modify?" checks inside the other `if` so they're only executed if the user lacks the write permission.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -211,7 +210,6 @@ abstract class DatabaseTest {
   protected val timeseriesDao: TimeseriesDao by lazyDao()
   protected val usersDao: UsersDao by lazyDao()
   protected val withdrawalsDao: WithdrawalsDao by lazyDao()
-  protected val notificationsDao: NotificationsDao by lazyDao()
```
sgrimm: If you end up needing this for a test class for `NotificationStore`, it should go in alphabetical order.

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
```diff
@@ -31,6 +31,7 @@ import org.springframework.web.bind.annotation.RestController
 @RestController
 class NotificationController(private val notificationStore: NotificationStore) {
 
+  /** API to retrieve a notification by its id */
```
sgrimm: A style thing: The convention for method KDoc comments is for the first sentence to be written as if it started with, "This method." So for example, this one would be, "Retrieves a notification by its ID."

src/main/kotlin/com/terraformation/backend/customer/db/NotificationStore.kt
```diff
@@ -20,6 +20,13 @@ class NotificationStore(
     private val clock: Clock,
 ) {
 
+  /**
+   * Fetches a notification by unique id
+   *
+   * @param notificationId ID of notification for which to retrieve information
+   * @return NotificationModel representation of the notification
```
sgrimm: It's fine to include these if you want, but you can leave out descriptions of parameters or return values where the parameter name or the type name already says pretty much the same thing the description would say. Basically my usual approach is that I'll only add param/return docs that tell you something you wouldn't already know by reading the method signature. And often I'll apply the same logic to the KDoc comment as a whole if the method name is already a clear description of what the method does.

That's all obviously super subjective, though! What's clear or obvious to one person might not be to another.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    every { user.canListNotifications(any()) } returns true
+    every { user.canUpdateNotification(any()) } returns true
+    every { user.canUpdateNotifications(any()) } returns true
+    every { user.canCountNotifications() } returns true
+
+    every { user.organizationRoles } returns mapOf(organizationId to Role.OWNER)
+    every { user.projectRoles } returns mapOf(projectId to Role.OWNER)
+
+    insertUser()
+    Assertions.assertEquals(
```
sgrimm: Minor nit: You could remove the `Assertions.` on this because there's already a static import of `assertEquals`.

Not-so-minor nit: Setup methods shouldn't contain assertions.

`insertOrganization()` can take the organization ID as an argument, but a better approach would be to call the `insertSiteData()` helper method, which inserts a canned organization and project. Lines 114-126 could be replaced with a call to that method.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+        projectId,
+        description = projectModel.description,
+        startDate = projectModel.startDate,
+        status = projectModel.status,
+        types = projectModel.types)
+  }
+
+  @Test
+  fun `should create a notification`() {
+    assert(dslContext.fetchCount(NOTIFICATIONS) == 0)
```
sgrimm: Use the JUnit assertion methods from the `Assertions` class instead of the `assert` built-in. The main reason is that the built-in won't output an error message that shows the expected and actual values.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    assert(dslContext.fetchCount(NOTIFICATIONS) == 1)
+  }
+
+  @Test
+  fun `should fetch a notification by id`() {
+    val toCreate = notificationModel()
+    val id = NotificationId(1)
+    val expected = toModel(toCreate, id)
+    store.create(toCreate, organizationId)
+    assertEquals(store.fetchById(id), expected)
```
sgrimm: I know it feels kind of backwards, and the assertion will fail either way, but the argument order for the assertion methods is `expected, actual` rather than `actual, expected`.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  @Test
+  fun `should list notifications scoped to an organization for user`() {
+    // create a few notification model representations
+    val toCreate = Array(5, { notificationModel() })
```
sgrimm: Idiomatic style is to put the lambda function outside the parentheses if it's the last argument. In other words, `Array(5) { notificationModel() }`. IntelliJ should show a little tooltip about that if you hover over this statement (and it can make the change for you if you hit the hotkey the tooltip mentions).

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    val expected =
+        toCreate.mapIndexed { index, model -> toModel(model, NotificationId(index + 1L)) }
+
+    // persist the notificatios
+    toCreate.forEach { notification -> store.create(notification, organizationId) }
```
sgrimm: An alternate way this could be structured, since the `NotificationModel` doesn't vary across entries:

```kotlin
val model = notificationModel()
val expected = (1..5).map { id -> toModel(model, NotificationId(id) }

repeat(5) { store.create(model, organizationId) }

assertEquals(expected, store.fetchByOrganization(organizationId))
```

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+
+    assertEquals(store.fetchByOrganization(organizationId), expected)
+  }
+
+  @Test
+  fun `should list notifications scoped globally for user`() {
```
sgrimm: Is the list of notifications supposed to be in a particular order (time order, etc.)? Might be good to explicitly check for that here too.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    store.markRead(true, id)
+
+    // confirm it is read
+    assertTrue(store.fetchById(id).isRead)
+
+    // mark as unread
+    store.markRead(false, id)
+
+    // confirm it is unread
+    assertFalse(store.fetchById(id).isRead)
```
sgrimm: When there are multiple assertions of the same type in a test method, it's useful to add brief descriptions to all of them so you can quickly tell which one failed by looking at the error message.

In this case I'd probably put descriptions on all the assertions; the descriptions could take the place of some/all of the code comments.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    // confirm they are unread
+    assertFalse(store.fetchById(id1).isRead)
+    assertFalse(store.fetchById(id2).isRead)
+  }
+
+  @Test
+  fun `should return count information on unread notifications`() {
+    val id = NotificationId(1)
+
+    // create 2 notifications
```
sgrimm: Should this say 2 of each type?

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+    store.create(notificationModel(), organizationId)
+    store.create(notificationModel(), organizationId)
+    store.create(notificationModel(true), organizationId)
+    store.create(notificationModel(true), organizationId)
+
+    // confirm unread count (2 for org, 2 for global)
+    var result = store.count()
+
+    // get unread count for org
+    var forOrg = result.filter { r -> r.organizationId == organizationId }[0]
```
sgrimm: Could do `result.firstOrNull { r -> r.organizationId == organizationId }` (yeah, there are a zillion of those little collection-related methods in Kotlin's standard library)

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -0,0 +1,289 @@
...
+
+  @Test
+  fun `should fetch a notification by id`() {
+    val toCreate = notificationModel()
+    val id = NotificationId(1)
+    val expected = toModel(toCreate, id)
+    store.create(toCreate, organizationId)
+    assertEquals(store.fetchById(id), expected)
+  }
+
```
sgrimm, line 117: Probably worth throwing in test cases to make sure it throws an exception if you try to fetch someone else's (or some other organization's) notifications.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
karthikbtf: Currently not explicitly ordered by any column (inherently by id I suppose).
UI will be making 2 calls (one with org id and one without).. combine the results and sort by created time (descending). Would it help to add an explicit ordering?

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
karthikbtf: i agree, will make that change.. this is how I would usually do it.

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
karthikbtf: good idea, will do

src/main/kotlin/com/terraformation/backend/customer/api/NotificationController.kt
```diff
@@ -85,7 +85,7 @@ class NotificationController(private val notificationStore: NotificationStore) {
   }
 
   /**
-   * Mark all user's notifcation as read or unread, scoped by organization. If organization id is
+   * Marks all user's notifcation as read or unread, scoped by organization. If organization id is
```
sgrimm: (Nit) Spelling error: "notifcation" should be "notification" (or I guess "notifications" would make sense)

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
```diff
@@ -175,115 +130,123 @@ internal class NotificationStoreTest : DatabaseTest(), RunsAsUser {
...
+    val otherUserId = UserId(3)
+    insertUser(otherUserId)
+    assertThrows<NotificationNotFoundException> {
+      val notification = notificationModel(userId = otherUserId)
+      store.create(notification, organizationId)
+      store.fetchById(NotificationId(1))
+    }
+  }
+
+  private fun listNotificationsTest(globalNotifications: Boolean) {
```
sgrimm: There's another way to do this, FYI, though what you have is fine: you can use a parameterized test rather than a helper method that gets called with different arguments.

It'd look something like
```kotlin
@ParameterizedTest
@ValueSource(booleans=[true, false])
fun `fetchByOrganization should return notifications with correct scope`(globalNotifications: Boolean) {
  // same body as here
}
```
and the test runner will call the test once with each of the values in the annotation. Then you don't need the two explicit test methods to call this with `true` and `false`.

src/main/resources/db/migration/common/V83__NotificationTypes.sql
```diff
@@ -0,0 +1,22 @@
...
+VALUES (1, 'Info'),
+       (2, 'Warning'),
+       (3, 'Error'),
+       (4, 'Success');
+
+INSERT INTO notification_types (id, name, notification_criticality_id)
+VALUES (1, 'User Added to Organization', 1),
+       (2, 'Facility Idle', 2),
+       (3, 'Facility Alert Requested', 3)
+ON CONFLICT (id) DO UPDATE SET name = excluded.name, notification_criticality_id = excluded.notification_criticality_id;
```
sgrimm: Should have caught this earlier, but it's important: these two `INSERT` statements should go in `R__TypeCodes.sql` instead of here.

Numbered migrations like this one are run exactly once, and it gets flagged as an error if you make changes to a numbered migration that has already been run. Basically, the migration library (Flyway) treats numbered migrations as immutable.

Repeatable (`R__`) migrations are run after all the numbered migrations are finished, and they run every time the server gets started. They're allowed to change over time.

We're definitely going to be adding notification types over time, and by putting them in `R__TypeCodes.sql`, we can do it by editing that file rather than having to add a new numbered migration for each one.

src/main/resources/db/migration/common/V84__NotificationsTable.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE notifications (
```
sgrimm: Might as well combine these two migrations into one (in other words, put this `CREATE TABLE` in the `V83` migration).

src/test/kotlin/com/terraformation/backend/customer/db/NotificationStoreTest.kt
karthikbtf: That's pretty cool, I will use this... saw parameterized tests in one of your reviews, I should have applied the learnings here.

src/main/resources/db/migration/common/V83__NotificationTypes.sql
karthikbtf: will do 👍  yes i have been clearing the row in flyway table every time i make changes to this file.

src/main/resources/db/migration/common/V84__NotificationsTable.sql
karthikbtf: > 

Thank you so much. No hurry. At some point when approved, I should maybe merge this into a feature branch. The notifications part of the UI and the backend effort need to go out at the same time, similar to species list changes.

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -18,4 +18,9 @@ class WebAppUrls(private val config: TerrawareServerConfig) {
         .queryParam("organizationId", organizationId)
         .build()
   }
+
+  /** Generates a relative path of organization home within the web app */
+  fun inAppOrganizationHome(organizationId: OrganizationId): URI {
```
karthikbtf: Not sure if this function name is friendly enough. Objective is to generate just the app routes without app host url.
React single page web-app works better with internal route semantics, also for security reasons (no external/full urls). Also helps us test UI in dev mode against say staging backend server.

sgrimm: Local URLs are definitely the way to go here, agreed. Kind of wish we could do this without needing two methods per target URL but I don't love any of the alternate approaches I can think of.

But as for the name: maybe `local` rather than `inApp` here, or alternately, maybe put a `remote` or `full` prefix on the existing method and this one can be `organizationHome`? Shorter method name returns the shorter URL seems like a decent heuristic.

karthikbtf:
> I think this is good to merge now. I'll leave it to you to decide if you want to change the method name you asked about.
> 
> Oh, I guess one thing: update the PR title and description so we can use them as the commit message (I just updated https://terraformation.atlassian.net/wiki/spaces/FT/pages/329613338/Git+and+GitHub+conventions to bring it up to date, FYI.)
> 
> I don't think this needs to go on a feature branch. As far as I can tell it isn't changing any existing client-facing functionality, just adding a new endpoint. Old clients just won't use the new endpoint but it shouldn't break them. Or am I missing a non-backward-compatible aspect of this?

Thanks for the review! I like the `fullOrganizationHome` / `organizationHome` for full/short uri.. suggestion. Will make that change.

Re: merging this PR, I will follow the wiki doc on merge comments.

The changes/api here are completely new and should not affect existing clients.

Is there a chance we might push latest code to production before the Q2 features are shipped? 
If so.. I don't have the final copy for title/body, I wouldn't want this change rolled out to production and end up persisting 'lorem ipsum..' in notification messages. Perhaps I should get that resolved first.

sgrimm:
Ah, good point about persisting the placeholder text. Yeah, maybe good to get a first draft of real text in place before merging this. Doesn't have to be perfect, just good enough for us to be okay with users seeing it.

PR #303, opened 2022-04-26
WIP: Overhaul species data model

sgrimm:
What's here at the moment:

Client-visible changes:

* Species are now purely per-organization. There is no more concept of a species ID that's shared across orgs.
* Species API payload fields have changed to reflect the new requirements:
  * Scientific and common names are separate fields
  * Endangered and rare are optional boolean values
  * Plant form is replaced with growth form
  * New field for seed storage behavior
  * New field for family name
  * Scientific name is required; all other fields are optional
* Removed the concept of "species names" as a distinct thing from "species." All the name-related endpoints are gone.
* There is a new `/api/v1/species/lookup/names` endpoint that returns a list of scientific names that match a list of prefixes. This is to support the typeahead in the "Add Species" dialog.
* The admin UI now allows uploading the GBIF dataset. README.md has a section describing how to do it.

Internal changes of note:

* Consolidated species data into a single `species` table; `species_names` and `species_options` no longer exist
* Accession `species` field uses scientific name (previously it used the primary name, which might or might not have been scientific)
* Species-related permission checks now take species ID arguments for consistency with the rest of the permissions API (previously they took organization ID arguments)

PR #308, opened 2022-04-28
Overhaul species data model

sgrimm:
Replace the previous set of species tables with a simplified version that matches the requirements for the upcoming changes to the species management features in the app.

Client-visible changes:

* Species are now purely per-organization. There is no more concept of a species ID that's shared across orgs.
* Species API payload fields have changed to reflect the new requirements:
  * Scientific and common names are separate fields
  * Endangered and rare are optional boolean values
  * Plant form is replaced with growth form
  * New field for seed storage behavior
  * New field for family name
  * Scientific name is required; all other fields are optional
* Removed the concept of "species names" as a distinct thing from "species." All the name-related endpoints are gone.

Internal changes of note:

* Consolidated species data into a single `species` table; `species_names` and `species_options` no longer exist
* Accession `species` field uses scientific name (previously it used the primary name, which might or might not have been scientific)
* Species-related permission checks now take species ID arguments for consistency with the rest of the permissions API (previously they took organization ID arguments)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -303,15 +303,12 @@ data class IndividualUser(
     return canReadOrganization(organizationId)
   }
 
-  override fun canDeleteSpecies(speciesId: SpeciesId): Boolean {
-    val organizationId = parentStore.getOrganizationId(speciesId) ?: return false
-    return canCreateSpecies(organizationId)
-  }
-
   override fun canUpdateSpecies(speciesId: SpeciesId): Boolean {
     val organizationId = parentStore.getOrganizationId(speciesId) ?: return false
     return canCreateSpecies(organizationId)
   }
+
+  override fun canDeleteSpecies(speciesId: SpeciesId): Boolean = canUpdateSpecies(speciesId)
 
   override fun canCreateTimeseries(deviceId: DeviceId): Boolean {
     val facilityId = parentStore.getFacilityId(deviceId) ?: return false


sgrimm:
* main
  * **PR #308** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/308" target="_blank">Graphite</a>  (this PR)
    * **PR #309** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/309" target="_blank">Graphite</a> 
      * **PR #311** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/311" target="_blank">Graphite</a> 
        * **PR #303** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/303" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/308?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -298,34 +298,19 @@ data class IndividualUser(
...
+  override fun canDeleteSpecies(speciesId: SpeciesId): Boolean {
+    val organizationId = parentStore.getOrganizationId(speciesId) ?: return false
+    return canCreateSpecies(organizationId)
   }
 
-  override fun canDeleteSpeciesName(speciesNameId: SpeciesNameId): Boolean {
-    return canUpdateSpeciesName(speciesNameId)
+  override fun canUpdateSpecies(speciesId: SpeciesId): Boolean {
+    val organizationId = parentStore.getOrganizationId(speciesId) ?: return false
+    return canCreateSpecies(organizationId)
```
karthikbtf, line 308: wondering if canUpdateSpecies and canDeleteSpecies can be extracted out into a reusable function..
or not sure if this is kotlin friendly:
`override fun canUpdateSpecies(speciesId: SpeciesId) : Boolean = canDeleteSpecies`

sgrimm: Yeah, I can reduce the repetition here. Thanks.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -119,174 +75,82 @@ class SpeciesStore(
...
+            .and(SPECIES.DELETED_TIME.isNotNull)
+            .fetchOneInto(SpeciesRow::class.java)
+
+    return if (existingRow?.deletedTime != null) {
+      log.info("Reusing existing species ${existingRow.id}")
+
+      val rowWithExistingId =
+          row.copy(
+              createdBy = existingRow.createdBy,
+              createdTime = existingRow.createdTime,
```
karthikbtf, line 102: just a question, no real suggestion..
this resurrection, do we care if the createdBy/Time should reflect old state,, or should it reflect current user/time.. since we are 'creating' and maybe not 'un-deleting'

sgrimm: It's a good question! I was on the fence about this and couldn't think of a compelling reason to pick either one. The not-super-compelling reason I went with the existing timestamp is just that it's marginally more consistent if there are references to the species (say, in the seeds data) from before it was deleted.

karthikbtf: Sounds good to me.

PR #309, opened 2022-04-28
Add endpoint to look up species names matching a list of prefixes

sgrimm:
`/api/v1/species/lookup/names` returns a list of scientific names that match a
list of prefixes. This is to support the typeahead in the "Add Species" dialog.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
@@ -44,6 +44,10 @@ class GbifStore(private val dslContext: DSLContext) {
     // AND gbif_names.is_scientific = TRUE
     // AND LOWER(gbif_names.name) LIKE '%abc%def%'
     // ORDER by gbif_names.name
+    //
+    // The LOWER(gbif_names.name) LIKE '%abc%def%' is needed because we want the prefixes to be
+    // order-sensitive, but we don't require them to start at the beginning of the name. That is,
+    // this search should match species 'Abc def' and 'Xyz abc var. def' but not species 'Def abc'.
 
     val selectFrom = dslContext.selectDistinct(GBIF_NAMES.asterisk()).from(GBIF_NAMES)
 


sgrimm:
* species
  * **PR #309** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/309" target="_blank">Graphite</a>  (this PR)
    * **PR #311** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/311" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/309?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    // Dynamically construct a query that has a separate join to the gbif_name_words table for each
+    // prefix string. For prefixes == listOf("abc", "def") we want to end up with a query like
+    //
+    // SELECT name
+    // FROM gbif_names
+    // JOIN gbif_name_words gnw_0 ON gbif_names.id = gnw_0.id
+    // JOIN gbif_name_words gnw_1 ON gbif_names.id = gnw_1.id
+    // WHERE gnw_0.word LIKE 'abc%'
+    // AND gnw_1.word LIKE 'def%'
```
karthikbtf, line 43: are these words at rest in lowercase?

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
```diff
@@ -0,0 +1,75 @@
...
+    // AND gbif_names.is_scientific = TRUE
+    // AND LOWER(gbif_names.name) LIKE '%abc%def%'
```
karthikbtf, line 45: do we need the leading '%' ?

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
sgrimm: Yes, they get folded to lower case as part of the import process in `GbifImporter`.

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
sgrimm: Yes, because it's valid to search for the second (and third) words in a species name without the first word. That is, if you search for "xy qr" it should match species "Abc xyz var. qrst".

sgrimm: This won't run into the typical problem of leading `%`s (forcing a full scan of the whole table) since it'll act as a filter on the list of IDs that come from the per-word searches.

karthikbtf: Thanks for the clarification, makes sense.

PR #311, opened 2022-04-28
Allow uploading GBIF backbone zipfile to admin UI

sgrimm:
The existing API endpoint for importing GBIF backbone data requires unpacking
the zipfile from the GBIF website and then constructing a correct URL to point
to the location of the unpacked files.

That's workable, but kind of a hassle, especially in dev environments.

Add a section to the admin UI to allow uploading the original zipfile without
having to unpack it or construct a URL for it. (I chose to do this as a file
upload rather than having the server directly download from the GBIF site
because we'll need to import the data over and over again in dev environments
and it's not nice to eat GBIF's bandwidth repeatedly downloading the same big
file.)

Importing species data is only permitted for super-admin users. There wasn't
previously any documentation describing how to set someone as a super-admin,
so add a section to `README.md` about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,32 @@ terraware:
     subject-prefix: "[DEV]"
 ```
 
+## Super-Admins
+
+Some operations are not available to regular users of the system; they must be requested by privileged administrative users. Internally, these are referred to as "super-admins".
+
+Marking a user as a super-admin currently requires connecting to the server's database and modifying the users table directly. The exact command you'll need to run to connect to the database will vary a bit depending on where it's running (local host, Docker container, managed cloud database service, etc.). For a database running on your local host, `psql terraware` will usually work.
+
+Once you're connected to the database, run the following query to change the user with a particular email address to a super-admin:
+
+```sql
+UPDATE users SET user_type_id = 2 WHERE email = 'your_email@your.domain';
+```
+
+## Importing GBIF backbone data
+
+The server uses a public species database from the [Global Biodiversity Information Facility (GBIF)](https://gbif.org/). If you're working with species-related parts of the code, you'll need to import the data into your local database.
+
+This must be done by a super-admin; see above for more information on that.
+
+1. Download `backbone.zip` from [the GBIF backbone dataset page](https://www.gbif.org/dataset/d7dddbf4-2cf0-4f39-9b2a-bb099caae36c) (it's the "Source archive" item in the download menu at the top of the page).
+2. Log into the admin interface on your terraware-server instance. For a locally-running server instance, it'll typically be [http://localhost:8080/admin/](http://localhost:8080/admin/).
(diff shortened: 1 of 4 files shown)


sgrimm:
* species
  * **PR #311** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/311" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/311?utm_source=stack-comment).

README.md
```diff
@@ -164,6 +164,32 @@ terraware:
...
+This must be done by a super-admin; see above for more information on that.
+
+1. Download `backbone.zip` from [the GBIF backbone dataset page](https://www.gbif.org/dataset/d7dddbf4-2cf0-4f39-9b2a-bb099caae36c) (it's the "Source archive" item in the download menu at the top of the page).
+2. Log into the admin interface on your terraware-server instance. For a locally-running server instance, it'll typically be [http://localhost:8080/admin/](http://localhost:8080/admin/).
+3. If you are a super-admin, you'll see an "Import GBIF species data" section on the admin home page. If you don't see that section, make sure you've set your user type properly as described in the "Super-Admins" section above.
+4. Select the zipfile you downloaded from the GBIF site, and hit the "Upload Zipfile" button.
+5. Watch the server logs to monitor the progress of the import. It will take several minutes to finish.
+
+Once it's done, the species lookup endpoints under `/api/v1/species/lookup` should start returning real results.
+
```
karthikbtf, line 192: guessing this is something we have to do in production prior to launch of new species feature

src/main/kotlin/com/terraformation/backend/species/db/GbifImporter.kt
```diff
@@ -79,22 +81,30 @@ class GbifImporter(
...
+      val entry = zipFile.getEntry("backbone/$fileName") ?: throw FileNotFoundException(fileName)
+      zipFile.getInputStream(entry)
+    }
+  }
+
+  fun import(openFile: (String) -> InputStream) {
     requirePermissions { importGlobalSpeciesData() }
 
     try {
       dslContext.transaction { _ ->
```
karthikbtf, line 98: ignore if this comment is invalid.. wondering if we should lock some row as part of this transaction, to avoid supporting parallel uploads by multiple users (not realistic for us currently i agree)

sgrimm: That's probably a reasonable sanity check. I'm not worried about multiple users doing this simultaneously by coincidence, but it's not hard to imagine someone double-submitting the form by accident.

At the moment there's no obvious thing that one would hold a lock on as part of this, but that's probably a sign I need to add one, rather than a sign locking isn't needed.

sgrimm: I've added a locking mechanism to prevent double importing, but it is substantial enough to deserve its own PR. So I'm going to merge this one and add the locking as a followup.

PR #317, opened 2022-05-03
Prevent overlapping GBIF imports

sgrimm:
Previously, if you double-clicked the submit button in the admin UI's
GBIF import form, the server would run both imports at the same time.
This could result in duplicate data being imported.

Add a locking mechanism so the second attempt will fast-fail before
making any changes.

The locking mechanism is implemented as a general-purpose service since
this is unlikely to be the last time we'll need application-wide locking.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/LockService.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/LockService.kt
@@ -29,7 +29,7 @@ class LockService(private val dslContext: DSLContext) {
    *
    * Recommended usage:
    * ```
-   * lockService.tryExclusiveLockNonTransactional(key).use { ... }
+   * lockService.lockExclusiveNonTransactional(key).use { ... }
    * ```
    *
    * @return An object that can be closed to release the lock; you'll typically want to use the


sgrimm:
* species
  * **PR #317** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/317" target="_blank">Graphite</a>  (this PR)
    * **PR #318** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/318" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/317?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/db/LockService.kt
```diff
@@ -0,0 +1,93 @@
...
+    lockBlocking("pg_advisory_xact_lock", lockType)
+  }
+
+  /**
+   * Acquires an exclusive lock on the given key. The lock is held until it is released or the
+   * current database session is closed.
+   *
+   * Recommended usage:
+   * ```
+   * lockService.tryExclusiveLockNonTransactional(key).use { ... }
```
karthikbtf: should be lockService.lockExclusiveNonTransactional...

PR #319, opened 2022-05-03
SW-907 Add endpoint to fetch GBIF species details

sgrimm:
When the user selects a preexisting species name while adding a new species, we
need to fill out some of the fields in the add-species form with values from the
GBIF database. Add an endpoint to return the relevant details if available.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1668,6 +1668,43 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/CreateSpeciesResponsePayload'
+  /api/v1/species/lookup/details:
+    get:
+      tags:
+      - Customer
+      summary: Gets more information about a species with a particular scientific
+        name.
+      operationId: getSpeciesDetails
+      parameters:
+      - name: scientificName
+        in: query
+        required: true
+        schema:
+          type: string
+          description: Exact scientific name to look up. This name is case-sensitive.
+      - name: language
+        in: query
+        required: false
+        schema:
+          type: string
+          description: "If specified, only return common names in this language or\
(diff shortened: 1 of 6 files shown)


sgrimm:
* species
  * **PR #319** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/319" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/319?utm_source=stack-comment).

sgrimm:
The fact that this endpoint requires you to pass in the full scientific name as opposed to a shorter ID feels kind of weird to me, but it's kind of an inevitable consequence of wanting the "look up matching names" endpoint to return as short a response as possible.

karthikbtf:
> The fact that this endpoint requires you to pass in the full scientific name as opposed to a shorter ID feels kind of weird to me, but it's kind of an inevitable consequence of wanting the "look up matching names" endpoint to return as short a response as possible.
Seems reasonable given the use-case. Would it be awkward to treat the scientific name as the unique ID `/details-endpoint/<scientific name>?language=<code>`   .. just thinking out aloud, not a request.

openapi.yaml
```diff
@@ -1668,6 +1668,43 @@ paths:
...
+          example: en
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SpeciesLookupDetailsResponsePayload'
+        "404":
+          description: The scientific name was not found in the server's taxonomic
```
karthikbtf, line 1702: mention ...for the specific language if provided...?
A scientific name may exist but not for an input language.

PR #320, opened 2022-05-04
SW-905 Add species to search API

sgrimm:
Add a `species` sublist to the organizations search table, and add some scalar
fields that were previously missing. This allows clients to search the species
lists of their organizations using the generic search API.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
@@ -28,16 +28,21 @@ class SpeciesTable(tables: SearchTables, fuzzySearchOperators: FuzzySearchOperat
 
   override val fields: List<SearchField> =
       listOf(
-          idWrapperField("id", "Species ID", SPECIES.ID) { SpeciesId(it) },
           textField("commonName", "Species common name", SPECIES.COMMON_NAME),
-          booleanField("endangered", "Species endangered", SPECIES.ENDANGERED),
+          booleanField("endangered", "Species is endangered", SPECIES.ENDANGERED),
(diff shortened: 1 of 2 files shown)


sgrimm:
* species
  * **PR #320** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/320" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/320?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
```diff
@@ -18,15 +20,32 @@ class SpeciesTable(tables: SearchTables, fuzzySearchOperators: FuzzySearchOperat
...
+              SPECIES.SCIENTIFIC_NAME,
+              nullable = false),
+          enumField(
+              "seedStorageBehavior",
+              "Species seed storage behavior",
+              SPECIES.SEED_STORAGE_BEHAVIOR_ID),
       )
+
+  override fun conditionForPermissions(): Condition {
+    return SPECIES.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)
```
karthikbtf, line 49: learned something new.. had to look up back-tick notation, understand this is to handle function names that are reserved keywords

sgrimm: Yeah, it's an unfortunate artifact of the fact that jOOQ is a Java library and `in` is a SQL keyword but not a Java one.

The backticks are annoying to have to use but I think the alternative (adding an alias that isn't a Kotlin keyword) isn't any better since it'd break the "jOOQ's query-building DSL looks just like the resulting SQL" feature.

PR #329, opened 2022-05-09
SW-954 Send app and email notification to user when added to a project

karthikbtf:
Allow user to receive notifications in the webapp and email, when added to a project.
Introduced ProjectService which handles adding a user to a project, and publishes a relevant event.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
@@ -93,7 +93,7 @@ class AppNotificationService(
             title = message.title,
             body = message.body,
             localUrl = projectHomeUrl,
-            notificationType = NotificationType.UserAddedToProject)
+            notificationType = NotificationType.UserAddedtoProject)
     notificationStore.create(notification, organization.id)
   }
 }
--- a/src/main/resources/db/migration/common/R__TypeCodes.sql
+++ b/src/main/resources/db/migration/common/R__TypeCodes.sql
@@ -141,5 +141,5 @@ INSERT INTO notification_types (id, name, notification_criticality_id)
 VALUES (1, 'User Added to Organization', 1),
        (2, 'Facility Idle', 2),
        (3, 'Facility Alert Requested', 3),
-       (4, 'User Added To Project', 1)
+       (4, 'User Added to Project', 1)
 ON CONFLICT (id) DO UPDATE SET name = excluded.name, notification_criticality_id = excluded.notification_criticality_id;
--- a/src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
@@ -124,29 +124,25 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
 
     service.on(UserAddedToOrganizationEvent(otherUserId, organizationId, user.userId))
 
-    assertEquals(
-        1,
-        notificationsDao.count(),
-        "There should be one notification after user is added to an organization")
+    val expectedNotifications =
+        listOf(
+            NotificationsRow(
+                id = NotificationId(1),
+                notificationTypeId = NotificationType.UserAddedtoOrganization,
+                userId = otherUserId,
+                organizationId = null,
+                title = "organization title",
+                body = "organization body",
+                localUrl = webAppUrls.organizationHome(organizationId),
+                createdTime = Instant.EPOCH,
+                isRead = false))
 
-    val expectedNotification =
-        NotificationsRow(
-            id = NotificationId(1),
-            notificationTypeId = NotificationType.UserAddedtoOrganization,
-            userId = otherUserId,
-            organizationId = null,
-            title = "organization title",
-            body = "organization body",
-            localUrl = webAppUrls.organizationHome(organizationId),
-            createdTime = Instant.EPOCH,
-            isRead = false)
-
-    val actualNotification = notificationsDao.fetchById(NotificationId(1)).firstOrNull()!!
+    val actualNotifications = notificationsDao.findAll()
 
     assertEquals(
-        expectedNotification,
-        actualNotification,
-        "Notification should match that of user added to an organization")
+        expectedNotifications,
+        actualNotifications,
+        "Notifications should match that of a single user added to an organization")
   }
 
   @Test
@@ -162,28 +158,24 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
 
     service.on(UserAddedToProjectEvent(otherUserId, projectId, user.userId))
 
-    assertEquals(
-        1,
-        notificationsDao.count(),
-        "There should be one notification after user is added to a project")
+    val expectedNotifications =
+        listOf(
+            NotificationsRow(
+                id = NotificationId(1),
+                notificationTypeId = NotificationType.UserAddedtoProject,
+                userId = otherUserId,
+                organizationId = organizationId,
+                title = "project title",
+                body = "project body",
+                localUrl = webAppUrls.organizationProject(projectId),
+                createdTime = Instant.EPOCH,
+                isRead = false))
 
-    val expectedNotification =
-        NotificationsRow(
-            id = NotificationId(1),
-            notificationTypeId = NotificationType.UserAddedToProject,
-            userId = otherUserId,
-            organizationId = organizationId,
-            title = "project title",
-            body = "project body",
-            localUrl = webAppUrls.organizationProject(projectId),
-            createdTime = Instant.EPOCH,
-            isRead = false)
-
-    val actualNotification = notificationsDao.fetchById(NotificationId(1)).firstOrNull()!!
+    val actualNotifications = notificationsDao.findAll()
 
     assertEquals(
-        expectedNotification,
-        actualNotification,
-        "Notification should match that of user added to a project")
+        expectedNotifications,
+        actualNotifications,
+        "Notification should match that of a single user added to a project")
   }
 }
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3784,7 +3784,7 @@ components:
           - User Added to Organization
           - Facility Idle
           - Facility Alert Requested
-          - User Added To Project
+          - User Added to Project
         notificationCriticality:
           type: string
           enum:


src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -0,0 +1,40 @@
...
+    private val dslContext: DSLContext,
+    private val publisher: ApplicationEventPublisher,
+    private val permissionStore: PermissionStore,
+    private val projectStore: ProjectStore,
+) {
+
+  fun addUser(projectId: ProjectId, userId: UserId) {
+    requirePermissions { addProjectUser(projectId) }
+
+    val project = projectStore.fetchById(projectId) ?: throw ProjectNotFoundException(projectId)
```
karthikbtf, line 27: Not sure if the throw is needed, since the require permissions check captures this.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -65,4 +69,32 @@ class AppNotificationService(
...
+  fun on(event: UserAddedToProjectEvent) {
+    userStore.fetchById(event.addedBy) ?: throw UserNotFoundException(event.addedBy)
+    val user = userStore.fetchById(event.userId) ?: throw UserNotFoundException(event.userId)
+    val project =
+        projectStore.fetchById(event.projectId) ?: throw ProjectNotFoundException(event.projectId)
+    val organization =
+        organizationStore.fetchById(project.organizationId)
+            ?: throw OrganizationNotFoundException(project.organizationId)
+
+    val projectHomeUrl = webAppUrls.projectHome(event.projectId).toString()
```
sgrimm: Could probably remove the `.toString()` here and then there's no need for the `URI.create()` below.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -65,4 +69,32 @@ class AppNotificationService(
...
+            "${event.projectId}.")
+
+    val notification =
+        CreateNotificationModel(
+            userId = user.userId,
+            organizationId = organization.id,
+            title = message.title,
+            body = message.body,
+            localUrl = URI.create(projectHomeUrl),
+            notificationType = NotificationType.UserAddedtoOrganization)
```
sgrimm: Should this be a different notification type since it's a project, not an organization?

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -0,0 +1,40 @@
...
+    val organizationId = project.organizationId
+    if (organizationId !in permissionStore.fetchOrganizationRoles(userId)) {
+      throw IllegalArgumentException(
+          "Cannot add user that does not belong to project's organization")
+    }
+
+    return dslContext.transactionResult { _ ->
```
sgrimm: This method isn't returning anything, so this could be `dslContext.transaction { _ ->` without the `return`.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -68,4 +68,10 @@ class Messages {
       NotificationMessage(
           title = "You've been added to a new organization!",
           body = "You are now a member of $orgName. Welcome!")
+
+  /** Title and body to use for "user added to project" app notification */
+  fun userAddedToProjectNotification(projectName: String, orgName: String): NotificationMessage =
+      NotificationMessage(
+          title = "You've been added to a new project!",
+          body = "You are now a member of $projectName, a $orgName project.")
```
sgrimm: If the org name starts with a vowel, this is going to be incorrect (you'd need "an" instead of "a"). Rather than trying to construct the correct sentence, could we find alternate wording that gets rid of that problem?

Do we need to include the organization name at all? If this is a per-org notification, I think the user will only see it if they're looking at the org in question.

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
sgrimm: I'd keep it just because `!!` isn't considered great form to use outside of test code, but yeah, this is a little redundant.

We should probably add non-nullable fetch methods to the various store classes so this "throw if the thing isn't found" logic doesn't need to be repeated all over the place. But that's out of scope for this PR, of course; what you have here is fine.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
karthikbtf: oops, yes, completely missed changing this. thanks!

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
karthikbtf: Copy came from Product, let me counter this. Wondering if adding quotes \"org name\"  might resolve the a vs an issue.

sgrimm: I don't think so, or at least this doesn't sound good to me:

```
You are now a member of Some Project, a "Eden" project.
```

karthikbtf: Yes, looks uncomfortable. I will look at how to fix the article.

karthikbtf: Removed org name for now just to make progress. Will follow up separately once Nicole updates the table. Looked online for utilities, no luck finding one. Tried creating my own `withArticle(noun)` function but realized there are so many corner cases like a silent-h (an-hour vs a-hour) and those like 'University' (a Univ..).

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -61,8 +64,36 @@ class AppNotificationService(
...
+            "${event.projectId}.")
+
+    val notification =
+        CreateNotificationModel(
+            userId = user.userId,
+            organizationId = organization.id,
+            title = message.title,
+            body = message.body,
+            localUrl = projectHomeUrl,
+            notificationType = NotificationType.UserAddedToProject)
```
karthikbtf: Tried writing a test for AppNotificationService but ran into a dead-end, wasn't sure how to initialize the event publisher without mocking it.

karthikbtf:
added email support as well, with placeholder text.
tested email using ses (images looked blank, will check that out)

DEV  k b1 has added you to project test1 on Terraware - karthik balach@gmail com - Gmail 2022-05-10 09-48-32

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
sgrimm: So first of all, I think it wouldn't be a disaster to lack test coverage here; these methods are pretty simple. Better to have tests than not, of course! But this is unlikely to be a particularly brittle or risky part of the code base.

That said, a couple approaches you could take to test these event listeners:

* Call the listener method directly from the test. For purposes of verifying what the method does, it doesn't matter if it's called by Spring's event infrastructure or not. The only thing that won't get tested this way is whether or not the method is annotated with `@EventListener`.
* To test whether the method is annotated, use reflection.
* If you want to make sure Spring really calls it, you could use `ApplicationContextRunner`. [This SO answer](https://stackoverflow.com/a/60058016) has an example.

I'd probably do the first two just to make the test lighter-weight; no point testing Spring's ability to call event listeners. Calling the handler method directly is hopefully straightforward. Checking an annotation on a method requires using the Java reflection API:

```kotlin
fun assertIsEventListener(obj: Any, method: String, eventClass: Class<*>) {
  assertNotNull(
      obj.javaClass.getMethod(method, eventClass).getAnnotation(EventListener::class.java),
      "$method(${eventClass.simpleName} missing EventListener annotation")
}

// ... in the test method
assertIsEventListener(service, "on", YourEvent::class.java)
```

Or, if you want to get fancier, you could use a reified type parameter for the event class:

```kotlin
inline fun <reified T> assertIsEventListener(obj: Any, method: String = "on") {
  assertNotNull(
      obj.javaClass.getMethod(method, T::class.java).getAnnotation(EventListener::class.java),
      "$method(${T::class.simpleName}) missing EventListener annotation")
}

// and then the call looks like this, assuming the method is called "on"
assertIsEventListener<YourEvent>(service)
```

If you do either of those, might be worth making it a top-level function (and maybe putting it in its own file a la the `mockUser()` function) so we can use it from other tests too.

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -31,7 +31,7 @@ class ProjectService(
           "Cannot add user that does not belong to project's organization")
     }
 
-    return dslContext.transactionResult { _ ->
+    return dslContext.transaction { _ ->
```
sgrimm: I don't think you need the `return` here.

src/main/resources/db/migration/common/R__TypeCodes.sql
```diff
@@ -134,10 +134,12 @@ INSERT INTO notification_criticalities (id, name)
 VALUES (1, 'Info'),
        (2, 'Warning'),
        (3, 'Error'),
-       (4, 'Success');
+       (4, 'Success')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
```
sgrimm, line 138: Oops! I should have caught that in my last review round.

sgrimm:
Don't worry about the blank images. That's probably just Gmail not knowing what to do with `localhost` URLs.

sgrimm:
Email notification part of this looks good too! Glad you figured out all the MJML stuff.

karthikbtf:
Thanks so much for the test code snippets, will leverage that.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -0,0 +1,153 @@
...
+  fun `should have event listener for User Added To Project event`() {
+    assertIsEventListener<UserAddedToProjectEvent>(service)
+  }
+
+  @Test
+  fun `should store a notification of type User Added To Organization`() {
+    insertUser(otherUserId)
+    insertOrganizationUser(otherUserId, organizationId, Role.CONTRIBUTOR)
+
+    assertEquals(0, notificationsDao.count())
```
sgrimm: These assertions could use messages.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -0,0 +1,153 @@
...
+    service.on(UserAddedToOrganizationEvent(otherUserId, organizationId, user.userId))
+
+    assertEquals(1, notificationsDao.count())
+    val notification = notificationsDao.fetchById(NotificationId(1)).firstOrNull()!!
+    assertEquals(NotificationType.UserAddedtoOrganization, notification.notificationTypeId)
+    assertEquals(otherUserId, notification.userId)
+    assertNull(notification.organizationId)
+    assertEquals("organization title", notification.title)
+    assertEquals("organization body", notification.body)
+    assertEquals(webAppUrls.organizationHome(organizationId), notification.localUrl)
```
sgrimm: It might be clearer to have the test construct a `NotificationsRow` with the expected values, then call `assertEquals` on the whole object. That way if more than one value doesn't match, the assertion failure will show all of them.

karthikbtf: Yes, makes sense, will update this.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -117,18 +117,36 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
...
+            notificationTypeId = NotificationType.UserAddedtoOrganization,
+            userId = otherUserId,
+            organizationId = null,
+            title = "organization title",
+            body = "organization body",
+            localUrl = webAppUrls.organizationHome(organizationId),
+            createdTime = Instant.EPOCH,
+            isRead = false)
+
+    val actualNotification = notificationsDao.fetchById(NotificationId(1)).firstOrNull()!!
```
sgrimm: The DAO classes have a `fetchOneById` method you can call to avoid the `firstOrNull()` here. Also, since the assertion now compares the whole object, you could get rid of the `!!` since `assertEquals` will fail if the object is null.

But there's a better option, I think: wrap the expected `NotificationsRow` in `listOf()` and call `notificationsDao.findAll()`. Then you wouldn't need the assertion on `count()` above, and if the count is greater than 1, the output of `assertEquals` will include a dump of the unexpected extra objects.

sgrimm: One other FYI, though it's not relevant if you use a list as the expected value: There is a `first()` method that throws an exception if the list is empty. You can call that instead of `firstOrNull()!!`.

karthikbtf: I like the list suggestion, it covers many cases and simplifies the test. Thanks very much.

karthikbtf:
> Looks good! I won't merge this in case there are other changes you're about to make; feel free to merge it at your leisure if not.

Thanks, Steve. I will merge it. (edited comment, removed stale text that shouldn't have been there)

PR #331, opened 2022-05-12
SW959 - App and Email notifications for Accession date notifications

karthikbtf:
Allow customers to receive in-app and email notifications for the various accession date notifications.
Leveraged existing code to publish relevant events.
Updated daily task to also look for accessions whose end date is today.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
@@ -18,7 +18,6 @@ import com.terraformation.backend.db.GerminationTestType
 import com.terraformation.backend.db.OrganizationNotFoundException
 import com.terraformation.backend.db.ProjectNotFoundException
 import com.terraformation.backend.db.UserNotFoundException
-import com.terraformation.backend.db.tables.daos.OrganizationsDao
 import com.terraformation.backend.email.model.AccessionDryingEnd
 import com.terraformation.backend.email.model.AccessionGerminationTest
 import com.terraformation.backend.email.model.AccessionMoveToDry
@@ -42,7 +41,6 @@ class EmailNotificationService(
     private val emailService: EmailService,
     private val facilityStore: FacilityStore,
     private val messages: Messages,
-    private val organizationsDao: OrganizationsDao,
     private val organizationStore: OrganizationStore,
     private val parentStore: ParentStore,
     private val projectStore: ProjectStore,
@@ -129,39 +127,33 @@ class EmailNotificationService(
   @EventListener
   fun on(event: AccessionMoveToDryEvent) {
     val organizationId = parentStore.getOrganizationId(event.accessionId)
-    val organization =
-        organizationsDao.fetchOneById(organizationId)
-            ?: throw OrganizationNotFoundException(organizationId)
+    val facilityName = parentStore.getFacilityName(event.accessionId)
     val accessionUrl = webAppUrls.fullAccession(event.accessionId, organizationId).toString()
     getRecipients(event.accessionId).forEach { user ->
       emailService.sendUserNotification(
           user,
           "accessionMoveToDry",
-          AccessionMoveToDry(config, event.accessionNumber, organization, accessionUrl))
+          AccessionMoveToDry(config, event.accessionNumber, facilityName, accessionUrl))
     }
   }
 
   @EventListener
   fun on(event: AccessionDryingEndEvent) {
     val organizationId = parentStore.getOrganizationId(event.accessionId)
-    val organization =
-        organizationsDao.fetchOneById(organizationId)
-            ?: throw OrganizationNotFoundException(organizationId)
+    val facilityName = parentStore.getFacilityName(event.accessionId)
     val accessionUrl = webAppUrls.fullAccession(event.accessionId, organizationId).toString()
     getRecipients(event.accessionId).forEach { user ->
       emailService.sendUserNotification(
           user,
           "accessionDryingEnd",
-          AccessionDryingEnd(config, event.accessionNumber, organization, accessionUrl))
+          AccessionDryingEnd(config, event.accessionNumber, facilityName, accessionUrl))
     }
   }
 
   @EventListener
   fun on(event: AccessionGerminationTestEvent) {
     val organizationId = parentStore.getOrganizationId(event.accessionId)
-    val organization =
-        organizationsDao.fetchOneById(organizationId)
-            ?: throw OrganizationNotFoundException(organizationId)
+    val facilityName = parentStore.getFacilityName(event.accessionId)
     val accessionUrl =
         webAppUrls
             .fullAccessionGerminationTest(event.accessionId, event.testType, organizationId)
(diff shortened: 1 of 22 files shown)


karthikbtf:
I have asked Nicole for updated copy, what's in this review is just placeholder material.

karthikbtf:
I have yet to test this off the running backend, that is pending.
State summary notifications is not covered in this review, will be a follow up.

karthikbtf:
was able to receive emails / notifications for one of the events. need to test the remaining 3 events.

DEV  Accession is scheduled to be moved from racks to drying cabinets today! - karthik balach@gmail com - Gmail 2022-05-13 09-56-22

Terraware App 2022-05-13 09-55-58

karthikbtf:
`/api/test/clock` worked great for testing. Thanks so much for setting this up!!

openapi.yaml
```diff
@@ -3785,6 +3785,10 @@ components:
           - Facility Idle
           - Facility Alert Requested
           - User Added to Project
+          - Accession Scheduled for Drying
```
sgrimm: Is the client actually using the notification type for anything? I'm a little nervous that enumerating the types in the API definition will make rolling out new notifications more of a heavyweight process since it'll require updating both the FE and BE (and maybe the mobile app too, if/when it's able to read notifications).

To be more concrete: Can we remove `notificationType` from `NotificationPayload`?

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -79,21 +88,114 @@ class AppNotificationService(
...
+    if (emailR.isEmpty()) {
+      return
+    }
+    val recipients =
+        projectStore.fetchEmailRecipients(projectId, false).toSet().mapNotNull {
+          userStore.fetchByEmail(it)
+        }
+    recipients.forEach { user ->
+      insert(notificationType, user.userId, organizationId, message, localUrl, organizationId)
+    }
```
sgrimm: Probably worth wrapping this in `dslContext.transaction` so we don't end up with notifications for some users but not others if there's an error.

src/main/kotlin/com/terraformation/backend/customer/event/AccessionDryingEndEvent.kt
```diff
@@ -0,0 +1,5 @@
+package com.terraformation.backend.customer.event
```
sgrimm: I'd put these events under the `seedbank` package rather than the `customer` package since they're related to the seed banking part of the app rather than to organization/user management.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -108,7 +125,90 @@ class EmailNotificationService(
...
+  }
+
+  private fun getProjectId(accessionId: AccessionId): ProjectId {
+    val facilityId = parentStore.getFacilityId(accessionId)!!
+    return parentStore.getProjectId(facilityId)!!
+  }
+
+  private fun getOrganizationId(accessionId: AccessionId): OrganizationId {
+    return parentStore.getOrganizationId(getProjectId(accessionId))!!
+  }
```
sgrimm: These could go in `ParentStore` since they may be useful elsewhere.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -108,7 +125,90 @@ class EmailNotificationService(
...
+    getRecipients(event.accessionId).forEach { user ->
+      emailService.sendUserNotification(
+          user,
+          "accessionWithdrawal",
+          AccessionWithdrawal(config, event.accessionNumber, organization, accessionUrl))
+    }
+  }
+
+  private fun getProjectId(accessionId: AccessionId): ProjectId {
+    val facilityId = parentStore.getFacilityId(accessionId)!!
```
sgrimm: Maybe `?: throw AccessionNotFoundException(accessionId)` instead of `!!`? Obviously we should never run into this in practice but if we do, seeing that exception in the logs will make it a bit quicker to figure out what's messed up.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -108,7 +125,90 @@ class EmailNotificationService(
...
+    return parentStore.getProjectId(facilityId)!!
+  }
+
+  private fun getOrganizationId(accessionId: AccessionId): OrganizationId {
+    return parentStore.getOrganizationId(getProjectId(accessionId))!!
+  }
+
+  private fun getRecipients(accessionId: AccessionId): List<IndividualUser> {
+    return projectStore.fetchEmailRecipients(getProjectId(accessionId)).toSet().mapNotNull {
```
sgrimm: Curious why you needed the `toSet()` here.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -74,4 +74,31 @@ class Messages {
       NotificationMessage(
           title = "You've been added to a new project!",
           body = "You are now a member of project $projectName.")
+
+  fun accessionMoveToDryNotification(accessionNumber: String): NotificationMessage =
```
sgrimm, line 78: More of a product question, but do we need to include the facility name in these messages? If you have 6 seed banks and you see "Accession 123 scheduled for drying" will you know where accession 123 is? I guess you can just click on the notification to find out, so maybe this is fine.

src/main/kotlin/com/terraformation/backend/seedbank/daily/DateNotificationTask.kt
```diff
@@ -47,18 +54,44 @@ class DateNotificationTask(
   private fun moveToDryingCabinet(after: TemporalAccessor, until: TemporalAccessor) {
     accessionStore.fetchDryingMoveDue(after, until).forEach { (number, id) ->
       insert(id, messages.dryingMoveDateNotification(number))
+      try {
+        eventPublisher.publishEvent(AccessionMoveToDryEvent(number, id))
+      } catch (e: Exception) {
+        log.error("Error handling AccessionMoveToDryEvent", e)
```
sgrimm: Might be worth including the accession ID in the exception message.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -749,6 +749,22 @@ class AccessionStore(
     }
   }
 
+  fun fetchDryingEndDue(
+      after: TemporalAccessor,
+      until: TemporalAccessor
+  ): Map<String, AccessionId> {
```
sgrimm, line 755: Out of scope for this PR, but I realize I got this API wrong. Should really be `Map<AccessionId, String>` since accession IDs are guaranteed to be globally unique but accession numbers aren't. I'd keep it as is here, though, so it's consistent with the existing code.

PR #332, opened 2022-05-13
Support unknown-length files in FileStore

sgrimm:
Previously, mostly due to the fact that we were using the simple `PutObject` S3
API, callers had to tell `FileStore.write()` the length of the file they wanted
to store.

Unfortunately, that doesn't play well with streaming uploads, where the incoming
request doesn't specify a content length.

Change `S3FileStore` to use the S3 multipart upload API, which supports uploading
a series of chunks and then explicitly marking the upload as complete after the
last chunk is written.

With that change in place, remove the `size` argument from `FileStore.write()`
and update callers accordingly.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/file/S3FileStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/file/S3FileStore.kt
@@ -1,7 +1,6 @@
 package com.terraformation.backend.file
 
 import com.terraformation.backend.config.TerrawareServerConfig
-import com.terraformation.backend.log.debugWithTiming
 import com.terraformation.backend.log.perClassLogger
 import java.io.InputStream
 import java.net.URI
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #332** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/332" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/332?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/file/S3FileStore.kt
```diff
@@ -89,11 +93,54 @@ class S3FileStore(config: TerrawareServerConfig, private val pathGenerator: Path
...
+        val completedMultipartUpload =
+            CompletedMultipartUpload.builder().parts(completedParts).build()
+        val completeMultipartUploadRequest =
+            CompleteMultipartUploadRequest.builder()
+                .bucket(bucketName)
+                .key(s3Key)
+                .multipartUpload(completedMultipartUpload)
+                .uploadId(uploadId)
+                .build()
+        s3Client.completeMultipartUpload(completeMultipartUploadRequest)
```
karthikbtf, line 134: guessing this step simply instructs s3 to assemble previously uploaded chunks into one object, basically telling s3 all data has been uploaded for this uploadId?

sgrimm: Correct, that's how it works.

PR #333, opened 2022-05-13
SW-909 Data model and permissions for uploaded files

sgrimm:
Add tables to represent files that are uploaded by users and processed
asynchronously by the server.

This doesn't include any logic specific to species list CSV uploads; it's just the
file-type-agnostic scaffolding that we can also use for other kinds of files
(accession lists, etc.) down the road.

Currently, the permissions model is simply, "You have permission to do anything
to your own uploads, but no permission to do anything with anyone else's." That's
what we want for the current product requirements; down the road we might want to
allow organization admins to work with their users' uploaded files, so this change
adds the usual fine-grained permission methods to let us do that later on.


sgrimm:
* main
  * **PR #333** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/333" target="_blank">Graphite</a>  👈
    * **PR #334** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/334" target="_blank">Graphite</a> 
      * **PR #335** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/335" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/333?utm_source=stack-comment).

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -51,6 +60,16 @@ val ENUM_TABLES =
             "storage_conditions",
             listOf("accessions\\.target_storage_condition", "storage_locations\\.condition_id")),
         EnumTable("timeseries_types", "timeseries\\.type_id"),
+        EnumTable("upload_problem_types", "upload_problems\\.type_id"),
+        EnumTable(
+            "upload_statuses",
+            listOf("uploads\\.status_id"),
```
karthikbtf, line 66: is there a preference when to use listOf vs no list, for a single expression?

sgrimm:
Gotta rebase this before deleting the species branch, apparently

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
sgrimm: The non-listOf constructor doesn't let you specify the enum name, though I guess it'd be harmless to add that.

PR #339, opened 2022-05-17
SW-964 App and Email notifications for Accession State Summaries

karthikbtf:
Allow user to receive in-app and email notifications for existing accession state summaries, generated by the daily task job.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
@@ -251,12 +251,12 @@ class EmailNotificationService(
   }
 
   @EventListener
-  fun on(@Suppress("UNUSED_PARAMETER") event: DateNotificationTask.PeriodProcessedStartEvent) {
+  fun on(@Suppress("UNUSED_PARAMETER") event: DateNotificationTask.StartedEvent) {
     accessionDatePendingEmails.clear()
   }
 
   @EventListener
-  fun on(@Suppress("UNUSED_PARAMETER") event: DateNotificationTask.PeriodProcessedFinishedEvent) {
+  fun on(@Suppress("UNUSED_PARAMETER") event: DateNotificationTask.SucceededEvent) {
     accessionDatePendingEmails.forEach { sendEmail(it) }
     accessionDatePendingEmails.clear()
   }
@@ -267,16 +267,12 @@ class EmailNotificationService(
   }
 
   @EventListener
-  fun on(
-      @Suppress("UNUSED_PARAMETER") event: StateSummaryNotificationTask.PeriodProcessedStartEvent
-  ) {
+  fun on(@Suppress("UNUSED_PARAMETER") event: StateSummaryNotificationTask.StartedEvent) {
     accessionStatePendingEmails.clear()
   }
 
   @EventListener
-  fun on(
-      @Suppress("UNUSED_PARAMETER") event: StateSummaryNotificationTask.PeriodProcessedFinishedEvent
-  ) {
+  fun on(@Suppress("UNUSED_PARAMETER") event: StateSummaryNotificationTask.SucceededEvent) {
     accessionStatePendingEmails.forEach { sendEmail(it) }
     accessionStatePendingEmails.clear()
   }
--- a/src/main/kotlin/com/terraformation/backend/seedbank/daily/DateNotificationTask.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/daily/DateNotificationTask.kt
@@ -42,7 +42,7 @@ class DateNotificationTask(
 
   override fun processPeriod(since: Instant, until: Instant) {
     log.info("Generating date update notifications for due dates since $since")
-    eventPublisher.publishEvent(PeriodProcessedStartEvent())
+    eventPublisher.publishEvent(StartedEvent())
 
     dslContext.transaction { _ ->
       moveToDryingCabinet(since, until)
@@ -51,7 +51,7 @@ class DateNotificationTask(
       withdraw(since, until)
     }
 
-    eventPublisher.publishEvent(PeriodProcessedFinishedEvent())
+    eventPublisher.publishEvent(SucceededEvent())
   }
 
   private fun moveToDryingCabinet(after: TemporalAccessor, until: TemporalAccessor) {
@@ -88,13 +88,13 @@ class DateNotificationTask(
   }
 
   /** Published when the period processed task begins */
-  class PeriodProcessedStartEvent
+  class StartedEvent
 
   /**
    * Published when the period processed task ends successfully, this event will not be published if
    * there are errors
    */
-  class PeriodProcessedFinishedEvent
+  class SucceededEvent
 
   /**
    * Published when the system has finished generating notifications for individual accessions
--- a/src/main/kotlin/com/terraformation/backend/seedbank/daily/StateSummaryNotificationTask.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/daily/StateSummaryNotificationTask.kt
@@ -49,7 +49,7 @@ class StateSummaryNotificationTask(
 
   override fun processPeriod(since: Instant, until: Instant) {
     log.info("Generating state update notifications")
-    eventPublisher.publishEvent(PeriodProcessedStartEvent())
+    eventPublisher.publishEvent(StartedEvent())
 
     val zonedSince = ZonedDateTime.ofInstant(since, clock.zone)
 
@@ -65,7 +65,7 @@ class StateSummaryNotificationTask(
           }
     }
 
-    eventPublisher.publishEvent(PeriodProcessedFinishedEvent())
+    eventPublisher.publishEvent(SucceededEvent())
   }
 
   private fun pending(facilityId: FacilityId, lastNotificationTime: ZonedDateTime) {
@@ -166,13 +166,13 @@ class StateSummaryNotificationTask(
   }
 
   /** Published when the period processed task begins */
-  class PeriodProcessedStartEvent
+  class StartedEvent
 
   /**
    * Published when the period processed task ends successfully, this event will not be published if
    * there are errors
    */
-  class PeriodProcessedFinishedEvent
+  class SucceededEvent
 
   /**
    * Published when the system has finished generating notifications with summaries of accession


src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -101,8 +112,7 @@ class EmailNotificationService(
     emailService.sendUserNotification(
         user,
         "userAddedToOrganization",
-        UserAddedToOrganization(config, admin, organization, organizationHomeUrl),
-        requireOptIn = false)
+        UserAddedToOrganization(config, admin, organization, organizationHomeUrl))
```
karthikbtf: This one change is unrelated.. should we skip sending this email if user opted out of email comms?

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -165,19 +169,59 @@ class AppNotificationService(
...
+      notificationType: NotificationType,
+      message: NotificationMessage,
+      localUrl: URI
+  ) {
     val projectId = parentStore.getProjectId(facilityId)!!
     val organizationId = parentStore.getOrganizationId(projectId)!!
-    val emailR = projectStore.fetchEmailRecipients(projectId, false)
-    if (emailR.isEmpty()) {
-      return
-    }
```
karthikbtf, line 180: debug code that got checked in earlier

src/main/resources/templates/email/accessionsAwaitingProcessing/body.ftlh.mjml
```diff
@@ -0,0 +1,29 @@
...
+<mjml>
+    <mj-include path="../head.ftlh.mjml" />
+
+    <mj-body>
+        <mj-include path="../logo.ftlh.mjml" />
+
+        <mj-section padding="0px">
+            <mj-column mj-class="body-wrapper">
+                <mj-text mj-class="text-headline06-bold">
+                    <!-- htmlmin:ignore --><#if numAccessions == 1><!-- htmlmin:ignore -->
```
karthikbtf, line 11: The  `<!-- htmlmin:ignore -->` padding is needed to enable FreeMarker conditional templating.
See https://documentation.mjml.io/#mj-raw

karthikbtf: All emails were tested, with count=1 and count != 1

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -64,6 +65,18 @@ class WebAppUrls(private val config: TerrawareServerConfig) {
     return UriBuilder.fromPath(accessionGerminationTestPath(accessionId, testType)).build()
   }
 
+  fun fullAccessions(organizationId: OrganizationId, state: AccessionState): URI {
+    return UriBuilder.fromUri(config.webAppUrl)
+        .path("/accessions")
+        .queryParam("stage", state)
```
karthikbtf, line 76: stage=<state> param is for UI to show corresponding views in the table. To be implemented in the frontend by me shortly (currently unsupported).

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
sgrimm: No, this one is supposed to be sent whether or not the user opted out (in fact, it's the reason there's a `requireOptIn` parameter at all!)

sgrimm: One reason is that adding a nonexistent user to an organization causes their account to be created, and they won't have been able to opt in at that point.

src/main/kotlin/com/terraformation/backend/seedbank/daily/StateSummaryNotificationTask.kt
```diff
@@ -128,13 +133,30 @@ class StateSummaryNotificationTask(
...
+              eventPublisher.publishEvent(
+                  AccessionsReadyForTestingEvent(facilityId, count, weeks, state))
+            }
+          }
+          AccessionState.Dried ->
+              eventPublisher.publishEvent(AccessionsFinishedDryingEvent(facilityId, count, state))
+          else -> log.warn("Unsupported state $state for notification events.")
+        }
+      } catch (e: Exception) {
+        log.error("Error handling notification event for state summary $state", e)
```
sgrimm: The fact that this is sending email makes it less obvious what the right behavior is here, but this is going to be a behavior difference from the previous version of the code if there's an exception while publishing the notifications.

Previously, the exception would have bubbled up to `DailyTaskRunner.runTask()`, which would cause it to not mark the current day as successfully processed. So the next daily task run would try again (that is, it'd be asked to scan for timestamps over a 48-hour time period instead of the usual 24-hour one).

But this exception handler will cause the run to look successful, so there'll be no retry.

Like I said, though, I'm not actually sure which way is better. If it were just creating notifications in our database, we could have logic to prevent duplicates, so letting the exception bubble up and retrying the next day would be better. But with email, doing that would run the risk of spamming people with redundant notifications if we'd already successfully gotten partway through the list.

src/main/resources/templates/email/accessionsAwaitingProcessing/body.ftlh.mjml
sgrimm: Ah, that's useful to know for the future. I bet that was a fun one to figure out.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -446,4 +458,105 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
...
+  fun `should store accessions awaiting processing notification`() {
+    // add a second user to check for multiple notifications
+    insertUser(otherUserId)
+    insertOrganizationUser(user.userId, organizationId, Role.CONTRIBUTOR)
+    insertProjectUser(user.userId, projectId, user.userId)
+
+    assertEquals(
+        0,
+        notificationsDao.count(),
+        "There should be no notifications before any notification event")
```
sgrimm: It occurs to me that this is probably not necessary; if the test database is in the wrong initial state, we'll get tons of failures all over the place.

But if it _were_ necessary here, it might be better to do `assertEquals(emptyList<NotificationsRow>(), notificationsDao.findAll(), "...")` so you'd see what the offending data was.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
karthikbtf: Got it, I will revert this change.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
karthikbtf: I will remove this check since the table is cleared before every test anyways.

src/main/kotlin/com/terraformation/backend/seedbank/daily/StateSummaryNotificationTask.kt
karthikbtf: I see what you mean. `DateNotificationTask` follows a similar pattern.
Looks like notifications are not rolled back either, in the current implementation, that could be fixed.
We could potentially queue up email requests and send them out at the end of the task run (if there are no errors) - spikes in SES sends are usually flagged (would need to throttle the sends), also not sure how we should handle email errors in this case (don't bubble up to avoid spamming users that already had emails sent).
Perhaps a per-user task but that would be overkill?

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -226,10 +239,50 @@ class EmailNotificationService(
...
+              AccessionsFinishedDrying(
+                  config, event.numAccessions, organization.name!!, accessionsUrl)))
+    }
+  }
+
+  @EventListener
+  fun on(@Suppress("UNUSED_PARAMETER") event: DailyTaskTimeArrivedEvent) {
+    accessionDatePendingEmails.clear()
+    accessionStatePendingEmails.clear()
+  }
```
sgrimm: I think this makes the intended order of execution a little fuzzy. Event listeners are called synchronously and by default they're called in an undefined order. This could be called before the notifications are generated or after all the processing is done.

Maybe it would make sense to add `DateNotificationTask.StartedEvent` (and similarly for the state summaries) and publish it before doing any real work. Then the code here could clear the pending list based on that event rather than `DailyTaskTimeArrivedEvent`.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -253,4 +306,11 @@ class EmailNotificationService(
       userStore.fetchByEmail(it)
     }
   }
+
+  companion object data
```
sgrimm: Don't need `companion object` here. I assume the intent here was just to define a data class, but FYI, this will end up creating a companion object named `data` that has no methods or properties, and then defining a (non-data) class called `EmailRequest`.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
karthikbtf: I was going to do that,, I should have done that.. thanks for confirming this approach. will do.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
karthikbtf: oops, not what i wanted w.r.t. an empty object named `data`. Thanks for this info.

src/main/kotlin/com/terraformation/backend/seedbank/daily/DateNotificationTask.kt
```diff
@@ -86,11 +87,14 @@ class DateNotificationTask(
     accessionNotificationStore.insertDateNotification(accessionId, message)
   }
 
+  /** Published when the period processed task begins */
+  class PeriodProcessedStartEvent
```
sgrimm: These names feel a little awkward to me. Maybe `StartedEvent` and `SucceededEvent`?

karthikbtf: `StartedEvent` and `SucceededEvent` sound better/cleaner to me. Thanks for the suggestion.

PR #343, opened 2022-05-20
SW-999 Stop autocreating seed banks for new orgs

sgrimm:
Now that we're adding the ability for users to manage their own seed banks, there
is no longer a need to automatically create a seed bank when a user creates a new
organization.

However, since the autocreated seed bank project is special (it is configured as
an organization-wide project) and we don't currently allow clients to create that
kind of project, we still want to autocreate the project and the site that clients
will use when they create new seed bank facilities.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
@@ -76,7 +76,8 @@ class FacilityStore(
       type: FacilityType,
       name: String,
       description: String? = null,
-      maxIdleMinutes: Int = defaultMaxIdleMinutes
+      maxIdleMinutes: Int = defaultMaxIdleMinutes,
+      createStorageLocations: Boolean = true,
   ): FacilityModel {
     requirePermissions { createFacility(siteId) }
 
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #343** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/343" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/343?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
```diff
@@ -76,7 +76,8 @@ class FacilityStore(
       type: FacilityType,
       name: String,
       description: String? = null,
-      maxIdleMinutes: Int = defaultMaxIdleMinutes
+      maxIdleMinutes: Int = defaultMaxIdleMinutes,
+      createStorageLocations: Boolean = true,
```
karthikbtf, line 80: Just for my understanding - when would/should this be false? Does this need to be exposed in FacilitiesController or AdminController? Guessing the UI would use defaults (so sensor kit onboarding makes sense).

sgrimm: My guess is that we will realize that if a user is setting up a non-Terraformation seed bank, they might not have precisely 3 fridges and precisely 3 freezers and we will have a quick followup project to add a "manage storage locations" section to the UI, which will involve not auto-creating these locations for non-TF seed banks. I could definitely add this to the API but I didn't want to add a useless setting the FE needed to worry about in the meantime.

sgrimm:
Yes, the site ID should always be the default one that gets created for a new organization. Its name will always be "Seed Bank". (We don't have a concept of site types.)

src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
karthikbtf: Thanks for the clarification. No point adding this to the REST API unless there's a use-case.

PR #351, opened 2022-05-23
Limit the size of timeseries values requests

sgrimm:
We had an incident last week where a seed bank lost Internet connectivity for a
while and tried to send all its sensor readings when it came back online. This
resulted in a huge `/api/v1/timeseries/values` request which timed out and got
retried repeatedly, leading to high database load.

The device manager is getting updated to discard some of its sensor data when it
loses connectivity for an extended period, but to mitigate against a similar
situation in the future, we also need to limit the number of values in a single
request on the server side.

Longer term, we'll probably want more sophisticated rate limiting, but this should
at least prevent a repeat of the recent incident.

Implementation note: this checks the payload size explicitly and throws an
exception rather than using Spring's built-in validation feature because the
built-in validation doesn't have a special case for "request too large" and
returns HTTP 400. HTTP 413 is a more approprate status code to return here.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.device.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.terraformation.backend.api.ApiResponse413
 import com.terraformation.backend.api.ApiResponseSimpleSuccess
 import com.terraformation.backend.api.DeviceManagerAppEndpoint
 import com.terraformation.backend.api.ErrorDetails
@@ -17,6 +18,9 @@ import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Schema
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import java.time.Instant
+import javax.validation.constraints.Size
+import javax.ws.rs.WebApplicationException
+import javax.ws.rs.core.Response
 import org.springframework.dao.DuplicateKeyException
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #351** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/351" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/351?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
```diff
@@ -182,7 +195,16 @@ data class TimeseriesValuePayload(
 
 data class CreateTimeseriesRequestPayload(val timeseries: List<CreateTimeseriesEntry>)
 
-data class RecordTimeseriesValuesRequestPayload(val timeseries: List<TimeseriesValuesPayload>)
+data class RecordTimeseriesValuesRequestPayload(
+    @Size(
+        max = MAX_VALUES,
+    )
```
karthikbtf, line 201: There is an explicit check in the controller api, just curious... how the Size validation works here or in conjunction with the explicit check.

sgrimm: Not a silly question at all. This annotation is what you'd use if you wanted to tell Spring to enforce this limit. However, Spring only does validation on a payload if the parameter to the controller method is annotated with `@Valid`.

But the OpenAPI schema generator pays attention to this annotation. So the effect of putting this here but not putting `@Valid` on the method parameter is that the payload schema has a `maxItems` property:
```yaml
    RecordTimeseriesValuesRequestPayload:
      required:
      - timeseries
      type: object
      properties:
        timeseries:
          maxItems: 1000
          minItems: 0
          type: array
          items:
            $ref: '#/components/schemas/TimeseriesValuesPayload'
```
In other words: this is for documentation purposes.

PR #352, opened 2022-05-23
SW-1004 add facilityName to accessions search fields

karthikbtf:
Allow searching accessions by facility name and viewing facility name in the accession response.
This will be used in the accessions UI view.


sgrimm:
Do we really need this? The client can pass in `facility_name` as a search field, no need for the alias, I think.

The existing `Name` aliases are mostly for backward compatibility with older versions of the search API that didn't have the ability to navigate the data model.

sgrimm:
I realize I promised I would write up something on how to transition searches over to the generic search API -- will go do that now.

karthikbtf:
Thanks for the info. I am able to retrieve this with the `facility_name` field.

PR #353, opened 2022-05-23
SW-1000 Add device templates

sgrimm:
To support user-selectable PV systems at containerized seed banks, we need to have
a list of the available options, and we need to know the specific device settings
to use when the user picks one.

Introduce the concept of a "device template" which has all the same settings as a
real device, plus a category and a human-readable description. The UI will get a
list of all the PV templates using the new `/api/v1/devices/templates` endpoint,
and when the user picks one, it will call the existing `/api/v1/devices` endpoint
to create a new device with the specified settings.

Device templates can be created and edited using a new page in the admin UI. Only
super admins are allowed to modify the list of templates.


sgrimm:
* main
  * **PR #353** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/353" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/353?utm_source=stack-comment).

PR #354, opened 2022-05-24
SW-1014 Drop collectors table

sgrimm:
Previously, collectors (both primary and secondary) were tracked per facility, the
idea being that a collector in a particular location could collect multiple
accessions and we'd want to associate them with a single identity.

In practice, though, we treat the collector name as a free-form text field and
there are no features in the app that expose the concept of a collector having a
unique identity. Even the accession APIs present collectors as strings rather than
as separate entities.

This is going to be a problem if/when we decide to support accessions that aren't
associated with specific facilities. It's also a problem if accessions are moved
between facilities (the collectors will be created on the old facility initially).

Update the data model to reflect the way the app works: the primary collector is a
plain text field on the accession, and the secondary collectors are per-accession
text values.

Since the APIs already worked that way, this shouldn't require any changes on the
client side. There is a client-visible difference in the way secondary collectors
are addressed in the search API, but the web app doesn't support searching on
secondary collectors at the moment so no change is needed to account for that.


sgrimm:
* main
  * **PR #354** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/354" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/354?utm_source=stack-comment).

PR #356, opened 2022-05-25
SW-1017 New summary endpoint for all seedbanks in an organization

karthikbtf:
Allow UI to retrieve a comprehensive summary of all seedbanks within an organization.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
@@ -73,7 +73,6 @@ class SummaryController(
             accessionStore.countInState(
                 facilityId, AccessionState.Withdrawn, sinceAfter = startOfWeek))
   }
-
   @GetMapping
   @Operation(
       summary =
@@ -86,17 +85,14 @@ class SummaryController(
       @Schema(description = "If set, return summary on that specific seedbank.")
       facilityId: FacilityId?
   ): SummaryResponse {
-    if (facilityId != null) {
-      if (organizationId != null) {
-        throw IllegalArgumentException("Specify either one of organizationId or facilityId.")
-      }
-      return getSummary(facilityId)
+    return when {
+      facilityId != null && organizationId == null -> getSummary(facilityId)
+      facilityId == null && organizationId != null -> getSummary(organizationId)
+      else -> throw IllegalArgumentException("...")
     }
+  }
 
-    if (organizationId == null) {
-      throw IllegalArgumentException("One of organizationId or facilityId is required.")
-    }
-
+  private fun getSummary(organizationId: OrganizationId): SummaryResponse {
     val now = ZonedDateTime.now(clock)
     val startOfDay = now.atMostRecent(config.dailyTasks.startTime)
     val startOfWeek = startOfDay.atMostRecent(DayOfWeek.MONDAY)
--- a/src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
@@ -1625,7 +1625,7 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     @Test
     fun `counts active accessions by facility`() {
       store.create(AccessionModel(facilityId = facilityId))
-      assertEquals(store.countActive(facilityId, ZonedDateTime.now(clock)), 1)
+      assertEquals(1, store.countActive(facilityId, ZonedDateTime.now(clock)))
     }
 
     @Test
@@ -1640,7 +1640,7 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     @Test
     fun `counts accession families by facility`() {
       store.create(AccessionModel(facilityId = facilityId, family = "test"))
-      assertEquals(store.countFamilies(facilityId, ZonedDateTime.now(clock)), 1)
+      assertEquals(1, store.countFamilies(facilityId, ZonedDateTime.now(clock)))
     }
 
     @Test
@@ -1655,8 +1655,8 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     @Test
     fun `counts accessions state by facility`() {
       store.create(AccessionModel(facilityId = facilityId, state = AccessionState.AwaitingCheckIn))
-      assertEquals(store.countInState(facilityId, AccessionState.AwaitingCheckIn), 1)
-      assertEquals(store.countInState(facilityId, AccessionState.Dried), 0)
+      assertEquals(1, store.countInState(facilityId, AccessionState.AwaitingCheckIn))
+      assertEquals(0, store.countInState(facilityId, AccessionState.Dried))
     }
 
     @Test
@@ -1675,17 +1675,19 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
 
       store.create(AccessionModel(facilityId = facilityId, state = AccessionState.AwaitingCheckIn))
       assertEquals(
+          0,
           store.countInState(
               facilityId,
               AccessionState.AwaitingCheckIn,
               sinceAfter = ZonedDateTime.now(clock).plusDays(1)),
-          0)
+      )
       assertEquals(
+          1,
           store.countInState(
               facilityId,
               AccessionState.AwaitingCheckIn,
               sinceBefore = ZonedDateTime.now(clock).plusDays(1)),
-          1)
+      )
     }
 
     @Test
@@ -1700,7 +1702,7 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
     @Test
     fun `counts active accessions by organization`() {
       store.create(AccessionModel(facilityId = facilityId))
-      assertEquals(store.countActive(organizationId, ZonedDateTime.now(clock)), 1)
+      assertEquals(1, store.countActive(organizationId, ZonedDateTime.now(clock)))
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


karthikbtf:
This was mostly a copy/paste of existing summary code with some changes to the query. Worked correctly in my local dev environment - a bit unsure on efficiency of queries.

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
```diff
@@ -71,6 +72,46 @@ class SummaryController(
             accessionStore.countInState(
                 facilityId, AccessionState.Withdrawn, sinceAfter = startOfWeek))
   }
+
+  @GetMapping("/organization/{organizationId}")
```
sgrimm: In terms of the client-facing API, I wonder if it'd be cleaner to combine this with the per-facility endpoint and move the ID to the query string. That is, `/api/v1/seedbank/summary?facilityId=X` and `/api/v1/seedbank/summary?organizationId=X`. We could keep the existing facility-in-the-path API for backward compatibility until the client was updated.

The implementation might still involve two separate code paths, but from the client's point of view, it's asking for the same information in both cases, just with different scopes.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -637,6 +641,38 @@ class AccessionStore(
...
 
+  fun countActive(organizationId: OrganizationId, asOf: TemporalAccessor): Int {
+    val statesByActive = AccessionState.values().groupBy { it.toActiveEnum() }
+
+    val query =
+        dslContext
+            .select(DSL.count())
+            .from(ACCESSIONS)
+            .join(PROJECTS)
+            .on(PROJECTS.ORGANIZATION_ID.eq(organizationId))
```
sgrimm: The join criterion should be the thing that links two joined tables together. If there were multiple projects in an organization, I believe this query would count each accession against each project. In other words, the count would be multiplied by the number of projects. It just happens that right now, the number of projects is always 1, so the multiplication is a no-op.

You'd want to restrict the organization ID in the `WHERE` clause instead:

```
SELECT COUNT(*)
FROM accessions
JOIN facilities ON accessions.facility_id = facilities.id
JOIN sites ON facilities.site_id = sites.id
JOIN projects ON sites.project_id = projects.id
WHERE projects.organization_id = X
```

However! There is a little jOOQ trick you can use to make this simpler: [implicit joins](https://www.jooq.org/doc/latest/manual/sql-building/sql-statements/select-statement/implicit-join/). For tables with non-nullable foreign key columns, jOOQ generates methods to jump across the relationships into the other table, and it'll automatically add the correct `JOIN`s to the query. So this could look like

```kotlin
dslContext
    .select(DSL.count())
    .from(ACCESSIONS)
    .where(ACCESSIONS.facilities().sites().projects().ORGANIZATION_ID.eq(organizationId)
    ...
```

(The method names are plural because they're just the table names, which are plural.) That will produce the same SQL as what I wrote above.

sgrimm: And if you take this approach, then you could extract the facility/org condition out into a parameter, like

```kotlin
private fun countActive(condition: Condition, asOf: TemporalAccessor): Int {
  dslContext.select(DSL.count()).from(ACCESSIONS).where(condition) [...rest of query...]
}

fun countActive(facilityId: FacilityId, asOf: TemporalAccessor): Int {
  return countActive(ACCESSIONS.FACILITY_ID.eq(facilityId), asOf)
}
```

so the rest of the query construction doesn't need to be duplicated.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -637,6 +641,38 @@ class AccessionStore(
     return log.debugWithTiming("Active accessions query") { query.fetchOne()?.value1() ?: 0 }
   }
 
+  fun countActive(organizationId: OrganizationId, asOf: TemporalAccessor): Int {
+    val statesByActive = AccessionState.values().groupBy { it.toActiveEnum() }
```
sgrimm: Apparently I neglected to add a permission check in the existing methods! This should require read permission on the organization (and the existing methods should require it on the facility).

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
karthikbtf: Yes, that does look much cleaner. I will update the client-facing API.

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
```diff
@@ -71,6 +73,66 @@ class SummaryController(
...
+      @RequestParam("organizationId", required = false)
+      @Schema(description = "If set, return summary on all seedbanks for that organization.")
+      organizationId: OrganizationId?,
+      @RequestParam("facilityId", required = false)
+      @Schema(description = "If set, return summary on that specific seedbank.")
+      facilityId: FacilityId?
+  ): SummaryResponse {
+    if (facilityId != null) {
+      if (organizationId != null) {
+        throw IllegalArgumentException("Specify either one of organizationId or facilityId.")
```
karthikbtf: tad ugly?

sgrimm: Yeah, kind of. Here's the best alternative I can come up with off the top of my head:

```kotlin
return if (facilityId != null && organizationId == null) {
  getSummary(facilityId)
} else if (facilityId == null && organizationId != null) {
  getSummary(organizationId)
} else {
  throw IllegalArgumentException("Must specify organization or facility ID but not both")
}

private fun getSummary(organizationId: OrganizationId): SummaryResponse {
  // extract the body of the existing method here
}
```

sgrimm: But what you have is fine too. It's a little ugly but is easy enough to understand.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -928,4 +792,91 @@ class AccessionStore(
 
     return "%08d%03d".format(todayAsLong, suffix)
   }
+
+  private fun countActive(condition: Condition, asOf: TemporalAccessor): Int {
```
sgrimm, line 796: Cool, I think this ends up being cleaner. I guess it also gives us an easy way to do summaries across arbitrary sets of facilities if we ever need to group them together using tags or whatever.

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -1608,6 +1611,159 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      assertThrows<FacilityNotFoundException> {
+        store.countActive(facilityId, ZonedDateTime.now(clock))
+      }
+    }
+
+    @Test
+    fun `counts active accessions by facility`() {
+      store.create(AccessionModel(facilityId = facilityId))
+      assertEquals(store.countActive(facilityId, ZonedDateTime.now(clock)), 1)
```
sgrimm: Expected value should come first in `assertEquals`.

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -1608,6 +1611,159 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
+    }
+
+    @Test
+    fun `rejects count accession families by facility when user cannot read facility`() {
+      every { user.canReadFacility(any()) } returns false
+
+      assertThrows<FacilityNotFoundException> {
+        store.countFamilies(facilityId, ZonedDateTime.now(clock))
```
sgrimm, line 1636: Good call using `ZonedDateTime` here since it's what the controllers pass in. In theory it shouldn't matter what specific time type the caller uses but better to test it.

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -1608,6 +1611,159 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      assertThrows<FacilityNotFoundException> {
+        store.countInState(
+            facilityId, AccessionState.Pending, sinceBefore = ZonedDateTime.now(clock))
+      }
+    }
+
+    @Test
+    fun `counts accessions state by facility for duration`() {
+      every { clock.instant() } returns Instant.now()
```
sgrimm, line 1674: FYI, you will want to be careful calling `Instant.now()` in tests. It is totally fine here, but in cases where you're populating a `TIMESTAMP WITH TIME ZONE` column in the database and then reading it back and comparing for equality, you can run into precision problems where the local system's timestamps are higher-precision than PostgreSQL's representation and the equality test fails because the value that gets read back from the database was truncated.

Even worse: the precision is platform-dependent. So this manifests as tests passing fine locally (on OS X, where the system clock has microsecond precision just like PostgreSQL) but failing in CI (on Linux, where the clock has nanosecond precision).

All that said, you're only using the system clock to do range comparisons here, so I think this should be fine; the `sinceAfter` value will get truncated the same as the timestamp on the accession so it should be in the correct range. The above is more for your future reference.

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
sgrimm: Oh, I guess my alternative might be better as a `when`:

```kotlin
return when {
  facilityId != null && organizationId == null -> getSummary(facilityId)
  facilityId == null && organizationId != null -> getSummary(organizationId)
  else -> throw IllegalArgumentException("...")
}
```

karthikbtf: this is cleaner, i will use this, thanks!

PR #357, opened 2022-05-26
SW-1018 add organizationId scope to ValuesController.ListAllFieldValues

karthikbtf:
List all field values needs to return values relevant to an organization. Added support for this.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
@@ -2,7 +2,6 @@ package com.terraformation.backend.search
 
 import com.terraformation.backend.log.debugWithTiming
 import com.terraformation.backend.log.perClassLogger
-import com.terraformation.backend.search.SearchTable.SearchScope
 import com.terraformation.backend.search.field.SearchField
 import javax.annotation.ManagedBean
 import org.jooq.Condition
@@ -187,8 +186,8 @@ class SearchService(private val dslContext: DSLContext) {
    */
   fun fetchAllValues(
       fieldPath: SearchFieldPath,
-      limit: Int = 50,
       searchScopes: List<SearchScope>,
+      limit: Int = 50,
   ): List<String?> {
     if (searchScopes.isEmpty()) {
       throw IllegalArgumentException("No search scopes specified.")
@@ -209,7 +208,7 @@ class SearchService(private val dslContext: DSLContext) {
           val permsCondition = conditionForPermissions(fieldPath.searchTable)
           val searchConditions =
               searchScopes.mapNotNull { fieldPath.searchTable.conditionForScope(it) }
-          val conditions = listOfNotNull(permsCondition, *searchConditions.toTypedArray())
+          val conditions = listOfNotNull(permsCondition) + searchConditions
 
           val fullQuery =
               dslContext
--- a/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
@@ -1,9 +1,7 @@
 package com.terraformation.backend.search
 
 import com.terraformation.backend.db.EnumFromReferenceTable
-import com.terraformation.backend.db.FacilityId
 import com.terraformation.backend.db.FuzzySearchOperators
-import com.terraformation.backend.db.OrganizationId
 import com.terraformation.backend.search.field.AliasField
 import com.terraformation.backend.search.field.BigDecimalField
 import com.terraformation.backend.search.field.BooleanField
@@ -292,8 +290,4 @@ abstract class SearchTable(val fuzzySearchOperators: FuzzySearchOperators) {
   ) =
       UpperCaseTextField(
           fieldName, displayName, databaseField, this, nullable, fuzzySearchOperators)
-
-  sealed interface SearchScope
-  data class OrganizationIdScope(val organizationId: OrganizationId) : SearchScope
-  data class FacilityIdScope(val facilityId: FacilityId) : SearchScope
 }
--- a/src/main/kotlin/com/terraformation/backend/search/table/AccessionsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/AccessionsTable.kt
@@ -14,7 +14,10 @@ import com.terraformation.backend.db.tables.references.GERMINATION_TESTS
 import com.terraformation.backend.db.tables.references.SPECIES
 import com.terraformation.backend.db.tables.references.STORAGE_LOCATIONS
 import com.terraformation.backend.db.tables.references.WITHDRAWALS
+import com.terraformation.backend.search.FacilityIdScope
 import com.terraformation.backend.search.FieldNode
+import com.terraformation.backend.search.OrganizationIdScope
+import com.terraformation.backend.search.SearchScope
 import com.terraformation.backend.search.SearchTable
 import com.terraformation.backend.search.SublistField
 import com.terraformation.backend.search.field.SearchField
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
@@ -5,9 +5,10 @@ import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.FacilityId
 import com.terraformation.backend.db.OrganizationId
 import com.terraformation.backend.db.StorageCondition
+import com.terraformation.backend.search.FacilityIdScope
+import com.terraformation.backend.search.OrganizationIdScope
 import com.terraformation.backend.search.SearchFieldPrefix
 import com.terraformation.backend.search.SearchService
-import com.terraformation.backend.search.SearchTable
 import com.terraformation.backend.search.api.HasSearchNode
 import com.terraformation.backend.search.api.SearchNodePayload
 import com.terraformation.backend.search.table.SearchTables
@@ -75,11 +76,8 @@ class ValuesController(
       @RequestBody payload: ListAllFieldValuesRequestPayload
   ): ListAllFieldValuesResponsePayload {
     val limit = 100
-    val organizationIdScope =
-        if (payload.organizationId != null) SearchTable.OrganizationIdScope(payload.organizationId)
-        else null
-    val facilityIdScope =
-        if (payload.facilityId != null) SearchTable.FacilityIdScope(payload.facilityId) else null
+    val organizationIdScope = payload.organizationId?.let { OrganizationIdScope(it) }
+    val facilityIdScope = payload.facilityId?.let { FacilityIdScope(it) }
     val searchScopes = listOfNotNull(organizationIdScope, facilityIdScope)
     if (searchScopes.isEmpty()) {
       throw IllegalArgumentException("One of organizationId or facilityId is required.")
@@ -88,7 +86,7 @@ class ValuesController(
     val values =
         payload.fields.associateWith { fieldName ->
           val searchField = rootPrefix.resolve(fieldName)
-          val values = searchService.fetchAllValues(searchField, limit, searchScopes)
+          val values = searchService.fetchAllValues(searchField, searchScopes, limit)
 
           val partial = values.size > limit
           AllFieldValuesPayload(values.take(limit), partial)
--- a/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
@@ -33,16 +33,19 @@ import com.terraformation.backend.db.tables.pojos.StorageLocationsRow
 import com.terraformation.backend.db.tables.references.ACCESSIONS
 import com.terraformation.backend.mockUser
 import com.terraformation.backend.search.AndNode
+import com.terraformation.backend.search.FacilityIdScope
 import com.terraformation.backend.search.FieldNode
 import com.terraformation.backend.search.NoConditionNode
 import com.terraformation.backend.search.NotNode
 import com.terraformation.backend.search.OrNode
+import com.terraformation.backend.search.OrganizationIdScope
 import com.terraformation.backend.search.SearchDirection
 import com.terraformation.backend.search.SearchFieldPath
 import com.terraformation.backend.search.SearchFieldPrefix
 import com.terraformation.backend.search.SearchFilterType
 import com.terraformation.backend.search.SearchNode
 import com.terraformation.backend.search.SearchResults
+import com.terraformation.backend.search.SearchScope
 import com.terraformation.backend.search.SearchService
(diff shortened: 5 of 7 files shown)


src/main/kotlin/com/terraformation/backend/search/SearchFieldPath.kt
```diff
@@ -211,10 +211,6 @@ class SearchFieldPath(val prefix: SearchFieldPrefix, val searchField: SearchFiel
       throw IllegalArgumentException("$otherPrefix is not a prefix of $this")
     }
 
-    if (otherPrefix.isRoot) {
-      return this
-    }
-
```
karthikbtf, line 217: this is already on line 206

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -181,9 +182,14 @@ class SearchService(private val dslContext: DSLContext) {
...
    * by at most 1 element, which callers can use to detect that the number of values exceeds the
    * limit.
+   * @param condition Condition to narrow down scope of search values
    * @return A list of values, which may include `null` if the field is not mandatory.
    */
-  fun fetchAllValues(fieldPath: SearchFieldPath, limit: Int = 50): List<String?> {
+  fun fetchAllValues(
+      fieldPath: SearchFieldPath,
+      limit: Int = 50,
+      searchScope: SearchScope? = null
```
karthikbtf: open to suggestions on how better to do this.

sgrimm:
Ah, man, this has been on my should-do-it-some-day list for months: https://terraformation.atlassian.net/browse/SW-552

I've actually tried to do it a couple times and stopped because it turned out to be trickier than I felt like I could justify spending time on given we didn't have an actual need for it previously. But now we do, of course.

I'll look at this change in detail later today.

The approach I was trying to take previously was to make it super general so you could basically pass in an arbitrary search query and ask for the possible values of a given field. Maybe being more restrictive about it by defining a fixed set of scopes like you're doing here is a good simplifying idea.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -181,9 +182,14 @@ class SearchService(private val dslContext: DSLContext) {
    * @param limit Maximum number of results desired. The return value may be larger than this limit
    * by at most 1 element, which callers can use to detect that the number of values exceeds the
    * limit.
+   * @param condition Condition to narrow down scope of search values
```
sgrimm: I guess you tried this a different way initially!

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -197,12 +203,23 @@ class SearchService(private val dslContext: DSLContext) {
...
+          val scopeCondition =
+              if (searchScope != null) fieldPath.searchTable.conditionForScope(searchScope)
+              else null
+          val conditions =
+              when {
+                (permsCondition != null && scopeCondition != null) ->
+                    permsCondition.and(scopeCondition)
+                permsCondition != null -> permsCondition
+                else -> scopeCondition
+              }
```
sgrimm: You can pass a list of conditions to `where()` so this could be `listOfNotNull(permsCondition, scopeCondition)`.

src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
```diff
@@ -287,4 +291,7 @@ abstract class SearchTable(val fuzzySearchOperators: FuzzySearchOperators) {
   ) =
       UpperCaseTextField(
           fieldName, displayName, databaseField, this, nullable, fuzzySearchOperators)
+
+  open class SearchScope
```
sgrimm: Maybe make this a sealed class so we can get exhaustiveness checks on `when` expressions.

Or maybe better, a sealed interface.

src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
```diff
@@ -77,7 +79,13 @@ class ValuesController(
...
         payload.fields.associateWith { fieldName ->
           val searchField = rootPrefix.resolve(fieldName)
-          val values = searchService.fetchAllValues(searchField, limit)
+          val values =
+              searchService.fetchAllValues(
+                  searchField,
+                  limit,
+                  if (payload.organizationId != null)
+                      SearchTable.OrganizationIdScope(payload.organizationId)
+                  else null)
```
sgrimm: I find this a little hard to scan visually because one branch is on multiple lines and the other isn't. Maybe extract it out into a variable?

src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
```diff
@@ -138,7 +147,11 @@ data class AllFieldValuesPayload(
     val partial: Boolean
 )
 
-data class ListAllFieldValuesRequestPayload(val facilityId: FacilityId, val fields: List<String>)
+data class ListAllFieldValuesRequestPayload(
+    val facilityId: FacilityId?,
+    val fields: List<String>,
+    val organizationId: OrganizationId?
```
sgrimm, line 154: [nitpick] The order of these properties is inconsistent with the other payload class. Personally I prefer alphabetical order like you're doing here, but either way it'd be nice if they were consistent.

karthikbtf:
> I am still thinking about this change at a high level, but in the meantime, a few things I noticed as I was reading through the code.

Got it, if there's a better way we can close this. I will follow up on the feedback in the meantime. Thanks for looking!

src/main/kotlin/com/terraformation/backend/search/table/AccessionsTable.kt
```diff
@@ -154,6 +154,14 @@ class AccessionsTable(
     return ACCESSIONS.FACILITY_ID.`in`(currentUser().facilityRoles.keys)
   }
 
+  override fun conditionForScope(scope: SearchScope): Condition? {
+    return when (scope) {
+      is OrganizationIdScope ->
+          ACCESSIONS.facilities().sites().projects().ORGANIZATION_ID.eq(scope.organizationId)
```
sgrimm, line 163: We'll need to be careful about where this gets called. It's not a problem in `fetchAllValues` since the query is otherwise just a simple single-table SELECT, but if we call this in `NestedQueryBuilder` we'll be in for some pain because these implicit joins may overlap with the joins that get explicitly added by the query construction code.

sgrimm:
Initially when I looked at this, my mental model was that it was widening the results from per-facility to per-organization. But in fact, it's narrowing the results, because the default behavior is to include everything you have permission to see. So if we go this way, I wonder if we need another scope for per-facility values. I guess we'd want that to take a list of facilities to support the user selecting a set of facilities to search.

karthikbtf:
looks like i messed up this merge, will take a look

karthikbtf:
> Initially when I looked at this, my mental model was that it was widening the results from per-facility to per-organization. But in fact, it's narrowing the results, because the default behavior is to include everything you have permission to see. So if we go this way, I wonder if we need another scope for per-facility values. I guess we'd want that to take a list of facilities to support the user selecting a set of facilities to search.

Sounds good, I can add that.

From a UI perspective, this is what happens:
1) UI calls list all values for 'state' and 'facility_name' (for the organization) so it can populate the filter menu values
2) UI calls list values (filtered by selected facilities, if applicable) to enable/disable filter menu values
 Example: Under state filter, all the states would be disabled except for say 'Drying' - if the selected facilities have accessions which are all Drying.

The only list all values we can do by facility id is perhaps for 'primary collectors' - in the create accession form, to narrow down primary collectors specific to the seed bank. This would perhaps prohibit user from selecting a primary collector from another seed bank?

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceTest.kt
```diff
@@ -1356,6 +1356,62 @@ class SearchServiceTest : DatabaseTest(), RunsAsUser {
...
+              facilityId = FacilityId(2200),
+              modifiedBy = user.userId,
+              modifiedTime = Instant.now(),
+          ))
+
+      val expectedUnscoped = listOf("ABCDEFG", "OtherProject", "OtherProject22", "XYZ")
+      val expectedScopedOrg = listOf("ABCDEFG", "OtherProject", "XYZ")
+      val expectedScopedOtherOrg = listOf("OtherProject22")
+
+      assertEquals(expectedUnscoped, searchService.fetchAllValues(accessionNumberField))
```
sgrimm: These assertions could use messages to differentiate them.

sgrimm:
Collectors are no longer facility-based as of commit 0b2f82b, so maybe there's no actual use for a facility-scoped version of this. Storage locations were my other example, but those are kind of fuzzy right now because the storage location names are identical across facilities, so you get the exact same result from "list all storage location names" regardless of scope. We can figure that one out once we start letting users edit their storage locations.

sgrimm:
Given how this is used, maybe the unscoped behavior isn't useful. Should we make this always require a scope?

sgrimm:
(Sorry for the comment spam!) I almost wonder if we even need this at all, given the use cases you listed.

1. Facilities can be queried using the regular search API. Alternately, I believe the frontend already has the facility list at hand (from the `/api/v1/organizations` response) and can use that. Not sure if that would introduce too much special-case logic in the frontend code, though.
2. Accession states are a fixed list. They're included in the OpenAPI schema so we could use code generation to output a list of them in a form the client code could reference.

karthikbtf:
> (Sorry for the comment spam!) I almost wonder if we even need this at all, given the use cases you listed.
> 
> 1. Facilities can be queried using the regular search API. Alternately, I believe the frontend already has the facility list at hand (from the `/api/v1/organizations` response) and can use that. Not sure if that would introduce too much special-case logic in the frontend code, though.
> 2. Accession states are a fixed list. They're included in the OpenAPI schema so we could use code generation to output a list of them in a form the client code could reference.

I made a mistake in my explanation, educated myself a bit more.
Looks like the UI lists all values for multi-selectable filters - which by default are 'state' and 'facility_name'.
User can customize the filter/columns.
I selected all the filter/columns and here's what the list-all-values request payload looks like.

```
{
  "fields": [
    "state",
    "facility_name",
    "active",
    "endangered",
    "rare",
    "sourcePlantOrigin",
    "viabilityTestType",
    "targetStorageCondition",
    "storageCondition",
    "storageLocationName",
    "withdrawalDestination",
    "withdrawalPurpose",
    "germinationTestType",
    "germinationSeedType",
    "germinationTreatment",
    "germinationSubstrate"
  ],
  "organizationId": 1
}
```

Most of them seem like pre-determined values and the others can be fetched using facilities API.
The list all values API is definitely super useful.

karthikbtf:
> Given how this is used, maybe the unscoped behavior isn't useful. Should we make this always require a scope?

Yes, we can definitely make this required and relax it later if there's such a use-case.

src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
```diff
@@ -73,11 +75,20 @@ class ValuesController(
       @RequestBody payload: ListAllFieldValuesRequestPayload
   ): ListAllFieldValuesResponsePayload {
     val limit = 100
+    val organizationIdScope =
+        if (payload.organizationId != null) SearchTable.OrganizationIdScope(payload.organizationId)
+        else null
```
karthikbtf: I didn't make this mandatory for now, older clients are still passing in facilityId

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -203,10 +207,9 @@ class SearchService(private val dslContext: DSLContext) {
               field.selectFields + listOf(field.orderByField.`as`(DSL.field("order_by_field")))
           val searchTable = fieldPath.searchTable
           val permsCondition = conditionForPermissions(fieldPath.searchTable)
-          val scopeCondition =
-              if (searchScope != null) fieldPath.searchTable.conditionForScope(searchScope)
-              else null
-          val conditions = listOfNotNull(permsCondition, scopeCondition)
+          val searchConditions =
+              searchScopes.mapNotNull { fieldPath.searchTable.conditionForScope(it) }
+          val conditions = listOfNotNull(permsCondition, *searchConditions.toTypedArray())
```
sgrimm: This can be `listOfNotNull(permsCondition) + searchConditions` since `searchConditions` already has nulls filtered out.

src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
```diff
@@ -294,4 +295,5 @@ abstract class SearchTable(val fuzzySearchOperators: FuzzySearchOperators) {
 
   sealed interface SearchScope
   data class OrganizationIdScope(val organizationId: OrganizationId) : SearchScope
+  data class FacilityIdScope(val facilityId: FacilityId) : SearchScope
```
sgrimm: What do you think of moving these to the top level? They aren't really associated with search tables, but rather with search operations.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -182,14 +182,18 @@ class SearchService(private val dslContext: DSLContext) {
...
    * limit.
-   * @param searchScope Optional scoping data for the search
+   * @param searchScope Scoping data for the search
    * @return A list of values, which may include `null` if the field is not mandatory.
    */
   fun fetchAllValues(
       fieldPath: SearchFieldPath,
       limit: Int = 50,
-      searchScope: SearchScope? = null
+      searchScopes: List<SearchScope>,
```
sgrimm, line 189: Could move this up one position so call sites can be `fetchAllValues(path, scopes)` rather than `fetchAllValues(path, searchScopes = scopes)`. If you move it using IntelliJ's method arguments refactoring tool (command-F6) it will automatically update all the call sites.

src/main/kotlin/com/terraformation/backend/seedbank/api/ValuesController.kt
```diff
@@ -75,15 +75,20 @@ class ValuesController(
       @RequestBody payload: ListAllFieldValuesRequestPayload
   ): ListAllFieldValuesResponsePayload {
     val limit = 100
-
-    val searchScope =
+    val organizationIdScope =
         if (payload.organizationId != null) SearchTable.OrganizationIdScope(payload.organizationId)
         else null
+    val facilityIdScope =
+        if (payload.facilityId != null) SearchTable.FacilityIdScope(payload.facilityId) else null
```
sgrimm: Could express this as `payload.facilityId?.let { FacilityIdScope(it) }`

src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
karthikbtf: Should these be in a separate file perhaps?

sgrimm: Yeah, a separate file would make sense.

PR #358, opened 2022-05-26
Remove unused dependency from IndividualUser

sgrimm:
We moved the accessor functions for notification permissions to `ParentStore`,
but we were still passing a `NotificationStore` argument to `IndividualUser`.

Remove it, which also means we can remove it from `UserStore`.


sgrimm:
* main
  * **PR #358** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/358" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/358?utm_source=stack-comment).

PR #360, opened 2022-05-27
SW-1003 Update copy for FacilityIdle app and email notifications

karthikbtf:
Allow user to get updated copy of app and email notifications when facility is idle.

Screenshots:

#### email
email

#### app
app

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
@@ -302,6 +302,7 @@ class ProjectStore(
         .on(ORGANIZATION_USERS.ORGANIZATION_ID.eq(PROJECTS.ORGANIZATION_ID))
         .leftJoin(PROJECT_USERS)
         .on(USERS.ID.eq(PROJECT_USERS.USER_ID))
+        .and(PROJECTS.ID.eq(PROJECT_USERS.PROJECT_ID))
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -311,5 +311,6 @@ class ProjectStore(
         .and(optedInCondition)
         .fetch(USERS.EMAIL)
         .filterNotNull()
+        .distinct()
```
karthikbtf: without this, i get duplicate emails.. was trying to parse the query here to see if something got folded out

sgrimm: I think I see a bug in the query that might explain this: the join condition on `PROJECT_USERS` should check both the user and project IDs. So

```
.leftJoin(PROJECT_USERS)
.on(USERS.ID.eq(PROJECT_USERS.USER_ID))
```

should be

```
.leftJoin(PROJECT_USERS)
.on(USERS.ID.eq(PROJECT_USERS.USER_ID))
.and(PROJECTS.ID.eq(PROJECT_USERS.PROJECT_ID))
```

Give that a try and see if it gets rid of the duplicates.

karthikbtf: That worked, thanks, makes sense. Will update the code.

PR #361, opened 2022-05-27
SW-1007 Add device managers to data model and API

sgrimm:
Add a table for tracking device manager information. The table has a
combination of locally-maintained data and data from the Balena API, though the
actual Balena API calls are not included in this change.

In addition, the facilities table gains a "connection state" column that indicates
whether or not a device manager has been associated with the facility. This is
included in the facility information returned by `/api/v1/organizations` and
`/api/v1/facilities`. A later change will add a way for clients to tell the server
that device maanger configuration is finished.

`GET /api/v1/devices/managers` allows searching for device managers by short code.
The search is currently case-sensitive; we don't know what the short codes are
going to actually look like yet, so it's unclear if they should be case-sensitive.

`POST /api/v1/devices/managers/{id}/connect` allows connecting a device manager to
a facility. This also creates a new Keycloak user and updates the connection state
of the facility. A later change will also make this configure the device via the
Balena API, but for now there's no Balena interaction.

One thing of note here is the way permissions work. Device managers are an unusual
case in that they need to be visible to users before the system knows which
organizations own them. So the permission logic here is basically that a device
manager is visible to all users if it's not connected yet, but then once it's
connected, it's only visible to users who have permission to see its assigned
facility.

Also addresses #SW-1008


sgrimm:
* main
  * **PR #361** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/361" target="_blank">Graphite</a>  👈
    * **PR #362** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/362" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/361?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -396,6 +401,32 @@ data class IndividualUser(
...
+  }
+
+  override fun canUpdateDeviceManager(deviceManagerId: DeviceManagerId): Boolean {
+    val facilityId = parentStore.getFacilityId(deviceManagerId)
+    return if (facilityId != null) {
+      when (facilityRoles[facilityId]) {
+        Role.OWNER,
+        Role.ADMIN -> true
+        else -> false
+      }
```
karthikbtf, line 424: reuse `canUpdateFacility(facilityId)`?

src/main/kotlin/com/terraformation/backend/device/DeviceManagerService.kt
```diff
@@ -0,0 +1,60 @@
...
+      }
+
+      facilityStore.updateConnectionState(
+          facilityId, FacilityConnectionState.NotConnected, FacilityConnectionState.Connected)
+
+      val apiUser = userStore.createApiClient(organizationId, "Balena ${manager.balenaUuid}")
+      val userId = apiUser.userId
+      val token = userStore.generateOfflineToken(userId)
+
+      balenaClient.configureDeviceManager(balenaId, facilityId, token)
```
karthikbtf, line 50: just curious, how does balena client recognize the api token, is there some api/registration as part of the seed bank setup? I'm guessing that's what was referred to by creating a  keyclock user for the device manager/?

src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
```diff
@@ -0,0 +1,103 @@
...
+    @Schema(
+        description =
+            "The facility this device manager is connected to, or null if it is not connected.")
+    val facilityId: FacilityId?,
+    @Schema(
+        description =
+            "If an update is being downloaded or installed, its progress as a percentage. Not " +
+                "present if no update is in progress.",
+        minimum = "0",
+        maximum = "100")
```
karthikbtf, line 76: is this set as part of the connect api?

src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
```diff
@@ -0,0 +1,103 @@
...
+    private val deviceManagerStore: DeviceManagerStore,
+) {
+  @GetMapping
+  fun getDeviceManagers(
+      @RequestParam("shortCode") shortCode: String
+  ): GetDeviceManagersResponsePayload {
+    val manager =
+        deviceManagerStore.fetchOneByShortCode(shortCode)?.let { DeviceManagerPayload(it) }
+
+    return GetDeviceManagersResponsePayload(listOfNotNull(manager))
```
karthikbtf, line 37: thoughts on `GET /devices/managers/shortcode/<code>` returning a single device manager?

src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
karthikbtf: wondering if maybe the connect api triggers an async device manager update.

sgrimm: Ah, yes, I should update the Confluence page to talk about how this is going to work. Thanks for asking this.

src/main/kotlin/com/terraformation/backend/device/DeviceManagerService.kt
sgrimm: So first off, "token" here is an OAuth2 "offline refresh token" which is basically a credential you can pass to Keycloak to get a valid bearer token to pass in an `Authentication` header. High level, it's basically an API key the device manager uses to authenticate to the backend. Keycloak (like any OAuth2 identity provider) has a way to generate these tokens for users. So we create the user in Keycloak, then immediately ask Keycloak to generate a token. This is already implemented in `UserStore` if you're curious about the mechanics.

Token in hand, we'll make a call to the Balena API to put the token in an environment variable which will then be visible to the device manager.

karthikbtf: Got it, this is a refresh-token/api-token for the device manager to leverage (to fetch bearer tokens) when making api requests to terraware. I mistook this is as the token for authenticating with balena in order to make the connect request. Thanks for the clarification.

src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
karthikbtf: ok this is a generic search by params api, with just one supported param for the moment. ignore my previous comment.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/361?utm_source=stack-comment).


Job Reference: lrLBwCbWCoXI94e7HKST

PR #362, opened 2022-05-27
SW-1009 Add device managers and devices to admin UI

sgrimm:
Update the admin UI to include device managers and devices. The initial use case
here is dev environments: developers can create device managers locally to test
the sensor kit setup workflow. It thus doesn't include the ability to view and
edit absolutely everything about devices, just the fields that will be needed
for sensor kit setup testing.

The basic flow will look something like:

1. Create a new seed bank facility, either via API or in the admin UI.
2. Create a new device manager in the admin UI's device managers page.
3. Connect the device manager to the facility, either via API or by editing the
   device manager's facility ID in the admin UI. (The behavior is a a little
   different: the admin UI won't create a Keycloak user.)
4. Add devices from the admin UI's facility page. There are convenience links
   to prefill the device settings for a couple different device types.

The device manager page can be used to manipulate some values that will come
from Balena in the real implementation. The two most useful are the "Online?"
setting and the "Update Progress" percentage.

There's currently no way to disconnect a device manager from a facility since it's
not a product requirement and would involve figuring out what the behavior should
be (e.g., what happens to the Keycloak user?) In dev environments, it should be
simple enough to create a new facility and device manager to test with.

Also addresses #SW-1010

### DIFF
--- a/src/main/resources/templates/admin/facility.html
+++ b/src/main/resources/templates/admin/facility.html
@@ -1,6 +1,48 @@
 <!DOCTYPE HTML>
 <html xmlns:th="http://www.thymeleaf.org">
-<head th:replace="/admin/header :: head"/>
+<head th:replace="/admin/header :: head">
+    <script th:fragment="additionalScript">
+        function fillOut(event, settings) {
+            event.preventDefault();
+            document.getElementById("deviceType").value = settings.type;
+            document.getElementById("make").value = settings.make;
+            document.getElementById("model").value = settings.model;
+            document.getElementById("deviceName").value = settings.name || '';
+            document.getElementById("address").value = settings.address || '';
+            document.getElementById("protocol").value = settings.protocol || '';
+        }
+
+        addEventListener('load', () => {
+            const configurations = {
+                blueIon: {
+                    type: "BMU",
+                    make: "Blue Ion",
+                    model: "LV",
+                    name: "BMU-1",
+                    address: "192.168.10.40",
+                    protocol: "modbus"
+                },
+                omniSense: {
+                    type: "sensor",
+                    make: "OmniSense",
+                    model: "S-11"
+                }
+            };
+
+            for (let elementName in configurations) {
+                document.getElementById(elementName)
+                        .addEventListener('click', (event) => fillOut(event, configurations[elementName]));
+            }
+        });
+    </script>
+
+    <style th:fragment="additionalStyle">
+        label {
+            display: inline-block;
+            min-width: 5em;
+        }
+    </style>
+</head>
 <body>
 
 <span th:include="/admin/header :: top"/>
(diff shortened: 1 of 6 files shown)


sgrimm:
* main
  * **PR #361** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/361" target="_blank">Graphite</a> 
    * **PR #362** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/362" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/362?utm_source=stack-comment).

src/main/resources/templates/admin/facility.html
```diff
@@ -1,6 +1,48 @@
...
+            document.getElementById("deviceType").value = settings.type;
+            document.getElementById("make").value = settings.make;
+            document.getElementById("model").value = settings.model;
+            document.getElementById("deviceName").value = settings.name || '';
+            document.getElementById("address").value = settings.address || '';
+            document.getElementById("protocol").value = settings.protocol || '';
+        }
+
+        addEventListener('load', () => {
+            const configurations = {
```
sgrimm, line 16: At some point it'd maybe make sense to pull this from the `device_templates` table but this was already getting complex enough as is.

sgrimm:
SW-1010 is also covered here (let's see if Jira picks up this comment automatically!)

src/main/resources/templates/admin/facility.html
```diff
@@ -45,12 +87,98 @@ <h2 th:text="|Facility ${facility.name} (${facility.id})|">Facility (id)</h2>
     <span th:text="${facility.type.displayName}">Seed Bank</span>
 </div>
 
+<h3>Devices</h3>
+
+<p th:if="${deviceManager} != null">
+    <a th:href="|${prefix}/deviceManagers/${deviceManager.id}|"
+       th:text="|Device manager ${deviceManager.id} connected.|">Device manager 1 connected.</a>
```
karthikbtf, line 94: Does the `Device manager 1 connected` get replaced by what's evaluated in `th:text="|....\|"` ?

sgrimm: Yeah, it's a Thymeleaf thing. The idea is that you could build a mocked-up HTML page in a WYSIWYG editor with dummy content, and then add the `th:` attributes to the tags to turn it into a working template that still looks correct in the editor. I'm not really sure why I put in the dummy content, honestly, since I'm writing the templates by hand anyway. Just habit.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/362?utm_source=stack-comment).


Job Reference: lrLBwCbWCoXI94e7HKST

PR #365, opened 2022-05-31
SW-1012 Add endpoint to mark facility as configured

sgrimm:
When the user has finished assigning sensor locations, the web app needs to
mark the facility as fully configured so that timeseries data and alerts from
the device manager will be processed.

Add connection state as an editable facility field in the admin UI.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/FacilitiesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/FacilitiesController.kt
@@ -29,6 +29,7 @@ import java.time.Instant
 import java.time.temporal.ChronoUnit
 import javax.ws.rs.InternalServerErrorException
 import javax.ws.rs.WebApplicationException
+import javax.ws.rs.core.Response
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #365** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/365" target="_blank">Graphite</a>  👈
    * **PR #366** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/366" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/365?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/api/FacilitiesController.kt
```diff
@@ -189,6 +191,27 @@ class FacilitiesController(
...
+              "and alerts from the device manager. Only valid if the facility's connection " +
+              "state is `Connected`.")
+  @PostMapping("/{facilityId}/configured")
+  fun postConfigured(@PathVariable facilityId: FacilityId): SimpleSuccessResponsePayload {
+    try {
+      facilityStore.updateConnectionState(
+          facilityId, FacilityConnectionState.Connected, FacilityConnectionState.Configured)
+      return SimpleSuccessResponsePayload()
+    } catch (e: IllegalStateException) {
+      throw WebApplicationException("Facility's devices are not being configured.")
```
karthikbtf: Response.Status.CONFLICT?

sgrimm: Oof, you're right. As I started typing this line of code, I fully intended to add that, and apparently forgot that intention by the time I finished. Thanks for catching it!

PR #366, opened 2022-05-31
SW-1019 Ignore device manager data during configuration

sgrimm:
When the user is in the process of placing sensors, we don't want to record bogus
temperature and humidity data or broadcast useless alerts about temperatures being
out of range.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
@@ -9,7 +9,9 @@ import com.terraformation.backend.api.ResponsePayload
 import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessOrError
 import com.terraformation.backend.customer.db.FacilityStore
+import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.db.DeviceId
+import com.terraformation.backend.db.FacilityConnectionState
 import com.terraformation.backend.db.TimeseriesType
 import com.terraformation.backend.db.tables.pojos.TimeseriesRow
 import com.terraformation.backend.device.db.TimeseriesStore
@@ -22,6 +24,7 @@ import javax.validation.constraints.Size
 import javax.ws.rs.WebApplicationException
 import javax.ws.rs.core.Response
 import org.springframework.dao.DuplicateKeyException
+import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #365** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/365" target="_blank">Graphite</a> 
    * **PR #366** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/366" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/366?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
```diff
@@ -72,6 +81,18 @@ class TimeseriesController(
...
 
+    val facilityNotConfigured =
+        payload.timeseries
+            .asSequence()
+            .map { it.deviceId }
+            .distinct()
+            .any {
+              parentStore.getFacilityConnectionState(it) != FacilityConnectionState.Configured
+            }
+    if (facilityNotConfigured) {
```
karthikbtf, line 92: novice question - who calls this api (specific device manager or some balena integration)? will we ever get a payload relevant to multiple device ids? if so, should we only ignore those timeseries values that are for unconfigured device ids?

src/test/kotlin/com/terraformation/backend/device/api/TimeseriesControllerTest.kt
```diff
@@ -85,22 +92,26 @@ internal class TimeseriesControllerTest : RunsAsUser {
...
+                        "ts2",
+                        valuePayloads("v20"),
+                        "Unexpected error while saving value"),
+                    TimeseriesValuesErrorPayload(
+                        deviceId1,
+                        "ts2",
+                        valuePayloads("v21"),
+                        "Already have a value with this timestamp"),
+                    TimeseriesValuesErrorPayload(
+                        deviceId2, "ts1", valuePayloads("v30", "v31"), "Timeseries not found"))))
```
karthikbtf, line 114: this (specifically L#91) answers my question about requests containing mutiple device ids.

src/main/kotlin/com/terraformation/backend/device/api/TimeseriesController.kt
sgrimm: This is called by the device manager.

Whether to ignore the whole request or not is a good question. I opted for ignoring the whole thing just so the HTTP 202 always means "nothing got recorded." In _theory_, a device manager could submit a request whose devices span multiple facilities, but that should never happen with our current product so in practice, we should never run into a situation where we'd only want to ignore part of a request.

karthikbtf: Makes sense.

PR #378, opened 2022-06-07
SW-971 Add client for Balena API

sgrimm:
In preparation for syncing device manager status information from Balena, add a
client that can make the necessary Balena API requests.

It must be explicitly enabled in the server configuration, and the configuration
must include a valid API key and the ID of an existing Balena fleet. If it's not
enabled, terraware-server will not attempt to communicate with Balena.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
@@ -219,7 +219,7 @@ internal class LiveBalenaClientTest {
     val response =
         client.sendRequest<Map<String, Any?>>(
             LiveBalenaClient.FLEET_PATH,
-            body = mapOf("app_name" to fleetName, "device_type" to "raspberrypi4-64"))
+            body = mapOf("app_name" to fleetName, "device_type" to RASPBERRY_PI_DEVICE_TYPE_SLUG))
     val fleetId = response.body().get()["id"].toString().toLong()
 
     log.info("Created fleet $fleetId")


sgrimm:
* main
  * **PR #378** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/378" target="_blank">Graphite</a>  👈
    * **PR #379** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/379" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/378?utm_source=stack-comment).

sgrimm:
Fixed the test. I'd been running it with the Balena environment variables and neglected to make sure it still worked without them after reorganizing the setup/shutdown code.

src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
```diff
@@ -0,0 +1,275 @@
...
+  private var raspberryPiDeviceTypeId: Long = -1
+
+  private var createdDevice: Boolean = false
+  private var createdFleet: Boolean = false
+
+  @BeforeAll
+  fun setUp() {
+    // Skip all the tests if no API key is available.
+    val apiKey =
+        System.getenv("TEST_BALENA_API_KEY") ?: throw AssumptionViolatedException("No API key")
```
karthikbtf, line 73: learned something new here, `AssumptionViolatedException`

src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
```diff
@@ -0,0 +1,275 @@
...
+import org.junit.jupiter.api.assertThrows
+
+/**
+ * End-to-end tests for the Balena client. This talks to the actual Balena API. To run it, you will
+ * need a valid Balena API key. The key doesn't have to be associated with an organization and can
+ * be created from a personal Balena account.
+ *
+ * Basic process for getting a key (note that Balena's UI may have changed since this was written):
+ *
+ * 1. Go to https://balena.io/ and click the "Sign up" button to sign up for an account.
```
karthikbtf, line 30: will we have one for our github ci/cd at some point?

src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
```diff
@@ -0,0 +1,275 @@
...
+        "access_token",
+        client.getDeviceEnvironmentVar(deviceId, LiveBalenaClient.TOKEN_ENV_VAR_NAME),
+        "Refresh token after retry")
+  }
+
+  private fun createFleet(fleetName: String): Long {
+    val response =
+        client.sendRequest<Map<String, Any?>>(
+            LiveBalenaClient.FLEET_PATH,
+            body = mapOf("app_name" to fleetName, "device_type" to "raspberrypi4-64"))
```
karthikbtf: use constant `RASPBERRY_PI_DEVICE_TYPE_SLUG`?

src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
sgrimm: FYI, usually you wouldn't do it explicitly like this. The more common idiom for checking whether to skip a test is to use the `assume` functions, e.g.,
```kotlin
assumeNotNull(whatever)
```
but in this case I also wanted to assign the value somewhere.

src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
sgrimm: It's a reasonable question. We could! But that means our CI builds would spuriously fail whenever Balena is having a hiccup, which, from my past experience with tests that depend on external services, can get super irritating if it happens with any regularity. I don't really know how often Balena has hiccups, though.

We could also add logic to the CI workflow to only run this test for real on PRs that contain edits to `LiveBalenaClient`.

I'll file a task to follow up on that. To start off with, I'm inclined to leave it a dev-environment-only thing.

karthikbtf: Sounds good.

PR #382, opened 2022-06-08
Upgrade to Spring Boot 2.7.0

sgrimm:
This is a manual upgrade rather than a Renovate bot one because there are
several Spring-related dependencies that need to be upgraded at the same time, and
also because the upgrade introduces a backward incompatibility with existing login
sessions.

Add a test so that future Spring upgrades will fail in CI if they would cause
login sessions to become unusable.


sgrimm:
* main
  * **PR #382** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/382" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/382?utm_source=stack-comment).

PR #385, opened 2022-06-10
SW-1034 Automatically create automations

sgrimm:
The device manager needs to know the acceptable ranges of various timeseries so it
can generate notifications if they go out of range.

The correct ranges, and the relevant timeseries, depend on various attributes of
the device, including its name. In the future, we'll likely also want to let users
edit the ranges on a per-device basis. So the server needs to be the source of
truth for the ranges.

Add logic to create and remove automations based on changes to device data.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/device/DeviceServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/device/DeviceServiceTest.kt
@@ -1,0 +1,258 @@
+package com.terraformation.backend.device
+
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.terraformation.backend.RunsAsUser
+import com.terraformation.backend.customer.db.AutomationStore
+import com.terraformation.backend.customer.db.ParentStore
+import com.terraformation.backend.customer.model.AutomationModel.Companion.DEVICE_ID_KEY
+import com.terraformation.backend.customer.model.AutomationModel.Companion.LOWER_THRESHOLD_KEY
+import com.terraformation.backend.customer.model.AutomationModel.Companion.SENSOR_BOUNDS_TYPE
+import com.terraformation.backend.customer.model.AutomationModel.Companion.TIMESERIES_NAME_KEY
+import com.terraformation.backend.customer.model.AutomationModel.Companion.TYPE_KEY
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #385** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/385" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/385?utm_source=stack-comment).

sgrimm:
The test class here is kind of ugly. Let me know if it's so bad you can't follow it. The alternative I thought of was to make it configuration-driven, reading the test cases from a YAML file or something, but I wasn't sure if that would be worth the effort.

sgrimm:
Also, the hardwired implementation here is not what we'll want long-term, but for the initial release it's hopefully good enough.

src/test/kotlin/com/terraformation/backend/device/DeviceServiceTest.kt
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  @Test
+  fun `create does not create automations for sensor with unrecognized name`() {
+    service.create(omniSenseRow)
+
+    assertAutomationConfigsEqual(emptyMap())
+  }
+
+  @MethodSource("getKnownSensors")
```
karthikbtf, line 166: is this an implied getter on `knownSensors`?

karthikbtf:
> The test class here is kind of ugly. Let me know if it's so bad you can't follow it. The alternative I thought of was to make it configuration-driven, reading the test cases from a YAML file or something, but I wasn't sure if that would be worth the effort.

Test class is readable to me.

src/test/kotlin/com/terraformation/backend/device/DeviceServiceTest.kt
sgrimm: Yeah. Sadly JUnit (being a Java framework) doesn't know about Kotlin properties, so we have to point it at the getter.

sgrimm:
Automations are indeed the things that trigger alerts. Some docs: https://github.com/terraware/terraware-devices/tree/HW-13#automations

karthikbtf:
> Automations are indeed the things that trigger alerts. Some docs: https://github.com/terraware/terraware-devices/tree/HW-13#automations
Awesome, thanks!

PR #394, opened 2022-06-15
SW-1050 Embed sensor id in monitoring urls specific to sensor

karthikbtf:
Allow user to land on dashboard with sensor pre-selected.

DEV  Ambient Middle cannot be detected  - karthik balach@gmail com - Gmail 2022-06-15 16-53-50

### DIFF
--- a/src/main/resources/templates/email/accession/link.ftlh.mjml
+++ b/src/main/resources/templates/email/accession/link.ftlh.mjml
@@ -1,9 +1,6 @@
-<!-- <#-- @ftlvariable name="accessionUrl" type="String" --> -->
-<mjml>
-    <mj-text mj-class="text-body03">
-        Please click the button below to go to your Terraware account where you can view
-        the accession.
-    </mj-text>
-    <mj-button mj-class="btn-productive-primary-md" href="${accessionUrl?no_esc}">Go to Terraware
-    </mj-button>
-</mjml>
+<mj-text mj-class="text-body03">
+    Please click the button below to go to your Terraware account where you can view
+    the accession.
+</mj-text>
+<mj-button mj-class="btn-productive-primary-md" href="${accessionUrl?no_esc}">Go to Terraware
+</mj-button>
--- a/src/main/resources/templates/email/accessions/link.ftlh.mjml
+++ b/src/main/resources/templates/email/accessions/link.ftlh.mjml
@@ -1,9 +1,6 @@
-<!-- <#-- @ftlvariable name="accessionsUrl" type="String" --> -->
-<mjml>
-    <mj-text mj-class="text-body03">
-        Please click the button below to go to your Terraware account where you can view
-        the accessions.
-    </mj-text>
-    <mj-button mj-class="btn-productive-primary-md" href="${accessionsUrl?no_esc}">Go to Terraware
-    </mj-button>
-</mjml>
+<mj-text mj-class="text-body03">
+    Please click the button below to go to your Terraware account where you can view
+    the accessions.
+</mj-text>
+<mj-button mj-class="btn-productive-primary-md" href="${accessionsUrl?no_esc}">Go to Terraware
+</mj-button>


src/main/resources/templates/email/facility/link.ftlh.mjml
```diff
@@ -1,11 +1,7 @@
...
-
-</mjml>
\ No newline at end of file
+<mj-text mj-class="text-body03">
+    Please click the button below to go to your Terraware account where you can view
+    the seed bank.
+</mj-text>
+<mj-button mj-class="btn-productive-primary-md" href="${facilityMonitoringUrl?no_esc}">
+    Go to Terraware
+</mj-button>
```
karthikbtf, line 7: So... this may not be the best change. This file inclusion is not coming through otherwise (no text/button in emails).
Something to do with including this file within an existing `<mj-column>`. With the `mjml` tags the inclusion doesn't work, without the `mjml` tags the markup gets commented out. I went crazy trying to figure out what's going on... will need to spend more time maybe.
Didn't test this pattern on accession/s yet.

karthikbtf: All the `link.ftlh.mjml` files are included within the context of a parent level `EmailTemplateModel` that contains `facilityMonitoringUrl`. That's probably why it works without the explicit directive.

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -89,15 +90,27 @@ class WebAppUrls(private val config: TerrawareServerConfig) {
...
-  fun facilityMonitoring(facilityId: FacilityId): URI {
-    return UriBuilder.fromPath("/monitoring/${facilityId.value}").build()
+  fun facilityMonitoring(facilityId: FacilityId, device: DevicesRow? = null): URI {
+    return UriBuilder.fromPath("/monitoring/${facilityId.value}")
+        .let { devicePath(it, device) }
+        .build()
+  }
+
+  private fun devicePath(uriBuilder: UriBuilder, device: DevicesRow?): UriBuilder {
+    return if (device?.deviceType == "sensor") {
```
sgrimm, line 109: I guess this special case is needed because sensors are treated specially in the UI? So, for example, if the notification is from the PV system, you wouldn't want to go to a page that just showed the PV device's graphs?

src/main/resources/templates/email/facility/link.ftlh.mjml
sgrimm: FYI, that directive is just for IntelliJ's benefit so it can give you autocomplete and detect typos in variable names. The MJML and FreeMarker renderers ignore it. IntelliJ's editor doesn't know whether a given file is only ever included from elsewhere, so I added the directive here.

As long as this way works, though, it's fine.

I guess we really ought to add a unit test for the email rendering like you did for the in-app notifications. Should have done that from the get-go. Obviously it can't (or at least shouldn't) test for the exact HTML but it could at least verify that there's a link to the expected URL somewhere in the email body.

(Not an appropriate thing to add to this PR, though.)

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
karthikbtf: Correct, this change is special-cased for sensors. 
The dashboard shows two charts today, the PV chart and the sensors chart (both always shown).
The chart that shows sensors defaults to the first sensor in the sensors list (dropdown selection), this change allows us to pre-select the sensor in the sensors chart (based on url param).

SBD - Sensor Kit Display V2 – Figma 2022-06-16 13-50-47

We don't have a page showing just the PV chart or sensor chart today (that is a good idea though, I will bring it up for UX cleanup).

src/main/resources/templates/email/facility/link.ftlh.mjml
karthikbtf: I agree about the tests, I can create a JIRA for this.

PR #396, opened 2022-06-16
SW-1055 Remove duplicate species names from lookup

sgrimm:
When we import the GBIF data, we calculate the scientific name for each
taxon such that it is formatted in a consistent way. That calculation can
result in the same scientific name for more than one GBIF species entry
if, e.g., we are stripping off the publication date or the name of the
person who discovered the species.

Update `/api/v1/species/lookup/names` to remove the duplicate names.


sgrimm:
* main
  * **PR #396** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/396" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/396?utm_source=stack-comment).

PR #408, opened 2022-06-20
SW-1062 Auto-create devices when facility is connected

sgrimm:
Containerized seed banks will always have an OmniSense hub to talk to the
temperature/humidity sensors, but the device manager can't autodetect the hub.
But other facility types, such as the desal/RO installations at PacFlight,
don't have OmniSense hubs, so the device manager can't just assume a hub exists.

Add a concept of "default devices" that get created when a sensor kit is connected
to a facility. The list of default devices can vary by facility type. Use the
existing "device templates" feature: each facility type that needs default devices
(currently just seed banks) has a corresponding template category, so we can
manage the default devices using the admin UI.

Fix a bug in the admin UI's handling of template category names with whitespace.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/FacilityModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/FacilityModel.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.customer.model
 
+import com.terraformation.backend.db.DeviceTemplateCategory
 import com.terraformation.backend.db.FacilityConnectionState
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #408** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/408" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/408?utm_source=stack-comment).

PR #409, opened 2022-06-20
SW-1051 Move test clock management to admin UI

sgrimm:
Previously, the UI to mess with the test clock was served from a one-off
administrative endpoint that was only enabled in dev environments. Move it
to the admin UI and add permission checks so it's only available to super
admins.

PR #410, opened 2022-06-20
SW-1051 Add ability to reset test clock

sgrimm:
For QA testing in the staging environment, we need to be able to set the system
clock back to real time when the tests are finished. Add the ability to do that.

To allow people to do another test run where they mess with the clock to generate
notifications, resetting the clock also resets the timestamps on the various daily
tasks (which are used for generating accession notifications) so the system won't
skip over time periods that it has already processed thanks to previous
adjustments of the clock. To simplify:

1. Current date (in the real world) is January 10
2. Next time daily task runs, it scans for accessions modified on January 10
3. Test clock is advanced 1 day
4. Daily task runs immediately and scans for accessions modified on January 11
5. Test clock is reset
6. Next time daily task runs, it scans for January 10 again

Absent the second part of this change, step 6 wouldn't happen because the system
would have recorded that it had already processed everything up to January 11.

This will potentially cause the system to generate repeat notifications, but that
side effect seems less bad than the side effect of not generating notifications
for newly-created accessions after the clock is reset.


sgrimm:
* main
  * **PR #409** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/409" target="_blank">Graphite</a> 
    * **PR #410** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/410" target="_blank">Graphite</a>  👈
      * **PR #411** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/411" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/410?utm_source=stack-comment).

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/410?utm_source=stack-comment).


Job Reference: DPISzEg7t69kL2XISo4c

PR #412, opened 2022-06-21
SW-1069 Don't check names of deleted species

sgrimm:
If an organization has a deleted species that was deleted before its name
was checked, the "check all the species that haven't been checked yet" logic
that runs after CSV upload was trying to check the deleted species, which
failed because a deleted species is treated as not existing by most of the
data access code.


sgrimm:
* main
  * **PR #412** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/412" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/412?utm_source=stack-comment).

sgrimm:
Ah, yes, we do indeed. Thanks.

PR #413, opened 2022-06-21
SW-1064 Disable unwanted health checks

sgrimm:
Spring Boot has a built-in component called Actuator. It sets up health checks
based on what components the application is using, e.g., if you're using a
database, it will make sure the database is reachable. We want most of the
default checks, but two of them are not useful for us: the free disk space check
(which wouldn't work right from inside a Docker container) and the mail server
connectivity check (we don't want to mark terraware-server as down if AWS's mail
service is having an outage).

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -73,6 +73,17 @@ spring:
   session:
     store-type: jdbc
 
+management:
+  health:
+    diskspace:
+      # Disable checking of disk space; the server runs in a Docker container so this would be
+      # measuring the free space in the container's filesystem which isn't too useful.
+      enabled: false
+    mail:
+      # Disable checking health of mail server; we don't want to mark terraware-server as down
+      # if the mail service is having an unrelated outage.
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #413** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/413" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/413?utm_source=stack-comment).

karthikbtf:
If the mail service is down, some of our API wouldn't work (such as inviting new users to an org). Should we capture it as a less severe alert?

sgrimm:
We definitely want to know about mail outages, but I think we shouldn't take the whole site offline if we can't deliver mail.

karthikbtf:
> We definitely want to know about mail outages, but I think we shouldn't take the whole site offline if we can't deliver mail.

Makes sense.

PR #414, opened 2022-06-21
SW-1069 Handle duplicate names from GBIF in name check

sgrimm:
Since we reformat scientific names to remove unwanted components such as
authorship information, we can end up with cases where there are two GBIF
taxa with the same scientific name. This was causing the name checking
code to bomb out because it expected scientific names to be unique.


sgrimm:
* main
  * **PR #414** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/414" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/414?utm_source=stack-comment).

PR #416, opened 2022-06-22
SW-1055 Sort first-word matches to the top of species name list

sgrimm:
The code that populates the species name typeahead in the UI lets the user
search for any word in a scientific name, including subspecies and so forth.
Previously, it returned all results alphabetically, which meant that if you
started typing a scientific name that happened to begin with a string that
also appears in other positions for lots of species, your species might be
sorted off the bottom of the typeahead.

Update the query so that it prioritizes search results that start with the
first word of the search phrase, but still includes matches for other words.

For example, if you search for "visco", previously, the first result would
have been "Abutilon viscosum". Now that will be the second result, and the
first will be "Viscoides pendula".

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
@@ -48,7 +48,7 @@ class GbifStore(
     // Dynamically construct a query that has a separate join to the gbif_name_words table for each
     // prefix string. For prefixes == listOf("abc", "def") we want to end up with a query like
     //
-    // SELECT name
+    // SELECT *
     // FROM gbif_names
     // JOIN gbif_name_words gnw_0 ON gbif_names.id = gnw_0.id
     // JOIN gbif_name_words gnw_1 ON gbif_names.id = gnw_1.id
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #416** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/416" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/416?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
```diff
@@ -48,22 +48,36 @@ class GbifStore(
...
+        DSL.case_()
+            .`when`(DSL.lower(GBIF_NAMES.NAME).like(firstWordPattern), 1)
+            .else_(2)
+            .`as`("first_word_sort_position")
 
     val selectFrom =
-        dslContext.selectDistinct(GBIF_NAMES.asterisk()).on(GBIF_NAMES.NAME).from(GBIF_NAMES)
+        dslContext
+            .selectDistinct(GBIF_NAMES.asterisk(), firstWordSortPosition)
+            .on(firstWordSortPosition, GBIF_NAMES.NAME)
```
karthikbtf, line 79: curious how this works, my basic understanding wonders why we need this in more than just the orderBy clause.

sgrimm: This is a good question, and the answer is one part "it's supposed to work that way" and another part "it is an annoying workaround."

The "supposed to" part is why it's in both the `ON` and `ORDER BY` clauses. PostgreSQL's `SELECT DISTINCT ON` feature requires that the uniqueness criterion consist of columns that are at the beginning of the `ORDER BY` clause. Otherwise the semantics are a little weird because the correct position of a set of unique values in the final search results might be ambiguous. See the "DISTINCT clause" section here: https://www.postgresql.org/docs/current/sql-select.html

The "annoying workaround" part is why it also needs to be in the `SELECT` column list and why I added a column alias (`as`) to it. The problem there is bind variables. If the same `CASE` clause is included verbatim in the `DISTINCT ON` and `ORDER BY` arguments, it turns into a prepared statement with bind variables for all the values. PostgreSQL's query parser can't guarantee that I'm not going to pass a different set of values for the two copies of the `CASE`, which would violate the requirement above. So I add a column alias, which means the query can reference the `CASE` by its alias. But aliased columns have to appear in the `SELECT` clause, so I have to add it there too even though nothing looks at it.

I considered instead rendering the `CASE` as plaintext SQL so there wouldn't be any variable binding, but that didn't seem any less ugly and would lose a smidgen of type safety.

Also, I should say I only discovered the annoying workaround after I'd written the query without the workaround and PostgreSQL spat out an error message. It took me a few minutes to figure out what was making it complain!

karthikbtf: Thanks for the clarification 👍

PR #418, opened 2022-06-23
SW-1059 Rename short code to sensor kit ID

sgrimm:
We decided to use the term "sensor kit ID" as the official name for the ID code
that gets printed on sensor kits and that users enter to connect a sensor kit to a
seed bank, so update the code to use the official terminology.

This affects the data model, the implementation, and the API. For now, the API
continues to also accept `shortCode` parameters in requests and continues to
include `shortCode` values in responses, alongside `sensorKitId`, to give clients
a migration path. We'll follow up and remove support for `shortCode` once the
client code has shifted to the new name.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
@@ -30,21 +30,24 @@ class DeviceManagersController(
 ) {
   @GetMapping
   fun getDeviceManagers(
+      @RequestParam("sensorKitId") sensorKitIdParam: String?,
+      // TODO: Remove this once frontend is updated to use sensorKitId
       @RequestParam("shortCode") shortCode: String?,
       @RequestParam("facilityId") facilityId: FacilityId?,
   ): GetDeviceManagersResponsePayload {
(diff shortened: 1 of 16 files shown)


sgrimm:
* main
  * **PR #418** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/418" target="_blank">Graphite</a>  👈
    * **PR #419** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/419" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/418?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/device/api/DeviceManagersController.kt
```diff
@@ -30,21 +30,24 @@ class DeviceManagersController(
 ) {
   @GetMapping
   fun getDeviceManagers(
+      @RequestParam("sensorKitId") sensorKitIdParam: String?,
+      // TODO: Remove this once frontend is updated to use sensorKitId
```
karthikbtf, line 34: Thanks for the backwards compat. I will update the FE

PR #420, opened 2022-06-24
SW-1111 Make text field searches case-insensitive

sgrimm:
Previously, the search API did case-sensitive matching for `Exact` search
filters on text fields and was inconsistent about case sensitivity for `Fuzzy`
searches based on the length of the search string. The latter was a side effect
of the fact that we were using a particular fuzzy match operator from the
`pg_trgm` extension.

Change it to always do case-insensitive searches and to use a more relaxed
fuzzy match operator that supports shorter search strings.


sgrimm:
* main
  * **PR #420** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/420" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/420?utm_source=stack-comment).

PR #421, opened 2022-06-24
SW-1117 Filter deleted species out of search results

sgrimm:
Previously, the search API wasn't paying attention to whether or not a species was
deleted. One option is to let the client decide whether or not to include deleted
species. But users shouldn't be able to see deleted species at all, so instead,
treat it as a permissions filter that always gets applied.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
@@ -51,5 +51,6 @@ class SpeciesTable(tables: SearchTables, fuzzySearchOperators: FuzzySearchOperat
 
   override fun conditionForPermissions(): Condition {
     return SPECIES.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)
+        .and(SPECIES.DELETED_TIME.isNull)
   }
 }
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #421** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/421" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/421?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/table/SpeciesTable.kt
```diff
@@ -51,5 +51,6 @@ class SpeciesTable(tables: SearchTables, fuzzySearchOperators: FuzzySearchOperat
 
   override fun conditionForPermissions(): Condition {
     return SPECIES.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)
+        .and(SPECIES.DELETED_TIME.isNull)
```
karthikbtf, line 54: A thought, if we want to scale the concept of deleted entries (perhaps for later) .. we could introduce an `isDeleted(): Condition` in SearchTable with a default implementation of no-op. Leverage that in SearchService.

sgrimm: Just wait until you see my next PR! I had a similar thought.

karthikbtf: nice, i was thinking along `visibility` initially. will take a full look.

PR #422, opened 2022-06-24
Use "visibility" instead of "permission" in search code

sgrimm:
Recently we've started to use `SearchTable.conditionForPermissions` to add
filtering that isn't exactly related to the permissions system, e.g., filtering
out deleted species.

Rather than thinking about those criteria as being weird sorts of permissions,
we can think of the permissions rules as a subset of visibility rules: ultimately
they are determining whether or not you can see a particular thing in search
results for whatever reason.

Update the internal terminology to reflect this broader perspective.

The code changes here are all "rename" refactorings; there are no manual edits
outside of comment text.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -636,26 +636,26 @@ class NestedQueryBuilder(
 
       val selectWithConditions = selectWithLateral.where(conditions)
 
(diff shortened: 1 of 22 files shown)


sgrimm:
* main
  * **PR #422** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/422" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/422?utm_source=stack-comment).

PR #429, opened 2022-06-28
SW-1026 Refactor default automation creation

sgrimm:
In preparation for some changes to the automation data model, refactor the code
that automatically creates automations when devices are created or edited.

No behavior change here (existing tests continue to pass) but it will make
upcoming behavior changes cleaner to implement.

PR #430, opened 2022-06-28
SW-1026 Move automation settings into schema

sgrimm:
The automations table has a general-purpose `configuration` column that, as its
name implies, holds the configuration settings for each automation. This is
arbitrary data from the server's point of view, but the device manager always uses
it for a consistent set of information.

Update the data model so that all of the values for sensor bounds automations (the
most common kind, and the only kind the server knows how to configure) are
first-class values in the automations table rather than elements of an opaque JSON
blob.

The `type` and `verbosity` settings are always required for all automation types;
other settings are optional since they may not apply to all automations.

Since the device manager can support automations the server doesn't know about,
there's still a place for arbitrary configuration values. It is renamed to
`settings` for consistency with the opaque JSON configuration blob in the
devices table.

The existing automation-related API endpoints under `/api/v1/facilities` remain
unchanged from the client's point of view for backward compatibility; the code
now converts the old-style configuration object to the new schema and back.

SW-1048 is also addressed here. New endpoints under `/api/v1/automations`
expose the updated data model more directly: the config values in question are
top-level payload fields. There is still a place for arbitrary settings for
unknown automation types, but for sensor bounds automations it won't be used.

Once the device manager is updated to use the new endpoints, the old ones and
their backward-compatibility code can be removed.


sgrimm:
* main
  * **PR #429** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/429" target="_blank">Graphite</a> 
    * **PR #430** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/430?utm_source=stack-comment).

sgrimm:
This isn't driven by a feature requirement; it just tightens up the data model a bit. Down the road it will make it a bit easier to do things like let the user configure the temperature thresholds on a per-sensor basis.

karthikbtf:
Did you want me to look at this @sgrimm ?

sgrimm:
Yeah, sorry, was waiting for the build to succeed before adding you, and then got distracted!

src/main/kotlin/com/terraformation/backend/customer/api/FacilitiesController.kt
```diff
@@ -245,7 +266,44 @@ data class AutomationPayload(
...
+    setOf(
+        DEVICE_ID_KEY,
+        LOWER_THRESHOLD_KEY,
+        TIMESERIES_NAME_KEY,
+        TYPE_KEY,
+        UPPER_THRESHOLD_KEY,
+        VERBOSITY_KEY,
+    )
+
+private fun AutomationModel.backwardCompatibleConfiguration(): Map<String, Any?> {
```
karthikbtf, line 294: Curious to understand how this pattern works, extending the model with a function but in a different context.

sgrimm: This isn't 100% accurate, but you can think of

```kotlin
fun MyClass.someFunction(): SomeResult {
  return SomeResult(myClassProperty)
}
```

as syntax sugar for

```kotlin
fun someFunction(obj: MyClass): SomeResult {
  return SomeResult(obj.myClassProperty)
}
```

And then it follows the scoping and visibility rules of all functions: if it's declared private at the top level, it's only visible to code in the same source file.

In this case I declared it as an extension method mostly because I was referencing a bunch of properties from `AutomationModel` and it would have been more cluttered to declare in the second style. I also wanted to keep all the backward-compatibility logic limited to the controller so it'll be easier to rip out later.

PR #431, opened 2022-06-29
SW-1026 Update settings in devices table

sgrimm:
Currently, the device manager always uses a device setting called `diagnosticMode`
to control the logging verbosity level for a device. Pull that value up into the
standard device schema.

For now, the verbosity level is returned both as a top-level payload field and as
a settings value called `diagnosticMode`, though it is no longer stored in the
settings JSON blob in the database. Once the device maanger is updated to use the
`verbosity` field, this backward-compatibility code can be removed.

The device manager doesn't use the "polling interval" value at all (the interval
depends on the device type and isn't user-adjustable), so remove that from the
table and the API.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
@@ -174,6 +174,12 @@ private fun DevicesRow.settingsWithVerbosity(objectMapper: ObjectMapper): Map<St
   }
 }
 
+private fun Map<String, Any?>.toJsonbWithoutVerbosity(objectMapper: ObjectMapper): JSONB? {
+  return minus(VERBOSITY_KEY)
+      .ifEmpty { null }
+      ?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) }
+}
+
 data class CreateDeviceRequestPayload(
     @Schema(description = "Identifier of facility where this device is located.")
     val facilityId: FacilityId,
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #431** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/431" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/431?utm_source=stack-comment).

sgrimm:
Scratch my previous comment -- this is good to go as-is.

src/main/resources/db/migration/common/V99__DevicesSettings.sql
```diff
@@ -0,0 +1,17 @@
...
+                    WHEN settings - 'diagnosticMode' = '{}' THEN NULL
+                    ELSE settings - 'diagnosticMode' END
+WHERE settings ? 'diagnosticMode';
+
+ALTER TABLE device_templates DROP COLUMN polling_interval;
+ALTER TABLE device_templates ADD COLUMN verbosity INTEGER;
+
+UPDATE device_templates
+SET verbosity = (settings ->> 'diagnosticMode')::INTEGER,
+    settings  = settings - 'diagnosticMode'
```
karthikbtf: Do we need a similar check for empty json?

src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
```diff
@@ -262,9 +288,13 @@ data class UpdateDeviceRequestPayload(
...
         port = port,
-        pollingInterval = pollingInterval,
+        verbosity = verbosity ?: settings?.get(VERBOSITY_KEY)?.toString()?.toInt(),
         enabled = true,
-        settings = settings?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
+        settings =
+            settings
+                ?.minus(VERBOSITY_KEY)
+                ?.ifEmpty { null }
+                ?.let { JSONB.jsonb(objectMapper.writeValueAsString(it)) },
```
karthikbtf: extract out as reusable code? not a big deal, i understand this will be ripped out soon.

src/main/resources/db/migration/common/V99__DevicesSettings.sql
sgrimm: Yeah, probably a good call. None of the device templates even have settings values at all right now but better to make the migration robust. Thanks.

PR #432, opened 2022-06-29
Remove support for non-PostgreSQL fuzzy text search

sgrimm:
Previously, fuzzy text search was implemented in such a way that it would be
possible for the code to support more than one database engine. But we've
decided to only target PostgreSQL, so there's no need for the added indirection.

Turn the `likeFuzzy` and `similarity` methods into top-level extension methods
and get rid of the classes that used to be needed to support runtime-selectable
implementations.


sgrimm:
* main
  * **PR #432** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/432" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/432?utm_source=stack-comment).

PR #433, opened 2022-06-29
Remove support for non-PostgreSQL migrations

sgrimm:
Previously, we thought we might want to run the server on more than one database
engine, so the Flyway migrations were structured to allow per-engine variants.
But we've decided to only target PostgreSQL, so that structure just makes the
migrations more complicated for no good reason.

Consolidate all the migrations into a single directory and remove the dummy
migrations that would have been used on engines other than PostgreSQL. Those
migrations were never used anyway.

Flyway keeps track of the checksums of migration scripts, so these are all
straight renames, no content changes. That means some of the scripts still use
placeholders for JSONB and UUID columns. Going forward we can stop using those
placeholders but for now they need to remain in the existing scripts or the
file checksums would change.


sgrimm:
* main
  * **PR #433** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/433" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/433?utm_source=stack-comment).

PR #434, opened 2022-06-29
SW-1049 Make AccessionStore.fetchOneById non-nullable

sgrimm:
Rename `fetchById` to `fetchOneById` for consistency with other store classes.

All the call sites were throwing exceptions on failure anyway, so consolidate the
exception throwing into the fetch method and make the call sites less cluttered.
This also eliminates the need for the "skip permission check" parameter which was
always kind of screwy.


sgrimm:
* main
  * **PR #434** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/434" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/434?utm_source=stack-comment).

PR #436, opened 2022-06-30
SW-1128 Generate report of dependency licenses

sgrimm:
In preparation for releasing the server as open source, add documentation of
the licenses of all its dependencies. This will make it easier to confirm that
the license we choose is compatible with the licenses of all our dependencies.

The license report can be generated using `./gradlew generateLicenseReport`
and will be automatically generated and published to the repo's GitHub Pages
when dependencies are modified.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,4 +1,5 @@
 # THIS FILE GETS REMOVED AS PART OF THE CI BUILD!
 
-# Don't allow generated schema docs to be committed to the main branch.
+# Don't allow generated docs to be committed to the main branch.
+license-report
 schema
--- a/src/docs/license-normalizer-bundle.json
+++ b/src/docs/license-normalizer-bundle.json
@@ -1,0 +1,19 @@
+{
+  "bundles": [
+    {
+      "bundleName": "apache2",
+      "licenseName": "Apache License, Version 2.0",
+      "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0.html"
+    }
+  ],
+  "transformationRules": [
+    {
+      "bundleName": "apache2",
+      "licenseUrlPattern": "http://www.apache.org/licenses/LICENSE-2.0*"
+    },
+    {
+      "bundleName": "apache2",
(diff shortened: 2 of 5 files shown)


sgrimm:
* main
  * **PR #436** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/436" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/436?utm_source=stack-comment).

sgrimm:
Gradle plugin I'm using for this:  https://github.com/jk1/Gradle-License-Report

src/docs/license-normalizer-bundle.json
```diff
@@ -0,0 +1,19 @@
...
+  ],
+  "transformationRules": [
+    {
+      "bundleName": "apache2",
+      "licenseUrlPattern": "http://www.apache.org/licenses/LICENSE-2.0*"
+    },
+    {
+      "bundleName": "apache2",
+      "licenseUrlPattern": "https://www.apache.org/licenses/LICENSE-2.0*"
+    }
```
karthikbtf, line 17: looks like a redundant entry

docs/.gitignore
```diff
@@ -1,4 +1,5 @@
 # THIS FILE GETS REMOVED AS PART OF THE CI BUILD!
 
-# Don't allow generated schema docs to be committed to the main branch.
+# Don't allow generated docs to be committed to the main branch.
+license-report
```
karthikbtf, line 4: index.html output won't be checked into the repo?

src/docs/license-normalizer-bundle.json
sgrimm: One is `http` and the other is `https` (which is why this rule needs to exist; otherwise the plugin thinks the URLs don't match so treats the two as being different licenses).

docs/.gitignore
sgrimm: Correct, everything in that directory will be excluded. The plugin also creates a bunch of subdirectories with the license files for each dependency.

PR #439, opened 2022-06-30
Use centralized IDs in DatabaseTest insert methods

sgrimm:
Now that the IDs for site data are centrally managed, there's no longer any point
forcing tests to pass those IDs into helper methods like `insertFacility()`.
Update those methods to use the standard IDs by default, and update the tests to
stop passing them in.

In many cases, this means the insert methods can be called with no arguments at
all and will do the right thing. To support that goal, update `insertSiteData()`
to stop passing in PacFlight-themed names for the site data; instead, it uses the
default names that include the object IDs, e.g., `Facility 100`.


sgrimm:
* main
  * **PR #439** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/439" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/439?utm_source=stack-comment).

PR #441, opened 2022-07-01
SW-965 Delete notifications after a configured number of days

karthikbtf:
Allow product to be able to clean up notifications, current default for expiration of notifications is 30 days.
Created a NotificationCleanupTask that piggy backs onto DailyTaskRunner's DailyTaskArrivedEvent.
Wondering if maybe this should be its own cron job that runs at a higher frequency.
Tested on my local dev server setup and adjusting the clock in the admin interface.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
@@ -56,8 +56,9 @@ internal class NotificationsCleanupTaskTest : DatabaseTest() {
     insertNotification(NotificationId(2), UserId(1), createdTime = now.minus(Duration.ofDays(6)))
 
     val beforeCleanup = notificationsDao.findAll()
-    assertTrue(
-        setOf(1L, 2L) == beforeCleanup.map { it.id?.value }.toSet(),
+    assertEquals(
+        setOf(1L, 2L),
+        beforeCleanup.map { it.id?.value }.toSet(),
         "Expected notification IDs 1 and 2")
 
     notificationsCleanupTask.cleanup(DailyTaskTimeArrivedEvent())


sgrimm:
To the question in your PR description: what would be the benefit of running this more than daily? I can't think of one but maybe I'm missing it.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -101,6 +101,9 @@ class TerrawareServerConfig(
 
     /** Configures how the server interacts with the Balena cloud service to manage sensor kits. */
     val balena: BalenaConfig = BalenaConfig(),
+
+    /** Configures notifications processing. */
+    val notificationsCleanup: NotificationsCleanupConfig = NotificationsCleanupConfig(),
```
sgrimm: I wonder if this should be a more general `notifications` section. I could see us wanting non-cleanup-related notification settings in the future. (Of course, we could always just restructure it at that point.)

karthikbtf:
> To the question in your PR description: what would be the benefit of running this more than daily? I can't think of one but maybe I'm missing it.

No real benefit except that the cleanup will be more current on a per notification basis. Running it once a day could leave some notifications living almost an extra day longer. Which isn't really a big issue right now.

src/main/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTask.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.terraformation.backend.db.tables.references.NOTIFICATIONS
+import com.terraformation.backend.log.perClassLogger
+import java.time.Clock
+import java.time.Duration
+import javax.annotation.ManagedBean
+import org.jooq.DSLContext
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.context.event.EventListener
+
+@ConditionalOnProperty(TerrawareServerConfig.NOTIFICATIONS_CLEANUP_ENABLED_PROPERTY)
```
sgrimm: I think you might need `havingValue = "true"` here or it will just check for whether or not the property exists, rather than on what its value is. But maybe I'm wrong. The three cases I think need to be covered:

* Property is set to true in config
* Property is set to false in config
* Property is not listed in config at all

src/main/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTask.kt
```diff
@@ -0,0 +1,48 @@
...
+@ManagedBean
+class NotificationsCleanupTask(
+    private val clock: Clock,
+    private val terrawareServerConfig: TerrawareServerConfig,
```
sgrimm: Small consistency thing: the rest of the code uses `config` as the variable name for this.

src/main/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTask.kt
```diff
@@ -0,0 +1,48 @@
...
+        val rowsDeleted =
+            dslContext
+                .deleteFrom(NOTIFICATIONS)
+                .where(
+                    NOTIFICATIONS.CREATED_TIME.le(
+                        clock
+                            .instant()
+                            .minus(
+                                Duration.ofDays(
+                                    terrawareServerConfig.notificationsCleanup.retentionDays)),
```
sgrimm: This is totally fine as-is, but FYI, another API for this is `minus(whatever, ChronoUnit.DAYS)`.

src/main/resources/application-default.yaml
```diff
@@ -7,6 +7,8 @@ terraware:
   email:
     always-send-to-override-address: true
     subject-prefix: "[DEV]"
+  notifications-cleanup:
+    enabled: true
```
sgrimm: `application-default.yaml` is, confusingly, only used in dev environments, so this will only enable cleanup in dev. We probably want this in `application.yaml` which is the one that's always loaded even in prod/staging. (Unless your intent was to not enable it in prod/staging, of course, in which case never mind this comment.)

If we do put this in `application.yaml`, I wonder if it makes more sense to default the feature to enabled in `TerrawareServerConfig`, so no config change is needed at all.

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,76 @@
...
+  fun setUp() {
+    every { clock.instant() } returns now
+    every { config.notificationsCleanup.enabled } returns true
+    every { config.notificationsCleanup.retentionDays } returns 1
+
+    notificationsCleanupTask = NotificationsCleanupTask(clock, config, dslContext)
+  }
+
+  @Test
+  fun `Does not delete notifications if there are none expired`() {
```
sgrimm: Style nit: the existing tests all start with lower-case letters. I am totally open to the idea that we should capitalize all of them across the board, if we want to change that.

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,76 @@
...
+    every { config.notificationsCleanup.retentionDays } returns 5
+
+    insertNotification(NotificationId(1), UserId(1), createdTime = now)
+
+    val expected = notificationsDao.findAll()
+    assertTrue(
+        expected.find { row -> row.id == NotificationId(1) } != null,
```
sgrimm: Maybe `expected.any { it.id == NotificationId(1) }`

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        "Expected to find notification with id 1.")
+
+    notificationsCleanupTask.cleanup(DailyTaskTimeArrivedEvent())
+
+    val actual = notificationsDao.findAll()
+    assertFalse(actual.isEmpty(), "Did not find any notifications but expected to.")
```
sgrimm: Don't think you need this `assertFalse` -- the `assertEquals` will fail if the list is empty.

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,76 @@
...
+    insertNotification(NotificationId(1), UserId(1), createdTime = now)
+    insertNotification(NotificationId(2), UserId(1), createdTime = now.minus(Duration.ofDays(6)))
+
+    val before = notificationsDao.findAll()
+    assertTrue(
+        before.find { row -> row.id == NotificationId(1) } != null,
+        "Expected to find notification with id 1.")
+    assertTrue(
+        before.find { row -> row.id == NotificationId(2) } != null,
+        "Expected to find notification with id 2.")
```
sgrimm: This might be taking the bikeshedding too far, but maybe `assertEquals(setOf(1,2), before.map { it.id.value }.toSet(), "Expected notification IDs")`?

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    val expected = before.filter { row -> row.id != NotificationId(2) }
```
sgrimm: This is fine, but alternate idea: call `notificationsDao.findAll()` after inserting the first notification but before the second  one.

sgrimm:
> Running it once a day could leave some notifications living almost an extra day longer. Which isn't really a big issue right now.

My guess is no user would notice this, so yeah, probably not a big deal right now like you say.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
karthikbtf: Good idea, I will make this a general notifications section.

src/main/resources/application-default.yaml
karthikbtf: Figured this being a destructive job, might be safer to default this to false unless specified. 
Most definitely overthinking it, I will make this as default true in the config, we can disable it explicitly if needed.

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
karthikbtf: Not sure why I can't get this to work
```
Expected notification IDs 1 and 2 ==> expected: java.util.LinkedHashSet@1a2b23f2<[1, 2]> but was: java.util.LinkedHashSet@1d8e9f22<[1, 2]>
```

karthikbtf: seems to be checking object addresses maybe

sgrimm: Oh, oops, my bad. It's probably just `Int` vs. `Long`. Change `setOf(1,2)` to `setOf(1L,2L)` and maybe it'll work.

karthikbtf: Yes, I got it to work with setOf(1L, 2L), missed updating here. Thanks for confirming.

src/main/resources/application-default.yaml
```diff
@@ -7,8 +7,8 @@ terraware:
   email:
     always-send-to-override-address: true
     subject-prefix: "[DEV]"
-  notifications-cleanup:
-    enabled: true
+  notifications:
+    cleanup-enabled: true
```
karthikbtf: oops this should not be here, was from my testing

src/test/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTaskTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    insertNotification(NotificationId(1), UserId(1), createdTime = now)
+    val expected = notificationsDao.findAll()
+
+    insertNotification(NotificationId(2), UserId(1), createdTime = now.minus(Duration.ofDays(6)))
+
+    val beforeCleanup = notificationsDao.findAll()
+    assertTrue(
+        setOf(1L, 2L) == beforeCleanup.map { it.id?.value }.toSet(),
+        "Expected notification IDs 1 and 2")
```
sgrimm: `assertEquals` would be better here since it'll show the expected and actual values.

karthikbtf: I will make the changes, had switched to == when i was debugging the Int vs Long comparsion

PR #457, opened 2022-07-05
SW-966 Remove references to previous accession notifications code and tables

karthikbtf:
Referencing code in the web app was removed earlier, except for some dead-code which is being cleaned up.


karthikbtf:
> PRs with lots of deleted code are the best!
> 
> I see you just merged my other PR so you'll need to renumber the migration here before merging to avoid a collision.

Thanks Steve, sounds good, will do that.

PR #458, opened 2022-07-05
SW-1061 Move mail delivery to a separate class

sgrimm:
In preparation for adding tests to verify the contents of email notifications,
refactor `EmailService` to extract the functions that interact with actual
email services into a separate class `EmailSender` that we can stub out in the
test code.

No behavior change.

PR #459, opened 2022-07-05
SW-1061 Add unit tests for notification email

sgrimm:
Add a test suite to confirm that the rendered email notification messages
include the expected content. This doesn't test the boilerplate text in the
template files since those are subject to change, but it attempts to verify
that the links are present and pointing to the right place and that the various
bits of dynamic content specified in the requirements docs are included.

It is still possible for someone to mess up an MJML template and cause the
notifications to look illegible, so when we revise notification content, we
will still want to do manual testing to confirm the results have the correct
appearance. But this set of tests will at least confirm that the templates
have no syntax errors and use the expected set of variable substitutions.

Writing this test suite turned up a bug: the notification for unknown
automations wasn't linking to the specific device on the monitoring dashboard.
That bug is fixed here.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/email/EmailNotificationServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/email/EmailNotificationServiceTest.kt
@@ -75,17 +75,15 @@ internal class EmailNotificationServiceTest {
 
   private val webAppUrls: WebAppUrls = WebAppUrls(config)
 
+  private val freeMarkerConfig =
+      Configuration(Configuration.VERSION_2_3_31).apply {
+        // Load the email template files from the templates folder in the build output.
+        setClassLoaderForTemplateLoading(
+            EmailNotificationServiceTest::class.java.classLoader, "templates")
+      }
+
   private val emailService =
-      EmailService(
-          config,
-          Configuration(Configuration.VERSION_2_3_28).apply {
-            setClassLoaderForTemplateLoading(
-                EmailNotificationServiceTest::class.java.classLoader, "templates")
-          },
-          organizationStore,
-          parentStore,
-          projectStore,
-          sender)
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #458** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/458" target="_blank">Graphite</a> 
    * **PR #459** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/459" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/459?utm_source=stack-comment).

sgrimm:
This test has a lot of stubbing. I don't love that, but it's not clear to me that making this a database-backed test would be a net improvement; you'd replace the stub objects with a bunch of additional dependency wiring and of course it'd be slower.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -128,12 +125,13 @@ class EmailNotificationService(
...
   fun on(event: UnknownAutomationTriggeredEvent) {
     val automation = automationStore.fetchOneById(event.automationId)
+    val devicesRow = automation.deviceId?.let { deviceStore.fetchOneById(it) }
     val facility = facilityStore.fetchOneById(automation.facilityId)
     val organizationId =
         parentStore.getOrganizationId(facility.id) ?: throw FacilityNotFoundException(facility.id)
 
     val facilityMonitoringUrl =
-        webAppUrls.fullFacilityMonitoring(organizationId, facility.id).toString()
+        webAppUrls.fullFacilityMonitoring(organizationId, facility.id, devicesRow).toString()
```
karthikbtf, line 134: good catch with the tests

karthikbtf:
@sgrimm this file has conflicts with the previously merged diff (though i understand the changes are the same).

PR #460, opened 2022-07-06
Analyze GBIF tables after import

sgrimm:
It seems PostgreSQL doesn't reliably autoanalyze tables after big bulk
delete/insert operations, so it was trying to plan queries on the GBIF tables
without the benefit of statistics about the distribution of values in various
columns.

Currently, someone has to manually run `ANALYZE` on all the tables after an
import, but that's easy to forget to do.

Make the code automatically do it at the end of the import process.


sgrimm:
* main
  * **PR #460** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/460" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/460?utm_source=stack-comment).

sgrimm:
> Just out of curiosity: would this slow down the import request tangibly?

It's not instantaneous, but compared to the import itself, it's pretty much noise. It takes on the order of a few seconds. (It doesn't analyze the entire table, just picks a selection of rows at random.)

PR #462, opened 2022-07-07
Add organization ID to facilities table

sgrimm:
In preparation for moving away from projects and sites, update the data model to
link facilities to their organizations.

This does not change the existing structure with facilities living under sites;
it is purely additive. There are a couple client-visible changes:

* The facility details payload now includes an organization ID in addition to a
  site ID.
* The search API allows navigating from organizations directly to facilities and
  back, without having to go through projects and sites. But going through
  projects and sites continues to work.

Down the road, we'll remove the site ID from facilities, but that'll be a breaking
change.


sgrimm:
* main
  * **PR #462** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/462" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/462?utm_source=stack-comment).

PR #464, opened 2022-07-08
SW-1183 Add facilities to organization payload

sgrimm:
In preparation for removing projects and sites from our data model, add a
`facilities` field to the response payloads of `/api/v1/organizations` and
`/api/v1/organizations/{id}` that contains a list of all the facilities the
user has access to. It is populated if the `depth` query parameter is set to
`Facility`.

The facilities continue to also be included under sites so as not to break
existing clients.


sgrimm:
* main
  * **PR #464** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/464" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/464?utm_source=stack-comment).

PR #487, opened 2022-07-12
Remove redundant features from admin UI

sgrimm:
The admin UI doesn't need to support doing things that you can already do in the
customer-facing web app.

Remove all the user management features (these are available via the People page
in the web app) and the ability to create an organization (which can be done via
the web app's navbar).

Creating and editing facilities is still supported since the admin UI exposes
some settings that aren't available in the web app.


sgrimm:
* main
  * **PR #487** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/487" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/487?utm_source=stack-comment).

PR #489, opened 2022-07-13
SW-1124 Rename "germination test" to "viability test"

sgrimm:
In preparation for updating the seed management code, bring the terminology we use
internally closer to the user-facing terminology by changing the term "germination
test" to "viability test."

This is only an update to the names the server uses internally. The API continues
to use the existing field names, and the term "germination test" still appears as
a value in some enums because they are included in API payloads and we don't want
to use different values the client isn't ready for. A subsequent change will update
the API to accept both kinds of names, and then a third change will remove support
for the old names.

Rename the following tables and their corresponding classes in the code:

* `accession_germination_test_typess` to `accession_viability_test_types`
* `germinations` to `viability_test_results`
* `germination_seed_types` to `viability_test_seed_types`
* `germination_substrates` to `viability_test_substrates`
* `germination_tests` to `viability_tests`
* `germination_test_types` to `viability_test_types`
* `germination_treatments` to `viability_test_treatments`

Rename fields in model classes to reflect the new terminology.

Rename viability-test-related methods and local variables in various classes.

Rename the `germinationTest` MJML template directory to `viabilityTest`.

Update references to "germination test" in comments, log messages, and test names
to use "viability test" instead.

All the code renames were done using IntelliJ "rename" refactoring. In cases
where the code had fields or parameters in alphabetical order, the order was
updated using IntelliJ "change method signature" refactoring.


sgrimm:
* main
  * **PR #489** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/489" target="_blank">Graphite</a>  👈
    * **PR #490** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/490" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/489?utm_source=stack-comment).

sgrimm:
This is kind of a monster PR. I thought about splitting it up into smaller steps (say, each PR renaming a single table) but I felt like it'd be even worse to have the code be in an internally inconsistent state after each step but the last one. If this is too overwhelming to review, though, I can do that easily enough. This was all pretty mechanical.

karthikbtf:
> This is kind of a monster PR. I thought about splitting it up into smaller steps (say, each PR renaming a single table) but I felt like it'd be even worse to have the code be in an internally inconsistent state after each step but the last one. If this is too overwhelming to review, though, I can do that easily enough. This was all pretty mechanical.

This should be fine, the nature of change makes it manageable.

src/main/kotlin/com/terraformation/backend/search/table/AccessionsTable.kt
```diff
@@ -36,23 +36,22 @@ class AccessionsTable(private val tables: SearchTables) : SearchTable() {
...
-              "germinationTests", ACCESSIONS.ID.eq(GERMINATION_TESTS.ACCESSION_ID)),
+          viabilityTests.asMultiValueSublist(
+              "germinationTests", ACCESSIONS.ID.eq(VIABILITY_TESTS.ACCESSION_ID)),
           species.asSingleValueSublist(
               "species", ACCESSIONS.SPECIES_ID.eq(SPECIES.ID), isRequired = false),
           storageLocations.asSingleValueSublist(
               "storageLocation",
               ACCESSIONS.STORAGE_LOCATION_ID.eq(STORAGE_LOCATIONS.ID),
               isRequired = false),
+          accessionViabilityTestTypes.asMultiValueSublist(
```
karthikbtf, line 53: lexicographical order broken?

karthikbtf:
Looking at the changes, if you prefer to add the API accepting new/old names in this reviiew - it won't break the client since it's an additive change. No preference on my part, was just wondering if breaking the client was a concern.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
```diff
@@ -558,9 +557,9 @@ data class WithdrawalPayload(
     val staffResponsible: String? = null,
     @Schema(
         description =
-            "If this withdrawal is of type \"Germination Testing\", the ID of the test it is " +
+            "If this withdrawal is of purpose \"Germination Testing\", the ID of the test it is " +
```
karthikbtf, line 560: "Germination Testing" as a purpose makes sense here. Calling this out in case you didn't intend to keep it as Germination Testing.

sgrimm:
Oops! I just published the API changes as a separate PR before I noticed your comment. Agreed that it won't break the client.

src/main/resources/templates/email/accession/viabilityTest/body.txt.ftl
```diff
@@ -1,4 +1,4 @@
-<#-- @ftlvariable name="" type="com.terraformation.backend.email.model.AccessionGerminationTest" -->
+<#-- @ftlvariable name="" type="com.terraformation.backend.email.model.AccessionViabilityTest" -->
 Time to test the viability of your seeds! ${accessionNumber} (located in ${facilityName}) is scheduled for a ${testType} germination test today!
```
karthikbtf, line 2: I'm a tad uncler on scope of renaming. Here 'germination test' makes sense, not sure if this is meant to be renamed as well.

karthikbtf:
> Oops! I just published the API changes as a separate PR before I noticed your comment. Agreed that it won't break the client.

No worries, all good :)

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
sgrimm: I definitely want to rename this but it's going to be a little more involved because this is an enum. We'll need to make the FE accept both values temporarily before we can switch this over. I'll be filing a Jira task about it.

src/main/resources/templates/email/accession/viabilityTest/body.txt.ftl
sgrimm: I didn't want to change any user-facing text as part of this so it would be behavior-neutral. Will do a followup PR for that.

karthikbtf: makes sense.

src/main/kotlin/com/terraformation/backend/search/table/AccessionsTable.kt
sgrimm: Oh, didn't respond to this: it is visually ambiguous, but this is sorted by search field name, not table name.

PR #490, opened 2022-07-13
SW-1124 Add viability test as alternate name in API

sgrimm:
Update the API payload definitions to allow clients to use either "germination
test" or "viability test" in various field names, and to include both names in
responses. If both are present, "viability test" takes precedence.

The search API is _partially_ updated here, to add alternate names for some
existing sublists and a field. The old names continue to work.

* `accessions.germinationTests` -> `accessions.viabilityTests`
* `accessions.latestGerminationTestDate` -> `accessions.latestViabilityTestDate`
* `germinationTests.germinations` -> `viabilityTests.viabilityTestResults`
* `germinations.germinationTest` -> `viabilityTestResults.viabilityTest`

There are some field aliases in the `accessions` search table that still use the
old names, and this change doesn't add equivalents for them. Instead, we'll want
to update the client to use the canonical field names, after which we can get
rid of the aliases entirely.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionController.kt
@@ -1,8 +1,10 @@
 package com.terraformation.backend.seedbank.api
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect
+import com.fasterxml.jackson.annotation.JsonCreator
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.annotation.JsonValue
 import com.terraformation.backend.api.ApiResponse404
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #489** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/489" target="_blank">Graphite</a> 
    * **PR #490** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/490" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/490?utm_source=stack-comment).

sgrimm:
SW-1224 has the list of search field name changes we'll need to make on the client side once this change is merged.

sgrimm:
Ah, oops, I realized I forgot to update the big comment at the top of `NestedQueryBuilder` that refers heavily to the viability test search fields. I'll go make that change.

PR #491, opened 2022-07-15
Move custom tasks out of build.gradle.kts

sgrimm:
Our Gradle build script is a little hard to navigate in part because it has a
bunch of one-off custom tasks to implement custom build logic. Turn these into
full-fledged classes that use Gradle's task API. This should declutter the build
script a bit and also make it easier to improve the custom logic in the future.

The task API lets us change MJML rendering to a single Gradle task rather than
a task per MJML file.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,13 +1,12 @@
-import com.github.gradle.node.yarn.task.YarnTask
 import com.github.jk1.license.filter.LicenseBundleNormalizer
 import com.github.jk1.license.render.InventoryHtmlReportRenderer
+import com.terraformation.gradle.PostgresDockerConfigTask
+import com.terraformation.gradle.VersionFileTask
 import com.terraformation.gradle.computeGitVersion
-import java.nio.file.Files
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.internal.deprecation.DeprecatableConfiguration
 import org.jetbrains.kotlin.gradle.internal.KaptGenerateStubsTask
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #491** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/491" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/491?utm_source=stack-comment).

build.gradle.kts
```diff
@@ -150,98 +150,17 @@ tasks.test {
...
-          }
-        }
+val renderMjmlTask = tasks.register<com.terraformation.gradle.RenderMjmlTask>("renderMjml")
 
 tasks {
   processResources {
-    processMjmlTasks.forEach { dependsOn(it.get()) }
-    exclude("**/*.mjml")
+    dependsOn(renderMjmlTask)
+    exclude("**/body.ftlh.mjml")
```
karthikbtf: just curious, what does this exclude do?

sgrimm: It stops Gradle from copying the MJML files to the build directory, since we don't need them at runtime. Normally everything under `src/main/resources` would be copied.

sgrimm: Actually, though, I think this exclude is wrong -- the previous version was correct. Not sure why I changed this here.

sgrimm: Updated the PR to put this back to `**/*.mjml`.

karthikbtf: Thanks for the explanation. The change looks good 👍

PR #493, opened 2022-07-15
Switch back to OpenJDK base Docker image

sgrimm:
The Eclipse Temurin image was running into a problem with detecting the amount of
available memory: https://github.com/adoptium/temurin-build/issues/2974

We still want to switch to that image, but for now, stop using it so the server
will start up successfully.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17-jre
+FROM openjdk:17-slim-buster
 
 ARG USER_ID=1001
 


sgrimm:
* main
  * **PR #493** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/493" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/493?utm_source=stack-comment).

PR #497, opened 2022-07-18
SW-1260 Remove accession status notifications

sgrimm:
Currently, the app sends per-accession notifications for various automated
status transitions as well as notifications that summarize the statues of all the
accessions at a seed bank.

We are simplifying the workflow, removing some of the transitions, and we are also
making many of the transitions manual rather than automated, so these
notifications will shortly become irrelevant.

Keep the per-accession notification for "scheduled drying end date has arrived"
since we are still going to want that when we move to the new set of statuses.


sgrimm:
* main
  * **PR #497** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/497" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/497?utm_source=stack-comment).

sgrimm:
@karthikbtf Let's hold off merging this until the product team confirms that it's okay to remove these notifications before we've implemented the rest of the changes to accession statuses. If not, we can hold onto this until later.

sgrimm:
Got the go-ahead to merge this.

PR #502, opened 2022-07-18
SW-1226 Rename "Germination Testing" withdrawal purpose

sgrimm:
Use the term "viability testing" in the list of withdrawal purposes to align with
the app terminology.

### DIFF
--- a/src/main/resources/db/migration/R__TypeCodes.sql
+++ b/src/main/resources/db/migration/R__TypeCodes.sql
@@ -154,7 +154,7 @@ VALUES (1, 'Propagation'),
        (4, 'Broadcast'),
        (5, 'Share with Another Site'),
        (6, 'Other'),
-       (7, 'Germination Testing')
+       (7, 'Viability Testing')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
 INSERT INTO notification_criticalities (id, name)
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #502** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/502" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/502?utm_source=stack-comment).

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -154,7 +154,7 @@ VALUES (1, 'Propagation'),
        (4, 'Broadcast'),
        (5, 'Share with Another Site'),
        (6, 'Other'),
-       (7, 'Germination Testing')
+       (7, 'Viability Testing')
```
karthikbtf, line 157: @sgrimm what's the process to update `dump/dump.sql` in the FE to capture this change?

sgrimm: Ah, good question! We should have a script for that, or at least docs. The general process is to restore the DB using this dump file, start the server to let it run migrations, then run `pg_dump` and overwrite the dump file.

You shouldn't _need_ to update the dump file, though; the value will be updated by the database migrations when the server starts up.

karthikbtf:
The build seems to be failing https://github.com/terraware/terraware-server/actions/runs/2693641947 in `generate-and-deploy`
```
found 0 vulnerabilities
/bin/sh: 1: mjml: not found
```

sgrimm:
Probably something to do with moving the MJML rendering to a separate source file. Luckily, this only affects the schema docs generation, not the server code. I'll take a look.

PR #503, opened 2022-07-18
Fix MJML rendering on docs branch

sgrimm:
On the docs branch, we weren't caching the `node_modules` directory as part of
the build, which exposed the fact that the build was using the wrong Gradle
task to set up the MJML renderer. Use the correct task, and also start caching
`node_modules` so docs get rendered more quickly.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -38,8 +38,9 @@ jobs:
         path: |
           ~/.gradle
           ~/.m2
+          node_modules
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #503** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/503" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/503?utm_source=stack-comment).

PR #507, opened 2022-07-19
SW-1291 Support multiple Balena fleet IDs

sgrimm:
Our internal device managers are running in a different Balena fleet than the
customer-owned ones, but we want both fleets to report sensor data to the prod
terraware-server instance. Allow multiple fleet IDs to be configured.

We'll update the server environments to include both the old-style and new-style
values before deploying this change, then remove the old-style values once this
is rolled out.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/device/balena/LiveBalenaClientTest.kt
@@ -75,7 +75,7 @@ internal class LiveBalenaClientTest {
     // Need to bootstrap the client without a fleet ID to create the fleet to define the config
     // that includes the fleet ID.
     every { config.balena } returns
-        TerrawareServerConfig.BalenaConfig(apiKey = apiKey, enabled = true, fleetId = 0)
+        TerrawareServerConfig.BalenaConfig(apiKey = apiKey, enabled = true, fleetIds = listOf(0))
 
     client =
         LiveBalenaClient(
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #507** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/507" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/507?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/device/balena/LiveBalenaClient.kt
```diff
@@ -162,15 +162,19 @@ class LiveBalenaClient(
    * Caution: Balena is picky about which characters are URL-encoded and which aren't. Make sure you
    * test this against the actual Balena service if you change the rendering logic.
    */
-  private fun filterTerm(field: String, value: Any?, operator: String = "eq"): String {
+  private fun filterTerm(field: String, value: Any?, operation: String = "eq"): String {
```
sgrimm, line 165: Renamed "operator" to "operation" here because "operator" is a Kotlin keyword (it's allowed here but kind of bad form to use keywords as variable names).

karthikbtf:
@sgrimm looks like some tests may need to be updated

sgrimm:
That was sloppy of me. I tested it by actually running it and pointing it at Balena but didn't do a test run. Will fix.

PR #508, opened 2022-07-19
Remove singular-noun accession endpoints

sgrimm:
The client code is now using the endpoints with plural names, so we don't need to
keep the singular versions around for backward compatibility any more.


sgrimm:
* main
  * **PR #508** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/508" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/508?utm_source=stack-comment).

PR #509, opened 2022-07-19
Remove some dead code

sgrimm:
Various methods and properties have become unused as the code has evolved, or were
only ever referenced by tests; remove a bunch of them.


sgrimm:
* main
  * **PR #509** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/509" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/509?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/file/UriNotSupportedException.kt
```diff
@@ -1,8 +0,0 @@
-package com.terraformation.backend.file
-
-import java.io.IOException
-import java.net.URI
-
-/** A file store was asked to accept a URL that refers to some other storage location. */
-class InvalidStorageLocationException(private val uri: URI, message: String? = null) :
```
sgrimm, line 7: This one is still referenced, so technically it doesn't quite fit the theme of the PR, but its source file had the wrong name and I figured better to just move it into the catch-all `Exceptions.kt`.

PR #514, opened 2022-07-28
SW-1310 Regenerate device manager credentials

sgrimm:
Upgrading Keycloak changed the format of authentication URLs, and since the URLs
are embedded in offline refresh tokens, existing tokens won't work.

Add a button to the device managers page in the admin UI to generate fresh tokens
for all device managers that are associated with users. This is only available for
super-admins since it cuts across all organizations.

Also add a button to generate a new token for a single device manager.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/DeviceManagerService.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/DeviceManagerService.kt
@@ -3,10 +3,15 @@ package com.terraformation.backend.device
 import com.terraformation.backend.customer.db.FacilityStore
 import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.model.SystemUser
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.BalenaDeviceId
 import com.terraformation.backend.db.DeviceManagerId
 import com.terraformation.backend.db.FacilityConnectionState
 import com.terraformation.backend.db.FacilityId
 import com.terraformation.backend.db.FacilityNotFoundException
+import com.terraformation.backend.db.UserId
+import com.terraformation.backend.db.tables.references.DEVICE_MANAGERS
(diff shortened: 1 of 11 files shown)


sgrimm:
* main
  * **PR #514** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/514" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/514?utm_source=stack-comment).

sgrimm:
@karthikbtf I merged this so I could update the token for the engineering sensor kit device manager, but it could use an after-the-fact review; I'll do a followup change to address any issues. Not urgent.

sgrimm:
See also followup commit 85f611ad

src/main/kotlin/com/terraformation/backend/device/DeviceManagerService.kt
```diff
@@ -62,4 +67,36 @@ class DeviceManagerService(
...
+      dslContext
+          .select(DEVICE_MANAGERS.USER_ID, DEVICE_MANAGERS.BALENA_ID, DEVICE_MANAGERS.FACILITY_ID)
+          .from(DEVICE_MANAGERS)
+          .where(DEVICE_MANAGERS.USER_ID.isNotNull)
+          .fetch()
+          .forEach { (userId, balenaId, facilityId) ->
+            if (userId != null && balenaId != null && facilityId != null) {
+              log.debug(
+                  "Generating new token for device manager at facility $facilityId (user $userId)")
+              generateOfflineToken(userId, balenaId, facilityId)
```
karthikbtf, line 97: should we try/catch and continue? in case a device manager is offline...

sgrimm: The "restart the device manager with the updated values" part of the flow here is asynchronous and happens completely on Balena's end. If a device manager is offline, Balena will send it the updated configuration next time it connects. So this only needs Balena's cloud service to be up.

PR #515, opened 2022-07-29
SW-693 Authenticate with AWS using OpenID Connect

sgrimm:
Update the GitHub Actions workflow to stop using an access key to authenticate
with AWS during deployment. Instead, GitHub will generate a signed OpenID
Connect token and use it to request temporary access to an AWS role that only
allows access to the specific operations we need for deployment.

This will eliminate the need to rotate our access key periodically.


sgrimm:
Some references for context:

* https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
* https://github.com/terraware/infra/commit/faf60b0bfcba6e9109fe800851c33e03ddc44b20

PR #523, opened 2022-08-03
SW-1321 Add search fields for accession age

sgrimm:
Add a new search field type that derives ages from date columns, and use it for
accession age based on collected dates. Ages can be in units of months or years;
these are represented as separate search columns so the correct values will be
included in exported CSV files.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/field/AgeField.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/field/AgeField.kt
@@ -1,0 +1,147 @@
+package com.terraformation.backend.search.field
+
+import com.terraformation.backend.search.FieldNode
+import com.terraformation.backend.search.SearchFilterType
+import com.terraformation.backend.search.SearchTable
+import java.time.Clock
+import java.time.LocalDate
+import java.util.*
+import org.jooq.Condition
+import org.jooq.Field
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #523** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/523" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/523?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/field/AgeField.kt
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  object MonthGranularity : AgeGranularity {
+    override fun truncate(date: LocalDate): LocalDate {
+      return LocalDate.of(date.year, date.month, 1)
+    }
+
+    override fun difference(startDate: LocalDate, endDate: LocalDate): Int {
+      return (endDate.year - startDate.year) * 12 + endDate.monthValue - startDate.monthValue
```
karthikbtf: nit: parens for readability?
` return ((endDate.year - startDate.year) * 12) + (endDate.monthValue - startDate.monthValue)`

PR #524, opened 2022-08-03
SW-1129 Add license file

sgrimm:
In preparation for our open-source release, add the Apache 2.0 license to the
repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #524** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/524" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/524?utm_source=stack-comment).

sgrimm:
We should first make sure we want to use the same license. I expect we'll want to use Apache everywhere, but in theory there could be reasons to choose some other license for a particular repo. We'll need to check with the legal team beforehand.

karthikbtf:
> We should first make sure we want to use the same license. I expect we'll want to use Apache everywhere, but in theory there could be reasons to choose some other license for a particular repo. We'll need to check with the legal team beforehand.

Thanks! cc @alexTerraformation I will bring this up in the 1:1

PR #526, opened 2022-08-03
Add README section about the open-source release

sgrimm:


sgrimm:
* main
  * **PR #526** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/526" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/526?utm_source=stack-comment).

sgrimm:
Adding a logo is a reasonable idea but here's a challenge with just using that SVG:

Screen Shot 2022-08-04 at 8 44 32 AM

sgrimm:
On second thought, maybe keeping it pure text isn't so bad. Looking around at some other GitHub repos from companies where they're releasing code that's strongly tied to their commercial products (as opposed to random utility libraries and such) it is surprisingly hard to find even one example of a company logo in a README. I hadn't really noticed until I went looking for it. Not to say we can't, but it doesn't seem to be the norm.

karthikbtf:
> On second thought, maybe keeping it pure text isn't so bad. Looking around at some other GitHub repos from companies where they're releasing code that's strongly tied to their commercial products (as opposed to random utility libraries and such) it is surprisingly hard to find even one example of a company logo in a README. I hadn't really noticed until I went looking for it. Not to say we can't, but it doesn't seem to be the norm.

Makes sense 👍

PR #532, opened 2022-08-08
SW-1322 Add viability test selection

sgrimm:
Add a `selected` field to the viability test objects in the accession API
payloads. Selecting a viability test doesn't actually do anything yet; this is
just the implementation of the ability to select and deselect tests. Subsequent
changes will add logic to use the results of the selected test as the accession's
overall viability.

At most one test may be marked as selected for a particular accession. It is
completely valid for none of an accession's tests to be selected.

PR #533, opened 2022-08-08
SW-1322 Add viability test selection as a search field

sgrimm:
Add a boolean search field that acts the same as the `selected` field on viability
test API payloads.

Under the covers, this is a new kind of search field that wraps an `EXISTS()` SQL
expression, since field selection is actually stored in a separate table rather
than as a simple boolean column on `viability_tests`. From the client's point of
view, though, it acts like any other boolean search field.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt
@@ -8,6 +8,7 @@ import com.terraformation.backend.search.field.BooleanField
 import com.terraformation.backend.search.field.DateField
 import com.terraformation.backend.search.field.DoubleField
 import com.terraformation.backend.search.field.EnumField
+import com.terraformation.backend.search.field.ExistsField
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #532** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/532" target="_blank">Graphite</a> 
    * **PR #533** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/533" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/533?utm_source=stack-comment).

karthikbtf (approved):
Looks good to me. 
Guessing we leverage this to search for accessions (in the table UI) that have a selected viability test?

sgrimm:
> Guessing we leverage this to search for accessions (in the table UI) that have a selected viability test?

My guess is it'll be more useful as a search result field than as a filter criterion, but yeah, you absolutely could use it that way.

PR #543, opened 2022-08-10
Remove unused summary endpoint

sgrimm:
The client code no longer calls the old `/api/v1/seedbank/summary/{facilityId}`
endpoint, so remove it from the API.

PR #544, opened 2022-08-10
SW-1331 Add seed counts to summary API

sgrimm:
To support showing the total seeds remaining on the accessions dashboard, add a
child object to the `/api/v1/seedbank/summary` response payload that has:

* The total number of seeds remaining in active count-based accessions
* The estimated total number of seeds remaining in active weight-based accessions
  that have subset weight/count
* The estimated total number of seeds remaining across all active accessions
  (this is just the sum of the first two numbers)
* The number of weight-based accessions that don't have subset weight/count and
  thus can't be included in the seed count estimate

The web app can use those four numbers to show a total seed count with indicators
to tell the user that the count is an estimate (`subtotalByWeightEstimate` is
nonzero) and that there are more seeds than indicated by the number
(`unknownQuantityAccessions` is nonzero).

The totals are in a separate payload class because a future change will add a way
to get the same summary statistics for search results, and we'll reuse this new
class in the response.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
@@ -17,7 +17,6 @@ import java.time.Clock
 import java.time.DayOfWeek
 import java.time.ZonedDateTime
 import org.springframework.web.bind.annotation.GetMapping
-import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #543** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/543" target="_blank">Graphite</a> 
    * **PR #544** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/544" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/544?utm_source=stack-comment).

karthikbtf (approved):
looks good to me

PR #548, opened 2022-08-11
Enable cascading deletes on accession child tables

sgrimm:
The initial version of the app didn't allow deleting accessions, so I erred on the
side of safety and didn't put `ON DELETE CASCADE` clauses on the foreign-key
constraints on the various child tables.

Now we support deleting accessions, though, so it makes sense to modify the
constraints to auto-delete rows in tables that exist purely to hold lists of
values in accessions. There is never a situation where, e.g., we'll want the
existence of a viability test on an accession to cause us to be unable to delete
the accession.

However, we _do_ want that behavior for accession photos: if we haven't deleted
an accession's photos from the file store yet, we shouldn't be able to delete the
accession since it would cause the files to be orphaned.

PR #547, opened 2022-08-11
SW-1398 Track user ID in accession state history 2/2

sgrimm:
Backfill user IDs into the existing state history rows as best we can, and make
the user ID mandatory.

Unfortunately, there's no way to accurately attribute past state changes to
specific users, especially since the system can automatically change accessions'
states based on scheduled events. So we attribute all state changes to the system
user, aside from the initial ones which we know will always have been caused by
the user who created the accession.

PR #546, opened 2022-08-11
SW-1398 Track user ID in accession state history 1/2

sgrimm:
We already keep track of the history of accession state changes; this was needed
to support the "changes since last week" summary statistics. Those statistics are
gone in the new version of the web app, but we're going to need to start showing
the history of changes to an accession, including which user made each change.

Add a user ID column to `accession_state_history` and start populating it.

The second change in this series will backfill the existing history entries and
make the column non-nullable. These changes need to be deployed in sequence to
maintain bidirectional compatibility during code deployment.

PR #549, opened 2022-08-11
SW-1228 Add endpoint to fetch accession history

sgrimm:
In the new version of the web app, we need to show a tab with the history of
certain kinds of changes to the accession. Add an endpoint that returns a
time-ordered list of items to show in that UI.

This initial implementation only includes accession creation and state change
events, but those should be enough to start building the UI. Additional types
of history will be added later, but the same API endpoint will return them.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
+++ b/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
@@ -1,8 +1,11 @@
 package com.terraformation.backend.i18n
 
+import com.terraformation.backend.db.AccessionState
 import com.terraformation.backend.db.GrowthForm
 import com.terraformation.backend.db.SeedStorageBehavior
 import com.terraformation.backend.db.tables.pojos.DevicesRow
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 import javax.annotation.ManagedBean
 
 /** Helper class to encapsulate notification message semantics */
@@ -102,7 +105,28 @@ class Messages {
           title = "$deviceName cannot be detected.",
           body = "$deviceName cannot be detected. Please check on it.")
 
+  fun historyAccessionCreated() = "created accession"
+
+  fun historyAccessionStateChanged(newState: AccessionState) =
+      "updated the status to ${newState.displayName}"
+
+  /**
+   * Returns the full name of a user based on their first and last names. It's possible for users to
+   * not have first or last names, e.g., if they were created by being added to an organization and
+   * haven't gone through the registration flow yet; returns null in that case. If the user has only
+   * a first name or only a last name, returns whichever name exists.
+   */
+  fun userFullName(firstName: String?, lastName: String?): String? =
+      if (firstName != null && lastName != null) {
+        "$firstName $lastName"
+      } else {
+        lastName ?: firstName
+      }
+
   private val validGrowthForms = GrowthForm.values().joinToString { it.displayName }
   private val validSeedStorageBehaviors =
       SeedStorageBehavior.values().joinToString { it.displayName }
+
+  /** Renders a date in the appropriate format. For now, this is always ISO YYYY-MM-DD format. */
+  private fun LocalDate.render() = DateTimeFormatter.ISO_LOCAL_DATE.format(this)
 }
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
@@ -1,17 +1,57 @@
 package com.terraformation.backend.seedbank.model
 
 import com.terraformation.backend.db.UserId
+import java.time.Instant
 import java.time.LocalDate
 
 enum class AccessionHistoryType {
   Created,
+  /** Manual edits to the accession's seed quantity, not including withdrawals. */
+  QuantityUpdated,
   StateChanged,
+  /** A withdrawal for the purpose of viability testing. */
+  ViabilityTesting,
+  /** A withdrawal for a purpose other than viability testing. */
+  Withdrawal,
 }
 
 data class AccessionHistoryModel(
+    /**
+     * When the underlying object representing the event was created. Used as a secondary sort key
+     * when multiple events happened on the same date.
+     */
+    val createdTime: Instant,
+    /**
+     * The effective date of the event. For example, if the event is a withdrawal, this would be the
+     * withdrawal date entered by the user (which might not be the same day as [createdTime].)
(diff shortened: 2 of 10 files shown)


sgrimm:
* main
  * **PR #546** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/546" target="_blank">Graphite</a> 
    * **PR #547** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/547" target="_blank">Graphite</a> 
      * **PR #548** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/548" target="_blank">Graphite</a> 
        * **PR #549** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/549" target="_blank">Graphite</a>  👈
          * **PR #550** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/550" target="_blank">Graphite</a> 
            * **PR #551** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/551" target="_blank">Graphite</a> 
              * **PR #552** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/552" target="_blank">Graphite</a> 
                * **PR #553** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/553" target="_blank">Graphite</a> 
                  * **PR #555** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/555" target="_blank">Graphite</a> 
                    * **PR #556** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/556" target="_blank">Graphite</a> 
                      * **PR #557** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/557" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/549?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -102,7 +105,28 @@ class Messages {
...
+  fun historyAccessionStateChanged(newState: AccessionState) =
+      "updated the status to ${newState.displayName}"
+
+  /**
+   * Returns the full name of a user based on their first and last names. It's possible for users to
+   * not have first or last names, e.g., if they were created by being added to an organization and
+   * haven't gone through the registration flow yet; returns null in that case. If the user has only
+   * a first name or only a last name, returns whichever name exists.
+   */
+  fun userFullName(firstName: String?, lastName: String?): String? =
```
karthikbtf, line 119: would using email be a fallback option?

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -102,7 +105,28 @@ class Messages {
           title = "$deviceName cannot be detected.",
           body = "$deviceName cannot be detected. Please check on it.")
 
+  fun historyAccessionCreated() = "created accession"
+
+  fun historyAccessionStateChanged(newState: AccessionState) =
+      "updated the status to ${newState.displayName}"
```
karthikbtf, line 111: would we need a `from` state since user will be able to update state to any value (doesn't have to follow logically)

karthikbtf (approved):
This looks great, was way simpler than I thought it would be. Thanks!
A few questions, mostly suggestions that can be ignored.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
sgrimm: The history list will always have all the previous transitions, so they should be able to infer the "from" state by looking at the preceding state change.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
sgrimm: The PRD just says to include full name. One hesitation I have with putting email in there too is a permission question: currently, you can have permission to view accession data but not permission to view the organization's user list. Exposing people's email addresses to non-admin users probably needs to be a product decision.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
karthikbtf: yes, makes sense

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
karthikbtf: good point, also may have eula or service terms disclosure implications. i keep hearing that there might be one signed in account that multiple users use.. maybe this won't be an issue after all.. history may show that one signed in user most of the time.

sgrimm:
I updated this to pull in some changes I ended up making when I added support for withdrawal events.

src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
```diff
@@ -1,17 +1,57 @@
...
+) : Comparable<AccessionHistoryModel> {
+  /**
+   * Compares two history models for sorting purposes. We sort history in reverse [date] order, with
+   * each day's history sorted in reverse [createdTime] order. We do it this way rather than just
+   * sorting by [createdTime] because some kinds of history can be backdated, e.g., on Tuesday you
+   * can tell the system that a withdrawal happened on Monday, and it should show up with Monday's
+   * date in the history records.
+   */
+  override fun compareTo(other: AccessionHistoryModel): Int {
+    // We sort history in reverse time order, so the comparison results are flipped here.
```
karthikbtf, line 48: Most recent event appears last?

karthikbtf:
> I updated this to pull in some changes I ended up making when I added support for withdrawal events.

Changes look good. I won't merge since target branch is something other than `main`.

src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
sgrimm: Most recent first, to match what's in the designs. That said: do you think it'd be better to return the results in chronological (earliest first) order and let the client display them in reverse order? I was trying to make the API match the UI here but maybe that's the wrong design choice.

karthikbtf: Ideally maybe .. since Ui will be the only client using this API, we could leave it as-is.

sgrimm: OK, will leave it for now and we can easily revisit later if we want.

PR #561, opened 2022-08-16
Remove accession search endpoints

sgrimm:
The last use of the old `/api/v1/seedbank/search` endpoints on the client side
went away a few months ago, so it's safe to remove from the API. The controller
class for those endpoints was the only place aside from test code where we were
using `AccessionSearchService`, so move the search construction logic from there
to `SearchServiceTest`.

Update `scripts/search.py` to call the generic search endpoints as well.


sgrimm:
* main
  * **PR #561** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/561" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/561?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #563, opened 2022-08-17
Simplify mapping of AccessionState to AccessionActive

sgrimm:
Commit eabb60b9 added the ability to include additional columns in the enum
classes that are generated from the database. Use it to expose the `active`
column on `accession_states`, and then map that to `AccessionActive` values
rather than using an exhaustive `when` statement.

PR #562, opened 2022-08-17
SW-1426 Add new accession states

sgrimm:
The new version of the app is going to switch to a simplified list of accession
states. Some of them are the same as existing states, but there are also a few
brand-new ones. Add them to the list of states.

No accessions should ever be in these states at the moment, but with this change,
they'll be included in the OpenAPI schema and will also be returned by API
endpoints that list all the values of enums.

PR #564, opened 2022-08-17
SW-1425 Add manual state flag to accessions

sgrimm:
The new version of the app will allow users to manually edit accession states.
There will still be some automatic state transitions but far fewer than before,
and they will follow a different set of rules.

Add support for manual state management in the data model, including the rules
for automatic transitions. This is tracked on a per-accession basis and an
accession can be moved back and forth between manual and automatic state
management, though the conversion is a bit lossy since the list of valid states
isn't the same in the two worlds.

This only changes the data model and the transition rules; the API behavior
remains the same as before. If you `GET` an accession whose state was manually
set to one of the newly-introduced states, it will be automatically converted
to one of the previously-existing states for backward compatibility.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -31,6 +31,7 @@ import com.terraformation.backend.seedbank.model.AccessionSummaryStatistics
 import com.terraformation.backend.seedbank.model.SeedQuantityModel
 import com.terraformation.backend.seedbank.model.ViabilityTestModel
 import com.terraformation.backend.seedbank.model.activeValues
+import com.terraformation.backend.seedbank.model.isV2Compatible
 import com.terraformation.backend.species.SpeciesService
 import com.terraformation.backend.time.toInstant
 import java.math.BigDecimal
@@ -141,6 +142,7 @@ class AccessionStore(
           founderId = record[FOUNDER_ID],
           geolocations = record[geolocationsField],
           id = accessionId,
+          isManualState = record[IS_MANUAL_STATE] ?: false,
           latestViabilityPercent = record[LATEST_VIABILITY_PERCENT],
           latestViabilityTestDate = record[LATEST_GERMINATION_RECORDING_DATE],
           numberOfTrees = record[TREES_COLLECTED_FROM],
@@ -185,6 +187,21 @@ class AccessionStore(
         accession.facilityId ?: throw IllegalArgumentException("No facility ID specified")
     val organizationId =
         parentStore.getOrganizationId(facilityId) ?: throw FacilityNotFoundException(facilityId)
+    val state =
+        when {
+          !accession.isManualState -> AccessionState.AwaitingCheckIn
+          accession.state == null -> AccessionState.AwaitingCheckIn
+          accession.state == AccessionState.UsedUp ->
+              throw IllegalArgumentException("Accessions cannot be set to Used Up at creation time")
+          accession.state.isV2Compatible -> accession.state
+          else -> throw IllegalArgumentException("Initial state must be v2-compatible")
+        }
+    val checkedInTime =
+        if (state != AccessionState.AwaitingCheckIn) {
+          clock.instant()
+        } else {
+          null
+        }
 
     requirePermissions { createAccession(facilityId) }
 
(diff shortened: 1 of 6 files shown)


sgrimm:
* main
  * **PR #562** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/562" target="_blank">Graphite</a> 
    * **PR #563** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/563" target="_blank">Graphite</a> 
      * **PR #565** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/565" target="_blank">Graphite</a> 
        * **PR #566** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/566" target="_blank">Graphite</a> 
          * **PR #564** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/564" target="_blank">Graphite</a>  👈
          * **PR #567** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/567" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/564?utm_source=stack-comment).

sgrimm:
Marking this as a draft because the list of new state transition rules still needs final approval from the product team. But if they ask for tweaks to the rules, the changes will most likely be limited to `AccessionModel.getManualStateTransition()` and the rest of the PR should remain unchanged.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -235,6 +252,7 @@ class AccessionStore(
                         .set(FAMILY_NAME, accession.family)
                         .set(FIELD_NOTES, accession.fieldNotes)
                         .set(FOUNDER_ID, accession.founderId)
+                        .set(IS_MANUAL_STATE, if (accession.isManualState) true else null)
```
sgrimm, line 244: Storing this as true/null rather than true/false mostly so it doesn't force me to add `isManualState = false` to all the `AccessionsRow` objects in various test classes.

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
```diff
@@ -151,6 +208,35 @@ data class AccessionModel(
...
+      clock: Clock
+  ): AccessionStateTransition? {
+    val seedsRemaining = newModel.calculateRemaining(clock)
+    val allSeedsWithdrawn = seedsRemaining != null && seedsRemaining.quantity <= BigDecimal.ZERO
+    val oldState = state ?: AccessionState.AwaitingProcessing
+    val newState = newModel.state ?: AccessionState.AwaitingProcessing
+    val alreadyCheckedIn = oldState != AccessionState.AwaitingCheckIn
+    val revertingToAwaitingCheckIn =
+        alreadyCheckedIn && newModel.state == AccessionState.AwaitingCheckIn
+    val addingSeedsWhenUsedUp =
```
karthikbtf, line 216: Just a weird corner case since the client supports saving accessions without a quantity set, if seedsRemaining is null.. we may flag this as addingSeedsWhenUsedUp - can only happen from incorrect usage of client so maybe it's ok.

karthikbtf (approved):
Looks good, thanks for this work, makes the UI effort easier.

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
sgrimm: I can't seem to figure out that corner case -- can you describe the sequence of changes that would end up going wrong? I definitely want this to be as airtight as possible.

karthikbtf: Saving an accession with no quantity set and updating state as 'used up'.
A subsequent update to this accession would move its state to 'in storage' (addingSeedsWhenUsedUp will be true).
This is a nonsensical corner case as it makes no sense but client would support saving such an accession with no quantity. I could be wrong about the 'used up' state being supported for such an accession.

sgrimm: Ah, got it. Yeah, that'd end up wrong. The real bug there is you shouldn't be able to set the state to `UsedUp` if there's no quantity. Just updated the PR with a fix for that (and a test case to cover it).

karthikbtf: Thanks @sgrimm , changes look good. I'll create a ticket for FE/UX to handle this case as well.

PR #569, opened 2022-08-19
Fix bug in sort order generation in search queries

sgrimm:
The search code generates `ORDER BY` clauses that use field positions to refer to
fields that are also in the `SELECT` clause, e.g., `SELECT a, b FROM c ORDER BY 2`
to sort by the `b` column. It thus keeps track of what order the fields are added
to the `SELECT` clause so it can use the correct position numbers.

Unfortunately, the code that was actually generating the `SELECT` clause would
always put all the scalar fields first in the list, before all the sublist fields,
even if the caller had specified that the sublist fields should come first. Thus
the `ORDER BY` clause would sometimes use the wrong positions, causing the results
to be incorrectly sorted.

Modify the code to put the fields in the positions the `ORDER BY` construction
code expects them to be, and add a test case that verifies the fix.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -903,15 +903,20 @@ class NestedQueryBuilder(
    * query, any fields that the parent query will need to be able to use in its `ORDER BY` clause.
    */
   private fun getSelectFields(): List<Field<*>> {
-    val selectFields = scalarFields.values.flatMap { it.selectFields }
+    val fieldsInSelectOrder =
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #569** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/569" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/569?utm_source=stack-comment).

sgrimm:
Found this one pretty indirectly: it's the reason https://github.com/terraware/terraware-server/pull/568 is failing.

karthikbtf (approved):

PR #570, opened 2022-08-19
Add a viability test type for cut tests

sgrimm:
The v2 API will represent cut tests as a third type of viability test rather than
as one-off accession attributes. But for v1 clients, we want to pretend there are
still only two valid test types. Update the data model to add the `Cut` type, and
add a shim to the v1 API to hide it as a valid type.

No behavior change here since there's currently no way to actually create a cut
test in the new style.


sgrimm:
* main
  * **PR #570** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/570" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/570?utm_source=stack-comment).

karthikbtf (approved):
looks good to me

PR #576, opened 2022-08-25
SW-1357 Add latest observed quantity/time to v2 API

sgrimm:
In order to display a warning to the user when a change in the date of a
withdrawal would affect the remaining quantity of an accession, the frontend
code needs to know the time of the most recent user-observed accession
quantity. That timestamp won't be shown to users, just used by the frontend code.

Add `latestObservedQuantity` and `latestObservedTime` fields to the v2
accession response payloads. These are computed values, so they're not included
in the request payloads.

The v2 API doesn't support directly updating the remaining quantity yet, so for
now, these two new fields are always calculated from the v1 quantity, similarly
to the way we do an on-the-fly mapping of v1 accession states to v2 ones.
Subsequent changes will start to introduce the write path for v2-style
quantities.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
@@ -24,6 +24,7 @@ import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Schema
 import io.swagger.v3.oas.annotations.responses.ApiResponse
 import java.time.Clock
+import java.time.Instant
 import java.time.LocalDate
 import javax.validation.Valid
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #576** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/576" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/576?utm_source=stack-comment).

sgrimm:
Failure is because of a precision mismatch between PostgreSQL and Linux timestamps; will fix.

karthikbtf (approved):
Looks good to me, I understand there's a fix for tests coming up.

karthikbtf (approved):

sgrimm:
Oops, I typo-d the issue number for this. Should be SW-1537.

PR #577, opened 2022-08-26
SW-1537 Define v2 withdrawals endpoints

sgrimm:
We want the v2 accessions API to be more fine-grained than the v1 one. Start
moving in that direction by introducing a separate set of endpoints for reading
and writing withdrawals.

`PUT /api/v2/accessions/{id}` will no longer create, delete, or update
withdrawals.

**The POST and PUT endpoints don't work yet!** They can't be implemented until
the broader updates to quantity tracking are in place. This change just defines
the URL paths and payload formats so client developers can see what they look
like.

The GET and DELETE endpoints don't depend on the new quantity tracking logic, so
they're implemented here, though the quantity values are copied verbatim from
the v1 data and may not be consistent with v2 semantics.

The payloads defined here differ from the ones that are included as children of
the accession object in the v1 API. There are now separate payloads for GET,
POST, and PUT operations, and some fields have been added and removed.

Added fields:

* `accessionId` (POST) = which accession to withdraw from.
* `estimatedCount` (GET) = the estimated number of seeds withdrawn.
* `estimatedWeight` (GET) = the estimated weight of seeds withdrawn.
* `withdrawnByName` (GET) = the full name of the person who withdrew the seeds.
  For withdrawals created with the v1 API, this will be the same as the v1
  `staffResponsible` field, but in the v2 world, it will be the full name of
  a Terraware user.
* `withdrawnByUserId` (GET, POST, PUT) = the ID of the user who withdrew the
  seeds. The way this field works can change depending on the permissions of
  the user who's making the request; see the schema docs.

Removed fields:

* `destination` is being removed from the product.
* `estimatedQuantity` is replaced by `estimatedCount` and `estimatedWeight`.
* `remainingQuantity` is no longer tracked on a per-withdrawal basis.
* `staffResponsible` is replaced by `withdrawnByName` (GET) and
  `withdrawnByUserId` (GET/POST/PUT).
* `weightDifference` isn't meaningful without `remainingQuantity`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
@@ -16,6 +16,7 @@ import com.terraformation.backend.db.SpeciesId
 import com.terraformation.backend.db.StorageLocationId
 import com.terraformation.backend.db.UploadId
 import com.terraformation.backend.db.UserId
+import com.terraformation.backend.db.WithdrawalId
 import com.terraformation.backend.db.tables.references.ACCESSIONS
 import com.terraformation.backend.db.tables.references.AUTOMATIONS
 import com.terraformation.backend.db.tables.references.DEVICES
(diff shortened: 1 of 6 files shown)


sgrimm:
* main
  * **PR #577** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/577" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/577?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #581, opened 2022-08-30
Support weight/count conversions in SeedQuantityModel

sgrimm:
We're going to start allowing count-based withdrawals from weight-based accessions
and vice versa, provided there's a subset weight/count available to convert
between the two.

Enhance `SeedQuantityModel.toUnits()` to support these conversions.

This also fixes imperial unit conversions; previously they worked by converting to
and from metric units, but since metric-to-imperial conversion is approximate, the
end result was incorrect. For example, if you converted 16 ounces to pounds, the
value would be 0.999997995 instead of 1.


sgrimm:
* main
  * **PR #581** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/581" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/581?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #583, opened 2022-08-31
SW-1252 Remove viability test selection

sgrimm:
Previously, we'd planned to let users select a viability test whose result would
be used as the accession's viability percentage; once a test was selected, the
accession's viability% would no longer be directly editable. The server thus
needed to keep track of which test, if any, was selected.

We've decided to change the behavior of the app such that the user can edit the
accession's viability% at any time, and finishing a test offers to copy the test's
percentage to the accession.

Since there's no more concept of selecting a test, we can get rid of the support
for test selection that was added in commit 5c239e5 and the support for searching
whether or not a test is selected, which was added in commit 939f3db.


sgrimm:
* main
  * **PR #583** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/583" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/583?utm_source=stack-comment).

karthikbtf (approved):
Sorry about the selected-test effort.

PR #585, opened 2022-09-01
SW-1537 Implement withdrawals POST, PUT endpoints

sgrimm:
Add implementations for the withdrawals POST (create) and PUT (update) operations.

Because updating withdrawals causes accession-level information to change, make
the write operations return the entire accession rather than just the information
about the updated withdrawal. In practice, we expect clients will always want the
latest accession details after updating a withdrawal, so no point sending back a
useless response payload and forcing the client to GET the one it actually wants.

Move the endpoints under `/api/v1/seedbank/accessions/{id}` to make it less
surprising that they now return accession payloads.

The new endpoints currently only work for count-based accessions; support for
weight-based accessions will be added as part of a larger change to quantity
management.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/AccessionService.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/AccessionService.kt
@@ -2,13 +2,22 @@ package com.terraformation.backend.seedbank
 
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.AccessionId
+import com.terraformation.backend.db.WithdrawalId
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #585** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/585" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/585?utm_source=stack-comment).

karthikbtf (approved):
Looks good to me.

PR #590, opened 2022-09-06
SW-1561 Honor search scopes on all tables

sgrimm:
`/api/v1/seedbank/values/all` has a concept of a "search scope" where we limit
results to the ones relevant to a particular organization or facility. The scope
was being honored for searches on the accessions table, but not for searches on
other tables. So if you asked for, e.g., all the values of `bags_number` for a
given facility ID, you'd actually get all the bag numbers across all the
facilities you had access to.

Eventually we're going to want to revamp how that endpoint works (SW-552) but
in the meantime, the API ought to work as advertised.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
@@ -216,9 +216,8 @@ class SearchService(private val dslContext: DSLContext) {
           val selectFields =
               field.selectFields + listOf(field.orderByField.`as`(DSL.field("order_by_field")))
           val searchTable = fieldPath.searchTable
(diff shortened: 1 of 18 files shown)


sgrimm:
* main
  * **PR #590** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/590" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/590?utm_source=stack-comment).

sgrimm:
Build failure looks like a GitHub Actions glitch; retrying.

karthikbtf (approved):
Thanks for this fix. 
Looks like my doing.. I solved a local problem and didn't extend SearchScope for all use-cases.

PR #593, opened 2022-09-07
Use fully qualified github ref name in concurrency check

karthikbtf:


sgrimm (approved):
Thanks! This just bit me yesterday and my thought was, "Maybe I should turn off cancel-in-progress on main," not, "Hey, I turned off cancel-in-progress on main! Why didn't it work?"

PR #596, opened 2022-09-07
SW-1566 Add viability test endpoints

sgrimm:
Add a set of endpoints under `/api/v1/seedbank/accessions/{id}/viabilityTests` to
read and write viability tests. These work the same way as the withdrawals
endpoints: they return the entire accession.

Remove the `viabilityTests` field from the v2 accession PUT payload; the new
endpoints are the only way to write viability tests in the v2 API.

The v2 accession GET response payload no longer has the fields related to the
v1-style viability percentage calculations, since they don't make sense in the
context of v2-style tests:

* `latestViabilityPercent`
* `latestViabilityTestDate`
* `totalViabilityPercent`

A future change will add a new field to hold the user-supplied viability
percentage.

### DIFF
--- a/scripts/client.py
+++ b/scripts/client.py
@@ -1,6 +1,6 @@
 from argparse import ArgumentParser, Namespace
 import requests
-from typing import Dict, Optional
+from typing import Optional
 
 DEFAULT_URL = "http://localhost:8080"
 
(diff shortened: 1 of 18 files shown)


sgrimm:
* main
  * **PR #596** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/596" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/596?utm_source=stack-comment).

sgrimm:
I should point out a design choice I made here, in case people disagree: I didn't add separate endpoints for updating test results. They are still an array of child objects here. There isn't really a technical reason; it was more of a gut feeling that test results are too tiny for it to be useful to individually address them. But obviously I can add endpoints for them easily enough.

karthikbtf:
> I should point out a design choice I made here, in case people disagree: I didn't add separate endpoints for updating test results. They are still an array of child objects here. There isn't really a technical reason; it was more of a gut feeling that test results are too tiny for it to be useful to individually address them. But obviously I can add endpoints for them easily enough.

I'm for your current design choice, given current requirements and UI views.

PR #598, opened 2022-09-08
SW-1573 Support notes in V2 accession create payload

karthikbtf:
This is needed to set notes when creating an accession in v2 seed management
Tested locally in my UI.


sgrimm:
No technical objection to this but I want to make sure we really need it. Just looked through the design again and I believe the API already has enough notes fields. I think it's hard to see because the design uses inconsistent field names (both internally inconsistent and inconsistent with the field list in the PRD) and I based my changes off the PRD.

* `collectionSiteNotes` = "Site Description" (in the create UI), "Direction or Description" (in the edit UI)
* `notes` = "Plant Description" (in the create UI), "Environmental Notes" (in the edit UI)

sgrimm:
Once we've settled on the user-facing field names, we can rename the fields in the API. A little reluctant to do it now since there seem to be competing sources of truth for what the names should be.

karthikbtf:
> No technical objection to this but I want to make sure we really need it. Just looked through the design again and I believe the API already has enough notes fields. I think it's hard to see because the design uses inconsistent field names (both internally inconsistent and inconsistent with the field list in the PRD) and I based my changes off the PRD.
> 
> * `collectionSiteNotes` = "Site Description" (in the create UI), "Direction or Description" (in the edit UI)
> * `notes` = "Plant Description" (in the create UI), "Environmental Notes" (in the edit UI)

Yes, I totally understand. The PRD should be the source of truth, I will go with `notes` as planned/intended. This is currently not exposed in the create payload - I can update this PR to support notes in the create payload.

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
```diff
@@ -191,6 +191,25 @@ internal class AccessionStoreTest : DatabaseTest(), RunsAsUser {
...
+            facilityId = facilityId,
+            createdBy = user.userId,
+            createdTime = clock.instant(),
+            dataSourceId = DataSource.Web,
+            modifiedBy = user.userId,
+            modifiedTime = clock.instant(),
+            number = accessionNumbers[0],
+            processingNotes = "test processing notes",
+            stateId = AccessionState.AwaitingCheckIn),
+        accessionsDao.fetchOneById(AccessionId(1)))
```
sgrimm: The intent might be clearer if this were
```
assertEquals(
    "test processing notes",
    accessionsDao.fetchOneById(AccessionId(1)).processingNotes)
```

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/seedbank/db/AccessionStoreTest.kt
karthikbtf: Good point, thanks.

karthikbtf:
@sgrimm looks like all the PRs are failing a type check in create_accession.py script.
```
create_accessions.py:185: error: List comprehension has incompatible type List[Optional[str]]; expected List[str]
```
Does the return type need to be `List[Optional[str]]` instead of `Optional[List[Optional[str]]]`? Not sure where the `Optional[str]` comes from - I don't see the code generating that.

sgrimm:
PR is out to fix that type check failure. Once that's in, we can rebase this change and checks should pass.

PR #605, opened 2022-09-08
SW-1571 v2->v1: Recalculate "initial" quantity

sgrimm:
In the v1 logic for count-based accessions, the remaining quantity is derived from
the initial (total) quantity and the quantities of all the withdrawals, and it has
to add up to zero or more. In the v2 logic, since the user can reset the remaining
quantity at will, the withdrawal quantities can add up to more than the initial
quantity. Previously, that situation would cause an `IllegalArgumentException`
in v2->v1 conversion because the calculated remaining quantity would be negative.

Fix it by recalculating the "initial" quantity in v2->v1 conversion so the numbers
all match up.

Weight-based accessions don't need this fix because the remaining quantity is user
input rather than a calculated value in v1.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
@@ -169,9 +169,24 @@ data class AccessionModel(
             else -> ProcessingMethod.Weight
           }
 
+      // For count-based accessions, v1 recomputes the remaining quantity as the total (initial)
+      // quantity minus the sum of the withdrawal amounts. Make the numbers line up by overriding
+      // the total quantity.
+      val newTotal =
+          if (remaining?.units == SeedQuantityUnits.Seeds) {
+            withdrawalsWithCorrectUnits
+                .filter { it.withdrawn != null }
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #605** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/605" target="_blank">Graphite</a>  👈
    * **PR #606** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/606" target="_blank">Graphite</a> 
      * **PR #607** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/607" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/605?utm_source=stack-comment).

karthikbtf (approved):

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/605?utm_source=stack-comment).


Job Reference: p2ethxCwfJskv5PPoFvH

PR #608, opened 2022-09-09
SW-1274 Make plantsCollectedFrom a single value

sgrimm:
We decided against changing this to a range in the new version of the mobile app.
However, we still want the field name to refer to "plants" rather than "trees."

The database hadn't yet been updated to store a range, so this is purely a change
in API payloads and search field names.


sgrimm:
* main
  * **PR #608** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/608" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/608?utm_source=stack-comment).

karthikbtf (approved):

PR #611, opened 2022-09-09
Test client: Pull error message from server response

sgrimm:
If an HTTP request fails, the server usually returns an error message. Use it as
the exception message rather than the default one from the requests library, which
just says what the HTTP response code was.

Before:

    $ echo '{"state":"xxxxx"}' | ./edit_accession.py -V 2 --bearer $BEARER 120
    [... stack trace ...]
    requests.exceptions.HTTPError: 400 Client Error:  for url: http://localhost:8080/api/v2/seedbank/accessions/120

After:

    $ echo '{"state":"xxxxx"}' | ./edit_accession.py -V 2 --bearer $BEARER 120
    [... stack trace ...]
    requests.exceptions.HTTPError: Field value invalid: state

### DIFF
--- a/scripts/client.py
+++ b/scripts/client.py
@@ -26,19 +26,19 @@ class TerrawareClient:
     def delete(self, url, **kwargs):
         kwargs_with_auth = self._add_auth_header(kwargs)
         r = requests.delete(self.base_url + url, **kwargs_with_auth)
-        r.raise_for_status()
+        self.raise_for_status(r)
         return r.json()
 
     def get(self, url, **kwargs):
         kwargs_with_auth = self._add_auth_header(kwargs)
         r = requests.get(self.base_url + url, **kwargs_with_auth)
-        r.raise_for_status()
+        self.raise_for_status(r)
         return r.json()
 
     def post_raw(self, url, **kwargs):
         kwargs_with_auth = self._add_auth_header(kwargs)
         r = requests.post(self.base_url + url, **kwargs_with_auth)
-        r.raise_for_status()
+        self.raise_for_status(r)
         return r
 
     def post(self, url, **kwargs):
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #611** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/611" target="_blank">Graphite</a>  👈
    * **PR #612** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/612" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/611?utm_source=stack-comment).

scripts/client.py
```diff
@@ -47,9 +47,18 @@ def post(self, url, **kwargs):
...
         kwargs_with_auth = self._add_auth_header(kwargs)
         r = requests.put(self.base_url + url, **kwargs_with_auth)
-        r.raise_for_status()
+        self.raise_for_status(r)
         return r.json()
 
+    @staticmethod
+    def raise_for_status(r: requests.Response):
+        if r.status_code > 399:
+            if r.headers["Content-Type"].startswith("application/json"):
```
karthikbtf, line 56: do we need to ignore case here?

karthikbtf (approved):
one question/comment

scripts/client.py
karthikbtf: i meant the header key

sgrimm: I believe the headers dict is case-insensitive but I'll double-check.

sgrimm: Yeah, it's case-insensitive (changed it to `"CONTenT-TyPe"` locally and it still worked).

PR #612, opened 2022-09-09
SW-1596 Substrates and treatments for v2 viability tests

sgrimm:
In v2, we are expanding the list of viability test treatments and changing the
substrates so that there are different lists of valid values depending on test
type. Most of the existing substrate values are also being renamed, e.g., "Agar
Petri Dish" is now "Agar."

Update the underlying data model so it uses the v2 values. Continue to accept
and return the existing values from the v1 accessions endpoints.

In cases where there is no unambiguous mapping between v1 and v2 treatments,
we remove the value rather than defaulting to a possibly-incorrect default.

The search API returns the v2 values; this doesn't break the existing search
frontend code since it already asks the server for the list of possible values.
But it does potentially cause search results to be different than before.

### DIFF
--- a/scripts/client.py
+++ b/scripts/client.py
@@ -26,19 +26,19 @@ class TerrawareClient:
     def delete(self, url, **kwargs):
         kwargs_with_auth = self._add_auth_header(kwargs)
         r = requests.delete(self.base_url + url, **kwargs_with_auth)
-        r.raise_for_status()
+        self.raise_for_status(r)
         return r.json()
 
     def get(self, url, **kwargs):
         kwargs_with_auth = self._add_auth_header(kwargs)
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #611** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/611" target="_blank">Graphite</a> 
    * **PR #612** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/612" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/612?utm_source=stack-comment).

karthikbtf (approved):
Looks good to me.

PR #613, opened 2022-09-10
SW-1600 Keep track of minimum mobile app versions

sgrimm:
To support forced updates of mobile apps if we make breaking API changes, as
well as notifying users when there significant new versions of mobile apps are
released, add an API endpoint that returns a list of our apps with a minimum
and recommended version number for each.

The list of app versions is maintained in a new repeatable database migration
so we can ensure it matches the server code as we update the API. But having to
update server-side database migrations would be annoying during mobile app
development, so the list is also editable in the admin UI for super-admin users.
Edits made in the admin UI are overwritten when the migration script is updated.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -627,6 +627,7 @@ class AdminController(
           AppVersionsRow(appName, platform, minimumVersion, recommendedVersion))
       redirectAttributes.successMessage = "App version updated."
     } catch (e: DuplicateKeyException) {
+      // User edited the appName/platform to collide with an existing one.
       redirectAttributes.failureMessage = "An entry for that app name and platform already exists."
     } catch (e: Exception) {
       log.error("Failed to update app version", e)
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -3,6 +3,7 @@ package com.terraformation.backend.customer.api
 import com.terraformation.backend.api.RequireExistingAdminRole
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.config.TerrawareServerConfig
+import com.terraformation.backend.customer.db.AppVersionStore
 import com.terraformation.backend.customer.db.FacilityStore
 import com.terraformation.backend.customer.db.OrganizationStore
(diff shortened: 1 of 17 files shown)


sgrimm:
* main
  * **PR #613** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/613" target="_blank">Graphite</a>  👈
    * **PR #617** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/617" target="_blank">Graphite</a> 
        * **PR #619** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/619" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/613?utm_source=stack-comment).

sgrimm:
This is feature-complete but is a draft for now because it implements a proposed API payload format. I think it's probably either correct as-is or close to correct, but will wait for open discussion on the Jira task to finish before making this a non-draft PR.

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -574,6 +590,63 @@ class AdminController(
...
+      @RequestParam recommendedVersion: String,
+      redirectAttributes: RedirectAttributes
+  ): String {
+    try {
+      appVersionStore.update(
+          AppVersionsRow(originalAppName, originalPlatform),
+          AppVersionsRow(appName, platform, minimumVersion, recommendedVersion))
+      redirectAttributes.successMessage = "App version updated."
+    } catch (e: DuplicateKeyException) {
+      redirectAttributes.failureMessage = "An entry for that app name and platform already exists."
```
karthikbtf, line 631: should be updatable yes?

karthikbtf (approved):
looks good to me.. @joshbuchea ?

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
sgrimm: This is if the user edits the app name or platform and it collides with an existing entry for that combination. Added a comment to clarify.

karthikbtf: Got it, thanks. Ugh yes we don't want the app/platform names embedded in the url path.. otherwise we could accept the modifiable values alone in the payload.
Makes sense now.

karthikbtf (approved):
Reviewed earlier.

PR #617, opened 2022-09-12
SW-1512 Add permissions for setting withdrawal user IDs

sgrimm:
We're adding a "withdrawn by" field to withdrawals; managers, admins, and owners
can set it to the identity of any user in the organization, but contributors can
only set it to themselves.

Add two permissions to support this logic:

* `canReadOrganizationUser` is true if you have permission to read the membership
  information for a given user in a given organization.
* `canSetWithdrawalUser` is true if you have permission to set the user ID on a
  withdrawal for an accession.

No application code uses these new permissions yet; this is groundwork for a
subsequent change.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -269,6 +269,14 @@ data class IndividualUser(
     return organizationId in organizationRoles
   }
 
+  override fun canReadOrganizationUser(organizationId: OrganizationId, userId: UserId): Boolean {
+    return if (userId == this.userId) {
+      canReadOrganization(organizationId)
+    } else {
+      canListOrganizationUsers(organizationId) && parentStore.exists(organizationId, userId)
+    }
+  }
+
   override fun canReadSpecies(speciesId: SpeciesId): Boolean {
     // If this logic changes, make sure to also change code that bakes this rule into SQL queries
     // for efficiency. Example: SpeciesStore.fetchUncheckedSpeciesIds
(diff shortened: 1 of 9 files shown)


sgrimm:
* main
  * **PR #613** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/613" target="_blank">Graphite</a> 
    * **PR #617** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/617" target="_blank">Graphite</a>  👈
        * **PR #619** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/619" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/617?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -316,6 +324,17 @@ data class IndividualUser(
   override fun canSetTestClock(): Boolean {
     return userType == UserType.SuperAdmin
   }
+
+  override fun canSetWithdrawalUser(accessionId: AccessionId): Boolean {
```
karthikbtf, line 328: Should this have userId semantics as well.. since contributors can assign themselves..?

karthikbtf (approved):
👍

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
sgrimm: My mental model here was the Unix `chown` command -- as a non-superuser you can create files and they'll be owned by your user ID, but that doesn't mean you have permission to set the ownership of files even if you have write permission.

If there's another name for this that would get that idea across better, I could definitely rename it.

karthikbtf: Thanks for the clarification, this name works just fine.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/617?utm_source=stack-comment).


Job Reference: HkgjNjnh76dIJKxenCO5

PR #618, opened 2022-09-12
Move full name concatenation to IndividualUser

sgrimm:
Previously, there were two places in the code where we concatenated users'
first and last names to construct full names. Move the logic to a single place
which should make it easier to update when it comes time to add localization.

PR #619, opened 2022-09-12
SW-1512 Track withdrawal user IDs

sgrimm:
Store the withdrawn-by user ID the client can optionally include in v2 withdrawal
API payloads, and return the user ID and full name in v2 withdrawal response
payloads. These fields were already present in the payloads, but didn't do
anything previously.

For viability tests, replace `testingStaffUserId` and `testingStaffName` with
`withdrawnByUserId` and `withdrawnByName`. Initially we'd understood the product
requirements to mean we needed to track separate user IDs for tests and
withdrawals, but it turns out there's only one. The new name should make the
semantics clearer at the API level.

Setting the withdrawn-by user ID via either API is subject to permissions:

* You can only set it to the ID of a user who is currently in the organization.
* If you are a contributor user, you can't set it to another user's ID; it will
  always default to you for newly-added withdrawals and it will always be ignored
  when you edit existing withdrawals.
* If you are editing an existing withdrawal, the withdrawal's existing user ID
  is always allowed, regardless of the other rules above. That is, you won't get
  a permission error if you leave it as is, even if the user in question is no
  longer in the organization or you're a contributor.

The history API is updated to return withdrawal user IDs where available and to
return the withdrawing user's full name rather than the `staffResponsible` field
when a user ID is present.

### DIFF
--- a/buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
+++ b/buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
@@ -109,11 +109,13 @@ val ID_WRAPPERS =
             "UserId",
             listOf(
                 "users\\.id",
+                "withdrawals\\.withdrawn_by",
                 ".*\\.user_id",
                 ".*\\.created_by",
                 ".*\\.deleted_by",
                 ".*\\.modified_by",
-                ".*\\.updated_by")),
+                ".*\\.updated_by",
+            )),
         IdWrapper(
             "ViabilityTestId",
             listOf(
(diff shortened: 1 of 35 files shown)


sgrimm:
* main
  * **PR #613** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/613" target="_blank">Graphite</a> 
    * **PR #617** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/617" target="_blank">Graphite</a> 
        * **PR #619** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/619" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/619?utm_source=stack-comment).

karthikbtf (approved):

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/619?utm_source=stack-comment).


Job Reference: HkgjNjnh76dIJKxenCO5

PR #621, opened 2022-09-14
SW-1644 Add endpoint to delete current user

sgrimm:
Add `DELETE /api/v1/users/me` to allow a user to delete their own account.

This does the following:

* Replaces their email address with an invalid non-personally-identifying value.
* Replaces their first and last names with "Deleted User".
* Replaces the Keycloak user ID in the Terraware database with an invalid value.
* Deletes their organization memberships so they no longer appear in the user
  lists of any organizations.
* Deletes them from Keycloak.
* Invalidates their login session.

After deletion:

* Any subsequent API requests will fail (except for ones that don't require
  authentication, e.g., listing the supported mobile app versions).
* They will be unable to log in with their old username and password.
* Any historical records such as accession quantity history will show a name of
  "Deleted User" instead of their name.
* If they sign up for an account again, it will be a fresh new account and will
  not inherit any data from their previous one.
* Similarly, if they log in with Google or Apple SSO, they will get a fresh new
  user identity with no access to their previous data.
* They won't get any email notifications from terraware-server.

What this doesn't do:

* Delete organization-owned data. If they uploaded an accession photo, it will
  remain, but it will show as uploaded by "Deleted User" rather than their name.
* Remove mentions of their names from freeform text fields such as accession
  notes.
* Prevent them from signing up again.


sgrimm:
* main
  * **PR #621** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/621" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/621?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -595,6 +616,100 @@ internal class PermissionTest : DatabaseTest() {
     permissions.andNothingElse()
   }
 
+  @Test
+  fun `system user can perform most operations`() {
```
sgrimm, line 620: This is maybe _slightly_ out of scope for the PR but not totally. I wanted to add a test to verify/document that the system user can't delete itself, and I realized there was no test for system user permissions at all.

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
```diff
@@ -57,7 +59,23 @@ class OrganizationService(
...
+
+  @EventListener
+  fun on(event: UserDeletedEvent) {
+    val user = userStore.fetchOneById(event.userId)
+
+    dslContext.transaction { _ ->
+      user.organizationRoles.keys.forEach { organizationId ->
+        organizationStore.removeUser(organizationId, user.userId, allowRemovingLastOwner = true)
+
+        val remainingUsers = organizationStore.countRoleUsers(organizationId)
```
karthikbtf: `remainingUsersByRole` ?

karthikbtf (approved):
Looks good to me 👍 
Just a question, should we disallow searching these defunct users in UserStore:fetchByEmail? There's no code path today that would allow this.. but just in case.

sgrimm:
Good call on adding the sanity check to `fetchByEmail`.

karthikbtf (approved):
Thanks 🎉

PR #622, opened 2022-09-15
Rename UserDeletedEvent for clarity

sgrimm:
The event class name implies the user has already been deleted, but the
event really means we're _about to start_ deleting the user. Rename it
to make this more self-evident.

PR #623, opened 2022-09-15
SW-931 Purge data when organization is deleted

sgrimm:
When the last user is removed from an organization, delete its data.

Currently, a connected device manager counts as a user; if the organization has a
facility with a connected device manager, its data will not be removed from the
system when the last human user is removed.

Deployment note: The second database migration may take a while to run on the
production database, which may cause the deploy to look like it has gotten stuck.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
@@ -1,6 +1,8 @@
 package com.terraformation.backend.customer.db
 
 import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.customer.event.OrganizationAbandonedEvent
+import com.terraformation.backend.customer.event.OrganizationDeletionStartedEvent
 import com.terraformation.backend.customer.model.FacilityModel
 import com.terraformation.backend.customer.model.OrganizationModel
 import com.terraformation.backend.customer.model.OrganizationUserModel
@@ -29,6 +31,7 @@ import javax.annotation.ManagedBean
 import org.jooq.Condition
 import org.jooq.DSLContext
 import org.jooq.impl.DSL
+import org.springframework.context.ApplicationEventPublisher
 import org.springframework.dao.DuplicateKeyException
 
 @ManagedBean
@@ -36,6 +39,7 @@ class OrganizationStore(
     private val clock: Clock,
     private val dslContext: DSLContext,
     private val organizationsDao: OrganizationsDao,
+    private val publisher: ApplicationEventPublisher,
 ) {
   private val log = perClassLogger()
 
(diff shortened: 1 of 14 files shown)


sgrimm:
* main
  * **PR #622** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/622" target="_blank">Graphite</a> 
    * **PR #623** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/623" target="_blank">Graphite</a>  👈
      * **PR #624** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/624" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/623?utm_source=stack-comment).

sgrimm:
Just realized I got something wrong with this one. Fine to review it as-is since most of it is correct, but I realize I need to treat uploads the same as photos since they also refer to URLs on the external file store.

src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
```diff
@@ -175,6 +211,16 @@ class OrganizationStore(
...
   }
 
+  fun fetchOrganizationIds(userId: UserId): List<OrganizationId> {
+    return dslContext
+        .select(ORGANIZATION_USERS.ORGANIZATION_ID)
+        .from(ORGANIZATION_USERS)
+        .where(ORGANIZATION_USERS.USER_ID.eq(userId))
+        .fetch(ORGANIZATION_USERS.ORGANIZATION_ID)
+        .filterNotNull()
+        .filter { currentUser().canListOrganizationUsers(it) }
```
karthikbtf: would this prevent contributors from deleting their org memberships?

karthikbtf (commented):
Looks good over all - do we need to look at how any accessions related jobs might be affected by the data being deleted?

src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
sgrimm: Good point. Fixed.

sgrimm:
As for accession-related jobs, not having to deal with that kind of thing in bits and pieces across the code is one of the reasons I made this use cascading deletes. The deletion of the organization and all its accessions will be atomic, so it should be impossible to be in a situation where a job does something that causes the accession to no longer be deleted.

The other way around is possible if the job, say, reads an accession into memory and then writes it later; the write would fail. Not ideal, but it's no different than someone deleting an accession from the UI just as the job is running. We'll get an error in the logs but nothing should break.

karthikbtf:
> As for accession-related jobs, not having to deal with that kind of thing in bits and pieces across the code is one of the reasons I made this use cascading deletes. The deletion of the organization and all its accessions will be atomic, so it should be impossible to be in a situation where a job does something that causes the accession to no longer be deleted.
> 
> The other way around is possible if the job, say, reads an accession into memory and then writes it later; the write would fail. Not ideal, but it's no different than someone deleting an accession from the UI just as the job is running. We'll get an error in the logs but nothing should break.

Got it,  makes sense.

karthikbtf (approved):
Looks good to me!

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/623?utm_source=stack-comment).


Job Reference: kjtinIlCwOITM24Is9QS

PR #626, opened 2022-09-16
SW-1694 Accept viability test results when creating test

sgrimm:
To handle the case where a user is entering details about a test that has
already happened, allow test results to be included when initially creating
a viability test in the v2 API.


sgrimm:
* main
  * **PR #626** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/626" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/626?utm_source=stack-comment).

sgrimm:
This is purely a controller change because we already supported this in the v1 API, so the logic to handle this is already there in `AccessionModel` and `ViabilityTestStore`.

karthikbtf (approved):

PR #630, opened 2022-09-16
SW-1577 Add cut test support to v2 API

sgrimm:
Change the internal representation of cut test results to the v2 style, where they
are represented as another kind of viability test. This new representation is used
even when accessions are written using the v1 API, but the v1 API continues to
present cut test results as a set of accession-level fields.

Existing cut test results are converted to viability tests by a database migration.

Clients can switch back and forth between the v1 and v2 APIs, but if an accession
has multiple cut tests (which is allowed in v2) they will be collapsed to a single
aggregate test when the accession is written using the v1 API.

The interaction between cut test results and accession quantity is a little
different than before: previously, we would only subtract seeds from the accession
if all three fields (compromised, empty, filled) were filled in with values. In
the new model, a withdrawal is created even if some of those values are still
null.

Deployment note: This change will require brief downtime. If a v1 PUT request with
cut test results is handled by an older server instance after the migration here is
finished, the edit will effectively be lost because the older code won't create a
new viability test.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
@@ -513,8 +513,19 @@ data class AccessionPayload(
       model.calculateTotalScheduledWithdrawalQuantity(clock)?.toPayload(),
       model.totalViabilityPercent,
       model.calculateTotalWithdrawalQuantity(clock)?.toPayload(),
-      model.viabilityTests.map { ViabilityTestPayload(it) }.orNull(),
-      model.withdrawals.map { WithdrawalPayload(it) }.orNull(),
+      model.viabilityTests
+          .filter { it.testType != ViabilityTestType.Cut }
+          .map { ViabilityTestPayload(it) }
+          .orNull(),
+      model.withdrawals
+          .filter { withdrawal ->
+            withdrawal.viabilityTestId == null ||
+                model.viabilityTests
+                    .firstOrNull { test -> test.id == withdrawal.viabilityTestId }
+                    ?.testType != ViabilityTestType.Cut
+          }
+          .map { WithdrawalPayload(it) }
+          .orNull(),
   )
 }
 
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
@@ -519,10 +519,16 @@ data class AccessionPayload(
           .orNull(),
       model.withdrawals
           .filter { withdrawal ->
-            withdrawal.viabilityTestId == null ||
-                model.viabilityTests
-                    .firstOrNull { test -> test.id == withdrawal.viabilityTestId }
-                    ?.testType != ViabilityTestType.Cut
+            // If this withdrawal is for a viability test, only include it if the test is something
+            // other than a cut test. Do a linear search to find the test with the right ID for each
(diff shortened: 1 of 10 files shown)


sgrimm:
* main
  * **PR #630** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/630" target="_blank">Graphite</a>  👈
    * **PR #631** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/631" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/630?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
```diff
@@ -513,8 +513,19 @@ data class AccessionPayload(
...
+      model.viabilityTests
+          .filter { it.testType != ViabilityTestType.Cut }
+          .map { ViabilityTestPayload(it) }
+          .orNull(),
+      model.withdrawals
+          .filter { withdrawal ->
+            withdrawal.viabilityTestId == null ||
+                model.viabilityTests
+                    .firstOrNull { test -> test.id == withdrawal.viabilityTestId }
+                    ?.testType != ViabilityTestType.Cut
```
karthikbtf: should this check happen before the firstOrNull - in case the first match is a cut test.. but there could be another potential match.?

src/main/resources/db/migration/V134__ViabilityTestsCut.sql
```diff
@@ -0,0 +1,55 @@
...
+       remaining_grams,
+       remaining_quantity,
+       remaining_units_id,
+       cut_test_seeds_compromised,
+       cut_test_seeds_empty,
+       cut_test_seeds_filled,
+       COALESCE(cut_test_seeds_compromised, 0)
+           + COALESCE(cut_test_seeds_empty, 0)
+           + COALESCE(cut_test_seeds_filled, 0),
+       3 -- Cut
```
karthikbtf: is this the equivalent of a limit 3?

karthikbtf (approved):
looks good to me, one comment but I could be mistaken

src/main/resources/db/migration/V134__ViabilityTestsCut.sql
sgrimm: Nope, but clearly my comment is too terse. 3 is the test type ID for cut tests, and `--` is the SQL comment delimiter. I'll make this clearer.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
sgrimm: Yes, good catch.

sgrimm: Looking at it again, this is actually right, but the reason there's a `firstOrNull` is not very obvious; will add a comment to clarify what's going on.

karthikbtf: Thanks for checking @sgrimm , comment sounds good 👍

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/630?utm_source=stack-comment).


Job Reference: AY6TDCXGXTR6yclJtp7E

PR #634, opened 2022-09-19
Make "seeds tested" required in v2 viability tests

sgrimm:
In v2, we don't allow users to create viability tests without saying how many
seeds they're testing. Enforce that requirement in the API.

The generated OpenAPI schema was not including the list of required fields in
the viability test payloads; this seems to be due to a bug in the OpenAPI library
that causes it to ignore the `required` attribute on write-only fields. Remove
the annotation that marked the POST and PUT payloads as write-only; it's more
important to have fields correctly marked as required.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/ViabilityTestsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/ViabilityTestsController.kt
@@ -1,7 +1,8 @@
 package com.terraformation.backend.seedbank.api
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonSetter
+import com.fasterxml.jackson.annotation.Nulls
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.AccessionId
 import com.terraformation.backend.db.UserId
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #634** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/634" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/634?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/api/ViabilityTestsController.kt
```diff
@@ -133,7 +135,7 @@ data class GetViabilityTestPayload(
       seedsCompromised = model.seedsCompromised,
       seedsEmpty = model.seedsEmpty,
       seedsFilled = model.seedsFilled,
-      seedsTested = model.seedsTested,
+      seedsTested = model.seedsTested ?: 1,
```
karthikbtf, line 138: is this to handle existing null data?

karthikbtf (approved):

src/main/kotlin/com/terraformation/backend/seedbank/api/ViabilityTestsController.kt
sgrimm: Yes. There's only one case in our existing data across prod and staging where this is null, and it's a test accession, so it didn't seem worth doing something more sophisticated here.

sgrimm: I should add that when we get rid of the v1-only parts of the data model, this will become non-nullable in the database and we'll clean up the existing data at that point. I considered doing that here but I don't want to break v1's contract which allows this to be null. Hopefully nobody else will actually create a viability test without a seed count between now and then.

PR #637, opened 2022-09-21
SW-1721 Remove "Cleaning" accession state

sgrimm:
Hazards of implementing a spec that's still in flux: originally the v2 seeds PRD
had a "Cleaning" state, and the code implements it. But that state was removed
from the PRD and replaced with the v1 "Processing" state.

Remove "Cleaning" from the list of valid states and make "Processing" valid in v2.


sgrimm:
* main
  * **PR #637** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/637" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/637?utm_source=stack-comment).

karthikbtf (approved):
We need to check FE for similar usage, cc @constanzauanini

PR #646, opened 2022-09-23
SW-1732 Preserve staff responsible in v2 API

sgrimm:
In v1, viability tests and withdrawals have a "Staff Responsible" field which is
an arbitrary text value. In v2, we've replaced that field with a Terraware user
ID. But we don't want to lose any text the users entered into the field in v1,
so add it to the "Notes" field.

If the "Notes" field already contains the text of the "Staff Responsible" field,
we don't modify it; this prevents us from adding it multiple times if an accession
repeatedly flips between v1 and v2 writes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/ViabilityTest.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/ViabilityTest.kt
@@ -120,7 +120,19 @@ data class ViabilityTestModel(
           }
         }
 
-    return copy(seedsTested = seedsTested ?: 1, substrate = newSubstrate)
+    val notesWithStaffResponsible =
+        when {
+          staffResponsible.isNullOrBlank() -> notes
+          notes.isNullOrBlank() -> "Staff responsible: $staffResponsible"
+          staffResponsible in notes -> notes
+          else -> "$notes\n\nStaff responsible: $staffResponsible"
+        }
+
+    return copy(
+        notes = notesWithStaffResponsible,
+        seedsTested = seedsTested ?: 1,
+        substrate = newSubstrate,
+    )
   }
 
   fun fieldsEqual(other: ViabilityTestModel): Boolean {
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/ViabilityTest.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/ViabilityTest.kt
@@ -124,7 +124,7 @@ data class ViabilityTestModel(
         when {
           staffResponsible.isNullOrBlank() -> notes
           notes.isNullOrBlank() -> "Staff responsible: $staffResponsible"
-          staffResponsible in notes -> notes
+          "Staff responsible: $staffResponsible" in notes -> notes
           else -> "$notes\n\nStaff responsible: $staffResponsible"
         }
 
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #646** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/646" target="_blank">Graphite</a>  👈
    * **PR #647** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/647" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/646?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/model/Withdrawal.kt
```diff
@@ -176,4 +176,16 @@ data class WithdrawalModel(
...
   }
+
+  fun toV2Compatible(): WithdrawalModel {
+    val notesWithStaffResponsible =
+        when {
+          staffResponsible.isNullOrBlank() -> notes
+          notes.isNullOrBlank() -> "Staff responsible: $staffResponsible"
+          staffResponsible in notes -> notes
+          else -> "$notes\n\nStaff responsible: $staffResponsible"
+        }
```
karthikbtf: thoughts on checking for `Staff responsible:` in notes? in case user edits staff responsible in the notes section - in which case we may append staff responsible again.
nit: this code fragment could be extracted out for reusability but fine as-is since this will be removed with v1 code

karthikbtf (approved):
Looks good, some comments but nothing required.

src/main/kotlin/com/terraformation/backend/seedbank/model/Withdrawal.kt
sgrimm: Good call on checking for the prefix too. I thought about making this a helper function but it's only repeated once and is short enough that I thought there wouldn't be much of a maintainability win (less repetition, but you have to bounce around the code an additional time to follow the logic).

karthikbtf: Sounds fair to me 👍

src/main/kotlin/com/terraformation/backend/seedbank/model/ViabilityTest.kt
```diff
@@ -124,7 +124,7 @@ data class ViabilityTestModel(
         when {
           staffResponsible.isNullOrBlank() -> notes
           notes.isNullOrBlank() -> "Staff responsible: $staffResponsible"
-          staffResponsible in notes -> notes
+          "Staff responsible: $staffResponsible" in notes -> notes
```
karthikbtf, line 127: I was thinking more along
```
"Staff responsible: " in notes -> notes
```
That way user can edit the actual staff name and not worry about us appending the staff responsible line again.
However, my suggestion could be less foolproof - maybe not ideal.

karthikbtf (approved):
Looks good to me.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/646?utm_source=stack-comment).


Job Reference: 0jaE9F1qlrJa2GBmJQwn

PR #655, opened 2022-09-26
Move seedbank tables to separate DB schema

sgrimm:
Move seedbank tables to separate DB schema

To get ahead of name collisions when we add nursery support to the code, update
the jOOQ configuration to support generating code from multiple database schemas,
and move the seed-related tables to a new `seedbank` schema. Other tables remain
in the default schema (`public`).

This adds a new level of package hierarchy to the generated jOOQ classes.
Seed-related tables are now under a `seedbank` package. The tables in the
`public` schema are under `default_schema` because the word `public` is a
keyword that isn't a valid package name. The code generation config now breaks
the lists of tables and enum classes out by schema name.

For the most part, the impact on application code and tests is limited to
updating all the `import` statements that import jOOQ classes and fields. A small
amount of test code that uses JDBC to query table metadata needed to be changed
to be explicit about which schema to use.


sgrimm:
* main
  * **PR #655** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/655" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/655?utm_source=stack-comment).

sgrimm:
I broke this out into two commits. The first one has all the code edits and will be more interesting to review. The second one has all the updates to `import` statements, and was done using IntelliJ refactoring. Can't break them into separate PRs because neither one will build successfully without the other.

karthikbtf:
> I broke this out into two commits. The first one has all the code edits and will be more interesting to review. The second one has all the updates to `import` statements, and was done using IntelliJ refactoring. Can't break them into separate PRs because neither one will build successfully without the other.

Thanks. I reviewed the first commit and a cursory glance over the second.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -14,115 +14,133 @@ import org.jooq.meta.jaxb.EmbeddableDefinitionType
...
+                EnumTable(
+                    "storage_conditions",
+                    listOf(
+                        "accessions\\.target_storage_condition",
+                        "storage_locations\\.condition_id")),
+                EnumTable("viability_test_seed_types", "viability_tests\\.seed_type_id"),
+                EnumTable("viability_test_substrates", "viability_tests\\.substrate_id"),
+                EnumTable("viability_test_treatments", "viability_tests\\.treatment_id"),
+                EnumTable("viability_test_types", listOf("viability_tests\\.test_type")),
+                EnumTable("withdrawal_purposes", "withdrawals\\.purpose_id"),
```
karthikbtf, line 86: guessing nursery as a withdrawal purpose would be part of this enum - even though it has nursery semantics.

karthikbtf (approved):
Looks good to me.

buildSrc/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
sgrimm: It already is, in fact! Which will be a compatibility thing we'll have to figure out.

PR #660, opened 2022-09-28
SW-1797 Allow moving out of Used Up if remaining quantity is set

sgrimm:
The user is supposed to be able to change an accession from Used Up to some other
state if they set the remaining quantity at the same time, but the server wasn't
accepting the state change.


sgrimm:
* main
  * **PR #660** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/660" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/660?utm_source=stack-comment).

karthikbtf (approved):

PR #666, opened 2022-09-30
Use fully qualified class names for Spring beans

sgrimm:
By default, when Spring registers a class as a bean due to the presence of an
annotation, it uses the unqualified class name as the bean name. We don't refer
to beans by name in our code, just by class, but internally Spring requires
each bean to have a unique name.

The default naming strategy is going to cause collisions once we have multiple
database tables with the same name in different schemas, because each table gets
a DAO class that is annotated as a Spring bean, e.g., we'll have two separate
`WithdrawalsDao` classes in different packages.

Tell Spring to base its default bean names on fully-qualified class names instead.


sgrimm:
* main
  * **PR #666** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/666" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/666?utm_source=stack-comment).

karthikbtf (approved):

PR #667, opened 2022-09-30
Add smoke test for JobRunr API

sgrimm:
Twice now, we've run into Kotlin version upgrades that broke JobRunr's "extract
the parameters from the body of a lambda function" logic, but since our test suite
was always stubbing out JobRunr, we only discovered the problem at runtime.

Add a simple sanity-check test to try to detect future incidents before they get
merged.


sgrimm:
* main
  * **PR #667** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/667" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/667?utm_source=stack-comment).

sgrimm:
Tested this by setting the Kotlin version to 1.7.20 locally; it fails with the same exception we were seeing in staging earlier today.

karthikbtf (approved):
👍 
a `.DS_Store` file got picked up

sgrimm:
Ah, right, though it's deleted here, not added. Could split that out into a separate PR but doesn't seem worth the bother.

PR #677, opened 2022-10-05
Use a Consistent Postgis Version

charlie-tf:
The postgis version specified for tests and github is inconsistent.
It should be `postgis:13-3.0` to match what is used for staging/prod
deployment in RDS. Also add `.DS_Store` files to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ out
 .idea
 openapi.yaml
 .env
+*.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,4 @@ out
 .idea
 openapi.yaml
 .env
-*.DS_Store
+.DS_Store
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -48,7 +48,7 @@ jobs:
 
     - name: Download Docker images
       run: |
-        docker pull postgis/postgis:12-3.1
+        docker pull postgis/postgis:13-3.0
         docker pull schemaspy/schemaspy:6.1.0
 
     - name: Compile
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -9,3 +9,4 @@ out
 .idea
 openapi.yaml
 .env
+*.DS_Store
```
sgrimm: Why the `*` prefix?

sgrimm (commented):
Not sure why the CI checks aren't running. Maybe related to the GitHub issues. This seems good (aside from one question) but I'll wait until we have a successful CI run before approving it.

.gitignore
charlie-tf: to catch when it shows up in subdirectories. I had looked at the repo in Finder, which generated a breadcrumb trail of `.DS_Store` files following my navigation.

charlie-tf:
> Not sure why the CI checks aren't running. Maybe related to the GitHub issues.

Yeah I wondered about that. I did run tests locally, but certainly want to wait for CI to come back online. Thanks!

.gitignore
sgrimm: Ah, okay, We shouldn't need that -- unqualified filenames should be matched in any directory at or under the level where the `.gitignore` file lives.

`*.DS_Store` would match `Foo.DS_Store` as a filename.

sgrimm: Err, to be more helpful: this can be `.DS_Store` rather than `*.DS_Store`.

charlie-tf: Oh, I did not know that. I'll change it. Thanks!

sgrimm (approved):
CI checks passed.

PR #678, opened 2022-10-05
SW-1911 Add nursery inventories to search API

sgrimm:
Add two new views to the nursery database schema that summarize nursery
inventories on both a per-organization and per-facility basis, and add them as
tables in the search API. This allows clients to search and filter the summary
data as needed for the Inventory List page in the nursery UI.


sgrimm:
* main
  * **PR #678** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/678" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/678?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/nursery/NurserySearchTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    )
+    insertBatch(
+        facilityId = facilityId,
+        germinatingQuantity = 512,
+        notReadyQuantity = 1024,
+        readyQuantity = 2048,
+        speciesId = speciesId2,
+    )
+    insertBatch(
+        organizationId = organizationId2,
```
charlie-tf, line 103: the intent of this batch is to make sure we aren't counting across organizationIds, right?

src/main/resources/db/migration/V140__NurseryInventories.sql
```diff
@@ -0,0 +1,22 @@
+CREATE OR REPLACE VIEW nursery.inventories AS
+    SELECT organization_id,
+           species_id,
+           SUM(ready_quantity)                      AS ready_quantity,
+           SUM(not_ready_quantity)                  AS not_ready_quantity,
+           SUM(germinating_quantity)                AS germinating_quantity,
+           SUM(ready_quantity + not_ready_quantity) AS total_quantity
```
charlie-tf, line 7: why doesn't total include germinating?

charlie-tf (commented):
looks good! just a couple questions to clarify my understanding

src/test/kotlin/com/terraformation/backend/nursery/NurserySearchTest.kt
sgrimm: Correct. Especially since these "tables" are actually the first views we've added to the search implementation, want to make sure there isn't a missing join condition anywhere.

src/main/resources/db/migration/V140__NurseryInventories.sql
sgrimm: > why doesn't total include germinating?

My understanding is that nursery managers think about not-yet-germinated seeds as a separate category of thing from plants that have sprouted but just aren't ready to plant yet, but I couldn't tell you exactly why that is. The PRD is pretty clear on this, though.

PR #679, opened 2022-10-05
SW-1749 Split up AccessionModelTest

charlie-tf:
- Divide the tests into several categories and place in new files
- Place shared code into an abstract superclass


src/test/kotlin/com/terraformation/backend/seedbank/model/accession/Helpers.kt
```diff
@@ -0,0 +1,172 @@
...
+val tomorrowInstant = tomorrow.atStartOfDay(ZoneOffset.UTC).toInstant()
+val tomorrowClock = Clock.fixed(tomorrowInstant, ZoneOffset.UTC)
+val yesterday = today.minusDays(1)
+val yesterdayInstant = yesterday.atStartOfDay(ZoneOffset.UTC).toInstant()
+val yesterdayClock = Clock.fixed(yesterdayInstant, ZoneOffset.UTC)
+
+var viabilityTestResultId = ViabilityTestResultId(1)
+var viabilityTestId = ViabilityTestId(1)
+var withdrawalId = WithdrawalId(1)
+var defaultState = AccessionState.Processing
```
sgrimm: Moving these to top-level scope is going to be problematic, I think. First, they'll become singletons that are shared across tests. So if, for example, a test sets a new value for `viabilityTestId`, the next test will also see that new value because the initializer here only runs once rather than once per test method. That could cause tests to fail depending on what order they're run. Related to that, these objects will all stay around in memory for the entire rest of the run of the test suite once they're referenced. They're small enough that that won't matter much in practice but it still means less test isolation.

The other, probably bigger, concern I have is that it's going to be too easy to use these by mistake. Right now, if I type `clock` in a test method body and I've forgotten to declare a `clock` in my test class, it's an error. With this change, IntelliJ might auto-import it from here, with me none the wiser. I chose `clock` as an example because in a lot of tests, we use a MockK stub instead of `Clock.fixed()` to allow tests to manipulate the time, so the static import would cause runtime failures.

What do you think about putting these variables/functions in an abstract superclass instead?

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelViabilityTest.kt
```diff
@@ -0,0 +1,420 @@
...
+import com.terraformation.backend.db.seedbank.ViabilityTestType
+import com.terraformation.backend.db.seedbank.WithdrawalPurpose
+import com.terraformation.backend.seedbank.grams
+import com.terraformation.backend.seedbank.model.ViabilityTestModel
+import com.terraformation.backend.seedbank.seeds
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class AccessionModelViabilityTest {
```
sgrimm: This is a tough one to name! Either `Test` does double duty as the suffix of the unit test class name and part of the name of the thing being tested, or you have `TestTest` which looks weird. (The choice you made here is fine.)

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
```diff
@@ -1982,149 +1140,4 @@ internal class AccessionModelTest {
       assertEquals(grams(5), accession.estimatedWeight)
     }
   }
-
-  @Nested
-  inner class V1ToV2Conversion {
```
sgrimm, line 1987: I'm not seeing where this moved. I see the v2-to-v1 tests but not these.

sgrimm (changes requested):
Restructuring mostly looks good, but I have a couple concerns.

sgrimm:
The PR title could be a bit more specific. "SW-1749 Split up AccessionModelTest" would make it easier to identify this change in a git log. (It's fine to have multiple PRs with the same Jira task ID.)

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/Helpers.kt
charlie-tf: Oh, very good point. I like the idea of an abstract superclass! I'll make that change.

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelViabilityTest.kt
charlie-tf: I almost did the `TestTest` thing, or `AccessionModelViabilityTestCalculationsTest` but that was getting awfully long 😂

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
charlie-tf: good catch! I thought that I'd moved them, but looks like I missed these 😞

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelTests.kt
```diff
@@ -0,0 +1,174 @@
...
+import com.terraformation.backend.seedbank.model.ViabilityTestResultModel
+import com.terraformation.backend.seedbank.model.WithdrawalModel
+import com.terraformation.backend.seedbank.seeds
+import java.time.Clock
+import java.time.Instant
+import java.time.LocalDate
+import java.time.ZoneOffset
+
+abstract class AccessionModelTests {
+  val today = january(2)
```
sgrimm: Very minor nitpick, but easy to change: Everything in here could have `protected` visibility.

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelTests.kt
```diff
@@ -0,0 +1,174 @@
...
+import com.terraformation.backend.seedbank.model.ViabilityTestModel
+import com.terraformation.backend.seedbank.model.ViabilityTestResultModel
+import com.terraformation.backend.seedbank.model.WithdrawalModel
+import com.terraformation.backend.seedbank.seeds
+import java.time.Clock
+import java.time.Instant
+import java.time.LocalDate
+import java.time.ZoneOffset
+
+abstract class AccessionModelTests {
```
sgrimm: Might be worth having a discussion about how we want to name this kind of thing, but the existing abstract test superclass in the code base uses a singular name (`DatabaseTest`) rather than a plural one. To stay consistent until we make a decision to use some other naming convention, can we call this `AccessionModelTest` instead? That name is free now!

Reusing that class name shouldn't cause confusion in git history since the original `AccessionModelTest` was in a different directory.

sgrimm (changes requested):
Getting close. Just a couple super minor things.

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelTests.kt
charlie-tf: can do! I also changed the visibilities of the new classes to `internal` which I'd neglected previously

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelTests.kt
charlie-tf: I'd thought plural since it's an umbrella for those other test classes, but of course each of those is itself an "umbrella" for multiple tests. Singular sounds good to me!

sgrimm (approved):
Super! Thanks for iterating on this.

PR #680, opened 2022-10-05
SW-1945 Add endpoint to get nursery species summary

sgrimm:
`GET /api/v1/nursery/species/{id}/summary` will return a summary of the inventory
and withdrawals for a particular species. This is the information the UI code will
need for the summary bar at the top of the nursery species details page.


sgrimm:
* main
  * **PR #680** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/680" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/680?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/nursery/db/BatchStoreTest.kt
```diff
@@ -164,6 +167,83 @@ internal class BatchStoreTest : DatabaseTest(), RunsAsUser {
...
+              notReadyQuantity = 1,
+              readyQuantity = 1,
+              speciesId = speciesId,
+          )
+      val withdrawalId = insertWithdrawal(purpose = WithdrawalPurpose.Dead)
+      insertBatchWithdrawal(
+          batchId = batchId,
+          germinatingQuantityWithdrawn = 20,
+          notReadyQuantityWithdrawn = 2,
+          readyQuantityWithdrawn = 3,
```
charlie-tf, line 203: I'm a bit confused on the logic here. It looks like the batch has 10/1/1 quantity to start with, but 20/2/3 is withdrawn. How does that work exactly? And below, wondering where the `7 total past + current seedlings` quantity comes from. Maybe if I read the PRD I'd understand; would you be able to point me to that?

sgrimm: The batch-level quantities are the current inventory. When we withdraw, we'll subtract the withdrawal amounts from them, though I haven't written that code yet. So the expected end state of a batch over time will be to have 0 for all three quantities (because all the plants have been withdrawn).

sgrimm: Possibly of use as you look at these nursery PRs: https://terraformation.atlassian.net/wiki/spaces/SW/pages/648544257/Nursery+v1+data+model

charlie-tf: Thanks

karthikbtf (approved):

PR #684, opened 2022-10-06
SW-1749 Split up AccessionStoreTest

charlie-tf:
- Create an abstract base class
  - inherits DatabaseTest
  - implements RunAsUser
  - supplies helper vals/funs
- Categorize existing tests into new, smaller test classes


charlie-tf:
I counted the tests before and after to make sure I didn't accidentally delete any again.
Before:
© Tests passed 926, ignored 29 of 955 tests - 35 sec 625 ms
After:
Pasted Graphic 1

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreBagTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  @Test
+  fun `bags are inserted and deleted as needed`() {
+    val initial =
+        store.create(AccessionModel(bagNumbers = setOf("bag 1", "bag 2"), facilityId = facilityId))
+    val initialBags = bagsDao.fetchByAccessionId(AccessionId(1))
+
+    // Insertion order is not defined by the API, so don't assume bag ID 1 is "bag 1".
+
+    Assertions.assertEquals(
```
sgrimm: This was probably IntelliJ's fault, but can we add static imports for all these assertion methods so we don't have to have `Assertions.` prefixes in all the test methods?

sgrimm (changes requested):
Will look at this in more detail after the team meeting but I saw one initial thing that'd be nice to change.

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreBagTest.kt
charlie-tf: oh dangit intelliJ, yup, i can definitely fix those

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStorePermissionTest.kt
```diff
@@ -0,0 +1,117 @@
...
+  @Test
+  fun `delete throws exception if user does not have permission`() {
+    every { user.canDeleteAccession(any()) } returns false
+    val initial = store.create(AccessionModel(facilityId = facilityId))
+
+    assertThrows<AccessDeniedException> { store.delete(initial.id!!) }
+  }
+
+  @Test
+  fun `update ignores received and collected date edits for accessions from web`() {
```
sgrimm: This one isn't permission-related.

sgrimm (changes requested):
This is nice! I like the choices you made in breaking this up.

There's one split that I'm not sure of my opinion on: pulling all the permission test cases out into a separate class. On the one hand, permission checking is obviously a well-defined class of functionality and doing it this way makes total sense. On the other hand, permission checking is part of the contract of each individual method on the class and the permission tests should live with the other tests for the various methods.

I'm totally fine merging this with the split you chose, though; it's not clear to me that either approach is more correct than the other.

Requesting changes here because one test method is in the wrong place.

charlie-tf:
>There's one split that I'm not sure of my opinion on: pulling all the permission test cases out into a separate class. On the one hand, permission checking is obviously a well-defined class of functionality and doing it this way makes total sense. On the other hand, permission checking is part of the contract of each individual method on the class and the permission tests should live with the other tests for the various methods.

That's a very good point. I'm not quite sure what's more correct either. If you are okay with merging this, though, perhaps we see if having the permission checks live separately from their related methods is too onerous and can reunite them later if so?

sgrimm:
Yes, let's try it this way and see how it works over time.

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStorePermissionTest.kt
charlie-tf: Not sure where is an obvious place for it. Maybe `AccessionDatabaseTest` since it is testing the database behavior upon receiving a web-originated accession? What do you think?

sgrimm: Sure, that seems reasonable.

sgrimm (approved):

PR #685, opened 2022-10-06
SW-1966 Add endpoint to update batch quantities

sgrimm:
`PUT /api/v1/nursery/batches/{id}/quantities` will update the seedling quantities
of an existing batch.

This endpoint introduces a new pattern to the API: optimistic locking using
versions. This will help us manage data consistency when multiple users are
working with the same batch. The client must pass in a version number and the
version number must match the current version of the batch. If the versions
don't match, it means some other client updated the batch already, and the
server returns HTTP 412.

Example flow to demonstrate how it works:

1. Batch 123 has a version number of 9
2. Client A: `GET /api/v1/nursery/batches/123` (returns version 9)
3. Client B: `GET /api/v1/nursery/batches/123` (returns version 9)
4. Client A: `PUT /api/v1/nursery/batches/123/quantities` with version=9
   (succeeds and returns updated batch with version 10)
5. Client B: `PUT /api/v1/nursery/batches/123/quantities` with version=9
   (fails with HTTP 412)
6. Client B: `GET /api/v1/nursery/batches/123` (returns version 10)
7. Client B: `PUT /api/v1/nursery/batches/123/quantities` with version=10
   (succeeds and returns updated batch with version 11)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
@@ -1,11 +1,16 @@
 package com.terraformation.backend.nursery.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonSetter
+import com.fasterxml.jackson.annotation.Nulls
+import com.terraformation.backend.api.ApiResponse404
+import com.terraformation.backend.api.ApiResponse412
 import com.terraformation.backend.api.NurseryEndpoint
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.default_schema.FacilityId
 import com.terraformation.backend.db.default_schema.SpeciesId
 import com.terraformation.backend.db.nursery.BatchId
+import com.terraformation.backend.db.nursery.BatchQuantityHistoryType
 import com.terraformation.backend.db.nursery.tables.pojos.BatchesRow
 import com.terraformation.backend.db.seedbank.AccessionId
 import com.terraformation.backend.nursery.db.BatchStore
@@ -13,9 +18,11 @@ import io.swagger.v3.oas.annotations.media.Schema
 import java.time.Instant
 import java.time.LocalDate
 import java.time.temporal.ChronoUnit
+import javax.validation.constraints.Min
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
@@ -26,6 +33,7 @@ import org.springframework.web.bind.annotation.RestController
 class BatchesController(
     private val batchStore: BatchStore,
 ) {
+  @ApiResponse404
   @GetMapping("/{id}")
   fun getBatch(@PathVariable("id") id: BatchId): BatchResponsePayload {
     val row = batchStore.fetchOneById(id)
@@ -36,6 +44,24 @@ class BatchesController(
   fun createBatch(@RequestBody payload: CreateBatchRequestPayload): BatchResponsePayload {
     val insertedRow = batchStore.create(payload.toRow())
     return BatchResponsePayload(BatchPayload(insertedRow))
+  }
+
+  @ApiResponse404
+  @ApiResponse412
+  @PutMapping("/{id}/quantities")
+  fun updateBatchQuantities(
+      @PathVariable("id") id: BatchId,
+      @RequestBody payload: UpdateBatchQuantitiesRequestPayload
+  ): BatchResponsePayload {
+    batchStore.updateQuantities(
+        batchId = id,
+        version = payload.version,
+        germinating = payload.germinatingQuantity,
+        notReady = payload.notReadyQuantity,
+        ready = payload.readyQuantity,
+        historyType = BatchQuantityHistoryType.Observed)
+
+    return getBatch(id)
   }
 }
 
@@ -86,12 +112,12 @@ data class BatchPayload(
 data class CreateBatchRequestPayload(
     val addedDate: LocalDate,
     val facilityId: FacilityId,
-    val germinatingQuantity: Int,
     val notes: String? = null,
-    val notReadyQuantity: Int,
     val readyByDate: LocalDate? = null,
-    val readyQuantity: Int,
     val speciesId: SpeciesId,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val germinatingQuantity: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val notReadyQuantity: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val readyQuantity: Int,
 ) {
   fun toRow() =
       BatchesRow(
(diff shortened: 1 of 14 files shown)


sgrimm:
* main
  * **PR #685** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/685" target="_blank">Graphite</a>  👈
    * **PR #686** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/686" target="_blank">Graphite</a> 
      * **PR #687** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/687" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/685?utm_source=stack-comment).

sgrimm:
I realize this PR (and earlier related ones) are busy building up a big test class while @charlie-tf is busy breaking up big test classes. Old habits die hard. I'll do a followup to break up `BatchStoreTest`.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -148,16 +161,104 @@ class BatchStore(
...
+   * @throws BatchStaleException The batch's version number in the database wasn't the same as
+   * [version].
+   */
+  private fun updateVersionedBatch(
+      batchId: BatchId,
+      version: Int,
+      func: Batches.(UpdateSetFirstStep<BatchesRecord>) -> UpdateSetMoreStep<BatchesRecord>
+  ) {
+    val rowsUpdated =
+        dslContext
```
karthikbtf, line 217: do we need to lock the row or will concurrency not be an issue.

karthikbtf (approved):
looks good, one silly question

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: The database has implicit locking that takes care of this. Basically if one transaction updates a row and another transaction is doing a write whose success or failure depends on one of the values updated by the first transaction, the second write will block until the first transaction either commits or rolls back.

You can play with this in `psql` by connecting to the same database in two windows and doing something like

```
(window 1)
create table foo (col integer);
insert into foo values (1);
begin;
update foo set col = 2 where col = 1;

(window 2)
update foo set col = 2 where col = 1;
-- this blocks even though you never asked for a lock anywhere

(window 1)
commit;

(window 2 unblocks and shows 0 rows updated)
```

karthikbtf: Thanks so much, figured something like this is going on.
From a write/read concurrency perspective - I guess it's ok for the read to not wait for the write (with the caveat that the read could potentially be stale .. but avoids slowing down all reads).
I'm overthinking all this for sure.

sgrimm: https://www.postgresql.org/docs/13/transaction-iso.html has a more detailed discussion of this stuff if you want to _really_ overthink it.

And it's fine if reads return stale data. No different from if the read came in a few milliseconds earlier and didn't happen to overlap with the write.

sgrimm: Oh, and the reason I did it this way rather than explicit locking is because of the "optimistic" part of the equation: 99.9% of the time, there won't be another user editing the same batch and this will just succeed without anything getting blocked. Grabbing an explicit lock means paying the performance penalty on every write (at minimum, an extra database round-trip).

karthikbtf: Makes total sense, thanks for the clarification and the document links @sgrimm

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -148,16 +161,104 @@ class BatchStore(
...
+            .update(BATCHES)
+            .let { BATCHES.func(it) }
+            .set(BATCHES.MODIFIED_BY, currentUser().userId)
+            .set(BATCHES.MODIFIED_TIME, clock.instant())
+            .set(BATCHES.VERSION, version + 1)
+            .where(BATCHES.ID.eq(batchId))
+            .and(BATCHES.VERSION.eq(version))
```
charlie-tf, line 224: So the `BATCHES.VERSION.eq(version)` constraint makes sure we only select and update the rows where the version matches, resulting in `rowsUpdated == 0` if the version is incorrect. Am I understanding correctly?

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -148,16 +161,104 @@ class BatchStore(
...
+   * @param func Function to add `set` statements to an `UPDATE` query. This is called with the
+   * [BATCHES] table as its receiver so that lambda functions can refer to column names without
+   * having to qualify them (that is, `set(FOO, value)` instead of `set(BATCHES.FOO, value)`).
+   * @throws BatchStaleException The batch's version number in the database wasn't the same as
+   * [version].
+   */
+  private fun updateVersionedBatch(
+      batchId: BatchId,
+      version: Int,
+      func: Batches.(UpdateSetFirstStep<BatchesRecord>) -> UpdateSetMoreStep<BatchesRecord>
```
charlie-tf, line 214: So this allows you to inject updates that are namespaced to the `BATCHES` table, right?

src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
```diff
@@ -37,6 +45,24 @@ class BatchesController(
     val insertedRow = batchStore.create(payload.toRow())
     return BatchResponsePayload(BatchPayload(insertedRow))
   }
+
+  @ApiResponse404
+  @ApiResponse412
```
charlie-tf, line 50: Am I correct in understanding that these annotation are meant to catch exceptions and turn them into error responses? One thing that's not connecting for me is how they know which exception to catch for each of the 4xx responses.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: Yes, that's correct. The update is atomic with the `WHERE`, so if two clients are trying to update the same batch at the same time, only one will succeed.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: Kind of. This is arguably me trying to be too clever, but the idea is that if this were just
```
func: (UpdateSetFirstStep<BatchesRecord>) -> UpdateSetMoreStep<BatchesRecord>
```
then the lambda functions would end up needing to do
```
it.set(BATCHES.READY_QUANTITY, readyQuantity)
```
whereas with `Batches` as the receiver type, they can do
```
it.set(READY_QUANTITY, readyQuantity)
```
which is both more concise and also hopefully subtly enforces the notion that the lambda function should only be setting columns on the `batches` table rather than referencing other random tables.

src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
sgrimm: Not quite; these annotations are for the benefit of the OpenAPI schema generator. With these annotations, the generated API schema will list HTTP 404 and 412 as possible response codes for this endpoint. The annotations don't have any effect on the actual exception handling.

However, you're asking the right question. There _is_ a place where exceptions get mapped to HTTP responses: https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/api/ControllerExceptionHandler.kt

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/685?utm_source=stack-comment).


Job Reference: UXFSxouIzptnllifG8jt

src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
charlie-tf: Awesome, thanks for the explanation and link!

PR #691, opened 2022-10-07
SW-1163 Remove Update and Delete Permission from Contributor

charlie-tf:
- Modify the canUpdateAccession method to deny for Contributor
- Note that canDeleteAccession is the same as canUpdateAccession
- Modify PermissionTest to reflect new permissions

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -365,9 +365,15 @@ data class IndividualUser(
   }
 
   override fun canUpdateAccession(accessionId: AccessionId): Boolean {
-    // All users in a project can write all accessions in the project's facilities, so this
-    // is the same as the read permission check.
-    return canReadAccession(accessionId)
+    // Updating accession is allowed only for Role.MANAGER and higher.
+    // The user must also have read permissions for the accession.
+    val facilityId = parentStore.getFacilityId(accessionId)
+    return when (facilityRoles[facilityId]) {
+      Role.OWNER,
+      Role.ADMIN,
+      Role.MANAGER -> true
+      else -> false
+    } && canReadAccession(accessionId)
   }
 
   override fun canUpdateAppVersions(): Boolean {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -520,7 +520,7 @@ internal class PermissionTest : DatabaseTest() {
   }
 
   @Test
-  fun `contributors have access to data associated with their organizations`() {
+  fun `contributors have access to data associated with their organizations, but cannot update or delete accessions`() {
```
sgrimm: This feels too specific to me -- when we make a similar change for nursery support, we'd kind of need to update this to also describe the nursery-related limitations.

How about something like, `contributors have full read and limited write access to data associated with their organizations`?

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -365,9 +365,15 @@ data class IndividualUser(
...
-    return canReadAccession(accessionId)
+    // Updating accession is allowed only for Role.MANAGER and higher.
+    // The user must also have read permissions for the accession.
+    val facilityId = parentStore.getFacilityId(accessionId)
+    return when (facilityRoles[facilityId]) {
+      Role.OWNER,
+      Role.ADMIN,
+      Role.MANAGER -> true
+      else -> false
+    } && canReadAccession(accessionId)
```
sgrimm: We don't need the read permission check here; it'll cause an extra `parentStore` data fetch on the happy path and will never be false when the `when` expression is true. The test suite should ensure the return value is always what we want.

sgrimm (changes requested):
This is the right idea.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
charlie-tf: Yeah, that sounds like good verbiage. I'll go with it!

sgrimm (approved):
👍

PR #696, opened 2022-10-10
SW-1970 Add endpoint to delete seedling batches

sgrimm:
`DELETE /api/v1/nursery/batches/{id}` will delete a batch from the system.

The ability to delete a batch means we have to relax some of our constraints.
For example, it becomes possible for there to be a nursery transfer withdrawal
where the destination batch no longer exists; we still want the withdrawal to
exist on the originating side.

PR #698, opened 2022-10-10
SW-1967 Add endpoint to withdraw from seedling batches

sgrimm:
`POST /api/v1/nursery/withdrawals` will withdraw from one or more batches at a
facility. If the withdrawal date is on or after the date (in UTC time zone) the
user most recently edited the batch's remaining quantities, the withdrawal
quantities will be subtracted from the batch.

Withdrawals that change batches' quantities will cause quantity history records
to be created. There is currently no API to query the history of a batch.

This initial version does not support nursery transfer withdrawals; those will
come in a followup revision.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
@@ -14,6 +14,7 @@ import com.terraformation.backend.nursery.model.BatchWithdrawalModel
 import com.terraformation.backend.nursery.model.ExistingWithdrawalModel
 import com.terraformation.backend.nursery.model.NewWithdrawalModel
 import io.swagger.v3.oas.annotations.media.ArraySchema
+import io.swagger.v3.oas.annotations.media.Schema
 import java.time.LocalDate
 import javax.validation.constraints.Min
 import org.springframework.web.bind.annotation.PostMapping
@@ -43,7 +44,7 @@ class WithdrawalsController(
 
 data class BatchWithdrawalPayload(
     val batchId: BatchId,
-    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val germinatingQuantityWithdrawn: Int,
+    @Schema(defaultValue = "0") @Min(0) val germinatingQuantityWithdrawn: Int? = null,
     @JsonSetter(nulls = Nulls.FAIL) @Min(0) val notReadyQuantityWithdrawn: Int,
     @JsonSetter(nulls = Nulls.FAIL) @Min(0) val readyQuantityWithdrawn: Int,
 ) {
@@ -59,7 +60,7 @@ data class BatchWithdrawalPayload(
   fun toModel() =
       BatchWithdrawalModel(
           batchId = batchId,
-          germinatingQuantityWithdrawn = germinatingQuantityWithdrawn,
+          germinatingQuantityWithdrawn = germinatingQuantityWithdrawn ?: 0,
           notReadyQuantityWithdrawn = notReadyQuantityWithdrawn,
           readyQuantityWithdrawn = readyQuantityWithdrawn,
       )
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
@@ -1,0 +1,114 @@
+package com.terraformation.backend.nursery.api
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonSetter
+import com.fasterxml.jackson.annotation.Nulls
+import com.terraformation.backend.api.NurseryEndpoint
+import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.db.default_schema.FacilityId
+import com.terraformation.backend.db.nursery.BatchId
+import com.terraformation.backend.db.nursery.WithdrawalId
+import com.terraformation.backend.db.nursery.WithdrawalPurpose
+import com.terraformation.backend.nursery.db.BatchStore
+import com.terraformation.backend.nursery.model.BatchWithdrawalModel
+import com.terraformation.backend.nursery.model.ExistingWithdrawalModel
+import com.terraformation.backend.nursery.model.NewWithdrawalModel
+import io.swagger.v3.oas.annotations.media.ArraySchema
+import java.time.LocalDate
+import javax.validation.constraints.Min
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@NurseryEndpoint
+@RequestMapping("/api/v1/nursery/withdrawals")
+@RestController
+class WithdrawalsController(
+    val batchStore: BatchStore,
+) {
+  @PostMapping
+  fun createBatchWithdrawal(
+      @RequestBody payload: CreateNurseryWithdrawalRequestPayload
+  ): CreateNurseryWithdrawalResponsePayload {
+    val model = batchStore.withdraw(payload.toModel())
+    val batchModels = model.batchWithdrawals.map { batchStore.fetchOneById(it.batchId) }
+
+    return CreateNurseryWithdrawalResponsePayload(
+        batchModels.map { BatchPayload(it) },
+        NurseryWithdrawalPayload(model),
+    )
+  }
+}
+
+data class BatchWithdrawalPayload(
+    val batchId: BatchId,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val germinatingQuantityWithdrawn: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val notReadyQuantityWithdrawn: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val readyQuantityWithdrawn: Int,
+) {
+  constructor(
+      model: BatchWithdrawalModel
+  ) : this(
+      batchId = model.batchId,
+      germinatingQuantityWithdrawn = model.germinatingQuantityWithdrawn,
+      notReadyQuantityWithdrawn = model.notReadyQuantityWithdrawn,
+      readyQuantityWithdrawn = model.readyQuantityWithdrawn,
+  )
+
+  fun toModel() =
+      BatchWithdrawalModel(
+          batchId = batchId,
+          germinatingQuantityWithdrawn = germinatingQuantityWithdrawn,
+          notReadyQuantityWithdrawn = notReadyQuantityWithdrawn,
+          readyQuantityWithdrawn = readyQuantityWithdrawn,
+      )
+}
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class NurseryWithdrawalPayload(
+    val batchWithdrawals: List<BatchWithdrawalPayload>,
+    val destination: String? = null,
+    val facilityId: FacilityId,
+    val id: WithdrawalId,
+    val purpose: WithdrawalPurpose,
+    val reason: String?,
+    val withdrawnDate: LocalDate,
+) {
+  constructor(
+      model: ExistingWithdrawalModel
+  ) : this(
+      model.batchWithdrawals.map { BatchWithdrawalPayload(it) },
+      model.destination,
+      model.facilityId,
+      model.id,
+      model.purpose,
+      model.reason,
+      model.withdrawnDate,
+  )
+}
+
+data class CreateNurseryWithdrawalRequestPayload(
+    @ArraySchema(minItems = 1) val batchWithdrawals: List<BatchWithdrawalPayload>,
+    val destination: String? = null,
+    val facilityId: FacilityId,
+    val purpose: WithdrawalPurpose,
+    val reason: String? = null,
+    val withdrawnDate: LocalDate,
+) {
+  fun toModel() =
+      NewWithdrawalModel(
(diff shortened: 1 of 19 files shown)


sgrimm:
* main
    * **PR #698** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/698" target="_blank">Graphite</a>  👈
            * **PR #704** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/704" target="_blank">Graphite</a> 
          * **PR #702** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/702" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/698?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/model/WithdrawalModel.kt
```diff
@@ -0,0 +1,81 @@
...
+      throw IllegalArgumentException("Withdrawals must come from at least one batch")
+    }
+  }
+}
+
+/** A withdrawal that hasn't been inserted into the database yet. */
+typealias NewWithdrawalModel = WithdrawalModel<Nothing?>
+
+/** A withdrawal that was loaded from the database. */
+typealias ExistingWithdrawalModel = WithdrawalModel<WithdrawalId>
```
sgrimm, line 60: This is an experiment I want to try out, a way to deal with the fact that you want the `id` field on this kind of model object to be non-nullable if the object was pulled from the database, but you want to reuse the same model class to pass in new data to be inserted (in which case there's no ID).

An alternative approach (other than doing `val id: WithdrawalId?` as we do in, e.g., `AccessionModel`) is to declare `NewWithdrawalModel` and `ExistingWithdrawalModel` as entirely separate classes, but then that becomes a maintenance headache as the data model evolves over time; the two could gradually fall out of sync.

sgrimm: In case you aren't familiar with it, Kotlin's `Nothing` class is kind of an odd duck: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-nothing.html

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
```diff
@@ -0,0 +1,114 @@
...
+          id = null,
+          purpose = purpose,
+          reason = reason,
+          withdrawnDate = withdrawnDate,
+      )
+}
+
+data class CreateNurseryWithdrawalResponsePayload(
+    val batches: List<BatchPayload>,
+    val withdrawal: NurseryWithdrawalPayload
```
karthikbtf, line 114: This is good, allows client to refresh it's comprehensive view.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
```diff
@@ -0,0 +1,114 @@
...
+    val destination: String? = null,
+    val facilityId: FacilityId,
+    val purpose: WithdrawalPurpose,
+    val reason: String? = null,
+    val withdrawnDate: LocalDate,
+) {
+  fun toModel() =
+      NewWithdrawalModel(
+          batchWithdrawals = batchWithdrawals.map { it.toModel() },
+          destination = destination,
```
karthikbtf, line 103: Is this the 'Destination/Notes' equivalent (in design)?

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
```diff
@@ -0,0 +1,114 @@
...
+          facilityId = facilityId,
+          id = null,
+          purpose = purpose,
+          reason = reason,
```
karthikbtf, line 107: Not seeing a use-case in the designs for 'reason'. Unless this is the notes.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -264,6 +285,142 @@ class BatchStore(
...
+                        // here so the query object is of the correct jOOQ type)
+                        it.set(emptyMap<Any, Any>())
+                      }
+                    }
+
+                    batchWithdrawalsDao.insert(batchWithdrawalsRow)
+
+                    succeeded = true
+                  } catch (e: BatchStaleException) {
+                    if (retriesRemaining > 0) {
```
karthikbtf, line 408: just curious on the retry behavior intention.. instead of failing on the first stale batch exception.
Guessing this is to support some real-world behavior for the nursery specialists?

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
```diff
@@ -0,0 +1,114 @@
...
+      BatchWithdrawalModel(
+          batchId = batchId,
+          germinatingQuantityWithdrawn = germinatingQuantityWithdrawn,
+          notReadyQuantityWithdrawn = notReadyQuantityWithdrawn,
+          readyQuantityWithdrawn = readyQuantityWithdrawn,
+      )
+}
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class NurseryWithdrawalPayload(
```
karthikbtf, line 70: Took a sneak at the fast-followup WIP designs, this structure certainly supports the intended direction.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -264,6 +285,142 @@ class BatchStore(
...
+                      log.debug(
+                          "Batch $batchId was updated concurrently; retrying withdrawal ($retriesRemaining)")
+                    } else {
+                      log.error("Batch $batchId was stale repeatedly; giving up")
+                      throw e
+                    }
+                  }
+                }
+
+                if (retriesRemaining < 0) {
```
karthikbtf, line 418: can't see this happening

src/main/kotlin/com/terraformation/backend/nursery/model/WithdrawalModel.kt
karthikbtf: interesting, sorta but not `Void`?

charlie-tf: Oh interesting, so in the case of `WithdrawalModel<Nothing?>` it's impossible to instantiate the `Nothing` type, so it can only ever be null?

sgrimm: That's the idea, yep.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
```diff
@@ -0,0 +1,114 @@
...
+        NurseryWithdrawalPayload(model),
+    )
+  }
+}
+
+data class BatchWithdrawalPayload(
+    val batchId: BatchId,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val germinatingQuantityWithdrawn: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val notReadyQuantityWithdrawn: Int,
+    @JsonSetter(nulls = Nulls.FAIL) @Min(0) val readyQuantityWithdrawn: Int,
```
karthikbtf: I added some comments in here but my review draft doesn't show it anymore. If these comments are redundant apologies..
A few questions.
Based on designs, none of the withdrawal scenarios equire specifying the germinating quantity - is this mostly looking ahead and providing full coverage?
Outplant withdrawal seems to ask for a 'withdrawn quantity' - should that be ready quantity (client usage)?

karthikbtf (commented):
Looks good, thanks for doing all this looking ahead.
A few comments/questions.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
karthikbtf: Checked with @chudo76 `reason` (PRD spec) is shown as `Destination / Notes` in figma.

sgrimm: Reason is in the PRD under "Fields for Nursery Inventory: Withdraw." I take it the design renamed that field to "Notes?"  There is no "Notes" field for withdrawals in the PRD.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
sgrimm: I based this on the PRD, which says the destination needs to be a freeform text field for withdrawal types other than "Nursery Transfer." Seems like maybe the PRD and the design disagree about the list of withdrawal fields; I'll review the design and see what requirements changes they imply.

karthikbtf: Got it. 
Wondering if the designs conflate Destination and Reason - when they should be two separate inputs (or only one - "Reason" - for Nursery Transfer).

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
chudo76: Correct, this was not clear in the PRD vs mockup. The desired outcome was that if **Purpose** = Nursery Transfer, then "Destination" field becomes a `<select>` (`facilityId`). For other **Purpose**s, we replace that field with a freeform text field labelled "Destination / Notes" (which is currently named "Reason" in the PRD). Apologies for the confusion. cc:@rachel-tf-design

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: You can think of this as the optimistic-locking equivalent of doing a `SELECT FOR UPDATE` and blocking until the competing transaction is finished. If two withdrawals are happening at the same time (and there's enough inventory for both), they should both succeed without the client needing to know there was a collision.

Oh, I guess maybe the thing that makes this unclear is that in this case, unlike the case of editing a batch's quantities, the client is not passing in any version numbers. Semantically, "change the remaining quantity to 16" and "change the remaining quantity to 18" are conflicting operations, but "withdraw 3 seedlings" and "withdraw 4 seedlings" don't conflict, so there's no need for client-side conflict resolution in the latter case.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
sgrimm: > Based on designs, none of the withdrawal scenarios equire specifying the germinating quantity - is this mostly looking ahead and providing full coverage?

Yeah, pretty much. It seems wrong to me that there's no way to, e.g., tell the system you've thrown away dead seeds that failed to germinate. But you raise a good point and I think the right fix is to make this default to 0 rather than being required, so the client can leave it out of the payloads to start with.

> Outplant withdrawal seems to ask for a 'withdrawn quantity' - should that be ready quantity (client usage)?

PRD says withdrawals always ask for not-ready and ready quantities, so I'm not 100% sure of the intent, but your thought seems like it's probably correct.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
karthikbtf: Thanks for the explanation - this makes total sense.

I will add a note in the FE task to refresh the view prior to showing errors - in case the withdrawal failed because another withdrawal happened same time, and current request led to insufficient inventory.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
chudo76: I've updated the [PRD](https://terraformation.atlassian.net/wiki/spaces/PM/pages/569606164/PRD+Nursery+Support) to match designs.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
karthikbtf: > > Based on designs, none of the withdrawal scenarios equire specifying the germinating quantity - is this mostly looking ahead and providing full coverage?
> 
> Yeah, pretty much. It seems wrong to me that there's no way to, e.g., tell the system you've thrown away dead seeds that failed to germinate. But you raise a good point and I think the right fix is to make this default to 0 rather than being required, so the client can leave it out of the payloads to start with.
> 
> > Outplant withdrawal seems to ask for a 'withdrawn quantity' - should that be ready quantity (client usage)?
> 
> PRD says withdrawals always ask for not-ready and ready quantities, so I'm not 100% sure of the intent, but your thought seems like it's probably correct.

Got it, thanks :) i should cross check with the PRD instead of designs when reviewing here. Re: design/PRD differences  - I will bring them up with design outside the scope of this review.

src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt
sgrimm: I no longer understand what behavior we're trying to implement here, but I'll comment on the PRD since it's really a product question.

karthikbtf (approved):

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/698?utm_source=stack-comment).


Job Reference: Jaq45FKacaObSUdvtsmF

PR #701, opened 2022-10-11
SW-1845 Stop fetching accessions by number in tests

sgrimm:
In preparation for removing the requirement that accession numbers be globally
unique, update various tests that use accession numbers as unique lookup keys.

No application behavior change here. Replaces `AccessionsDao.fetchOneByNumber`
calls with `AccessionsDao.fetchOneById`.

PR #700, opened 2022-10-11
SW-1968 Implement nursery transfer withdrawals

sgrimm:
Add support for the `Nursery Transfer` withdrawal purpose, which will cause the
system to create new batches at a destination nursery facility.

PR #702, opened 2022-10-11
SW-1845 Revamp accession and batch number generation

sgrimm:
Previously, accession numbers were globally unique, which was fine because we were
generating them on the server side. But now we allow users to upload CSV files
with their own arbitrary accession numbers, and we don't want one organization to
be able to make an accession number unavailable to another one.

Change the database constraint so that accession numbers are only required to be
unique per facility.

In addition, the fact that we encoded dates into accession numbers sometimes
causes user confusion: those dates _usually_ match up with several other accession
dates such as the collection date, but not always. Although the date selection was
consistent on the server side, users didn't have a good way of learning how it
worked and made assumptions which then broke some of the time.

To address that, we now allocate these identifiers as simple incrementing numbers,
tracked separately per organization, starting with 100000 so they'll stay a
consistent width on screen for all but the very largest of organizations.

Seed accession and seedling batch numbers are allocated from the same place; we
will never generate an accession number that we've generated as a batch number for
the same organization or vice versa. However, if the user generates their own
accession numbers and uploads them in a CSV file, we don't enforce that the
numbers have to be unique across the organization, just unique within a particular
seed bank.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
@@ -1,11 +1,10 @@
 package com.terraformation.backend.db
 
-import com.terraformation.backend.db.seedbank.sequences.ACCESSION_NUMBER_SEQ
+import com.terraformation.backend.db.default_schema.OrganizationId
+import com.terraformation.backend.db.default_schema.tables.references.IDENTIFIER_SEQUENCES
 import com.terraformation.backend.log.perClassLogger
 import com.terraformation.backend.seedbank.db.AccessionStore
 import java.time.Clock
-import java.time.LocalDate
-import java.time.format.DateTimeFormatter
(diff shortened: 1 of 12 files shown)


sgrimm:
* main
    * **PR #698** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/698" target="_blank">Graphite</a> 
            * **PR #704** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/704" target="_blank">Graphite</a> 
          * **PR #702** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/702" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/702?utm_source=stack-comment).

sgrimm:
Apparently we still want date-based accession numbers after all. Closing this.

PR #703, opened 2022-10-11
Consolidate and rename nursery withdrawal notes fields

sgrimm:
Remove the `destination` field and rename `reason` to `notes`; having two separate
freeform text fields was redundant.

PR #704, opened 2022-10-11
SW-1845 Revamp accession and batch number generation

sgrimm:
Previously, accession numbers were globally unique, which was fine because we were
generating them on the server side. But now we allow users to upload CSV files
with their own arbitrary accession numbers, and we don't want one organization to
be able to make an accession number unavailable to another one.

Change the database constraint so that accession numbers are only required to be
unique per facility, and change the number generation so that we generate numbers
per organization rather than globally. Note the difference between the constraint
(per-facility uniqueness) and the default (per-organization uniqueness) -- we
think the latter is a good idea, but if people are uploading their own inventory
spreadsheets, they might use the same numbering scheme at two seed banks and we
don't want to bomb out on the CSV upload.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
@@ -2,46 +2,68 @@ package com.terraformation.backend.db
 
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.tables.references.IDENTIFIER_SEQUENCES
-import com.terraformation.backend.log.perClassLogger
 import com.terraformation.backend.seedbank.db.AccessionStore
 import java.time.Clock
 import java.time.LocalDate
+import java.time.ZoneId
+import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import javax.annotation.ManagedBean
 import org.jooq.DSLContext
+import org.jooq.impl.DSL
 
+/**
+ * Generates user-facing identifiers. These are used in places where we need a unique value to
+ * identify a resource but it's not acceptable to display the underlying integer ID from the
+ * database, e.g., accession numbers.
+ *
+ * Identifiers are mostly-fixed-length numeric values of the form YYYYMMDDXXX where XXX is a numeric
+ * suffix of three or more digits that starts at 000 for the first identifier created on a
+ * particular date. The desired behavior is for the suffix to represent the order in which entries
+ * were added to the system, so ideally we want to avoid gaps or out-of-order values, though it's
+ * fine for that to be best-effort.
+ *
+ * In most cases, there should be fewer than 1000 items created by an organization on a given day,
+ * so the identifiers will all be the same length, but if an organization creates a big burst of
+ * data all at once, we let the identifiers grow by a digit. That is, the identifier after
+ * 20221025999 is 202210251000 (999 -> 1000).
+ *
+ * The implementation uses a database table that holds the next value for each organization. The
+ * values follow the same pattern as the identifiers, but the suffix is always 10 digits; it is
+ * rendered as a 3-or-more-digit value. Doing it that way means the SQL query that allocates the
+ * next value doesn't have to be aware of the "grow the identifier if the suffix overflows 3 digits"
+ * logic; it just adds 1 to the previous value. To take the above example, the value in the database
+ * would go from 202210250000000999 to 202210250000001000.
+ *
+ * If the date part of the sequence value doesn't match the current date, the value is reset to the
+ * zero suffix for the current date. That is, if the database says the next value is
+ * 202210310000000036 but it is now November 1, the value will be bumped to 202211010000000000 and
+ * the next identifier (with the suffix collapsed to 3 digits) will be 20221101000.
+ *
+ * Note that although this class is guaranteed to only ever return a given value once for a given
+ * organization ID, it is possible for the generated identifiers to collide with user-supplied
+ * identifiers. To guard against that, [AccessionStore.create] will retry a few times if it gets an
+ * identifier that's already in use.
+ */
 @ManagedBean
 class IdentifierGenerator(
     private val clock: Clock,
     private val dslContext: DSLContext,
 ) {
-  private val log = perClassLogger()
-
   /**
-   * Returns the next unused user-facing identifier. This is used in places where we need a unique
-   * value to identify a resource but it's not acceptable to display the underlying integer ID from
-   * the database, e.g., accession numbers.
+   * Generates a new identifier for an organization.
    *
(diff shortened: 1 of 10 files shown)


sgrimm:
* main
    * **PR #698** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/698" target="_blank">Graphite</a> 
            * **PR #704** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/704" target="_blank">Graphite</a>  👈
          * **PR #702** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/702" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/704?utm_source=stack-comment).

sgrimm:
CI build died in the setup phase; retrying it.

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -38,28 +39,41 @@ class IdentifierGenerator(
...
+            .where(ORGANIZATION_ID.eq(organizationId))
+            .and(NEXT_VALUE.lt(firstValueForToday))
+            .execute()
+      }
+
+      return generateIdentifier(organizationId)
+    }
 
-    return "%08d%03d".format(todayAsLong, suffix)
+    return "%08d%03d".format(datePart, suffixPart)
```
charlie-tf, line 84: Why are only three digits of `suffixPart` used if `suffixMultiplier` is `10000000000L`?

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -38,28 +39,41 @@ class IdentifierGenerator(
...
-          dslContext
-              .alterSequence(ACCESSION_NUMBER_SEQ)
-              .restartWith(firstValueForToday + 1)
-              .execute()
-          log.info("Resetting identifier sequence to $firstValueForToday")
-          0
-        } else {
-          suffixPart
-        }
+    if (todayAsLong != datePart) {
```
charlie-tf: When does this happen? It seems like if `suffixPart` is greater than `9999999999L` and "overflows", thus incrementing the `datePart`. If the identifier is reset to `firstValueForToday` does that generate a duplicate identifier?

charlie-tf (commented):
couple questions about `generateIdentifier`

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
sgrimm: In theory that should be answered in the method doc comment (which doesn't show as part of the PR) but looking at it, I now realize I didn't update it to reflect this change, so thanks for asking. Not trying to be obtuse here; I'll update it and then ask you to read it to see if it answers your question (if not, I'll edit it further) so the answer is there for future people to see.

sgrimm:
@charlie-tf Have a look at the updated version of this PR and see if it answers your questions. They ended up prompting me to make a couple useful improvements to the code, so thanks for asking them!

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -1,65 +1,86 @@
...
+              .set(NEXT_VALUE, DSL.greatest(NEXT_VALUE.plus(1), DSL.value(firstValueForToday)))
+              .returning(NEXT_VALUE)
+              .fetchOne(NEXT_VALUE)!!
         }
 
-    return "%08d%03d".format(todayAsLong, suffix)
+    val datePart = sequenceValue / suffixMultiplier
+    val suffixPart = sequenceValue.rem(suffixMultiplier)
+
+    return "%08d%03d".format(datePart, suffixPart)
```
charlie-tf, line 84: oh, gotcha, `%03d` does not do a mod 1000, it just makes sure to pad with zeros out to _at least_ 3 digits (had to go look that up 😅 )

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -1,65 +1,86 @@
...
-              .execute()
-          log.info("Resetting identifier sequence to $firstValueForToday")
-          0
-        } else {
-          suffixPart
+              .insertInto(IDENTIFIER_SEQUENCES)
+              .set(ORGANIZATION_ID, organizationId)
+              .set(NEXT_VALUE, firstValueForToday)
+              .onDuplicateKeyUpdate()
+              .set(NEXT_VALUE, DSL.greatest(NEXT_VALUE.plus(1), DSL.value(firstValueForToday)))
```
charlie-tf, line 76: nice update!

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -1,65 +1,86 @@
...
-              .restartWith(firstValueForToday + 1)
-              .execute()
-          log.info("Resetting identifier sequence to $firstValueForToday")
-          0
-        } else {
-          suffixPart
+              .insertInto(IDENTIFIER_SEQUENCES)
+              .set(ORGANIZATION_ID, organizationId)
+              .set(NEXT_VALUE, firstValueForToday)
+              .onDuplicateKeyUpdate()
```
charlie-tf, line 75: Does `onDuplicateKeyUpdate` perform the retry behavior?

charlie-tf (approved):
Looks good! Thanks so much for elaborating on the generation method in the method documentation.

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
sgrimm: This gets translated to PostgreSQL-specific syntax, `ON CONFLICT (organization_id) DO UPDATE`. The organization ID is this table's primary key. The end effect is to insert the starting value if there isn't already a row for an organization, but if there is, run the update part of the query.

The database ensures this is all done atomically, so if two server instances try this at the same time, one will block until the other is done and only then will it read the current value and add 1 to it.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/704?utm_source=stack-comment).


Job Reference: ji4aaGlSyAHNp1Vle6VT

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
charlie-tf: ooh, that makes sense. thanks for the detailed explanation!

PR #705, opened 2022-10-12
SW-1971 Add endpoint to transfer seeds to nursery

sgrimm:
`POST /api/v2/seedbank/accessions/{id}/transfers/nursery` will withdraw seeds from
an accession and create a new seedling batch.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/TransfersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/TransfersController.kt
@@ -1,0 +1,79 @@
+package com.terraformation.backend.seedbank.api
+
+import com.fasterxml.jackson.annotation.JsonSetter
+import com.fasterxml.jackson.annotation.Nulls
+import com.terraformation.backend.api.SeedBankAppEndpoint
+import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.db.default_schema.FacilityId
+import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.nursery.tables.pojos.BatchesRow
+import com.terraformation.backend.db.seedbank.AccessionId
+import com.terraformation.backend.nursery.api.BatchPayload
+import com.terraformation.backend.seedbank.AccessionService
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDate
+import javax.validation.constraints.Min
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #705** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/705" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/705?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/api/TransfersController.kt
```diff
@@ -0,0 +1,79 @@
...
+            accessionId, payload.toBatchesRow(), payload.withdrawnByUserId)
+    return CreateNurseryTransferResponsePayload(AccessionPayloadV2(accession), BatchPayload(batch))
+  }
+}
+
+data class CreateNurseryTransferRequestPayload(
+    val date: LocalDate,
+    val destinationFacilityId: FacilityId,
+    @JsonSetter(nulls = Nulls.FAIL)
+    @Min(0) //
```
karthikbtf, line 46: should the `//` be removed?

karthikbtf (approved):
lgtm

src/main/kotlin/com/terraformation/backend/seedbank/api/TransfersController.kt
sgrimm: It's there to stop ktfmt from putting the annotations and the `val` on a single line, which I thought made the class harder to read since there's a mix of annotated and non-annotated properties. I initially tried adding an explanation after the `//` but that also added visual clutter.

The other option would be to put the annotated properties all together and let ktfmt do its default thing. I kind of like having properties in alphabetical order but maybe it's a bad tradeoff here.

sgrimm: Though maybe if I did `// line-break` or something, that might look okay and look less like a mistake; I'll try that.

karthikbtf: Got it, this is fine, alphabetical order is easier to read/search.
Doe ktfmt have a config option to override default behavior?

sgrimm: ktfmt is deliberately not configurable. It's of the class of code formatters like gofmt and Black that take an approach of, "The point of an automated formatter is to stop people from wasting time arguing about formatting choices, and adding a bunch of fine-grained settings would cause people to waste time fussing with the settings instead." So its only configuration option is to control which of two Kotlin style guides it follows.

When I've reported formatting problems, they've often agreed that it isn't acting as advertised and fixed the problems. I just filed an issue earlier today, in fact: https://github.com/facebook/ktfmt/issues/360

Splitting on annotations has been an open request for a while and it sounds like the authors are on the fence about whether it's a good thing to change: https://github.com/facebook/ktfmt/issues/226

karthikbtf: Thanks for the info. Just so you know, this PR is now mentioned in the FB issues :)

PR #706, opened 2022-10-12
SW-1986 Generate shorter accession/batch numbers

sgrimm:
After getting feedback about the previous accession/batch numbering scheme,
we want to change two things: make the identifiers shorter and allow people
to visually differentiate between accession numbers and batch numbers.

This new format is 3 characters shorter and treats accessions and batches
as separate sequential number spaces: an accession number in the old format
looked like `20221012888` and in the new scheme it would be `22-1-888` where
the `1` indicates that it's an accession number.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
@@ -7,38 +7,31 @@ import java.time.Clock
 import java.time.LocalDate
 import java.time.ZoneId
 import java.time.ZoneOffset
-import java.time.format.DateTimeFormatter
 import javax.annotation.ManagedBean
 import org.jooq.DSLContext
-import org.jooq.impl.DSL
 
 /**
  * Generates user-facing identifiers. These are used in places where we need a unique value to
  * identify a resource but it's not acceptable to display the underlying integer ID from the
  * database, e.g., accession numbers.
  *
- * Identifiers are mostly-fixed-length numeric values of the form YYYYMMDDXXX where XXX is a numeric
- * suffix of three or more digits that starts at 000 for the first identifier created on a
- * particular date. The desired behavior is for the suffix to represent the order in which entries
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #706** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/706" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/706?utm_source=stack-comment).

src/main/resources/db/migration/V145__IdentifierSequencesPerPrefix.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE identifier_sequences;
```
charlie-tf, line 1: should be good since we haven't hit prod with https://github.com/terraware/terraware-server/pull/704 yet 👍

src/main/kotlin/com/terraformation/backend/db/IdentifierGenerator.kt
```diff
@@ -7,38 +7,31 @@ import java.time.Clock
...
  *
- * The implementation uses a database table that holds the next value for each organization. The
- * values follow the same pattern as the identifiers, but the suffix is always 10 digits; it is
- * rendered as a 3-or-more-digit value. Doing it that way means the SQL query that allocates the
- * next value doesn't have to be aware of the "grow the identifier if the suffix overflows 3 digits"
- * logic; it just adds 1 to the previous value. To take the above example, the value in the database
- * would go from 202210250000000999 to 202210250000001000.
+ * In the highly unlikely event this code still exists in the year 2122, the two-digit year from
+ * 2022 will be reused, but this won't cause a collision; the next sequence value from 2022 will
+ * still exist and will be used.
```
charlie-tf, line 30: our ancestors will thank you for this. oh, and for all the trees, too 😉

charlie-tf (approved):
looks good!

PR #707, opened 2022-10-13
SW-1488 Use property accessors for jOOQ implicit joins

sgrimm:
The Kotlin code generator in jOOQ 3.17 added a new way to do implicit joins.
Previously, implicit joins were done using method calls. But now they can also
be done using property references.

It only saves two characters, but may as well use the more compact syntax.


sgrimm:
* main
  * **PR #707** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/707" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/707?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #708, opened 2022-10-14
SW-1979 Move jOOQ configuration to a subproject

sgrimm:
Previously, our customization and configuration of jOOQ code generation (which
needs to be executed by Gradle as part of the build process, not by our
application code at runtime) lived in the `buildSrc` directory.

`buildSrc` has a problem: whenever it changes, the entire project is rebuilt.
Changing the jOOQ configuration would cause the MJML email templates to be
rebuilt even though they have no dependency relationship.

Move the jOOQ configuration to a separate subproject that produces a Gradle
plugin. Changes to the configuration will cause the plugin to be rebuilt, which
will trigger a fresh round of jOOQ code generation, but won't invalidate
unrelated parts of the project. Updating `Config.kt` won't cause MJML templates
to be rebuilt.


sgrimm:
* main
  * **PR #708** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/708" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/708?utm_source=stack-comment).

jooq/src/main/kotlin/com/terraformation/backend/jooq/IdWrapper.kt
```diff
@@ -3,10 +3,7 @@ package com.terraformation.backend.jooq
 import org.jooq.codegen.JavaWriter
 import org.jooq.meta.jaxb.ForcedType
 
-class IdWrapper(
-  private val className: String,
-  includeExpressions: List<String>
-) {
+class IdWrapper(private val className: String, includeExpressions: List<String>) {
```
sgrimm, line 6: No actual changes to this file, just ktfmt reformatting.

jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -33,7 +33,8 @@ class TerrawareGenerator : KotlinGenerator() {
       import com.terraformation.backend.db.EnumFromReferenceTable
       import org.jooq.impl.AbstractConverter
 
-    """.trimIndent())
+    """
+            .trimIndent())
```
charlie-tf, line 37: what's the reason for these extra spaces?

charlie-tf: is just from ktfmt?

charlie-tf (approved):
👍

jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
sgrimm: Yeah, this is ktfmt. Its handling of multiline strings leaves something to be desired.

PR #711, opened 2022-10-17
SW-2006 Restart JobRunr after database recovery

sgrimm:
If there's a database problem, JobRunr shuts down its background worker thread to
cut down on log spew. But it doesn't restart the thread when the database comes
back. This causes the server's health check to report unhealthy status even though
request handling is working fine.

Add a watchdog thread to start JobRunr back up again after the database becomes
available.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/daily/JobRunrRecoveryThread.kt
+++ b/src/main/kotlin/com/terraformation/backend/daily/JobRunrRecoveryThread.kt
@@ -1,0 +1,79 @@
+package com.terraformation.backend.daily
+
+import com.terraformation.backend.config.TerrawareServerConfig
+import com.terraformation.backend.log.perClassLogger
+import java.sql.SQLException
+import java.time.Duration
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import javax.annotation.ManagedBean
+import javax.annotation.PostConstruct
+import javax.annotation.PreDestroy
+import javax.sql.DataSource
+import org.jobrunr.server.BackgroundJobServer
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+
+/**
+ * Restarts JobRunr's background job processing thread after database problems. JobRunr shuts down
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #711** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/711" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/711?utm_source=stack-comment).

sgrimm:
Tested this by:

1. Running the server
2. Killing PostgreSQL
3. Waiting for the "JobRunr not running" log message from this new class
4. Verifying that `/actuator/health` returns unhealthy status
5. Restarting PostgreSQL
6. Waiting for the "Restarting JobRunr" log message from this new class and the log messages from JobRunr indicating its background thread has started
7. Verifying that `/actuator/health` returns healthy status

src/main/kotlin/com/terraformation/backend/daily/JobRunrRecoveryThread.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.jobrunr.server.BackgroundJobServer
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+
+/**
+ * Restarts JobRunr's background job processing thread after database problems. JobRunr shuts down
+ * when it can't talk to the database to avoid spewing errors to the log, but it doesn't come back
+ * up when the database becomes available again. We want the server to recover from transient
+ * database problems.
+ */
+@ConditionalOnProperty(TerrawareServerConfig.DAILY_TASKS_ENABLED_PROPERTY, matchIfMissing = true)
```
sgrimm, line 22: This is so we don't start this background thread in contexts where we're not running JobRunr at all, e.g., when you run `./gradlew generateOpenApiDocs` and the server starts up with no database configuration.

karthikbtf (approved):
Looks good to me 👍 
Thanks!!

PR #712, opened 2022-10-17
SW-2010 Remove v1 scheduled accession state transitions

sgrimm:
We were continuing to run the scheduled state transitions on existing accessions
even though, with the v2 UI rollout, the user no longer has any way to edit most
of the dates.

Remove the daily job that scans for accessions whose scheduled dates have arrived.

In addition, remove the date-based transition logic from `AccessionModel` to
prevent the state from being recalculated when a user reads an older accession
using the v2 API, since the first step of the v1->v2 conversion is to call the
v1 "recalculate all the computed fields" method.

The `AccessionModel` changes represent a change in behavior, since v1 accessions
now ignore the various user-supplied dates and retain the existing state in most
cases. Update unit tests that relied on the old behavior.


sgrimm:
* main
  * **PR #712** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/712" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/712?utm_source=stack-comment).

karthikbtf (approved):

PR #713, opened 2022-10-18
SW-1603 Detect collision of suggested species name

sgrimm:
If species 1 and 2 exist, and species 2 has an incorrect name (misspelled or
synonym) that should be changed to the same name as species 1, return HTTP 409
(Conflict) with a descriptive error message rather than an internal server error
when the user tries to accept the suggested edit.

Eventually we'll most likely want to add logic to merge the two species, but
until then, we should at least show a meaningful error message.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/species/db/SpeciesStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/species/db/SpeciesStoreTest.kt
@@ -366,18 +366,18 @@ internal class SpeciesStoreTest : DatabaseTest(), RunsAsUser {
   }
 
   @Test
-  fun `acceptProblemSuggestion throws ScientificNameExistsException on name collision`() {
+  fun `acceptProblemSuggestion throws exception if suggested scientific name is already in use`() {
     store.createSpecies(
-        SpeciesRow(organizationId = organizationId, scientificName = "Species name"))
-    val badSpeciesId =
+        SpeciesRow(organizationId = organizationId, scientificName = "Correct name"))
+    val speciesIdWithOutdatedName =
         store.createSpecies(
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #713** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/713" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/713?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/species/db/SpeciesStoreTest.kt
```diff
@@ -363,4 +364,23 @@ internal class SpeciesStoreTest : DatabaseTest(), RunsAsUser {
...
     assertThrows<OrganizationNotFoundException> { store.findAllSpecies(organizationId) }
   }
+
+  @Test
+  fun `acceptProblemSuggestion throws ScientificNameExistsException on name collision`() {
+    store.createSpecies(
+        SpeciesRow(organizationId = organizationId, scientificName = "Species name"))
+    val badSpeciesId =
+        store.createSpecies(
+            SpeciesRow(organizationId = organizationId, scientificName = "Bad name"))
```
charlie-tf: little confused here. Does this throw because there already exists a species with the scientific name "Bad name"?

sgrimm: No, because there's already one with the suggested name ("Species name"). I'll update this test to make that clearer.

charlie-tf: Ohh, i see the flow of things now

PR #715, opened 2022-10-18
Add response annotations to some api in batches controller.

karthikbtf:
- primarily used in the generated types for FE

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonSetter
 import com.fasterxml.jackson.annotation.Nulls
 import com.terraformation.backend.api.ApiResponse404
 import com.terraformation.backend.api.ApiResponse412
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/terraformation/backend/nursery/api/BatchesController.kt
```diff
@@ -36,25 +38,35 @@ import org.springframework.web.bind.annotation.RestController
...
   @GetMapping("/{id}")
   fun getBatch(@PathVariable("id") id: BatchId): BatchResponsePayload {
     val row = batchStore.fetchOneById(id)
     return BatchResponsePayload(BatchPayload(row))
   }
 
+  @ApiResponse(
+      responseCode = "200",
+      description =
+          "The batch was created successfully. Response includes fields populated by the server, including the batch ID.")
```
sgrimm: Can you split these descriptions so they're under the 100-character line length limit? (ktfmt doesn't auto-split long strings.)

sgrimm (changes requested):
Looks good aside from a style thing. I've been sloppy about including these.

sgrimm (approved):

PR #723, opened 2022-10-20
SW-1939 Remove summary calculations for scheduled withdrawals

sgrimm:
In v1, we allowed withdrawals to be scheduled for the future, and we had summaries
of future withdrawals in the UI. In v2, we don't allow scheduling and there are no
summaries. Remove the code that calculatees the summary data.

This code was already dead; it was only referenced by tests.

PR #724, opened 2022-10-20
SW-1939 Test v2 behavior in AccessionServiceTest

sgrimm:
This test was verifying that we can create withdrawals on v1 accessions using the
v2 API; update it to test a pure-v2 scenario since there should no longer be any
v1 accessions in the database.


sgrimm:
* main
  * **PR #723** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/723" target="_blank">Graphite</a> 
    * **PR #724** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/724" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/724?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/seedbank/AccessionServiceTest.kt
```diff
@@ -101,13 +101,17 @@ internal class AccessionServiceTest : DatabaseTest(), RunsAsUser {
         AccessionModel(
             id = accessionId,
             facilityId = FacilityId(1),
+            isManualState = true,
```
charlie-tf, line 104: What does `isManualState` represent?

charlie-tf (approved):
just one question

src/test/kotlin/com/terraformation/backend/seedbank/AccessionServiceTest.kt
sgrimm: Literally: it means the accession's state is manually set by the user, as opposed to the system automatically figuring out the state based on things like whether or not the quantity has been entered.

But in practice: it means this accession was created or updated using the v2 API.

The name made sense when we were first starting to implement the new seeds logic; at that point we hadn't decided to split it off into a completely separate set of API endpoints and we thought that "can the user manually edit the state?" was going to be a setting the client would need to know about. But as things turned out, we should have called this field `apiVersion` or `isV2` or something like that.

sgrimm: I haven't bothered changing the name because this field will go away once we've gotten rid of the v1 code.

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/724?utm_source=stack-comment).


Job Reference: WwnVlKhsYop0dlmGxH7t

src/test/kotlin/com/terraformation/backend/seedbank/AccessionServiceTest.kt
charlie-tf: Makes sense, thanks!

PR #725, opened 2022-10-20
Add missing descriptions to OpenAPI schema fields

sgrimm:
Our OpenAPI customization class has code to work around a limitation of the
Swagger core library's annotation processing that causes it to throw away payload
field descriptions if the fields are references to other payload classes.
Previously, the workaround required keeping an explicit list of all the payload
classes that needed to be fixed up. The list was, inevitably, out of date and
missing several classes that ought to have been included.

Switch to a classpath-scanning approach where we probe the code base using
reflection and find all the payload classes. This is slower but it's a one-time
operation that only happens the first time someone requests the OpenAPI schema
from a server instance.

In addition, we were using the wrong annotations on some of the payload fields
whose values are collections; fix those.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
@@ -9,10 +9,6 @@ import com.terraformation.backend.device.api.CreateDeviceRequestPayload
 import com.terraformation.backend.device.api.DeviceConfig
 import com.terraformation.backend.device.api.UpdateDeviceRequestPayload
 import com.terraformation.backend.search.api.SearchResponsePayload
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #725** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/725?utm_source=stack-comment).

sgrimm:
This started off as part of my "remove all references to the v1 payload classes" work but I realized the reference in `OpenApiConfig` shouldn't have needed to be there in the first place.

karthikbtf (approved):
Thanks for doing this!

PR #727, opened 2022-10-20
SW-1939 Migrate v1 accessions to v2

sgrimm:
Now that the accession v1 API endpoints are gone, all existing accessions will be
converted to v2 whenever they're accessed. There's thus no need to keep them
around in v1 form any more. Add a migration class that launches when the server
starts up and migrates all the v1 accessions.

Once this migration is finished, we will be able to remove the v1->v2 conversion
code since there will no longer be any situations where it'd be called.


sgrimm:
* main
  * **PR #732** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/732" target="_blank">Graphite</a> 
    * **PR #727** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/727" target="_blank">Graphite</a>  👈
      * **PR #728** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/728" target="_blank">Graphite</a> 
        * **PR #729** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/729" target="_blank">Graphite</a> 
          * **PR #733** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/733" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/727?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionV2Migrator.kt
```diff
@@ -0,0 +1,88 @@
...
+      try {
+        // Convert each accession in a separate transaction so failed conversions don't stop
+        // successful ones from being committed to the database.
+        dslContext.transaction { _ ->
+          nextAccessionId =
+              dslContext
+                  .select(ACCESSIONS.ID)
+                  .from(ACCESSIONS)
+                  .where(ACCESSIONS.IS_MANUAL_STATE.isNull)
+                  .and(nextAccessionId?.let { ACCESSIONS.ID.gt(it) } ?: DSL.trueCondition())
```
karthikbtf, line 62: is this to avoid looping on error?

karthikbtf (approved):
👍

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionV2Migrator.kt
sgrimm: Yes, and the `skipLocked()` is so that if both server instances are running this at once, they won't step on each other.

PR #756, opened 2022-10-26
SW-673 Add refresh token support to test scripts

sgrimm:
Previously, to run the test scripts, you had to manually get a valid session
cookie or access token and pass it as an argument. Add support for refresh
tokens: you can set a refresh token in an environment variable or pass it as
a command-line argument and the scripts will use it to fetch a new access
token as needed.

This is mostly useful for manual ad-hoc testing at the moment, but moves us a
little closer to being able to use these test scripts for end-to-end automated
tests.

### DIFF
--- a/scripts/client.py
+++ b/scripts/client.py
@@ -27,12 +27,10 @@ def _authenticated(func):
 class TerrawareClient:
     def __init__(
         self,
-        client_id: Optional[str] = None,
         refresh_token: Optional[str] = None,
         session: Optional[str] = None,
         base_url: Optional[str] = None,
     ):
-        self.client_id = client_id
         self.base_url = (base_url or DEFAULT_URL).rstrip("/")
         self.refresh_token = refresh_token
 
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #756** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/756" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/756?utm_source=stack-comment).

scripts/client.py
```diff
@@ -137,15 +170,56 @@ def list_species(self, organization_id):
...
+            # and Keycloak will reject it if the parameters are bogus.
+            decoded_token = jwt.decode(
+                self.refresh_token, options={"verify_signature": False}
+            )
+            base_url = decoded_token["iss"]
+
+            r = requests.post(
+                f"{base_url}/protocol/openid-connect/token",
+                data={
+                    "client_id": self.client_id,
```
karthikbtf: just in case it is possible.. can client id be inferred from `sub` in decoded jwt? i can't recall if that only shows up in the access token.
this is fine too.. but just in case it is possible.. and we want client_id to be optional in runtime args - when refresh token is provided.

karthikbtf (approved):
super cool!

scripts/client.py
sgrimm: Ah, good point. It can indeed.

sgrimm: For the record, Keycloak puts it in the `azp` (audience) field.

karthikbtf (approved):
looks great, thanks for the changes 👍

PR #760, opened 2022-10-27
Publish git log since last release with every commit

alexTerraformation:


charlie-tf:
I've pulled this branch and trying to figure out how to run this. Is there a gradlew command for this, or is it strictly through GH actions? How would I generate this locally?

charlie-tf:
Update: `bash .github/scripts/gitlog.sh ` runs it, and I get a `docs/unreleased.log` file out with commit info in it as expected, but it seems to struggle on figuring out the tags
```
fatal: No tags can describe '126fa2031195b3fd9f9f83548d5e049240e9abac'.
Try --always, or create some tags.
```
docs/unreleased.log:
```
126fa203 Publish git log since last release with every commit
16cf9a08 SW-673 Add refresh token support to test scripts (#756)
3632c0be SW-1885 Add readyByDate to nursery withdrawals (#755)
c1ba5cad Add imports for JUnit assertion methods (#740)
77c3420e SW-1875 Add batch version to search API (#739)
fa7c6ee1 SW-1939 Remove v1 accession payload classes (#733)
...
```

alexTerraformation:
> Update: `bash .github/scripts/gitlog.sh ` runs it, and I get a `docs/unreleased.log` file out with commit info in it as expected, but it seems to struggle on figuring out the tags
> 
> ```
> fatal: No tags can describe '126fa2031195b3fd9f9f83548d5e049240e9abac'.
> Try --always, or create some tags.
> ```

That would be because some of the tags in the server repo are on a branch.  I didn't realize that git describe would fail in that case.  Will look at an alternative.

karthikbtf:
> > Update: `bash .github/scripts/gitlog.sh ` runs it, and I get a `docs/unreleased.log` file out with commit info in it as expected, but it seems to struggle on figuring out the tags
> > ```
> > fatal: No tags can describe '126fa2031195b3fd9f9f83548d5e049240e9abac'.
> > Try --always, or create some tags.
> > ```
> 
> That would be because some of the tags in the server repo are on a branch. I didn't realize that git describe would fail in that case. Will look at an alternative.

We could probably skip the `git describe`.. list the tags and grab the latest.

alexTerraformation:
> > > Update: `bash .github/scripts/gitlog.sh ` runs it, and I get a `docs/unreleased.log` file out with commit info in it as expected, but it seems to struggle on figuring out the tags
> > > ```
> > > fatal: No tags can describe '126fa2031195b3fd9f9f83548d5e049240e9abac'.
> > > Try --always, or create some tags.
> > > ```
> > 
> > 
> > That would be because some of the tags in the server repo are on a branch. I didn't realize that git describe would fail in that case. Will look at an alternative.
> 
> We could probably skip the `git describe`.. list the tags and grab the latest.

I think the issue is the same.  git describe will not return tags on a branch AFAIK.  But it's easy enough to get a list of tags by creation date and pick the latest so I'll just to that.

.github/workflows/docs.yaml
```diff
@@ -5,14 +5,6 @@ on:
   push:
     branches:
       - main
```
sgrimm, line 7: It's unfortunate that this will mean we're going to regenerate the license report and the database schema diagrams for every commit. Probably not worth optimizing for, though; it just means more GitHub Actions minutes but if we start caring about conserving those, there are much bigger fish to fry than this.

.github/scripts/gitlog.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+LAST_TAG_DESCRIBE=`for tag in $(git tag | grep -E 'v[0-9]+\.[0-9]'); do
+    git describe --tags --long --match="$tag"
+done | sort -k2 -t"-" -n | head -n1`
+
+LAST_TAG=${LAST_TAG_DESCRIBE%%-*}  # retain the part before the first dash
+HEAD=${LAST_TAG_DESCRIBE##*g}
+
+git log $LAST_TAG..HEAD --oneline > docs/unreleased.log
```
sgrimm, line 5: This makes me wonder if we could generate a full release history or a history of the last N releases rather than clearing this out every time we push a tag. Out of scope for this task, though.

alexTerraformation: Yes, that should be doable if we decide we want to

sgrimm (approved):
This seems like a good starting point. Is there a Jira task for this? If so, can you put its task ID in the PR title?

PR #761, opened 2022-10-27
SW-1951 Refactor CSV import and validation code

sgrimm:
In preparation for adding a CSV importer for seedling batches, refactor the
species and accession CSV code to extract the common logic into superclasses.

No change in functionality here except that the error message for invalid header
rows is now consistent across accession and species CSVs.

PR #762, opened 2022-10-27
SW-1951 Add seedling batch CSV upload API

sgrimm:
Add a set of endpoints under `/api/v1/nursery/batches/uploads` to import seedling
batches from CSV files.

The API is similar to the CSV upload APIs for species and accessions, but in the
case of seedling batches, there are no warning conditions that require user
confirmation. There is thus no need for endpoints to resolve problems or cancel
an upload that is awaiting confirmation.


sgrimm:
* main
  * **PR #761** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/761" target="_blank">Graphite</a> 
    * **PR #762** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/762" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/762?utm_source=stack-comment).

sgrimm:
This may need a slight change depending on the outcome of SW-2060 but we can do that as a followup if needed.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
```diff
@@ -0,0 +1,126 @@
...
+        dslContext,
+        fileStore,
+        scheduler,
+        uploadProblemsDao,
+        uploadsDao,
+        uploadService,
+        uploadStore,
+        userStore) {
+  override val templatePath: String
+    get() = "/csv//batches-template.csv"
```
charlie-tf: why is there a double `//`?

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
```diff
@@ -0,0 +1,126 @@
...
+  ) {
+    val values = rawValues.map { it?.trim()?.ifEmpty { null } }
+
+    val scientificName = values[0]!!
+    val commonName = values[1]
+    val germinatingQuantity = values[2]?.toInt() ?: 0
+    val seedlingQuantity = values[3]?.toInt() ?: 0
+    val storedDate = LocalDate.parse(values[4])!!
+
+    val speciesId =
```
charlie-tf, line 107: If i understand correctly, this line creates a species row in the DB if it doesn't exist before?

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    requirePermissions {
+      createBatch(facilityId)
+      createSpecies(organizationId)
+    }
+
+    // Consume header row
+    csvReader.readNext()
+
+    dslContext.transaction { _ -> csvReader.forEach { importRow(it, organizationId, facilityId) } }
```
charlie-tf, line 91: if any single row fails, it will fail the whole transaction, correct? That is, it won't skip invalid rows and continue?

charlie-tf (approved):
Looks good to me!

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
sgrimm: Oops, that's a typo. Thanks.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
sgrimm: > If i understand correctly, this line creates a species row in the DB if it doesn't exist before?

Yes, that's correct.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchImporter.kt
sgrimm: > if any single row fails, it will fail the whole transaction, correct? That is, it won't skip invalid rows and continue?

Correct. We don't want to end up with a situation where the file is half-imported and the user has to go edit their CSV file to remove the successfully-imported lines before they try again.

charlie-tf: that makes sense!

charlie-tf (approved):

sgrimm:
**Graphite Merge Job**

Current status: ✅ Merged

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/762?utm_source=stack-comment).


Job Reference: eiHYvN0UCPVMZk8mLgrI

PR #764, opened 2022-10-27
SW-1842 Only do fuzzy matching if there are no exact matches

sgrimm:
We want users to be able to type nonexistent accession numbers into the accession
list UI and get back close matches. But we also want users to be able to paste
exact accession numbers into the same search field and not have the results
cluttered with accessions they didn't search for.

Try to thread this needle by making the fuzzy search code start off by trying an
exact search. If that finds matches, then we're probably in the "user entered an
exact value" situation; return those results. Otherwise do the fuzzy search and
return approximate matches.

This change means there is no way for a user to do a fuzzy search for a value that
also happens to be an exact match; we believe that will still be an improvement
over the current behavior.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
@@ -12,13 +12,31 @@ enum class SearchFilterType {
 
 interface SearchNode {
   fun toCondition(): Condition
+
+  /**
+   * Converts this node and all its descendents from fuzzy to exact filtering. If this node does not
+   * currently contain any fuzzy filter criteria, this method _must_ return an object that tests
+   * equal to `this` (e.g., by just returning `this`).
+   */
+  fun toExactSearch(): SearchNode
+
   fun referencedSublists(): Set<SublistField>
 }
 
-class OrNode(private val children: List<SearchNode>) : SearchNode {
+data class OrNode(private val children: List<SearchNode>) : SearchNode {
   override fun toCondition(): Condition {
     val conditions = children.map { it.toCondition() }
     return if (conditions.size == 1) conditions[0] else DSL.or(conditions)
+  }
+
+  override fun toExactSearch(): OrNode {
+    val exactChildren = children.map { it.toExactSearch() }
+
+    return if (exactChildren != children) {
+      OrNode(exactChildren)
+    } else {
+      this
+    }
   }
 
   override fun referencedSublists(): Set<SublistField> {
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #764** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/764" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/764?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
```diff
@@ -12,15 +12,33 @@ enum class SearchFilterType {
...
+   * currently contain any fuzzy filter criteria, this method _must_ return an object that tests
+   * equal to `this` (e.g., by just returning `this`).
+   */
+  fun toExactSearch(): SearchNode
+
   fun referencedSublists(): Set<SublistField>
 }
 
-class OrNode(private val children: List<SearchNode>) : SearchNode {
+data class OrNode(private val children: List<SearchNode>) : SearchNode {
```
charlie-tf, line 26: nice usage of data classes. Is a one of the benefits here that it provides equality checking?

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
```diff
@@ -59,7 +97,7 @@ class NotNode(val child: SearchNode) : SearchNode {
   }
 }
 
-class FieldNode(
+data class FieldNode(
```
charlie-tf, line 100: Are `FieldNode`s the leaf nodes in the search?

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceFuzzySearchTest.kt
```diff
@@ -43,4 +43,20 @@ internal class SearchServiceFuzzySearchTest : SearchServiceTest() {
...
     assertEquals(expected, result)
   }
+
+  @Test
+  fun `fuzzy search on text fields tries exact matching first`() {
+    accessionsDao.update(accessionsDao.fetchOneById(AccessionId(1000))!!.copy(number = "22-1-100"))
+    accessionsDao.update(accessionsDao.fetchOneById(AccessionId(1001))!!.copy(number = "22-1-101"))
+
+    val fields = listOf(accessionNumberField)
+    val searchNode = FieldNode(accessionNumberField, listOf("22-1-100"), SearchFilterType.Fuzzy)
```
charlie-tf, line 53: in this case, would a fuzzy search also return `22-1-101`?

charlie-tf (approved):
looks good! just a few questions to check my understanding

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
sgrimm: Yes, I did this to get the `equals` implementation.

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
sgrimm: > Are `FieldNode`s the leaf nodes in the search?

Yep!

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceFuzzySearchTest.kt
sgrimm: > in this case, would a fuzzy search also return `22-1-101`?

Yes. I'll update the test to make that more explicit.

charlie-tf (approved):

PR #765, opened 2022-10-28
SW-1842 Exact-then-fuzzy logic for search summaries

sgrimm:
Commit ee97dd57 updated the search code so that if you do a fuzzy search, it first
looks for exact matches and returns those if there are any. This had the desired
effect of letting people type exact accession numbers into the accession search
UI, but the summary statistics were still based on the fuzzy matches.

Update the summary endpoint so it uses the same "try an exact search first, and
fall back to fuzzy if that doesn't return results" logic so the summary stats will
match the detailed search results.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
@@ -4,20 +4,16 @@ import com.terraformation.backend.api.SeedBankAppEndpoint
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.default_schema.FacilityId
 import com.terraformation.backend.db.default_schema.OrganizationId
-import com.terraformation.backend.db.seedbank.AccessionId
 import com.terraformation.backend.db.seedbank.AccessionState
 import com.terraformation.backend.search.SearchFieldPrefix
-import com.terraformation.backend.search.SearchNode
-import com.terraformation.backend.search.SearchService
 import com.terraformation.backend.search.api.HasSearchNode
 import com.terraformation.backend.search.api.SearchNodePayload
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #765** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/765" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/765?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
```diff
@@ -68,17 +65,32 @@ class SummaryController(
   fun summarizeAccessionSearch(
       @RequestBody payload: SummarizeAccessionSearchRequestPayload
   ): SummarizeAccessionSearchResponsePayload {
+    val originalCriteria = payload.toSearchNode(accessionsPrefix)
+
+    // If there are fuzzy search criteria, first see if we get results by treating them as exact
```
karthikbtf: With these finer details.. should we create a summary service for testability?

karthikbtf (approved):
Looks good to me 👍 
One comment/question - not necessarily requesting changes.

src/main/kotlin/com/terraformation/backend/seedbank/api/SummaryController.kt
sgrimm: Yeah, I was being lazy here, but you're right. Will refactor this. We definitely want to stick with our "controllers are dumb wrappers around services and have so little logic that we're okay with minimal test coverage" approach which this definitely violates.

karthikbtf (approved):
Thanks for the changes. Looks great to me.

PR #766, opened 2022-10-28
SW-2003 Use jOOQ JSONB for arbitrary JSON objects

sgrimm:
In preparation for changing the default behavior of payload deserialization so
that it treats empty strings as null values, change the way we declare payload
fields that hold arbitrary client-supplied JSON objects. We will want such objects
to be preserved as-is, skipping the "treat empty strings as null" transformation,
which would be applied to `Map<String, Any?>` objects.

We could define our own class to hold arbitrary JSON objects, but since the values
will need to turn into jOOQ `JSONB` objects anyway, may as well just use that
class directly. In the payload classes, though, we'll refer to it by an alias both
to make the intent clearer and to reduce coupling between the API definition and
the database representation.

Update the user preferences API to use this new setup. This is purely an internal
change; it doesn't cause any client-visible behavior differences.


sgrimm:
* main
  * **PR #766** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/766" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/766?utm_source=stack-comment).

openapi.yaml-
```diff
@@ -0,0 +1,6541 @@
+openapi: 3.0.1
```
karthikbtf: openapi.yaml- got picked up

karthikbtf (approved):
looks good to me 👍 
an extraneous file got picked up by mistake I think

openapi.yaml-
sgrimm: Oops. Thanks.

karthikbtf (approved):

PR #770, opened 2022-10-31
SW-1951 Fix CSV upload job scheduling

sgrimm:
Commit 508c1f9f refactored the CSV upload code to extract common logic into a
superclass. Unfortunately, the refactor caused JobRunr to be unable to detect
which importer was enqueuing the asynchronous jobs to validate and import the
files.

Fix it by enqueuing the jobs explicitly from each subclass.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/importer/CsvImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/importer/CsvImporter.kt
@@ -18,13 +18,12 @@ import com.terraformation.backend.file.UploadStore
 import com.terraformation.backend.log.perClassLogger
 import java.io.InputStream
 import java.io.InputStreamReader
-import org.jobrunr.scheduling.JobScheduler
+import org.jobrunr.jobs.JobId
 import org.jooq.DSLContext
 
 abstract class CsvImporter(
     protected val dslContext: DSLContext,
     private val fileStore: FileStore,
-    private val scheduler: JobScheduler,
     private val uploadProblemsDao: UploadProblemsDao,
     private val uploadsDao: UploadsDao,
     private val uploadService: UploadService,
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #770** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/770" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/770?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/importer/CsvImporter.kt
```diff
@@ -68,7 +79,7 @@ abstract class CsvImporter(
     uploadStore.requireAwaitingAction(uploadId)
 
     dslContext.transaction { _ ->
-      scheduler.enqueue<CsvImporter> { importCsv(uploadId, overwriteExisting) }
```
karthikbtf, line 71: hoping this isn't a dumb question, would inferring the implementing class type work? something like KClass::class

karthikbtf (approved):
Thanks 👍

src/main/kotlin/com/terraformation/backend/importer/CsvImporter.kt
sgrimm: The underlying problem is that the "implementing" class that JobRunr detects is the class that contains the lambda function. The bug was that because the lambda was in `CsvImporter`, when JobRunr woke up and tried to run the job, it was trying to find a Spring bean of type `CsvImporter` and bombing out because there were three of them (the three importer classes) and no way to tell which one to use.

The class name in angle brackets is a bit of a red herring. It isn't actually used at runtime at all. Its presence (regardless of which class it specifies) is an input to the compiler's type inference and changes which overload of `enqueue()` ends up getting called. Kind of confusing, I know!

PR #773, opened 2022-11-02
SW-1890 Notifications for Nursery Workflows

charlie-tf:
- Generate notification events with daily task
  - create lookup method in BatchStore
- Add event listeners
  - Listen to event and insert application notification
  - Add notification type to database
  - Listen to event and send emails
- Create email templates

Local Testing (performed by running the service locally and setting the "daily" task cron to every 5 minutes):
image
image


src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -91,6 +91,16 @@ class Messages {
       NotificationMessage(
           title = "An accession has dried", body = "$accessionNumber has finished drying.")
 
+  fun nurseryBatchReadyNotification(
+      batchNumber: String,
+      facilityName: String
+  ): NotificationMessage =
+      NotificationMessage(
+          title = "$batchNumber has reached its scheduled ready by date.",
```
karthikbtf, line 99: I did not review the email copy. Should the title begin with `Seedlings batch`?

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryBatchEventData.kt
```diff
@@ -0,0 +1,11 @@
+package com.terraformation.backend.nursery.db
```
karthikbtf: should this be under the `model` package?

karthikbtf (commented):
Looks good to me.
A few comments. Re: email copy ignore my comment if it doesn't make sense.
We could add a test in here https://github.com/terraware/terraware-server/blob/main/src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt if possible.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
charlie-tf: this is per the PRD 🙂

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryBatchEventData.kt
charlie-tf: I can certainly move it there! that would make more sense

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
karthikbtf: got it, i added a comment in the PRD. this is fine for now.

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryBatchEventData.kt
karthikbtf: oh don't move it yet, if we have similar db models that aren't really persisted, this should be fine.

karthikbtf (approved):

sgrimm:
"SW-1980,1981" won't be detected as a list of two issue numbers in the SW project by Jira's GitHub integration. You'd want to pick just one of the two to include in the title and put the other one (with SW- prefix) in the body of the PR description.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -91,6 +91,16 @@ class Messages {
       NotificationMessage(
           title = "An accession has dried", body = "$accessionNumber has finished drying.")
 
+  fun nurseryBatchReadyNotification(
```
sgrimm: Minor naming nit: this should be `seedlingBatchReadyNotification` or `nurserySeedlingBatchReadyNotification` since the full term is "seedling batch" rather than "nursery batch."

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryBatchEventData.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.terraformation.backend.db.default_schema.SpeciesId
+import com.terraformation.backend.db.nursery.BatchId
+
+data class NurseryBatchEventData(
```
sgrimm: Can we directly use `NurserySeedlingBatchReadyEvent` instead of this? It doesn't seem like this class will ever be used for anything except constructing one of those events (that idea is even encoded into this class's name) so may as well skip the intermediate step.

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -74,6 +74,7 @@ VALUES (1, 'User Added to Organization', 1),
        (2, 'Facility Idle', 2),
        (3, 'Facility Alert Requested', 3),
        (6, 'Accession Scheduled to End Drying', 1),
+       (7, 'Nursery Seedling Batch Ready', 1),
```
sgrimm: I'm a little uneasy about reusing type IDs that used to exist but were deleted. Seems like this would run the risk of inadvertently matching existing values from before we removed an ID from this list. In theory we will always be careful to have a migration with `DELETE FROM notifications WHERE notification_type_id = 7` when we remove ID 7 from a list of type codes, but maybe we'll forget.

What do you think of using 15 instead of 7 for this?

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -218,6 +224,48 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
...
+    val speciesId = SpeciesId(100)
+    val nurseryName = "my nursery"
+
+    insertSpecies(speciesId)
+    insertBatch(
+        BatchesRow(
+            id = batchId,
+            batchNumber = batchNumber,
+            speciesId = speciesId,
+            facilityId = facilityId))
```
sgrimm, line 248: I don't think it will matter for purposes of this test, but this will be inserting a batch into a seed bank facility. No need to change anything here, just noting it for the future in case at some point we add an integrity constraint at the database level to enforce that the facility type is correct.

src/test/kotlin/com/terraformation/backend/nursery/db/batchStore/BatchStoreFetchEstimatedReadyTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.time.LocalDateTime
+import java.time.ZoneId
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+internal class BatchStoreFetchEstimatedReadyTest : BatchStoreTest() {
+  @Test
+  fun `returns rows within specified ready by date range`() {
+    val facilityId = FacilityId(1000)
+    insertFacility(id = facilityId, name = "baby plant nursery")
```
sgrimm: May as well set the facility type to `Nursery` here.

sgrimm (changes requested):
This looks good! A few minor things but it's pretty close.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
charlie-tf: Oh thank you for catching that. I was trying to be consistent with Nursery-Seedling-Batch everywhere but I missed this one.

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryBatchEventData.kt
charlie-tf: Seems like a reasonable change!

src/main/resources/db/migration/R__TypeCodes.sql
charlie-tf: Ah, I didn't realize that the gaps were due to deleted types. I will definitely change this!

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
charlie-tf: I can see about creating a separate facility 👍

sgrimm (approved):
Changes look good to me.

PR #775, opened 2022-11-03
SW-2161 Switch to more feature-rich geometry library

sgrimm:
In our initial stab at plant tracking, we weren't really doing anything with
geometry values on the server side, just storing them so the client could
retrieve them later on. We were therefore using a very simple library to work
with PostGIS values: basically just a type-safe wrapper that allowed navigating
their component parts.

For the next round of tracking work, we're going to need to be able to
manipulate the actual geometries, e.g., to validate that the plot boundaries in
an uploaded shapefile lie inside the site's boundaries.

Replace the PostGIS library with the recursively-named JTS Topology Suite,
which supports more sophisticated operations on geometry values.

Unfortunately, it wasn't possible to do this in such a way that the existing
test suites could remain unmodified; JTS has some behavior differences that
affect our tests.  For example, its validation logic detects polygons whose
edges intersect, and its text rendering of geometry values doesn't include the
PostGIS-specific extension that identifies the coordinate system.


sgrimm:
* main
  * **PR #775** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/775" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/775?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/db/GeometryDeserializerTest.kt
```diff
@@ -1,165 +0,0 @@
-package com.terraformation.backend.db
```
sgrimm, line 1: It turned out nearly all these cases were also covered by GeometrySerializerTest, which was added after this class. I added the couple of missing cases to that class and there's no more need for this one.

src/test/kotlin/com/terraformation/backend/db/GeometrySerializerTest.kt
```diff
@@ -74,56 +97,69 @@ internal class GeometrySerializerTest {
...
               "Polygon with long/lat SRID",
               """{"type":"Polygon","coordinates":[[[1,2,3],[2.1,1.1,0.1],[5,5,5],[1,2,3]]]}""",
-              "SRID=4326;POLYGON((1 2 3,2.1 1.1 0.1,5 5 5,1 2 3))"),
+              SRID.LONG_LAT,
+              "POLYGON((1 2 3,2.1 1.1 0.1,5 5 5,1 2 3))"),
           arguments(
               "MultiPolygon with long/lat SRID",
-              """{"type":"MultiPolygon","coordinates":[[[[1,2,3],[1.1,2.2,3.3],[0,0,0],[1,2,3]],[[0,-1,1],[1,0,1],[0,1,0],[0,-1,1]]]]}""",
-              "SRID=4326;MULTIPOLYGON(((1 2 3,1.1 2.2 3.3,0 0 0,1 2 3),(0 -1 1,1 0 1,0 1 0,0 -1 1)))"),
+              """{"type":"MultiPolygon","coordinates":[[[[1,2,3],[1,2.5,3],[0.0,2.5,3],[1,2,3]]],[[[11,2,13],[11,2.5,13],[10,2.5,13],[11,2,13]]]]}""",
```
sgrimm, line 125: The previous version of this multipolygon had bad geometry (it specified a hole that fell outside the bounds of the enclosing shape, and it had edges that crossed one another).

sgrimm:
The jOOQ binding part of this is likely not quite correct, though hopefully it's not too far off. We don't have any `GEOMETRY` columns in our database at the moment so I couldn't really write a test suite for it without a bunch of extra setup work, which seemed like a waste of time given I'm about to add the planting site tables which I can then write tests against.

sgrimm:
Docs for the library, if anyone is curious: http://locationtech.github.io/jts/jts-features.html

karthikbtf (approved):
Looks good to me, plan sounds good. Nice that jts implements the OpenGIS geometry model.

src/main/kotlin/com/terraformation/backend/db/GeometryDeserializer.kt
```diff
@@ -5,18 +5,18 @@ import com.fasterxml.jackson.core.JsonParser
...
+import org.locationtech.jts.geom.GeometryFactory
+import org.locationtech.jts.geom.PrecisionModel
+import org.locationtech.jts.io.ParseException
+import org.locationtech.jts.io.geojson.GeoJsonReader
+import org.locationtech.jts.operation.valid.IsValidOp
+
+/** Deserializes GeoJSON strings into JTS [Geometry] objects of the appropriate types. */
 class GeometryDeserializer : JsonDeserializer<Geometry>() {
+  private val geoJsonReader =
+      GeoJsonReader(GeometryFactory(PrecisionModel(PrecisionModel.FLOATING), SRID.LONG_LAT))
```
charlie-tf, line 18: well this (and the serializer) seems much simpler 🙂

charlie-tf (approved):
👍

PR #776, opened 2022-11-04
SW-2068 Use github action to automatically transition Jira issues at release

alexTerraformation:
Transition all released Jira issues to the "Released to Production" status

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,7 +127,7 @@ jobs:
         SSH_KEY: ${{ secrets[env.SSH_KEY_SECRET_NAME] }}
       run: ./.github/scripts/deploy.sh
 
-    - name: Jira Login
+    - name: Log into Jira
       if: env.TIER == 'PROD'
       uses: atlassian/gajira-login@master
       env:
@@ -142,10 +142,9 @@ jobs:
           grep -E 'SW-[0-9]+' -o | \
           sort -u > ./docs/jiralist.txt
 
-    - name: Transition Jiras
+    - name: Transition Jira Issues
       if: env.TIER == 'PROD'
       uses: terraware/gajira-transition-multiple@master
       with:
         issueList: ./docs/jiralist.txt
         transition: "Released to Production"
-  
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -128,12 +128,12 @@ jobs:
       run: ./.github/scripts/deploy.sh
 
     - name: Login
-        if: env.TIER == 'PROD'
-        uses: atlassian/gajira-login@master
-        env:
-          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
-          JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
-          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+      if: env.TIER == 'PROD'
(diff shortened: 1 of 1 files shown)


karthikbtf (approved):

.github/workflows/main.yml
```diff
@@ -126,3 +126,23 @@ jobs:
...
+      if: env.TIER == 'PROD'
+      uses: atlassian/gajira-login@master
+      env:
+        JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
+        JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+        JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+
+    - name: Create Jira Transition List
+      if: env.TIER == 'PROD'
+      run: curl https://terraware.github.io/terraware-web/unreleased.log | grep -E 'SW-[0-9]+' -o | sort -u > ./docs/jiralist.txt
```
sgrimm: Should this be looking at terraware-server instead of terraware-web?

Also, this would be easier to read if it were split into multiple lines or extracted to a script.

alexTerraformation: Good catch, bad copy/paste job

.github/workflows/main.yml
```diff
@@ -126,3 +126,26 @@ jobs:
         SSH_CONFIG: ${{ secrets[env.SSH_CONFIG_SECRET_NAME] }}
         SSH_KEY: ${{ secrets[env.SSH_KEY_SECRET_NAME] }}
       run: ./.github/scripts/deploy.sh
+
+    - name: Jira Login
```
sgrimm: I hesitate to even mention this because it feels nitpicky even to me, but all the other step names in the workflow start with verbs. Maybe change this to "Log into Jira" for consistency with the Docker Hub step?

.github/workflows/main.yml
```diff
@@ -126,3 +126,26 @@ jobs:
...
+        JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+
+    - name: Create Jira Transition List
+      if: env.TIER == 'PROD'
+      run: |
+        curl https://terraware.github.io/terraware-server/unreleased.log | \ 
+          grep -E 'SW-[0-9]+' -o | \
+          sort -u > ./docs/jiralist.txt
+
+    - name: Transition Jiras
```
sgrimm: Maybe "Transition Jira issues"?

.github/workflows/main.yml
```diff
@@ -126,3 +126,26 @@ jobs:
...
+      if: env.TIER == 'PROD'
+      uses: terraware/gajira-transition-multiple@master
+      with:
+        issueList: ./docs/jiralist.txt
+        transition: "Released to Production"
+
```
sgrimm: Huh, I wonder why GitHub is flagging this as not ending with a newline.

sgrimm (approved):
A couple cosmetic comments, but this generally seems fine.

It is unfortunate that Jira's GitHub integration already detects which exact commits are associated with which issues, but we have to go figure out the linkage ourselves from scratch. I guess there's no way to query that info from Jira's API?

Jira's integration is going to detect some links we won't, e.g., it supports mentioning multiple issues in a commit message. But those cases are pretty rare and this approach should cover 99% of our commits.

sgrimm (approved):

PR #789, opened 2022-11-10
SW-2179 Add endpoint to get Mapbox API token

sgrimm:
The `GET /api/v1/tracking/mapbox/token` endpoint will return a temporary Mapbox
API token that has permission to read map data.

It requires the server to be configured with a Mapbox API token that has the
`token:write` scope, e.g., in an `application.yaml` file:

```
terraware:
  mapbox:
    api-token: "YOUR_MAPBOX_API_TOKEN"
```

The implementation introduces the Ktor HTTP client library to the code base.
It is a Kotlin-native HTTP client that is a lot more convenient to use than the
Java 11 standard library HTTP client we've used previously. Existing code that
uses the Java client is left untouched here; it can be converted later.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
@@ -102,6 +102,9 @@ class TerrawareServerConfig(
     /** Configures how the server interacts with the Balena cloud service to manage sensor kits. */
     val balena: BalenaConfig = BalenaConfig(),
 
+    /** Configures how the server interacts with the Mapbox service. */
+    val mapbox: MapboxConfig = MapboxConfig(),
+
     /** Configures notifications processing. */
     val notifications: NotificationsConfig = NotificationsConfig(),
 ) {
@@ -242,6 +245,13 @@ class TerrawareServerConfig(
       const val BALENA_API_URL = "https://api.balena-cloud.com"
     }
   }
+
+  @ConstructorBinding
+  class MapboxConfig(
+      val apiToken: String? = null,
+      /** How long temporary Mapbox API tokens should last. */
+      @DefaultValue("30") val temporaryTokenExpirationMinutes: Long = 30,
+  )
 
   @ConstructorBinding
   class NotificationsConfig(
--- a/src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
@@ -1,0 +1,90 @@
+package com.terraformation.backend.tracking.mapbox
+
+import com.terraformation.backend.config.TerrawareServerConfig
+import com.terraformation.backend.log.perClassLogger
+import io.ktor.client.HttpClient
+import io.ktor.client.call.body
+import io.ktor.client.plugins.ClientRequestException
+import io.ktor.client.request.get
+import io.ktor.client.request.post
+import io.ktor.client.request.setBody
+import io.ktor.client.statement.bodyAsText
+import java.net.URL
+import java.time.Instant
+import java.time.InstantSource
+import java.time.temporal.ChronoUnit
+import javax.annotation.ManagedBean
+import kotlinx.coroutines.runBlocking
+
+@ManagedBean
+class MapboxService(
+    private val clock: InstantSource,
+    private val config: TerrawareServerConfig,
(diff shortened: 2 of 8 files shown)


sgrimm:
* main
  * **PR #789** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/789" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/789?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
```diff
@@ -0,0 +1,90 @@
...
+      } catch (e: ClientRequestException) {
+        log.error("Mapbox token request failed with HTTP ${e.response.status}")
+        log.info("Mapbox error response payload: ${e.response.bodyAsText()}")
+        throw MapboxRequestFailedException(e.response.status)
+      }
+    }
+  }
+
+  private fun mapboxUrl(endpoint: String): URL =
+      URL("https://api.mapbox.com/$endpoint?accss_token=${config.mapbox.apiToken}")
```
karthikbtf: `accss` intended?

src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
```diff
@@ -0,0 +1,90 @@
...
+  val enabled: Boolean = config.mapbox.apiToken != null
+
+  /**
+   * The Mapbox username of the account that owns the configured API token. Some Mapbox APIs require
+   * the username to be included explicitly in the request.
+   *
+   * We fetch this lazily the first time it's accessed by requesting information about our access
+   * token, which includes the username.
+   */
+  val username: String by lazy {
```
karthikbtf: should this be private?

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -243,6 +246,13 @@ class TerrawareServerConfig(
     }
   }
 
+  @ConstructorBinding
+  class MapboxConfig(
+      val apiToken: String? = null,
```
karthikbtf, line 251: would we need a way to override the token from the admin ui - in case token is compromised? i guess we can always re-roll the service.

karthikbtf (approved):
Thank you so much!
Some questions - mostly naive ones.

src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
```diff
@@ -0,0 +1,90 @@
...
+      throw MapboxNotConfiguredException()
+    }
+
+    val url = mapboxUrl("tokens/v2/$username")
+
+    val expirationTime =
+        clock.instant().plus(config.mapbox.temporaryTokenExpirationMinutes, ChronoUnit.MINUTES)
+    val requestPayload =
+        TemporaryTokenRequestPayload(
+            expires = expirationTime, scopes = listOf("styles:tiles", "styles:read", "fonts:read"))
```
karthikbtf, line 61: not sure why this comment didn't come through earlier.
would `styles:tiles` cover the uploaded polygon tileset?

sgrimm: Yes, that's my assumption, though I haven't gotten as far as actually creating a tileset yet so it's just a semi-educated guess. It's possible this list of scopes will need to change as we figure things out.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
sgrimm: I think restarting the server is fine. Takes a minute or two and we'd want to do it anyway to change the default in the environment variables even if there were a way to do a temporary override at runtime.

Maybe down the road we'll want to move some of these settings to the database or to a configuration store like Consul so we can change them on the fly.

src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
sgrimm: Thanks. I was testing the error handling and forgot to undo this change.

karthikbtf (approved):
🎉

PR #790, opened 2022-11-10
allow generating types for tracking api

karthikbtf:
- needed for FE api types generation


sgrimm (approved):
Oops! Good catch. I (obviously) completely neglected to look at the OpenAPI schema for this.

PR #800, opened 2022-11-14
Return mapbox tempory token expiration info

karthikbtf:
- just realized client needs this to decide when to refetch tokens
- didn't want to assume 30 minutes client-side

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/MapboxController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/MapboxController.kt
@@ -42,4 +42,5 @@ class MapboxController(
   }
 }
 
-data class GetMapboxTokenResponsePayload(val token: String) : SuccessResponsePayload
+data class GetMapboxTokenResponsePayload(val token: MapboxService.GeneratedTemporaryToken) :
+    SuccessResponsePayload
--- a/src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
@@ -47,7 +47,7 @@ class MapboxService(
   }
 
   /** Generates a temporary API token. */
-  fun generateTemporaryToken(): String {
+  fun generateTemporaryToken(): GeneratedTemporaryToken {
     if (!enabled) {
       throw MapboxNotConfiguredException()
     }
@@ -65,7 +65,7 @@ class MapboxService(
         val responsePayload: TemporaryTokenResponsePayload =
             httpClient.post(url) { setBody(requestPayload) }.body()
 
-        responsePayload.token
+        GeneratedTemporaryToken(responsePayload.token, requestPayload.expires.toEpochMilli())
       } catch (e: ClientRequestException) {
         log.error("Mapbox token request failed with HTTP ${e.response.status}")
         log.info("Mapbox error response payload: ${e.response.bodyAsText()}")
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/terraformation/backend/tracking/api/MapboxController.kt
```diff
@@ -42,4 +42,5 @@ class MapboxController(
   }
 }
 
-data class GetMapboxTokenResponsePayload(val token: String) : SuccessResponsePayload
+data class GetMapboxTokenResponsePayload(val token: MapboxService.GeneratedTemporaryToken) :
```
karthikbtf, line 45: current response looks like:
```
{
  token: {
    token: string;
    tokenExpirationMs: number;
  }
}
```
open to renaming if needed

sgrimm:
I assumed the client would fetch a new token when it got HTTP 401 from Mapbox’s API.

karthikbtf:
> I assumed the client would fetch a new token when it got HTTP 401 from Mapbox’s API.

Got it, client does not make direct requests to mapbox api. Uses a mapbox component with some options - one of which is the token. There must be some error feedback let me check.

src/main/kotlin/com/terraformation/backend/tracking/api/MapboxController.kt
sgrimm, line 45: It might be a little silly for this simple case, but can we decouple the internal data structure from the API payload here? Practically speaking, turn this into something along the lines of
```
data class GetMapboxTokenResponsePayload(
    val token: String,
    val expirationMs: Long,
) {
    constructor(model: GeneratedTemporaryToken) : this(model.token, model.expirationMs)
}
```
That way if the internal implementation changes, the API payload will be unaffected.

karthikbtf:
> > I assumed the client would fetch a new token when it got HTTP 401 from Mapbox’s API.
> 
> Got it, client does not make direct requests to mapbox api. Uses a mapbox component with some options - one of which is the token. There must be some error feedback let me check.

There is an error event callback which captures all errors, including 401.

src/main/kotlin/com/terraformation/backend/tracking/mapbox/MapboxService.kt
```diff
@@ -87,4 +87,6 @@ class MapboxService(
   data class RetrieveTokenResponsePayload(val code: String, val token: Token) {
     data class Token(val user: String)
   }
+
+  data class GeneratedTemporaryToken(val token: String, val tokenExpirationMs: Long)
```
sgrimm, line 91: The second field here won't be what you're looking for, I think, though I can't say for sure without seeing the client side. This is an `Instant`, not a `Duration`, that is, it is the number of milliseconds since 1/1/1970 rather than the number of future milliseconds until expiration.

karthikbtf: Was trying to capture the time at which the token expires.
For a token generated at  Mon Nov 14 9:32am, client saw expiration ms as 1668448953057, which is  Mon Nov 14 2022 10:02:33 GMT-0800 (Pacific Standard Time)
I closed this ticket, like you had thought - client can capture mapbox 401s to regenerate the token.

sgrimm: Ah, okay, sorry for the comment noise, then. I wasn't sure if you had intended this to be an offset or an absolute time. But glad it turned out to not be needed anyway.

PR #805, opened 2022-11-16
SW-2225 Ignore fractional remaining seeds after withdrawal

sgrimm:
When the user creates a nursery transfer or a viability test, the withdrawal
quantity is always denominated in seeds even if the accession is weight-based.
We estimate the withdrawal weight using the subset count and weight, and
subtract that estimated weight from the accession's remaining weight.

This is normally fine, but when the user is trying to withdraw all the
remaining seeds from the accession, the withdrawal ends up either failing (if
the estimated weight ended up being slightly greater than the accession's
remaining weight) or leaving the accession active even though there are no more
seeds in reality (if the estimated weight was slightly less than the
accession's remaining weight).

Add logic to handle this case: if the most recent withdrawal from a
weight-based accession is count-based, and the remaining weight would be less
than the estimated weight of one seed (positive or negative), round the
accession's remaining quantity to zero.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
@@ -226,9 +226,11 @@ data class AccessionModel(
             remaining != existing.remaining) {
           remaining
         } else {
-          newWithdrawals
-              .filter { it.isAfter(latestObservedTime) }
-              .fold(latestObservedQuantity) { runningRemaining, withdrawal ->
+          val withdrawalsAfterObservation = newWithdrawals.filter { it.isAfter(latestObservedTime) }
+          val observedMinusWithdrawals =
+              withdrawalsAfterObservation.fold(latestObservedQuantity) {
+                  runningRemaining,
+                  withdrawal ->
                 val withdrawn = withdrawal.withdrawn
                 if (withdrawn != null) {
                   runningRemaining -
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #805** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/805" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/805?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
```diff
@@ -237,6 +239,23 @@ data class AccessionModel(
...
+          // always measured in seeds) then the remaining weight might be slightly more or less than
+          // zero thanks to imprecision of the subset weight measurement. We will end up with an
+          // observedMinusWithdrawals value less than the weight of one seed. Treat the withdrawal
+          // as having emptied out the accession in that case.
+          if (observedMinusWithdrawals.units != SeedQuantityUnits.Seeds &&
+              withdrawalsAfterObservation.isNotEmpty() &&
+              withdrawalsAfterObservation.last().withdrawn?.units == SeedQuantityUnits.Seeds &&
+              subsetCount != null &&
+              subsetWeightQuantity != null &&
+              observedMinusWithdrawals.abs() * subsetCount < subsetWeightQuantity) {
```
charlie-tf, line 254: This looks like it will round to zero if the remaining quantity (in seeds) is less than one subset, rather than one seed. Am I understanding that correctly?

sgrimm: Multiplying by the subset count should make it compare the weight of one seed.

If it helps: originally I had this as
```
observedMinusWithdrawals.abs() < (subsetWeightQuantity / subsetCount)
```
but that was problematic because the division operation could produce very small values if the subset count was large, and they'd get rounded to zero and the comparison would fail. So I multiplied both sides by `subsetCount` to turn the division into a multiplication.

charlie-tf: Oh, that makes sense now 😄 Thanks!

charlie-tf (approved):
All makes sense now. Nice!

karthikbtf (approved):
Looks good to me, thanks for the explanation 👍

PR #806, opened 2022-11-17
SW-2225 Clean up inconsistent accession quantities

sgrimm:
Thanks to a combination of bugs in the v1 accession code and problems with the
v1->v2 migration, a handful of accessions ended up with withdrawals whose
quantities add up to more than the latest observed quantity. It's thus impossible
to do anything with the accessions because the quantity recalculation logic always
catches the problem and bails out.

Fix up the existing accessions in three ways:

* Set the correct withdrawal purpose for viability test withdrawals and add a
  database constraint to make sure it's always properly set going forward.
* Remove withdrawals that were created after the most recent quantity observation
  but don't have associated quantity history entries. These are all duplicates
  of withdrawals that _do_ have history entries, and they are causing withdrawal
  quantities to be double-counted.
* Make sure the above doesn't cause the remaining quantities (which users have
  already seen) to change suddenly, by calculating a latest observed quantity
  that causes the withdrawals to add up to the currently-displayed remaining
  quantity. (An alternative would have been to set the latest observed quantity
  to the current quantity, but that'd cause a behavior difference if a user
  edited an existing withdrawal's quantity.)

Once the cleanup has run, its class can be removed from the code base.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/QuantityRepairer.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/QuantityRepairer.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.seedbank.db
 
+import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.customer.model.SystemUser
 import com.terraformation.backend.db.seedbank.AccessionId
 import com.terraformation.backend.db.seedbank.SeedQuantityUnits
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #806** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/806" target="_blank">Graphite</a>  👈
    * **PR #807** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/807" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/806?utm_source=stack-comment).

karthikbtf (approved):
Looks good to me based on the explanation.
There seems to be a postgres connection issue in the openapi docs task of the build.

sgrimm:
Oops, forgot to make this conditional on there being a database connection. The event listener is causing Spring to fire up this new class even when lazy bean evaluation is enabled.

karthikbtf (approved):

PR #808, opened 2022-11-17
SW-2288 Fix multiline command in github action yaml

alexTerraformation:
Use yaml folded style which is supported by github actions to break long command across multiple lines.


sgrimm (approved):
Nice, that's cleaner.

charlie-tf (approved):
nice!

PR #827, opened 2022-11-23
SW-2003 Use ArbitraryJsonObject for device settings

sgrimm:
Device settings are arbitrary JSON objects defined by the device manager.
Previously, we were unpacking them because we needed to populate a setting for
backward compatibility with old device manager versions. Now that the device
managers in the field have been updated to not use that setting any more, we
can treat the settings as opaque values and there's no longer any need to
represent them as `Map` objects internally.

In addition to removing unnecessary logic, this is a prerequisite to changing
the JSON deserializer configuration to treat empty strings as null; in the
absence of this change, doing so would have prevented the device manager from
including empty strings as setting values, which would break the "device manager
can put whatever it wants in the settings" API contract.

This doesn't change the API schema, but it does change the behavior in that the
server no longer inserts a `diagnosticMode` value into the settings object.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/api/DeviceController.kt
@@ -1,10 +1,9 @@
 package com.terraformation.backend.device.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.module.kotlin.readValue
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #827** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/827" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/827?utm_source=stack-comment).

karthikbtf (approved):

PR #840, opened 2022-11-28
Simplify OpenAPI customization

sgrimm:
Remove the explicit list of arbitrary-JSON-object payload fields whose
`additionalProperties` values need to be cleared, in favor of detecting them
dynamically during the scan of all the payload fields we're doing anyway.

Move the GeoJSON schema configuration to a helper function that lives alongside
the schema configuration classes.

There's no change in the intended behavior here, but this does fix the definitions
of a couple of payload fields that weren't being properly adjusted previously.

PR #839, opened 2022-11-28
SW-2003 Use ArbitraryJsonObject for automation settings

sgrimm:
In preparation for changing the JSON deserialization settings to treat empty
strings as null values, update the automation payloads' settings fields to use
`ArbitraryJsonObject` instead of `Map<String, Any?>`.

Client-visible behavior remains the same as before.

PR #841, opened 2022-11-28
SW-2003 Treat blank strings in payloads as null

sgrimm:
Add a custom JSON deserialization module for string values that maps blank
(all-whitespace) and empty strings to null in request payloads. For example,

    {
      "a": "b",
      "c": "   "
    }

becomes equivalent to

    {
      "a": "b",
      "c": null
    }

This change means that frontend code no longer needs to check for blank form field
values; they'll be treated as missing values on the server side.

The previous behavior (treating nulls as distinct from empty strings) had an
undesirable side effect in the web app: searching for missing values using the
search API by passing in `null` as the search value would sometimes fail to return
expected results because the field in question had a non-null value that happened
to be an empty string.

Clean up all the existing blank values and add database constraints to guarantee
that we don't write any in the future.


sgrimm:
* main
  * **PR #839** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/839" target="_blank">Graphite</a> 
    * **PR #840** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/840" target="_blank">Graphite</a> 
      * **PR #841** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/841" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/841?utm_source=stack-comment).

src/main/resources/db/migration/V163__RemoveBlankStrings.sql
```diff
@@ -0,0 +1,148 @@
...
+ALTER TABLE seedbank.accessions ADD CHECK ("number" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.accessions ADD CHECK ("processing_notes" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.bags ADD CHECK ("bag_number" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.viability_tests ADD CHECK ("notes" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.viability_tests ADD CHECK ("staff_responsible" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.withdrawals ADD CHECK ("destination" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.withdrawals ADD CHECK ("notes" NOT SIMILAR TO ' *');
+ALTER TABLE seedbank.withdrawals ADD CHECK ("staff_responsible" NOT SIMILAR TO ' *');
+ALTER TABLE tracking.planting_sites ADD CHECK ("description" NOT SIMILAR TO ' *');
+ALTER TABLE tracking.plantings ADD CHECK ("notes" NOT SIMILAR TO ' *');
```
karthikbtf, line 148: Compiling this list must have been rather painstaking, thanks for doing this.

karthikbtf (approved):
👍

src/main/resources/db/migration/V163__RemoveBlankStrings.sql
sgrimm: Not too painstaking, happily -- see the comment at the top of the migration file to see how I did it.

PR #843, opened 2022-11-29
Ignore null fuzzy search values

sgrimm:
Commit fdd126a1 changed the behavior of the search API subtly: previously, you
could do a fuzzy search for an empty string to match all non-null values, but
now those empty strings get converted to null during payload deserialization
and the resulting searches change from matching non-null values to only matching
nulls.

The web app passes in empty-string fuzzy search values when the user hasn't
entered a value to search for. In that scenario, the previous behavior was also
wrong: we'd want to ignore the search criterion entirely and return both null
and non-null values.

Update the search implementation to ignore fuzzy search criteria if one or more
of the values is null (which, thanks to commit fdd126a1, could happen if the
client passed in an empty string).

Searching for non-null values is still possible by wrapping an exact search for
null with a `not` operation.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
@@ -105,22 +105,22 @@ data class FieldNode(
   override fun toCondition(): Condition {
     val conditions = field.searchField.getConditions(this)
     return when {
-      type == SearchFilterType.Fuzzy && values.any { it == null } -> DSL.trueCondition()
+      isFuzzySearchForNull -> DSL.trueCondition()
       conditions.size == 1 -> conditions[0]
       else -> DSL.and(conditions)
     }
   }
 
   override fun referencedSublists(): Set<SublistField> {
-    return if (field.searchField is AliasField) {
-      field.searchField.targetPath.sublists.toSet()
-    } else {
-      field.sublists.toSet()
+    return when {
+      isFuzzySearchForNull -> emptySet()
+      field.searchField is AliasField -> field.searchField.targetPath.sublists.toSet()
+      else -> field.sublists.toSet()
     }
   }
 
   override fun toExactSearch(): FieldNode {
-    return if (type == SearchFilterType.Fuzzy && values.none { it == null }) {
+    return if (type == SearchFilterType.Fuzzy && !isFuzzySearchForNull) {
       FieldNode(field, values)
     } else {
       this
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #843** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/843" target="_blank">Graphite</a>  👈
    * **PR #844** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/844" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/843?utm_source=stack-comment).

sgrimm:
Going to clean this up slightly; update coming.

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
```diff
@@ -116,7 +120,7 @@ data class FieldNode(
   }
 
   override fun toExactSearch(): FieldNode {
-    return if (type == SearchFilterType.Fuzzy) {
+    return if (type == SearchFilterType.Fuzzy && values.none { it == null }) {
```
karthikbtf: should this filter null values and use the non-empty result instead?

sgrimm: Ah, yeah, that's a good idea. Thanks.

karthikbtf: i think i may have misread this line, will look at the updates. my comment doesn't seem right.

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
```diff
@@ -126,6 +130,9 @@ data class FieldNode(
   override fun toString(): String {
     return "FieldNode($field $type [${values.joinToString()}])"
   }
+
+  private val isFuzzySearchForNull: Boolean
+    get() = type == SearchFilterType.Fuzzy && values.any { it == null }
```
karthikbtf, line 135: should we consider an empty values list as search for null?

karthikbtf (approved):
Looks good to me, I think I'm understanding this better now.

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
sgrimm: Empty values lists aren't legal in the API.

sgrimm: Though I guess right now that's only enforced in the controller, and there also ought to be a check for it in the implementation. Will follow up with a separate PR for that.

karthikbtf: Got it 👍

PR #847, opened 2022-12-02
Reduce boilerplate in PermissionRequirementsTest

sgrimm:
Nearly all the test cases in PermissionRequirementsTest check for behavior that
follows one of two patterns depending on whether they're checking a write
permission or a read permission. The only things that vary across the read
tests and across the write tests are

* The method under test
* The type of the ID of the permission's target
* The type of the exception that's thrown if the user doesn't have read access
  (this is always the same for a given ID type)
* The TerrawareUser method that controls whether or not the user has permission

Use lazy evaluation and extension methods to make a mini-DSL for these tests,
collapsing the vast majority of them to one-liners.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
@@ -195,153 +195,120 @@ internal class PermissionRequirementsTest : RunsAsUser {
   }
 
   @Test
-  fun addOrganizationUser() {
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #847** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/847" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/847?utm_source=stack-comment).

karthikbtf (approved):
Wow, this is pretty neat, nifty Kotlin! And this is finally tackled.. 🎉

PR #848, opened 2022-12-02
Ignore blank or null collector names

sgrimm:
Commit fdd126a1 inadvertently broke accession submission from the mobile app in
cases where the user didn't indicate who collected the seeds. In that case, rather
than submitting an empty list of collector names, the client is submitting a list
with a single collector name that happens to be a zero-length string. The server
is now treating that as a null, and the accession creation fails because collector
names are non-nullable in the database.

Update the code to automatically ignore null or empty collector names, and add a
database migration to remove the collectors with zero-length names that were
inserted into the database previously.


joshbuchea (approved):

joshbuchea:
I approved, but admit it's a language I'm unfamiliar with 😅

charlie-tf (approved):
Looks good to me!

PR #851, opened 2022-12-06
SW-1054 Support localhost HTTP redirects

sgrimm:
Attempt to fix the problem with the login flow that prevented us from pointing
local web app dev environments at the staging or production server instances.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/HttpLocalhostRedirectConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/HttpLocalhostRedirectConfig.kt
@@ -1,0 +1,109 @@
+package com.terraformation.backend.auth
+
+import javax.servlet.http.HttpServletRequest
+import org.keycloak.adapters.AdapterTokenStore
+import org.keycloak.adapters.KeycloakDeployment
+import org.keycloak.adapters.OAuthRequestAuthenticator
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #851** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/851" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/851?utm_source=stack-comment).

sgrimm:
Tested this by setting up a local terraware-server instance to listen for HTTPS requests using a different hostname, and setting the FE dev environment to proxy to that. I believe my setup should have been an adequate simulation of staging, but I didn't run behind an AWS load balancer so there may be other subtleties.

karthikbtf (approved):
Thanks for doing this.
Should we disallow this on production?

sgrimm:
The original request was to let people use this to try to reproduce FE bugs in either staging or prod, so I think we have to allow it in both.

sgrimm:
From a security point of view it doesn't hurt to allow it in prod; it's already the case that anyone with a valid login session can call any of our API endpoints with whatever arguments they want.

sgrimm:
Sadly, this didn't work on staging, so clearly my local setup wasn't close enough to the real thing. It gets one step closer than before but still generates a bogus redirect URL.

PR #852, opened 2022-12-06
SW-1054 Support localhost HTTP redirects, round two

sgrimm:
The hack in commit 1e987c94 failed to solve the problem in the actual staging
environment; the port number of the original request didn't make it through
the AWS load balancer plus the Nginx proxy. Try a different approach that uses
the redirect URL that's passed in by the client.


sgrimm:
* main
  * **PR #852** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/852" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/852?utm_source=stack-comment).

sgrimm:
This time I tested using a real AWS load balancer pointing at an Nginx proxy on an EC2 instance, which is how the real staging/prod environments work.

karthikbtf (approved):
🎉

PR #853, opened 2022-12-07
SW-2043 Log request and response payloads

sgrimm:
To make it easier to troubleshoot problems we discover during QA testing, add the
ability to log request and response bodies for a subset of users.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.auth
 
+import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.customer.model.IndividualUser
 import com.terraformation.backend.log.perClassLogger
 import io.ktor.http.ContentType
@@ -29,7 +30,9 @@ import org.springframework.web.util.ContentCachingResponseWrapper
  * Payloads are logged using [MDC] rather than included in the message text so that they show up as
  * discrete values in structured logs without requiring extra parsing.
  */
-class RequestResponseLoggingFilter(private val requestLogEmailRegex: Regex) : Filter {
+class RequestResponseLoggingFilter(
+    private val requestLogConfig: TerrawareServerConfig.RequestLogConfig
+) : Filter {
   /** Log up to this many bytes of each payload. */
   private val maxPayloadSize = 10000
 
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #853** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/853" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/853?utm_source=stack-comment).

sgrimm:
Once this lands, I'll update the staging configuration to set this new option to `.*@terraformation.com` so we log everything during internal testing.

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
```diff
@@ -0,0 +1,95 @@
...
+          wrappedResponse.copyBodyToResponse()
+        }
+      }
+    } else {
+      chain.doFilter(request, response)
+    }
+  }
+
+  private fun mdcPut(key: String, value: String?) {
+    if (value != null) {
```
karthikbtf, line 82: might allowing null values actually help ... depending on the debugging context..

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
```diff
@@ -0,0 +1,95 @@
...
+        chain.doFilter(wrappedRequest, wrappedResponse)
+      } finally {
+        val oldMdc = MDC.getCopyOfContextMap()
+        try {
+          mdcPut("request", payload(wrappedRequest.contentType, wrappedRequest.contentAsByteArray))
+          mdcPut(
+              "response", payload(wrappedResponse.contentType, wrappedResponse.contentAsByteArray))
+          mdcPut("email", user.email)
+          mdcPut("method", wrappedRequest.method)
+          mdcPut("uri", wrappedRequest.requestURI)
```
karthikbtf, line 68: would we need an ignore list or a regex for uri, in case it gets spammy

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
```diff
@@ -0,0 +1,95 @@
...
+
```
karthikbtf, line 69: mdcPut the email regex for audit purposes?

karthikbtf (approved):
👍 nice

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
sgrimm: I did it this way originally but I didn't think having `request=null` on all the requests without bodies (GETs and one or two POSTs that don't require payloads) was too useful. What use case do you have in mind where it'd be useful to have that?

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
sgrimm: How about logging it once at start time, since it will never change over the lifetime of a server instance? I don't think we need it logged on every request.

What would you use it for, given that our server configuration is version-controlled (meaning there's already a single source of truth for the value)?

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
karthikbtf: Mostly corner cases like a post that requires a payload but one wasn't provided.
Or null responses when we'd expect to see a non-null response.
I didn't think about the spammy case of GETs and non-entity POSTs.. yes, we certainly don't want those.

src/main/kotlin/com/terraformation/backend/auth/RequestResponseLoggingFilter.kt
karthikbtf: Since we are logging emails, just thought might be good to also identify the logging criteria for emails. I guess server configuration works.

karthikbtf: I meant to say that our version controlled server configuration works, ignore my comment on logging the regex.

karthikbtf (approved):
👍 Looks good to me, useful feature.

PR #862, opened 2022-12-12
Fix incorrect OpenAPI schema for search payloads

sgrimm:
The OpenAPI schema for search criteria was incorrect in two respects:

1. It incorrectly mapped the `or` operation to the payload for the `not`
   operation.
2. The payload for the `or` operation didn't include the `children` property.

The second item is due to a bug in the OpenAPI schema generator, which is
triggered when two payload classes in the same type hierarchy have an array
property of the same name with identical `@ArraySchema` annotations. Work around
it by making the annotations non-identical.

This didn't affect the web app thanks to the fact that its build process
overwrites the search payload definitions to avoid code generator bugs related to
circular references. But we should still generate correct schema definitions.


sgrimm:
* main
  * **PR #862** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/862" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/862?utm_source=stack-comment).

karthikbtf (approved):
👍 thanks for catching this

PR #870, opened 2022-12-21
SW-1307 Improve performance of duplicate timeseries values

sgrimm:
If an incoming timeseries values request has a lot of entries that are already in
the database, detecting the duplicates is currently pretty slow and leads to
request timeouts. Check for duplicates explicitly by querying for all of the
timestamps at once rather than relying on individual INSERT queries failing.

For a request whose values are entirely duplicates, this will result in the
server doing a single SELECT query rather than a ton of failing INSERTs.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
@@ -3,6 +3,7 @@ package com.terraformation.backend.device.db
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.TimeseriesNotFoundException
+import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.default_schema.DeviceId
 import com.terraformation.backend.db.default_schema.TimeseriesId
 import com.terraformation.backend.db.default_schema.TimeseriesType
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #870** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/870" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/870?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
```diff
@@ -238,6 +239,24 @@ class TimeseriesStore(private val clock: Clock, private val dslContext: DSLConte
...
+  fun checkExistingValues(
+      timeseriesId: TimeseriesId,
+      timestamps: Collection<Instant>
+  ): Set<Instant> {
+    return with(TIMESERIES_VALUES) {
+      dslContext
+          .select(CREATED_TIME)
+          .from(TIMESERIES_VALUES)
+          .where(TIMESERIES_ID.eq(timeseriesId))
+          .and(CREATED_TIME.`in`(timestamps))
```
karthikbtf, line 255: does this get optimized internally because of the nature of timestamps/range?

karthikbtf (approved):
Looks good to me 👍

src/main/kotlin/com/terraformation/backend/device/db/TimeseriesStore.kt
sgrimm: You mean on the database side? Not exactly. It's fast because there is a unique index on `(timeseries_id, created_time)` which means the database can run this query purely by scanning the index, no need to read the table data.

But that's also true of the duplicate key detection (it can detect duplicates using the index). What really makes this faster than the previous code is the fact that we're querying all the timestamps in a single query, rather than a separate query for each one. One database round-trip vs. 1000 of them, basically.

PR #871, opened 2022-12-22
SW-2602 Add time zones to database schema

sgrimm:
Create a new table to hold the list of valid time zone names, and add code to
keep it in sync with the Java standard library's list of time zones (which is
taken from the IANA time zone database).

No API changes here, just database changes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/TimeZoneConverter.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/TimeZoneConverter.kt
@@ -1,0 +1,16 @@
+package com.terraformation.backend.db
+
+import java.time.ZoneId
+import org.jooq.impl.AbstractConverter
+
+/**
+ * Converts text values from the database to and from ZoneId objects. This is for type safety, so we
+ * don't treat arbitrary string values as zone IDs.
+ *
+ * This is referenced in generated database classes.
+ */
+class TimeZoneConverter :
+    AbstractConverter<String, ZoneId>(String::class.java, ZoneId::class.java) {
+  override fun from(databaseObject: String?): ZoneId? = databaseObject?.let { ZoneId.of(it) }
+  override fun to(userObject: ZoneId?): String? = userObject?.id
+}
--- a/src/main/kotlin/com/terraformation/backend/db/TimeZoneConverter.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/TimeZoneConverter.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.db
 
+import java.time.DateTimeException
 import java.time.ZoneId
 import org.jooq.impl.AbstractConverter
 
@@ -11,6 +12,13 @@ import org.jooq.impl.AbstractConverter
  */
 class TimeZoneConverter :
     AbstractConverter<String, ZoneId>(String::class.java, ZoneId::class.java) {
+  /**
+   * Converts a time zone name to a ZoneId.
+   *
+   * @throws DateTimeException The zone name wasn't found in the list of zones recognized by the
+   * java.time package.
+   */
   override fun from(databaseObject: String?): ZoneId? = databaseObject?.let { ZoneId.of(it) }
+
   override fun to(userObject: ZoneId?): String? = userObject?.id
 }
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #872** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/872" target="_blank">Graphite</a> 
    * **PR #871** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/871" target="_blank">Graphite</a>  👈
      * **PR #873** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/873" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/871?utm_source=stack-comment).

sgrimm:
The failing test, at first glance, seems to have nothing to do with this change; will investigate further before merging.

src/main/kotlin/com/terraformation/backend/db/TimeZonePopulator.kt
```diff
@@ -0,0 +1,55 @@
...
+    if (valuesToDelete.isNotEmpty()) {
+      val timeZonesDeleted =
+          dslContext
+              .deleteFrom(TIME_ZONES)
+              .where(TIME_ZONES.TIME_ZONE.`in`(valuesToDelete))
+              .execute()
+
+      log.info("Deleted $timeZonesDeleted time zones")
+    }
+  }
```
karthikbtf, line 49: log no updates?

src/test/kotlin/com/terraformation/backend/db/TimeZonePopulatorTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        TimeZonesRow(validTimeZone),
+        timeZonesDao.fetchOneByTimeZone(validTimeZone),
+        "Time zone should have been inserted")
+  }
+
+  @Test
+  fun `deletes time zones that are no longer in the Java time zones list`() {
+    timeZonesDao.insert(TimeZonesRow(invalidTimeZone))
+
+    insertUser(userId, timeZone = invalidTimeZone)
```
karthikbtf: just curious why postgres doesn't complain on this.

karthikbtf (approved):
Looks good to me!!

src/main/kotlin/com/terraformation/backend/db/TimeZonePopulator.kt
karthikbtf: i guess just looking for a way to see that the populator ran

sgrimm: That's a good idea. Updated this to always log how many entries were inserted and deleted.

src/test/kotlin/com/terraformation/backend/db/TimeZonePopulatorTest.kt
sgrimm: Which part? You mean inserting the invalid time zone? It's because from Postgres's point of view, the time zone is just a string value with no particular meaning.

sgrimm:
@karthikbtf If you could take another look at this (no rush), I decided I should do this in a more type-safe way.

src/test/kotlin/com/terraformation/backend/db/TimeZonePopulatorTest.kt
karthikbtf: Thanks for the clarification. Yes, makes sense now.

karthikbtf (approved):

karthikbtf:
> @karthikbtf If you could take another look at this (no rush), I decided I should do this in a more type-safe way.

Oops, I didn't refresh my view, will do so now.

src/main/kotlin/com/terraformation/backend/db/TimeZoneConverter.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.jooq.impl.AbstractConverter
+
+/**
+ * Converts text values from the database to and from ZoneId objects. This is for type safety, so we
+ * don't treat arbitrary string values as zone IDs.
+ *
+ * This is referenced in generated database classes.
+ */
+class TimeZoneConverter :
+    AbstractConverter<String, ZoneId>(String::class.java, ZoneId::class.java) {
```
karthikbtf, line 14: mention that ZoneId.of throws an exception if the region is not found?

karthikbtf (approved):
The latest changes (zone id converter / type safety + log message) looks good to me.

karthikbtf (approved):
Thank you!!

PR #872, opened 2022-12-22
Fix failing BatchStore test

sgrimm:
The test was assuming the results of a database query would be returned in a
specific order, but the function under test does not guarantee ordering.

Update the test to ignore what order the results are in.

PR #873, opened 2022-12-22
Simplify non-nullable jOOQ queries

sgrimm:
Use the new `asNonNullable()` extension function to remove `filterNotNull()` calls
on jOOQ queries that were only there to inform Kotlin that the results were
non-nullable. This is really just for conciseness; there will be a performance
gain from not having to copy the list, but it'll be miniscule.

In a couple places we were doing `fetch().get(FIELD)`; simplify those to
`fetchOne(FIELD)`.

No behavior change here.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -261,8 +261,7 @@ class AccessionStore(
                         .set(TOTAL_VIABILITY_PERCENT, accession.totalViabilityPercent)
                         .set(TREES_COLLECTED_FROM, accession.numberOfTrees)
                         .returning(ID)
-                        .fetchOne()
-                        ?.get(ID)!!
+                        .fetchOne(ID)!!
                   }
 
               if (accession.remaining != null) {
--- a/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
@@ -100,7 +100,6 @@ class GbifStore(private val dslContext: DSLContext) {
         .orderBy(firstWordSortPosition, GBIF_NAMES.NAME)
         .limit(maxResults)
         .fetchInto(GbifNamesRow::class.java)
-        .filterNotNull()
   }
 
   /**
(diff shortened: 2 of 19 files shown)


sgrimm:
* main
  * **PR #872** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/872" target="_blank">Graphite</a> 
    * **PR #871** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/871" target="_blank">Graphite</a> 
      * **PR #873** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/873" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/873?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -261,8 +261,7 @@ class AccessionStore(
                         .set(TOTAL_VIABILITY_PERCENT, accession.totalViabilityPercent)
                         .set(TREES_COLLECTED_FROM, accession.numberOfTrees)
                         .returning(ID)
-                        .fetchOne()
-                        ?.get(ID)!!
+                        .fetchOne(ID)!!
```
karthikbtf, line 264: previous change `fetchOne()?.get(ID)!!` is more forgiving than `fetchOne(ID)!!` ?
Guessing fetchOne will always yield a row and the `?` was not needed.

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
```diff
@@ -100,7 +100,6 @@ class GbifStore(private val dslContext: DSLContext) {
         .orderBy(firstWordSortPosition, GBIF_NAMES.NAME)
         .limit(maxResults)
         .fetchInto(GbifNamesRow::class.java)
-        .filterNotNull()
```
karthikbtf, line 103: this removal intended?

karthikbtf (approved):
Very nice!
A couple comments but mostly from some ignorance on my part.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: The behavior should be the same. Old version:

1. `fetchOne()` returns null if no row matches
2. `?.get(ID)` passes the null through (evaluates to null)
3. `!!` operator throws an exception on the null

New version:

1. `fetchOne(ID)` returns null if no row matches
2. `!!` operator throws an exception on the null

The old version could have been written `fetchOne()!!.get(ID)!!` instead.

src/main/kotlin/com/terraformation/backend/species/db/GbifStore.kt
sgrimm: It was unnecessary. `fetchInto(T::class.java)` returns a list that never has null values. This call was probably left over from an earlier draft of the code.

PR #881, opened 2022-12-29
SW-2630 Add API endpoint to get time zone names

sgrimm:
`GET /api/v1/i18n/timeZones` returns a list of time zone IDs and human-readable 
names in the specified locale, US English by default. The list of time zones is
based on Chrome's supported time zone list, minus zones that appear to be
duplicates.

The end result is very close to the list of time zones in the Google Calendar
time zone menu (not that Google Calendar is the ultimate authority on time
zones, just a useful comparison point).


sgrimm:
* main
  * **PR #881** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/881" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/881?utm_source=stack-comment).

sgrimm:
Some examples. English:
```
    "Africa/Abidjan": "Greenwich Mean Time - Abidjan",
    "Africa/Nairobi": "East Africa Time",
    "Africa/Algiers": "Central European Standard Time - Algiers",
    "Africa/Lagos": "West Africa Standard Time - Lagos",
    "Africa/Bissau": "Greenwich Mean Time - Bissau",
    "Africa/Maputo": "Central Africa Time - Maputo",
    "Africa/Cairo": "Eastern European Standard Time - Cairo",
    "Africa/Casablanca": "Western European Time - Casablanca",
```
Spanish:
```
    "Africa/Abidjan": "hora del meridiano de Greenwich - Abiyán",
    "Africa/Nairobi": "hora de África oriental",
    "Africa/Algiers": "hora estándar de Europa central - Argel",
    "Africa/Lagos": "hora estándar de África occidental - Lagos",
    "Africa/Bissau": "hora del meridiano de Greenwich - Bisáu",
    "Africa/Maputo": "hora de África central - Maputo",
    "Africa/Cairo": "hora estándar de Europa oriental - El Cairo",
    "Africa/Casablanca": "hora de Europa occidental - Casablanca",
```
Simplified Chinese (this is not quite correct because it uses a space-separated ASCII hyphen; need to make the city name suffix format localizable too):
```
    "Africa/Abidjan": "格林尼治标准时间 - 阿比让",
    "Africa/Nairobi": "东部非洲时间",
    "Africa/Algiers": "中欧标准时间 - 阿尔及尔",
    "Africa/Lagos": "西部非洲标准时间 - 拉各斯",
    "Africa/Bissau": "格林尼治标准时间 - 比绍",
    "Africa/Maputo": "中部非洲时间 - 马普托",
    "Africa/Cairo": "东欧标准时间 - 开罗",
    "Africa/Casablanca": "西欧时间 - 卡萨布兰卡",
```

src/main/kotlin/com/terraformation/backend/i18n/TimeZones.kt
```diff
@@ -0,0 +1,503 @@
...
+            ?.distinct()
+            ?: emptyList()
+    return if (citiesForZonesOfSameName.size != 1) {
+      "$nameWithoutCity - $city"
+    } else {
+      nameWithoutCity
+    }
+  }
+
+  private val timeZonesWithoutDuplicateCities: List<ZoneId> by lazy {
```
karthikbtf: is this the cldr solution rewritten for us?

src/main/kotlin/com/terraformation/backend/i18n/TimeZones.kt
```diff
@@ -0,0 +1,503 @@
...
+import java.time.ZoneId
+import java.time.format.TextStyle
+import java.util.*
+import java.util.concurrent.ConcurrentHashMap
+import org.apache.commons.text.similarity.LevenshteinDistance
+
+object TimeZones {
+
+  private val chromeTimeZones =
+      listOf(
```
karthikbtf, line 66: should this be a separate data file of sorts?

src/main/kotlin/com/terraformation/backend/i18n/TimeZones.kt
```diff
@@ -0,0 +1,503 @@
...
+    val englishWithoutCity = getTimeZoneNameWithoutCity(zoneId, Locale.ENGLISH)
+    val calendars = CLDR.get(locale).Calendars
+    val city = calendars.timeZoneInfo(zoneId.id).city().name()
+    val citiesForZonesOfSameName =
+        timeZonesByEnglishNameWithoutCity[englishWithoutCity]
+            ?.map { calendars.timeZoneInfo(it.id).city().name() }
+            ?.distinct()
+            ?: emptyList()
+    return if (citiesForZonesOfSameName.size != 1) {
+      "$nameWithoutCity - $city"
```
karthikbtf: yes, i agree this may need further localization as well.. wonder how

karthikbtf (commented):
Thanks for doing this. Looking good.

karthikbtf:
Steve, can you help with a theoretical example where there are no duplicate cities in one language - but present in another language and the dupes are removed?
I'm thinking about a scenario where uses changes preferred locale and we can no longer find the tz in the new list.. how do we find the closest match.. etc.

sgrimm:
> I'm thinking about a scenario where uses changes preferred locale and we can no longer find the tz in the new list.. how do we find the closest match.. etc.

This won't be possible, though that wasn't clear in the first draft of this PR. The list of zone IDs is determined based on the English names, and then we localize the names of those IDs. The updated version of this includes a unit test to verify that the list of IDs is identical in every locale.

sgrimm:
To expand on that a little: whether or not we append a city name is, as of my recent update to the PR, determined on a per-locale basis (so it's possible for a localized name to have a city in one locale but not another), but the list of zone IDs is locale-independent.

src/main/kotlin/com/terraformation/backend/i18n/TimeZones.kt
sgrimm: I'd say it's more inspired by, or maybe informed by, the CLDR solution. They made different design tradeoffs but their code helped me think through some scenarios.

sgrimm:
The way I'm using resource bundles (the `.properties` files) here is kind of an initial stab. Once we get to the "support Spanish localization" project, I may end up wanting to structure the translations in a different way, but I haven't thought it through in enough detail yet and IMO this is a reasonable approach to start with.

src/main/kotlin/com/terraformation/backend/i18n/TimeZones.kt
sgrimm: I did end up moving this to a JSON file, yeah. Maybe at some point we'll set up a fancy CI pipeline that auto-updates it as new browser versions are released.

karthikbtf:
> The way I'm using resource bundles (the `.properties` files) here is kind of an initial stab. Once we get to the "support Spanish localization" project, I may end up wanting to structure the translations in a different way, but I haven't thought it through in enough detail yet and IMO this is a reasonable approach to start with.

This looks great for now and enables us to use curated values.

src/test/kotlin/com/terraformation/backend/i18n/TimeZonesTest.kt
```diff
@@ -0,0 +1,157 @@
...
+            .map { StringUtils.parseLocaleString(it) }
+
+    @JvmStatic fun nonEnglishLocales(): Stream<Locale> = locales().filter { it.language != "en" }
+
+    @JvmStatic
+    fun browsers(): Stream<String> =
+        Stream.of(
+            "Chrome 108.0.5359.124 (OS X 12.6)",
+            "Edge 108.0.1462.54 (Windows 11)",
+            "Safari 17614.1.25.9.10 (OS X 12.6)",
```
karthikbtf, line 154: guessing we periodically update this to catch unsupported time zones?

karthikbtf (approved):
thank you so much!! this is awesome!!

src/test/kotlin/com/terraformation/backend/i18n/TimeZonesTest.kt
sgrimm: Yeah, that's the theory. We'll see how well we keep up with this in practice, but I think it won't be a huge problem to fall behind (we'd run into trouble if a browser removed zones from its list, but not if it updates existing zones or adds new ones).

PR #890, opened 2023-01-06
SW-2634 Make email notifications thread-safe

sgrimm:
In preparation for sending notifications at different times of day for different
facilities based on their time zones, change `EmailNotificationService` so it
accumulates a list of pending messages per thread rather than per notification
type. This will let us safely generate notifications for two facilities at the
same time.

PR #891, opened 2023-01-06
Add test fixture for ApplicationEventPublisher

sgrimm:
Previously, when testing classes that can publish application events, we were
using a MockK stub in place of a real `ApplicationEventPublisher` implementation.

Replace that with a dedicated test fixture class that captures the list of
published events and provides a variety of assertion methods to verify what was
published. This is less awkward to interact with than using MockK capturing slots.


sgrimm:
* main
  * **PR #890** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/890" target="_blank">Graphite</a> 
    * **PR #891** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/891" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/891?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -119,7 +121,7 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
             organizationStore,
             ParentStore(dslContext),
             PermissionStore(dslContext),
-            mockk(),
+            publisher,
```
karthikbtf, line 124: use the publisher to publish events in the tests below - instead of current approach `service.on(event)` (sadly my doing)

karthikbtf (approved):
looks good, one suggestion but not required

src/test/kotlin/com/terraformation/backend/TestEventPublisher.kt
```diff
@@ -0,0 +1,76 @@
+package com.terraformation.backend
```
karthikbtf, line 1: thanks for adding this

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
sgrimm: The approach you used is the right one. This dummy "publisher" doesn't call any event listeners, and the listeners are what we're testing.

PR #892, opened 2023-01-06
SW-2634 Send notifications at 12:01AM local time

sgrimm:
Previously, notifications for scheduled seed bank and nursery events (e.g., for
seeds with drying due dates) were sent once a day and used the server's time zone
(UTC) when checking whether or not a scheduled date had arrived.

Introduce logic to use each facility's time zone instead. Specifically:

* Replace the `task_processed_times` table with two new `facilities` columns:
  `last_notification_date` is the date (in the facility's time zone) on which we
  most recently generated notifications, and `next_notification_time` is 12:01AM
  tomorrow in the facility's time zone.

* Daily tasks now run at 12:01AM rather than 8AM by default. Server-wide daily
  tasks (cleanup of old notifications, etc.) continue to use the server's time
  zone. Notifications are no longer generated by the server-wide daily task code.

* Add a new recurring JobRunr job that scans for facilities whose notifications
  need to be generated. This runs every 15 minutes, as well as immediately when
  the test clock is advanced.

* Replace the `TimePeriodTask` interface, which was only being used for tasks that
  generated notifications, with the `FacilityNotifier` interface, which is not
  only much simpler, but also allows exceptions to be caught by the notification
  scanner job.

* Publish an application event when the time zone of a facility is changed, either
  explicitly or as a side effect of the organization's time zone being changed.
  This can cause notifications to be sent immediately if the time zone change
  causes the facility's local time to cross the 12:01AM boundary.


sgrimm:
* main
  * **PR #894** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/894" target="_blank">Graphite</a> 
    * **PR #892** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/892" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/892?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/daily/NurseryNotifier.kt
```diff
@@ -0,0 +1,35 @@
...
+    if (facility.type == FacilityType.Nursery) {
+      log.info("Generating seedling batch notifications for facility ${facility.id}")
+
+      if (facility.lastNotificationDate != null) {
+        seedlingBatchReady(facility.id, facility.lastNotificationDate, todayAtFacility)
+      }
+    }
+  }
+
+  private fun seedlingBatchReady(facilityId: FacilityId, after: LocalDate, until: LocalDate) {
```
karthikbtf, line 30: seedlings plural?

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -117,10 +117,12 @@ class TerrawareServerConfig(
       val enabled: Boolean = true,
 
       /**
-       * What time of day the daily tasks are run. This is treated as a local time in the configured
-       * [timeZone]. Default is 8AM.
+       * What time of day the daily tasks are run. For server-wide tasks, the globally-configured
+       * server time zone ([timeZone]) is used. For tasks related to a particular facility,
+       * organization, etc., that entity's time zone is used. Default is 00:01 (1 minute past
```
karthikbtf, line 122: Not an issue since we check every 15minutes anyways - just calling out that technically 12:00am is start of day.. if that was the intent.

src/main/kotlin/com/terraformation/backend/daily/NotificationScanner.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.core.annotation.Order
+
+@ConditionalOnProperty(TerrawareServerConfig.DAILY_TASKS_ENABLED_PROPERTY, matchIfMissing = true)
+@Named
+class NotificationScanner(
+    private val clock: InstantSource,
+    private val config: TerrawareServerConfig,
+    private val eventPublisher: ApplicationEventPublisher,
+    private val facilityStore: FacilityStore,
+    private val notifiers: List<FacilityNotifier>,
```
karthikbtf, line 28: Embarrassing question.. do the named notifiers (seedbank and nursery) get bound to this list automagically?

karthikbtf (approved):
Looks good to me!

src/main/kotlin/com/terraformation/backend/daily/NotificationScanner.kt
sgrimm: Totally reasonable question. Yes, Spring looks at the type argument and passes in a list of all the objects it manages that are instances of that type. Did it this way to maintain roughly the same level of decoupling we had before with the event listeners: `NotificationScanner` doesn't need to know what the notifiers actually are.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
sgrimm: No technical reason for this; I just used 12:01 because it's what the product folks were saying in the PRD review. I guess if I wanted to invent a technical reason, I could talk about protecting against clock skew in the cluster or something, but really it's just what people said in the PRD review.

src/main/kotlin/com/terraformation/backend/nursery/daily/NurseryNotifier.kt
sgrimm: I think we use the term "seedling batch" (singular "seedling") in the UI, no? Or did we change to plural?

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
karthikbtf: Got it, this is totally fine! Yes, I recall 12:01 being picked during the review.

src/main/kotlin/com/terraformation/backend/nursery/daily/NurseryNotifier.kt
karthikbtf: We use `Seedlings Batches` as a title, and `Seedling Batch` as the column header (referring to single batches). I guess it is correct here, my mistake.

PR #893, opened 2023-01-06
SW-2637 Use seed bank time zone in accession logic

sgrimm:
If the user creates a viability test and doesn't give a start date, the test is
associated with a withdrawal on today's date. Use the seed bank's time zone to
determine what "today's date" is.

Also use the seed bank's time zone when determining whether a manual quantity
adjustment falls before or after a withdrawal date.

To implement these changes, `AccessionModel` now has a required `clock`
property, which replaces the `Clock` arguments that used to need to be passed
to all the calculation methods. The new property is populated with a clock that
uses the seed bank's time zone.

Another option would have been to keep the `Clock` arguments and make sure to
always pass in a clock with the correct time zone. But that would be easier to
screw up: there'd be no good way to ensure a caller could never accidentally
pass in the system `Clock`, which would potentially cause the calculations to
return different results. With the property approach, we only need to make sure
the clock is correct in the (relatively few) places in the code where we're
constructing an `AccessionModel` from scratch, and none of the callers of the
calculation methods need to worry about it.

Unfortunately, while the application code rarely constructs `AccessionModel`s,
the same can't be said for tests. Adding a new required constructor argument
meant a lot of tests had to be touched. Since they needed to change anyway, I 
took the opportunity to introduce a new helper method to populate both the
clock and the facility ID. The result is that most of the test changes end up
removing clutter, not adding more clutter.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/FacilityStore.kt
@@ -425,6 +425,10 @@ class FacilityStore(
     return facility.copy(nextNotificationTime = nextNotificationTime)
   }
 
+  fun getClock(facilityId: FacilityId): Clock {
+    return clock.withZone(fetchEffectiveTimeZone(fetchOneById(facilityId)))
+  }
+
   fun fetchEffectiveTimeZone(facility: FacilityModel): ZoneId {
     return fetchEffectiveTimeZone(facility.timeZone, facility.organizationId)
   }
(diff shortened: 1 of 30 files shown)


sgrimm:
* main
  * **PR #894** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/894" target="_blank">Graphite</a> 
      * **PR #893** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/893" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/893?utm_source=stack-comment).

karthikbtf (approved):
👍

sgrimm:
Oops, this is incomplete. I was correct that this is the only place in `AccessionModel` itself that needs the time zone, but I failed to notice that we need it in `WithdrawalModel` too, which is going to require a more invasive change. Will update this and rerequest review when it's ready.

sgrimm:
This turned out to need a lot more changes than the initial stab. Updated the PR description to reflect what it looks like now.

karthikbtf (approved):
Looks great!

PR #894, opened 2023-01-06
Add test fixture for Clock

sgrimm:
We've been using MockK to stub out `Clock` objects for testing. That works fine
when the application code just reads scalar values from the clock, but we're
shortly going to start calling `Clock.withZone()` in some places and that's
more awkward to stub out with MockK.

Replace the MockK stubs with a dedicated test class.

Most of the edits here are mechanical:

* Replace `val clock: Clock = mockk()` with `val clock = TestClock()`.
* Delete `every { clock.instant() } returns Instant.EPOCH` and
  `every { clock.zone } returns ZoneOffset.UTC` because they are the defaults.
* Replace `every { clock.instant() } returns` with `clock.instant =` for tests
  that need to change the clock on the fly.

In a few places, manual tweaking was required, e.g., passing an `Instant` to
the constructor in tests that don't use `Instant.EPOCH`.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/TestClock.kt
+++ b/src/test/kotlin/com/terraformation/backend/TestClock.kt
@@ -1,0 +1,18 @@
+package com.terraformation.backend
+
+import java.time.Clock
+import java.time.Instant
+import java.time.ZoneId
+import java.time.ZoneOffset
+
+/** Mutable clock for tests. Its time can be set on the fly; otherwise it acts as a fixed clock. */
+class TestClock(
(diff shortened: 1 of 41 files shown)


sgrimm:
* main
  * **PR #894** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/894" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/894?utm_source=stack-comment).

karthikbtf (approved):
This is nice!

PR #901, opened 2023-01-11
SW-2543 Support imperial weight units in search

sgrimm:
Generalize the support for querying accession and accession withdrawal weights in
grams. Each grams field is now replaced with a set of weight fields, one per unit
(including grams) that does unit conversions. For example, in addition to the
existing `remainingGrams` field, we now also have `remainingKilograms`,
`remainingMilligrams`, `remainingOunces`, and `remainingPounds`.

For exact and range searches, these behave the same as the grams fields used to,
in that you can append an optional units suffix to the search terms.

But exact searches are unlikely to work like users expect in the face of unit
conversions between imperial and metric: if you search for "1 pound" you will
expect it to match an accession whose weight is entered in grams and is the
equivalent of 1.0001 pounds. So the weight field supports fuzzy searches, which
work by matching values that equal the search term if they're rounded to the
same number of decimal places as the search term. If you do a fuzzy search for
1.0 pounds, you will get back a result that weighs 1.02 pounds but not one that
weighs 1.07 pounds (since if you round that to the same number of decimal places
as the search term, it'll be 1.1, not 1.0).

Existing client code shouldn't notice this change; it can continue to do exact
and range searches on grams fields and they should return the same results as
before.


sgrimm:
* main
  * **PR #901** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/901" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/901?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/field/WeightField.kt
```diff
@@ -0,0 +1,153 @@
...
+                  // Range that rounds to the search value, e.g., 432.1 -> BETWEEN(432.05, 432.15)
+                  val scale = quantityModel.quantity.scale()
+                  val maxDifference =
+                      SeedQuantityModel(BigDecimal(".5").movePointLeft(scale), quantityModel.units)
+                  val lowerBound = quantityModel - maxDifference
+                  val upperBound = quantityModel + maxDifference
+                  val lowerGrams = lowerBound.toUnits(SeedQuantityUnits.Grams).quantity
+                  val upperGrams = upperBound.toUnits(SeedQuantityUnits.Grams).quantity
+
+                  gramsField.between(lowerGrams, upperGrams)
```
charlie-tf, line 107: Is between inclusive of its endpoints?

charlie-tf (approved):
Looks great. Just a minor question

src/main/kotlin/com/terraformation/backend/search/field/WeightField.kt
sgrimm: Yes, it's inclusive. If that ever gets to be a problem we can change it to do an exclusive upper bound but it seemed okay to me to err on the side of returning more results.

PR #902, opened 2023-01-11
Properly mark arbitrary JSON objects in OpenAPI

sgrimm:
The OpenAPI schema's definitions of payload properties that allow arbitrary
JSON objects was inaccurate: it didn't include an `additionalProperties`
attribute to indicate that arbitrary keys and values were allowed.

This doesn't impact our existing client code but if we ever use the schema to
generate client code in statically-typed languages, the generated types would
be wrong.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
@@ -16,8 +16,10 @@ import javax.inject.Named
 import kotlin.reflect.KClass
 import kotlin.reflect.full.declaredMemberProperties
 import kotlin.reflect.full.findAnnotation
+import kotlin.reflect.full.isSubtypeOf
 import kotlin.reflect.full.primaryConstructor
 import kotlin.reflect.jvm.javaField
+import kotlin.reflect.typeOf
 import org.jooq.DSLContext
 import org.springdoc.core.SpringDocUtils
 import org.springdoc.core.customizers.OpenApiCustomiser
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #902** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/902" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/902?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/api/OpenApiConfig.kt
```diff
@@ -142,10 +144,16 @@ class OpenApiConfig(private val keycloakInfo: KeycloakInfo) : OpenApiCustomiser
...
+          if (propertySchema is ObjectSchema &&
+              property.returnType.isSubtypeOf(typeOf<ArbitraryJsonObject?>())) {
+            propertySchema.additionalProperties = true
+          }
+
           // Map<*,*> fields default to additionalProperties values that say the values are all
           // objects, which is wrong; values could be other JSON types too.
           if (propertySchema is ArraySchema && propertySchema.items is MapSchema) {
-            propertySchema.items.additionalProperties = null
+            propertySchema.items.additionalProperties = true
```
charlie-tf, line 156: Is the comment above this block still accurate? I'll admit to not being fully familiar with this, but it looks like justification for un-setting the additionalProperties field.

sgrimm: The comment is more accurate now than it was before. There are three possibilities for the `additionalProperties` option:

* If it's missing or null, the schema says you can't include properties other than the ones explicitly listed in the schema. That's what the previous version of this code was doing by setting it to null. Our TypeScript code doesn't care if an object has fields that aren't in the interface, but if we'd, say, generated a Java client from this schema, it would have had no way to add arbitrary key/value pairs.
* If it's `true`, you can include arbitrary properties and their values can be of any type.
* If it's an OpenAPI type (a child object with a schema definition, a reference to another type defined in the schema, or a built-in type name like `object` or `string`) it means you can include arbitrary properties but their values all have to be of the specified type. This is the default behavior (it uses `object` as the type).

https://swagger.io/docs/specification/data-models/dictionaries/ has more gory detail.

charlie-tf: ah, gotcha. Thanks for explaining!

charlie-tf (approved):

PR #913, opened 2023-01-23
SW-2771 Add locale to user details

sgrimm:
Update the `/api/v1/users` payloads to include a `locale` field.

PR #914, opened 2023-01-23
Use Accept-Language header to determine locale

sgrimm:
Configure Spring's web framework so it sets the current locale based on the value
of the `Accept-Language` header.

Use that header to determine the default locale for `/api/v1/i18n/timeZones`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/LocaleConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/LocaleConfig.kt
@@ -1,0 +1,24 @@
+package com.terraformation.backend.api
+
+import com.terraformation.backend.i18n.Locales
+import java.util.*
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.web.servlet.i18n.AcceptHeaderLocaleResolver
+
+@Configuration
+class LocaleConfig {
+  /**
+   * Make Spring's request handlers look at the HTTP Accept-Language header to set the locale for
+   * the current thread.
+   */
+  @Bean
+  fun localeResolver(): AcceptHeaderLocaleResolver {
+    val resolver = AcceptHeaderLocaleResolver()
+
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #913** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/913" target="_blank">Graphite</a> 
    * **PR #914** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/914" target="_blank">Graphite</a>  👈
      * **PR #915** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/915" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/914?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/api/LocaleConfig.kt
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Make Spring's request handlers look at the HTTP Accept-Language header to set the locale for
+   * the current thread.
+   */
+  @Bean
+  fun localeResolver(): AcceptHeaderLocaleResolver {
+    val resolver = AcceptHeaderLocaleResolver()
+
+    resolver.defaultLocale = Locale.ENGLISH
+    resolver.supportedLocales = Locales.supported
```
karthikbtf, line 20: just curious.. are unsupported locales ignored or an error condition?
do we need to set accept-language in the list of allowed headers somewhere or in a yaml?

karthikbtf (approved):
Looks clear to me!

src/main/kotlin/com/terraformation/backend/api/LocaleConfig.kt
sgrimm: Unsupported locales will be ignored and the default will be used instead. That's a standard header so it's passed through by default by Nginx.

PR #916, opened 2023-01-24
Support gibberish time zones

sgrimm:
If the locale is `gx, make `/api/v1/i18n/timeZones` return gibberish time zone
names for localization testing.

PR #917, opened 2023-01-24
SW-2786 Make Messages text localizable

sgrimm:
The `Messages` class is where we render human-readable text on the server side.
The intent was always that it'd be updated to return text in the user's language.

Move the English strings into a properties file and look them up based on the
current locale. The strings should be the same as before since we only have an
English properties file.

In addition, render the numeric parts of seed quantities using the number
formatter for the user's locale. This doesn't currently support the number
formatting rules for the gibberish locale; it just uses the built-in formatting
rules from the Java standard library.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/i18n/Gibberish.kt
+++ b/src/main/kotlin/com/terraformation/backend/i18n/Gibberish.kt
@@ -1,0 +1,16 @@
+import java.util.*
+
+/**
+ * Converts an English string to gibberish for localization testing:
+ * 1. Reverse the order of words
+ * 2. Replace each word with the base64 encoding of its UTF-8 representation, minus padding
+ *
+ * Note that this function does not preserve placeholder tokens.
+ */
+fun String.toGibberish(): String {
+  val encoder = Base64.getEncoder()
+
+  return split(' ').asReversed().joinToString(" ") { word ->
+    encoder.encodeToString(word.toByteArray()).trimEnd('=')
+  }
+}
--- a/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
+++ b/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.i18n
 
+import com.terraformation.backend.db.EnumFromReferenceTable
 import com.terraformation.backend.db.default_schema.GrowthForm
 import com.terraformation.backend.db.default_schema.SeedStorageBehavior
 import com.terraformation.backend.db.default_schema.tables.pojos.DevicesRow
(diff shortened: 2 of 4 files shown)


sgrimm:
* main
  * **PR #916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/916" target="_blank">Graphite</a> 
    * **PR #917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/917" target="_blank">Graphite</a>  👈
      * **PR #918** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/918" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/917?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/i18n/Gibberish.kt
```diff
@@ -0,0 +1,16 @@
+import java.util.*
+
+/**
+ * Converts an English string to gibberish for localization testing:
+ * 1. Reverse the order of words
+ * 2. Replace each word with the base64 encoding of its UTF-8 representation, minus padding
+ *
+ * Note that this function does not preserve placeholder tokens.
+ */
+fun String.toGibberish(): String {
```
karthikbtf, line 10: use this in TerrawareGenerator as well?

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -25,86 +26,93 @@ data class NotificationMessage(val title: String, val body: String)
...
 class Messages {
-  fun csvBadHeader() = "Incorrect column headings"
+  private val messageSource =
+      ResourceBundleMessageSource().apply {
+        setBasename("i18n.Messages")
+        setDefaultEncoding("UTF-8")
+      }
+
+  private fun getMessage(code: String, vararg args: Any): String {
+    return messageSource.getMessage(code, args, currentLocale())
```
karthikbtf, line 36: this may have been clarified before.. just curious how the locale is used to map to the correct message properties file (as we have new language translations).

karthikbtf (approved):
Looks good to me.

src/main/kotlin/com/terraformation/backend/i18n/Gibberish.kt
sgrimm: TerrawareGenerator runs as part of the build process; at the point it runs, `src/main` hasn't been compiled yet (and can't be, because it depends on the generated code).

I agree it's not ideal to have multiple copies of this logic but sharing a single copy across the build and the application would require adding another build stage which didn't seem worth it for a function this simple.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
sgrimm: It's a Java standard library thing: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ResourceBundle.html

src/main/kotlin/com/terraformation/backend/i18n/Gibberish.kt
karthikbtf: got it, thanks!

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
karthikbtf: Thanks, that helps. Looks like the we will have `_locale` suffixes. `Messages_de.properties` etc..

PR #918, opened 2023-01-24
Generate gibberish translations of English messages

sgrimm:
Add a build step that reads the default-locale properties files from
`src/main/resources/i18n` and produces gibberish versions of them.

For now, `src/main/resources/i18n/Messages.properties` is the only message
bundle; it is transformed to `build/resources/main/i18n/Messages_gx.properties`.

This causes the `Messages` class to support the gibberish locale.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,9 +78,6 @@ jobs:
     - name: Compile tests
       run: ./gradlew testClasses
 
-    - name: "DEBUG: List generated properties files"
-      run: ls -l build/resources/main/i18n
-
     - name: Run tests
       run: ./gradlew test
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,6 +78,9 @@ jobs:
     - name: Compile tests
       run: ./gradlew testClasses
 
+    - name: "DEBUG: List generated properties files"
+      run: ls -l build/resources/main/i18n
+
     - name: Run tests
       run: ./gradlew test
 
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/916" target="_blank">Graphite</a> 
    * **PR #917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/917" target="_blank">Graphite</a> 
      * **PR #918** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/918" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/918?utm_source=stack-comment).

sgrimm:
Not sure what's happening with the failure here; investigating.

sgrimm:
Kind of looks like the gibberish enum names property file isn't being generated correctly. Haven't been able to reproduce this locally yet but will keep looking.

sgrimm:
I think I see what's going on here; fix coming up.

sgrimm:
The problem, in case anyone is curious, was that the new Gradle task was just specifying an output directory, and Gradle was seeing that the directory already existed (because it was created during jOOQ code generation) and skipping the build step. The fix is to specify the individual output files so Gradle can check whether or not they need to be rebuilt.

sgrimm:
The way it is now isn't _quite_ ideal since it will rebuild all the properties files if any of them changes, but the build step takes milliseconds so it doesn't seem worth optimizing.

karthikbtf (approved):

PR #919, opened 2023-01-24
Implement gibberish number formatting

sgrimm:
To test localized number formats, the spec for the gibberish locale says it should
use ampersands as thousands separators and commas as decimal separators.

Implement a number formatter that uses those characters. Register it with Java's
localization system so call sites don't have to special-case the gibberish locale;
they can just call `NumberFormat` as usual.

Unfortunately, this requires adding a command-line argument to set a system
property when launching the server. Add it to the command line in the Docker image
and to the `generateOpenApiDocs` configuration, and add a check at start time to
tell people to include it in dev environments.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -162,6 +162,7 @@ tasks.register("downloadDependencies") {
 
 tasks.test {
   useJUnitPlatform()
+  systemProperty("java.locale.providers", "SPI,CLDR,COMPAT")
   testLogging { exceptionFormat = TestExceptionFormat.FULL }
 }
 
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #919** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/919" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/919?utm_source=stack-comment).

charlie-tf (approved):
Seems pretty clear to me. Too bad about the command-line argument hoop to jump through 🤷

PR #920, opened 2023-01-25
SW-2790 Support localized CSV uploads

sgrimm:
Update the CSV validators and importers for species lists, accessions, and
seedling batches to support localized CSV files:

* Return localized template files. The gibberish templates are in English for the
  accession and batch CSVs, but in gibberish for the species CSV since it
  validates that the column names in the header line are correct.
* Since CSV processing happens asynchronously, record the locale as part of the
  upload metadata and run the processing code in that locale.
* Accept localized enum values for enum fields. A special case here is booleans,
  which, in English, can be represented a variety of ways.
* Use the locale's number formatting rules to parse numeric values.
* Return localized validation failure messages, including field names. Localizing
  the field names required decoupling the field names from the validation
  functions.

Also implements SW-2791.


sgrimm:
* main
  * **PR #920** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/920" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/920?utm_source=stack-comment).

karthikbtf (approved):
this is great!

PR #925, opened 2023-01-27
SW-2797 Localize country names

sgrimm:
Make the search API return localized names for countries and country subdivisions.
The frontend uses the search API to get the list of available countries.

English country names are stored in the database, but they aren't turned into
Kotlin enums at build time, so the existing functionality for returning localized
strings for enum search fields won't work. But we still want to look up localized
names from resource bundles and generate the English and gibberish bundles from
the database, same as we do for enum tables.

To address this, introduce a new search field type, `LocalizedTextField`, that
works roughly the same way as `EnumField` but looks values up in resource bundles
explicitly rather than relying on the helper functions in
`EnumFromReferenceTable`.

### DIFF
--- a/jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
+++ b/jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
@@ -50,6 +50,37 @@ class TerrawareGenerator : KotlinGenerator() {
     closeJavaWriter(out)
     english.save()
     gibberish.save()
+
+    generateCountryProperties(schema, "countries", "Countries")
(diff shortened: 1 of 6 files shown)


sgrimm:
* main
  * **PR #925** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/925" target="_blank">Graphite</a>  👈
    * **PR #926** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/926" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/925?utm_source=stack-comment).

karthikbtf (approved):
Looks good to me!!

PR #927, opened 2023-01-28
SW-2800 Sort enums using locale's ordering rules

sgrimm:
Different locales have different rules for sorting text. When an enum field is
used as a sort field in a search API call, we need to sort based on the correct
rules.

We were already mapping enum values to localized display names for sorting
purposes, but that would have ended up sorting based on the database's default
rules. Make this more correct and more efficient by instead mapping enum values
to integers representing the correct sort order of the localized display names.

For gibberish, there's no behavior change here since gibberish uses English
sort ordering. The existing tests that cover this functionality continue to pass.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/field/EnumField.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/field/EnumField.kt
@@ -5,6 +5,7 @@ import com.terraformation.backend.i18n.currentLocale
 import com.terraformation.backend.search.FieldNode
 import com.terraformation.backend.search.SearchFilterType
 import com.terraformation.backend.search.SearchTable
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #927** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/927" target="_blank">Graphite</a>  👈
    * **PR #929** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/929" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/927?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #930, opened 2023-01-31
Expect canonical enum names in request payloads

sgrimm:
Commit a5055833 added the ability to render enum display names in different
locales, but it erroneously changed the parsing logic such that clients had to
pass localized JSON string values for enum fields in request payloads.  This
violates the OpenAPI schema which specifies that enums are always represented
using their English display names in the API.

The practical effect of this bug is that if you switch locales in the client and
try to call a write endpoint whose payload includes an enum field, you'll get a
serialization error because the server doesn't recognize the English display name.


sgrimm:
* main
  * **PR #930** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/930" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/930?utm_source=stack-comment).

karthikbtf (approved):

PR #932, opened 2023-01-31
SW-2826 Localize exported CSV column names

sgrimm:
When search results are exported to a CSV file, the field names in the header
row need to be rendered in the user's locale.

Move the names to `Messages.properties` and look them up when exporting a CSV.

PR #931, opened 2023-01-31
SW-2727 Calculate total accession withdrawn quantities

sgrimm:
Keep track of the total amount withdrawn from each accession, and expose the
totals as search fields.

Reduce a bit of repetition by including the raw quantity and units fields in the
output of `SearchTable.weightFields()`.

Add a migration to add the totals to existing accessions. Since this isn't the
first time we've needed a "go through all the accessions one by one and compute
something" migration, add a helper method to `AccessionStore` to handle the
iteration logic. Once this migration is done, we'll remove
`MigrateWithdrawnTotals` but we can keep the helper method.

PR #934, opened 2023-01-31
Clean up conditional bean declarations

sgrimm:
To stop database-related code from running at start time when there's no database
(e.g., when generating OpenAPI docs) we used a Spring Boot annotation that checked
whether the `terraware.dailyTasks.enabled` property was set to `false`. That
works, but it's both verbose and misleading: we had the annotation on classes
that weren't actually daily tasks at all.

Define two new annotations that make the intent clearer:

`@DisableIfNoDatabase` does what the name implies: disables the bean if there
isn't a database configured, regardless of whether or not daily tasks are
enabled.

`@DailyTask` encapsulates the configuration check and is used on actual daily
tasks. It implies `@DisableIfNoDatabase`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTask.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/daily/NotificationsCleanupTask.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.customer.daily
 
 import com.terraformation.backend.config.TerrawareServerConfig
+import com.terraformation.backend.daily.DailyTask
 import com.terraformation.backend.daily.DailyTaskTimeArrivedEvent
 import com.terraformation.backend.db.default_schema.tables.references.NOTIFICATIONS
 import com.terraformation.backend.log.perClassLogger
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #932** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/932" target="_blank">Graphite</a> 
    * **PR #936** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/936" target="_blank">Graphite</a> 
    * **PR #937** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/937" target="_blank">Graphite</a> 
    * **PR #931** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/931" target="_blank">Graphite</a> 
      * **PR #934** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/934" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/934?utm_source=stack-comment).

sgrimm:
This isn't quite right; will try a different approach.

PR #936, opened 2023-02-01
SW-2861 Generate Role enum from database

sgrimm:
In anticipation of possibly supporting user-defined roles in the future, we had
been manually maintaining the `Role` enum to make it easier to replace with a
dynamic implementation later. We're no longer in imminent danger of implementing
user-defined roles, but the one-off implementaton wasn't hurting anything, so
we left it alone.

But now, this special-case implementation means we don't get localized role
names in search results. Rather than adding more special-case code to address that
problem, convert `Role` to a code-generated enum. This simplifies working with
role columns in the database, since conversion between the integer IDs and the
enum is now done by a code-generated converter class.

`Role` was the only reason `MappedField` existed, so remove it as well.

PR #937, opened 2023-02-01
SW-2801 Use locale number formatting in search results

sgrimm:
Currently, the search API renders numeric values with no thousands separator and
a period as the decimal separator. That's fine in English, but will be wrong in
locales that use different decimal separators.

Add locale-specific number formatting to all the numeric search fields, both at
display time and in exported CSVs. Formatted numbers are accepted as search
criteria as well.

Though this change is meant to support non-English locales, it also affects
English users: we will now show `1,234` instead of `1234` in search results.


sgrimm:
* main
  * **PR #932** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/932" target="_blank">Graphite</a> 
    * **PR #936** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/936" target="_blank">Graphite</a> 
    * **PR #937** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/937" target="_blank">Graphite</a>  👈
    * **PR #931** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/931" target="_blank">Graphite</a> 
      * **PR #934** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/934" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/937?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/nursery/NurserySearchTest.kt
```diff
@@ -272,7 +274,8 @@ internal class NurserySearchTest : DatabaseTest(), RunsAsUser {
                       "notReadyQuantity" to "16",
                       "readyQuantity" to "32",
                       "totalQuantity" to "${16 + 32}",
-                      "totalQuantityWithdrawn" to "${8192 + 16384 + 65536 + 131072}",
+                      "totalQuantityWithdrawn" to
+                          numberFormat.format(8192 + 16384 + 65536 + 131072),
```
charlie-tf: Would it be better to put in the literal string here for the test? Using `numberFormat.format` feels a little circular.

charlie-tf (commented):
Just a question on the tests. Otherwise looks great!

src/test/kotlin/com/terraformation/backend/nursery/NurserySearchTest.kt
sgrimm: The reason I didn't put the literal string in there is that it would make it unclear where the value came from: `8192 + 16384 + 65536 + 131072` is clearly the sum of some specific numbers, but `"221,184"` looks arbitrary and I'd need to add a comment to explain it at which point I felt like it'd be more verbose than calling `numberFormat.format()`.

All that said, I can make the intent clearer here; will update it.

charlie-tf: Cool, thanks!

charlie-tf (approved):
Looks good! Thanks for the test updates

sgrimm:
> Looks good to me 👍 Do we need this formatting for non-search related API later?

This is only an issue for search because all the values are returned as strings that are (in theory) opaque to the client. Numeric fields in other API payloads are JSON numbers which have a fixed format.

PR #951, opened 2023-02-07
SW-1828 Ignore diacritics in species name lookup

sgrimm:
When looking up a species name from the GBIF database, we want to ignore
diacritics like we do for searches of text fields in our search API.

For efficiency, this includes a change to the GBIF import logic: we need to
remove diacritics from the words in the `gbif_name_words` table so they
can be looked up using exact matches on the index on that table.


sgrimm:
* main
  * **PR #951** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/951" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/951?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #952, opened 2023-02-07
SW-2896 Support custom storage location names

sgrimm:
Update the API to support user-supplied storage location names:

* `GET /api/v1/seedbank/values/storageLocations` includes storage location IDs.

* The `facilities.storageLocations` search table now includes an `id` field.

* A new set of endpoints under `/api/v1/seedbank/storageLocations` allows
  creating, reading, updating, and deleting storage locations.

* `POST /api/v1/facilities` has a new optional field to set the list of storage
  location names to create.

* The system continues to create default storage locations if the client doesn't
  specify a list of names to create. The default storage location names are now
  localizable and are based on the locale of the admin who creates the seed bank.

Names are required to be unique within a facility.

For now, the accession API continues to use storage location names rather than IDs.

We're about to remove the concept of storage conditions from the system, so the
new APIs only require (and return) names and IDs. For now, if you use the existing
APIs to query storage locations that were created with the new APIs, they will
have "Freezer" as the storage condition, but that will be removed once the client
code no longer refers to storage conditions.


sgrimm:
* main
  * **PR #952** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/952" target="_blank">Graphite</a>  👈
    * **PR #954** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/954" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/952?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/api/StorageLocationsController.kt
```diff
@@ -0,0 +1,91 @@
...
+  fun getStorageLocation(
+      @PathVariable("id") id: StorageLocationId
+  ): GetStorageLocationResponsePayload {
+    val location = facilityStore.fetchStorageLocation(id)
+
+    return GetStorageLocationResponsePayload(location)
+  }
+
+  @PostMapping
+  fun createStorageLocation(
```
karthikbtf, line 50: api response annotation for 409?

src/main/kotlin/com/terraformation/backend/seedbank/api/StorageLocationsController.kt
```diff
@@ -0,0 +1,91 @@
...
+  @PutMapping("/{id}")
+  fun updateStorageLocation(
+      @PathVariable("id") id: StorageLocationId,
+      @RequestBody payload: UpdateStorageLocationRequestPayload
+  ): SimpleSuccessResponsePayload {
+    facilityStore.updateStorageLocation(id, payload.name, StorageCondition.Freezer)
+
+    return SimpleSuccessResponsePayload()
+  }
+}
```
karthikbtf, line 82: do we need a delete?

karthikbtf (approved):
looks good..
client seems to refer to storage location condition in the accessions table today (customizable column)

src/main/kotlin/com/terraformation/backend/seedbank/api/StorageLocationsController.kt
sgrimm: 🤦

PR #953, opened 2023-02-07
SW-2904 Support CSV export of nested search fields

sgrimm:
Commit 07039f0e added the ability to import a species list CSV file with multiple
ecosystem types separated by line breaks.

Allow exporting the species list in the same format.

The actual change here is more general: it removes the nested-fields restriction
on CSV export of search results. Previously, only flattened fields could be
exported because there is no way to represent tree structures in a CSV file.
That's still the case, even though we now allow nested fields to be exported;
if there are multiple levels of nesting, or if the values contain line breaks,
the results will be ambiguous. But for simple "list of enumerated values" cases
like species ecosystem type, there's no ambiguity.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
@@ -16,4 +16,61 @@ data class SearchResults(
      * [results] contains the full set of results, this will be null.
      */
     val cursor: String?
-)
+) {
+  /**
+   * Turns nested field values into CRLF-delimited strings suitable for exporting to a CSV file.
+   *
+   * For example, if the client searched for fields `foo.bar` and `foo.oof` and the search results
+   * are
+   *
+   * ```
+   * [
+   *   {
+   *     "foo": [
+   *       { "bar": "bar 1", "oof": "oof 1" },
+   *       { "bar": "bar 2"  }
+   *     ]
+   *   }
+   * ]
+   * ```
+   *
+   * this function will return
+   *
+   * ```
+   * [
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #953** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/953" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/953?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
```diff
@@ -16,4 +16,61 @@ data class SearchResults(
...
+  }
+
+  /** Converts nested search results into a flat list of pairs of field names and values. */
+  private fun expandNestedFields(
+      result: Map<*, *>,
+      prefix: String = ""
+  ): List<Pair<String, String>> {
+    return result.entries.flatMap { (key, value) ->
+      when (value) {
+        is String -> listOf(prefix + key to value)
```
karthikbtf, line 67: just so i remembered correctly.. numbers would be returned as strings in search results.

karthikbtf (approved):
looks clear to me 👍

src/main/kotlin/com/terraformation/backend/search/SearchResults.kt
sgrimm: Yes, that's correct, and they'll be formatted according to the locale's number formatting rules (e.g., `"1,234,567"` in English).

PR #964, opened 2023-02-13
SW-2935 Support raw search fields

sgrimm:
To allow clients to more easily parse specific values in search results, add
support for "raw" variants of fields whose values can vary based on locale.

Specify that you want to use raw values rather than localized ones by
appending the string `(raw)` to the name of a search field. You can use raw
fields both for input (search criteria) and for output (list of fields to return).

These are treated as separate search fields with separate names. It is fully
supported to include both the raw and localized versions of a field in the same
search request, and indeed, that's expected to be a common usage pattern whenever
a client needs to both display a value to a user and use it internally for
business logic.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/field/SearchField.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/field/SearchField.kt
@@ -58,6 +58,10 @@ interface SearchField {
   val nullable: Boolean
     get() = true
 
+  /** If true, values should be localized to the current locale. */
+  val localize: Boolean
+    get() = true
+
   /**
    * Returns a list of conditions to include in a WHERE clause when this field is used to filter
    * search results. This may vary based on the filter type.
@@ -76,4 +80,12 @@ interface SearchField {
   fun getDisplayName(messages: Messages): String {
     return messages.searchFieldDisplayName(table.name, fieldName)
   }
+
+  /**
+   * Returns a copy of this field that accepts and returns raw values rather than localized ones, if
+   * relevant. The copy should have a field name with a suffix of "(raw)".
+   *
+   * Returns null if raw values are irrelevant for the field or if this field is already raw.
+   */
+  fun raw(): SearchField?
 }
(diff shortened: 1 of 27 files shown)


sgrimm:
* main
  * **PR #964** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/964" target="_blank">Graphite</a>  👈
    * **PR #965** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/965" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/964?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/field/SearchField.kt
```diff
@@ -76,4 +80,12 @@ interface SearchField {
...
     return messages.searchFieldDisplayName(table.name, fieldName)
   }
+
+  /**
+   * Returns a copy of this field that accepts and returns raw values rather than localized ones, if
+   * relevant. The copy should have a field name with a suffix of "(raw)".
+   *
+   * Returns null if raw values are irrelevant for the field or if this field is already raw.
+   */
+  fun raw(): SearchField?
```
karthikbtf, line 90: thoughts on a `rawFieldName() = "$fieldName(raw)"`

karthikbtf (approved):
Looks good.
Question, in cases where the default behavior changes (example AgeField is now localize=true)... do we need to change the client anywhere or just go with new default behavior?

sgrimm:
I didn't see any places in the frontend code where it was using the values that are localizable now for anything but display. Obviously it's possible I missed some.

karthikbtf:
> I didn't see any places in the frontend code where it was using the values that are localizable now for anything but display. Obviously it's possible I missed some.

oops yes my bad, i was still thinking about client side sorting.. we aren't doing that anymore in the tables for most cases.

karthikbtf (approved):

PR #965, opened 2023-02-13
SW-2939 Localize accession active field

sgrimm:
The "Active" search field on accessions is a one-off implementation rather than a
normal enum field, so it wasn't covered when `EnumField` gained localization
support. Update it to return localized values and to support a raw variant.


sgrimm:
* main
  * **PR #964** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/964" target="_blank">Graphite</a> 
    * **PR #965** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/965" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/965?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #966, opened 2023-02-13
SW-2877 API to edit partner reports

sgrimm:
Add a set of endpoints under `/api/v1/reports` to manage quarterly reports from 
partners. Partners can read and update reports, but can't create or submit them
(creation and submission will be implemented separately).

To enforce the requirement that only one user at a time can edit a report, 
reports must be locked before they can be updated. (Reading a locked report is
allowed.) The basic workflow will look something like this:

1. `GET /api/v1/reports?organizationId=123` to list the available reports. The 
   search API can also be used; see below.
2. `GET /api/v1/reports/456` to fetch a report for viewing. If the report is
   still in progress, this will include the latest values for all the system-
   generated fields in the report.
3. `POST /api/v1/reports/456/lock` to lock a report for editing. This will 
   return an error if the report is already locked by another user.
4. (Optionally) `POST /api/v1/reports/456/lock/force` to steal the lock from 
   another user, presumably after asking for confirmation in the UI.
5. `PUT /api/v1/reports/456` to save the user's edits. This doesn't release the 
   lock, so it can be called multiple times to incrementally save work in
   progress.
6. `POST /api/v1/reports/456/unlock` to release the lock.

Since we expect the report format to evolve over time but we want historical 
reports to be treated as immutable, the report body format is versioned. 
In-progress reports are always upgraded to the latest version, but submitted
reports are always returned in the format that was used when they were
submitted.

To allow local testing, the organization page in the admin UI has a button to 
create a report for the previous calendar quarter. It's only available to super
admins.

Internally, the report is represented as a set of version-independent metadata
fields and a variable-format body. The body is stored as JSON in the database.

Also covers SW-2879, SW-2880, and SW-2881.


sgrimm:
* main
  * **PR #966** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/966" target="_blank">Graphite</a>  👈
    * **PR #978** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/978" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/966?utm_source=stack-comment).

charlie-tf (commented):
Not much to comment here. It all seems pretty sensible to me 👍 Thanks for sharing!

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Fetches a report in whatever format it was written to the database.
+   *
+   * You probably want [ReportService.fetchOneById] instead of this.
+   */
+  fun fetchOneById(reportId: ReportId): ReportModel {
+    requirePermissions { readReport(reportId) }
+
+    val row = reportsDao.fetchOneById(reportId) ?: throw ReportNotFoundException(reportId)
+    val body = objectMapper.readValue<ReportBodyModel>(row.body!!.data())
```
charlie-tf, line 46: Not null because of the NOT NULL constraint in the SQL `body JSONB NOT NULL`? Why is it nullable on the `row` object?

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -0,0 +1,182 @@
...
+            .set(REPORTS.MODIFIED_BY, currentUser().userId)
+            .set(REPORTS.MODIFIED_TIME, clock.instant())
+            .where(REPORTS.ID.eq(reportId))
+            .and(REPORTS.LOCKED_BY.eq(currentUser().userId))
+            .execute()
+
+    if (rowsUpdated != 1) {
+      val row = reportsDao.fetchOneById(reportId) ?: throw ReportNotFoundException(reportId)
+      if (row.statusId == ReportStatus.Submitted) {
+        throw ReportAlreadySubmittedException(reportId)
```
charlie-tf, line 147: So this is enforced by the `CONSTRAINT cannot_submit_while_locked CHECK (locked_time IS NULL OR submitted_time IS NULL)` constraint, right? If the report is submitted, then it cannot be locked by the current user, and is thus uneditable by the current user?

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -0,0 +1,182 @@
...
+  fun create(organizationId: OrganizationId, body: ReportBodyModel): ReportMetadata {
+    requirePermissions { createReport(organizationId) }
+
+    // Quarter is calculated using the server's time zone.
+    val now = ZonedDateTime.now(clock)
+
+    val row =
+        ReportsRow(
+            organizationId = organizationId,
+            quarter = (now.monthValue + 3) / 4,
```
charlie-tf: Not sure if this formula is correct. It yields:
```
x | (x+3)/4
-----------
1 | 1
2 | 1
3 | 1
4 | 1
5 | 2
6 | 2
7 | 2
8 | 2
9 | 3
10| 3
11| 3
12| 3
```
I think `(now.monthValue - 1) / 3` may be correct, but it does give 0 for months 1, 2, 3, where it should go to Q4 of the previous year.

charlie-tf (commented):
Just a couple comments/questions. Looking good!

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
sgrimm: Not only did I get this wrong here, I also got it wrong in the test I wrote to make sure I wasn't getting it wrong! Thanks for catching it.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
sgrimm: Yes, that's it exactly.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
sgrimm: Correct, this is non-nullable in the database but Kotlin doesn't know that. All the fields in jOOQ generated classes are nullable because you can use them with arbitrary queries that might only be selecting a subset of the columns. This is a longstanding design problem and there have been a few experiments to try to figure out a good solution, but it's surprisingly tough to solve in a way that doesn't break as many things as it fixes.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
charlie-tf: You're welcome. Though is `(now.monthValue + 2) / 3` correct? It gives the current quarter instead of the previous one.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
charlie-tf: Ah I see, that makes sense 👍

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
sgrimm: Oh, man, I guess I need to go back to elementary school and re-take calendar math. You're completely right (and also, the year needs to be last year if it's currently Q1).

sgrimm: But the good news is, my original test was correct after all. It just passed for the wrong reason.

charlie-tf: 😄 Helping my kids with their homework pays off!

charlie-tf (approved):
Nice, thanks

PR #969, opened 2023-02-14
SW-2908 Allow deleting seedling batches with photos

sgrimm:
When you delete a seedling batch that is associated with a withdrawal, and the
withdrawal isn't also associated with other batches, the withdrawal is
automatically deleted as well. This was failing if the withdrawal had photos.

Add logic to delete the photos (including the files in the file store) before
attempting to delete the withdrawal from the database.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
@@ -7,6 +7,7 @@ import com.terraformation.backend.db.FacilityTypeMismatchException
 import com.terraformation.backend.db.IdentifierGenerator
 import com.terraformation.backend.db.IdentifierType
 import com.terraformation.backend.db.SpeciesNotFoundException
+import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.default_schema.FacilityId
 import com.terraformation.backend.db.default_schema.FacilityType
 import com.terraformation.backend.db.default_schema.SpeciesId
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #969** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/969" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/969?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -270,19 +274,25 @@ class BatchStore(
...
-              DSL.selectOne()
-                  .from(BATCH_WITHDRAWALS)
-                  .where(WITHDRAWALS.ID.eq(BATCH_WITHDRAWALS.WITHDRAWAL_ID))
-                  .and(BATCH_WITHDRAWALS.BATCH_ID.ne(batchId)))
-          .execute()
+                      .where(WITHDRAWALS.ID.eq(BATCH_WITHDRAWALS.WITHDRAWAL_ID))
+                      .and(BATCH_WITHDRAWALS.BATCH_ID.ne(batchId)))
+              .fetch(WITHDRAWALS.ID.asNonNullable())
+
+      if (withdrawalIds.isNotEmpty()) {
```
karthikbtf, line 291: we could skip this check since foreach over empty is no-op-ish anyways..?

karthikbtf (approved):
Thanks!!

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: I needed it for the `DELETE` statement (an empty `IN` clause is invalid SQL) and it seemed slightly cleaner to wrap the `forEach` at the same time.

karthikbtf: ok, got it,, makes total sense 👍

PR #984, opened 2023-02-21
Fix order dependency in ReportServiceTest

sgrimm:
ReportServiceTest was relying on the report ID sequence being reset, but wasn't
actually resetting it. It was thus passing when run in isolation, but failing if
another report-related test happend to allocate a report ID first.


sgrimm:
* main
  * **PR #984** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/984" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/984?utm_source=stack-comment).

karthikbtf (approved):

PR #983, opened 2023-02-21
SW-2986 Add endpoints to attach photos to reports

sgrimm:
Add endpoints to allow uploading and downloading report-related photos. These
endpoints follow the usual pattern for photos in our API (similar to the endpoints
for nursery withdrawal photos) including server-generated thumbnail images.

One difference from nursery withdrawal photos is that report photos need to
support captions; there is a POST endpoint to update a photo's caption, and the
captions are included in the `GET /api/v1/reports/{id}/photos` response.

PR #982, opened 2023-02-21
Extract common photo endpoint code

sgrimm:
In preparation for adding photo endpoints for reports, extract some OpenAPI
annotations and helper functions from the existing accession and seedling
withdrawal controllers.

No change in functionality here, just reduced code duplication.

PR #985, opened 2023-02-21
SW-2987 Rename photos to files

sgrimm:
In preparation for adding the ability to attach non-photo files to reports, rename
the `photos` table to `files` and update various class and parameter names to
reflect the name change. The table itself is already generic; it doesn't have any
information that only applies to photos.

Child tables such as `accession_photos` keep their existing names because they
only hold information about photos, not about arbitrary files.

There are no changes to the API here, and there should be no user-visible behavior
change.

The majority of these changes were made using IntelliJ refactoring operations, but
there were manual edits in a few places to update comments and to keep lists of
names in alphabetical order.

DEPLOYMENT NOTE: If someone uploads a photo while this is in the middle of being
deployed, the upload will fail if their request happens to be routed to an instance
running the older code.


sgrimm:
* main
  * **PR #982** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/982" target="_blank">Graphite</a> 
    * **PR #983** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/983" target="_blank">Graphite</a> 
      * **PR #985** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/985" target="_blank">Graphite</a>  👈
        * **PR #986** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/986" target="_blank">Graphite</a> 
          * **PR #988** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/988" target="_blank">Graphite</a> 
            * **PR #989** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/989" target="_blank">Graphite</a> 
              * **PR #990** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/990" target="_blank">Graphite</a> 
                * **PR #991** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/991" target="_blank">Graphite</a> 
                  * **PR #992** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/992" target="_blank">Graphite</a> 
                    * **PR #994** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/994" target="_blank">Graphite</a> 
                      * **PR #995** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/995" target="_blank">Graphite</a> 
                        * **PR #996** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/996" target="_blank">Graphite</a> 
                          * **PR #997** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/997" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/985?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/file/FileService.kt
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Throws(IOException::class)
+  fun readFile(
+      fileId: FileId,
+      maxWidth: Int? = null,
+      maxHeight: Int? = null,
+  ): SizedInputStream {
+    return if (maxWidth != null || maxHeight != null) {
+      thumbnailStore.getThumbnailData(fileId, maxWidth, maxHeight)
```
karthikbtf, line 85: just a thought, we could remove width/height semantics and directly call thumbnail store where needed.. or have a derived photo service (but that's certainly overkill)

karthikbtf (approved):
lgtm, one comment but I could just be overthinking/picky

src/main/kotlin/com/terraformation/backend/file/FileService.kt
sgrimm: Having separate thumbnail endpoints is a fair suggestion, but I think it's better if this PR is a pure refactor that doesn't also introduce client-visible changes.

karthikbtf: Makes sense, totally agree 👍

PR #987, opened 2023-02-21
SW-2994 Limit Jira transitions during prod release

alexTerraformation:
Only transition to "Released to Production" if the status is already "Done"


sgrimm (approved):
I'm not sure how to tell whether or not this is correct since I'm not an Atlassian admin, but it seems unlikely to break anything user-visible.

PR #996, opened 2023-02-23
SW-3000 Render reports as barebones HTML

sgrimm:
In preparation for exporting submitted reports to Google Drive, add logic to
render report data as simple HTML, and allow the rendered reports to be viewed
using the admin UI.

The HTML includes references to photos and files. These are assumed to live in
the same directory as the exported HTML file. The admin UI includes endpoints
that serve the correct contents for those filenames.

PR #994, opened 2023-02-23
Calculate nursery stats for reports

sgrimm:
Add server-calculated `mortalityRate` and `totalPlantsPropagated` fields to the
nursery information in reports.

PR #991, opened 2023-02-22
SW-2992 Automatically create reports

sgrimm:
Once a day, check whether or not there are organizations with the `Reporter`
internal tag that don't yet have a report for the previous calendar quarter. If
there are any, create reports for them.

For ad-hoc testing in dev environments, advancing the test clock using the admin
UI will cause daily tasks to run.

This requires a minor change to the permission logic in `OrganizationStore`,
which previously didn't allow the system user to fetch organization details
because the system user doesn't support enumerating its organization and
facility roles.

PR #990, opened 2023-02-22
Include annual details in Q4 reports

sgrimm:
Q4 reports should always include annual details; update the report creation
logic accordingly.

PR #997, opened 2023-02-23
SW-3021 Populate facility details from reports

sgrimm:
When a report is submitted, update the build/operation dates and (for nurseries)
the capacity on the facility if they aren't already set, so they'll be prefilled
when a subsequent report is generated.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/FacilityService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/FacilityService.kt
@@ -5,6 +5,8 @@ import com.terraformation.backend.customer.event.FacilityIdleEvent
 import com.terraformation.backend.customer.event.FacilityTimeZoneChangedEvent
 import com.terraformation.backend.customer.event.OrganizationTimeZoneChangedEvent
 import com.terraformation.backend.customer.model.SystemUser
+import com.terraformation.backend.log.perClassLogger
+import com.terraformation.backend.report.event.ReportSubmittedEvent
 import javax.inject.Named
 import org.jobrunr.jobs.annotations.Job
 import org.jobrunr.spring.annotations.Recurring
@@ -18,6 +20,8 @@ class FacilityService(
     private val publisher: ApplicationEventPublisher,
     private val systemUser: SystemUser,
 ) {
+  private val log = perClassLogger()
+
   /** Sends alert email when facilities go idle. Runs once per minute. */
   @Job(name = SCAN_FOR_IDLE_FACILITIES_JOB_NAME, retries = 0)
   @Recurring(id = SCAN_FOR_IDLE_FACILITIES_JOB_NAME, cron = "* * * * *")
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #982** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/982" target="_blank">Graphite</a> 
    * **PR #983** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/983" target="_blank">Graphite</a> 
      * **PR #985** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/985" target="_blank">Graphite</a> 
        * **PR #986** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/986" target="_blank">Graphite</a> 
          * **PR #988** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/988" target="_blank">Graphite</a> 
            * **PR #989** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/989" target="_blank">Graphite</a> 
              * **PR #990** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/990" target="_blank">Graphite</a> 
                * **PR #991** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/991" target="_blank">Graphite</a> 
                  * **PR #992** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/992" target="_blank">Graphite</a> 
                    * **PR #994** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/994" target="_blank">Graphite</a> 
                      * **PR #995** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/995" target="_blank">Graphite</a> 
                        * **PR #996** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/996" target="_blank">Graphite</a> 
                          * **PR #997** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/997" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/997?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/FacilityService.kt
```diff
@@ -41,6 +45,42 @@ class FacilityService(
...
+  fun on(event: ReportSubmittedEvent) {
+    event.body.nurseries.forEach { nurseryBody ->
+      val existing = facilityStore.fetchOneById(nurseryBody.id)
+      val updated =
+          existing.copy(
+              buildCompletedDate = existing.buildCompletedDate ?: nurseryBody.buildCompletedDate,
+              buildStartedDate = existing.buildStartedDate ?: nurseryBody.buildStartedDate,
+              capacity = existing.capacity ?: nurseryBody.capacity,
+              operationStartedDate = existing.operationStartedDate
+                      ?: nurseryBody.operationStartedDate,
```
charlie-tf, line 58: If a report changes these values (or the seedbank ones) after they have been set once, they won't be updated here. But the "new values" will still go out with the submitted report. Is there no way to update these and is that intended?

sgrimm: The flow here is a little weird because there's a bidirectional data dependency, but what would happen in that case would be that the fields wouldn't be considered editable any more, and the new values from the subsequent report would be overwritten by `ReportService.fetchOneById()` during the PUT request.

The idea (or rather, the requirement per the PRD) is that submitting a report can only set initial values for these fields on the facility. You can use the facility edit UI to update them afterwards, but once they're set, they're considered read-only in the report data.

charlie-tf: Ah that makes sense. Thank you.

charlie-tf (approved):

PR #1006, opened 2023-02-27
Move ID and storage URL into FileMetadata

sgrimm:
In preparation for implementing Google Drive export of report data, add fields for
the file ID and storage URL to `FileMetadata`. Since we use that class both for
new file uploads and for fetching existing files, introduce typealiases to make
the new fields null or non-nullable using the same scheme we use for other model
classes.

Remove redundant parameters and fields in places where we already have a
`FileMetadata` with the required information.

No change in client-visible behavior.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/file/model/FileMetadata.kt
+++ b/src/main/kotlin/com/terraformation/backend/file/model/FileMetadata.kt
@@ -1,33 +1,54 @@
 package com.terraformation.backend.file.model
 
 import com.terraformation.backend.db.asNonNullable
+import com.terraformation.backend.db.default_schema.FileId
 import com.terraformation.backend.db.default_schema.tables.pojos.FilesRow
 import com.terraformation.backend.db.default_schema.tables.references.FILES
+import java.net.URI
 import org.jooq.Record
 import org.springframework.http.MediaType
 
-data class FileMetadata(
-    val filename: String,
+data class FileMetadata<I : FileId?, U : URI?>(
     val contentType: String,
+    val filename: String,
+    val id: I,
     val size: Long,
+    val storageUrl: U,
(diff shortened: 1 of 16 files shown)


sgrimm:
* main
  * **PR #1006** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1006" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1006?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/file/model/FileMetadata.kt
```diff
@@ -1,33 +1,54 @@
...
+import com.terraformation.backend.db.default_schema.FileId
 import com.terraformation.backend.db.default_schema.tables.pojos.FilesRow
 import com.terraformation.backend.db.default_schema.tables.references.FILES
+import java.net.URI
 import org.jooq.Record
 import org.springframework.http.MediaType
 
-data class FileMetadata(
-    val filename: String,
+data class FileMetadata<I : FileId?, U : URI?>(
```
karthikbtf, line 11: The optional types to support Nothing (new file usecase)?

karthikbtf (approved):
Looks good to me.

src/main/kotlin/com/terraformation/backend/file/model/FileMetadata.kt
sgrimm: Yeah, exactly. `Nothing?` is a subtype of all nullable types, but there are no valid values of `Nothing`, so the only valid value of `Nothing?` is null. It pretty much exists for this exact "value must be null" use case.

PR #1008, opened 2023-02-28
SW-2388 Add instructions to species CSV template

sgrimm:
Include lists of supported enum values and other guidance in the template for
importing species list CSV files.

Previously, per the initial requirements, we were checking for exact matches of
the header column names, but this was causing problems for users, and with the
instructions included, it's now much more likely that people will upload otherwise
valid CSV files where the header row isn't byte-identical to our template. Remove
the strict checking in favor of just checking that the number of columns in the
header row is correct.

Add test cases to ensure that localized templates work as intended.

Also partially addresses SW-3071.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/db/SpeciesCsvValidator.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/SpeciesCsvValidator.kt
@@ -44,20 +44,6 @@ class SpeciesCsvValidator(
     return messages.speciesCsvColumnName(position)
   }
 
-  override fun validateHeaderRow(rawValues: Array<String?>?): Boolean {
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #1008** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1008" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1008?utm_source=stack-comment).

karthikbtf:
Looks like a csv validator test is failing.

karthikbtf (approved):
Changes look good. Might need a test update.

PR #1009, opened 2023-02-28
SW-3070 Add ExactOrFuzzy search type

sgrimm:
Allow clients to specify whether or not they want the "try an exact search
first, then fall back to fuzzy if there were no matches" behavior introduced in
commit 64c854c. After getting user feedback, it turns out we only want that
behavior on a few specific search fields, not across the board.

Search field criteria with a type of `Fuzzy` are now always fuzzy, with no
exact-then-fuzzy fallback. The new type `ExactOrFuzzy` will return exact matches
if there are any, or fuzzy matches if there are no exact ones.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
@@ -6,6 +6,7 @@ import org.jooq.impl.DSL
 
 enum class SearchFilterType {
   Exact,
+  ExactOrFuzzy,
   Fuzzy,
   Range
 }
@@ -14,9 +15,9 @@ interface SearchNode {
   fun toCondition(): Condition
 
   /**
-   * Converts this node and all its descendents from fuzzy to exact filtering. If this node does not
-   * currently contain any fuzzy filter criteria, this method _must_ return an object that tests
-   * equal to `this` (e.g., by just returning `this`).
+   * Converts this node and all its descendents from exact-or-fuzzy to exact filtering. If this node
+   * does not currently contain any exact-or-fuzzy filter criteria, this method _must_ return an
+   * object that tests equal to `this` (e.g., by just returning `this`).
    */
   fun toExactSearch(): SearchNode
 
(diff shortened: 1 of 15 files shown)


sgrimm:
* main
  * **PR #1009** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1009" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1009?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
```diff
@@ -126,7 +127,7 @@ data class FieldNode(
   }
 
   override fun toExactSearch(): FieldNode {
-    return if (type == SearchFilterType.Fuzzy && values.any { it != null }) {
+    return if (type == SearchFilterType.ExactOrFuzzy && values.any { it != null }) {
```
charlie-tf, line 130: If I understand correctly, now the `toExactSearch` conversion will only happen for `ExactOrFuzzy` nodes, whereas for `Fuzzy` node `this` is returned, meaning it will stay fuzzy?

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -176,13 +177,23 @@ class SearchService(private val dslContext: DSLContext) {
...
-            distinct = true,
-        )
+                rootPrefix,
+                listOf(fieldPath),
+                criteria.toExactSearch(),
+                listOf(SearchSortField(fieldPath)),
+                limit = limit,
+                distinct = true,
+            )
+            .ifEmpty {
```
charlie-tf: Will this effectively run the search twice for a fuzzy query that returns empty?

charlie-tf (commented):
Overall looks good, just a couple questions

src/main/kotlin/com/terraformation/backend/search/SearchFilter.kt
sgrimm: Yes, that's the idea. We're removing the "try an exact search first" behavior when the filter type is `Fuzzy`.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
sgrimm: Good catch. I should only do this if `toExactSearch()` makes changes.

charlie-tf (approved):
Neato

PR #1010, opened 2023-02-28
SW-3041 Render reports as CSV files

sgrimm:
Add rendering logic to produce a summary of the data in a report in CSV format.
The CSV file can be downloaded from the admin UI for local testing.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
@@ -1,10 +1,14 @@
 package com.terraformation.backend.report.render
 
+import com.opencsv.CSVWriter
+import com.terraformation.backend.api.writeNext
 import com.terraformation.backend.customer.db.OrganizationStore
 import com.terraformation.backend.db.default_schema.ReportId
 import com.terraformation.backend.report.ReportFileService
 import com.terraformation.backend.report.db.ReportStore
+import com.terraformation.backend.report.model.ReportBodyModel
 import com.terraformation.backend.report.model.ReportBodyModelV1
+import java.io.StringWriter
 import java.time.Month
 import java.time.format.TextStyle
 import java.util.Locale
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #1010** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1010" target="_blank">Graphite</a>  👈
    * **PR #1011** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1011" target="_blank">Graphite</a> 
      * **PR #1013** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1013" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1010?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
```diff
@@ -44,4 +48,83 @@ class ReportRenderer(
...
     }
   }
+
+  fun renderReportCsv(reportId: ReportId): String {
+    val report = reportStore.fetchOneById(reportId)
+
+    val body: ReportBodyModel = report.body
+    val columnValues: List<Pair<String, Any?>> =
+        when (body) {
+          is ReportBodyModelV1 -> {
```
karthikbtf, line 58: return empty list otherwise?

karthikbtf (approved):
lgtm
will this be exposed to the end user at some point?

src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
sgrimm: `ReportBodyModel` is a sealed interface, so it's a compile error if there are cases this `when` doesn't cover.

sgrimm:
This is going to be used for the Google Drive export. End users will never see this, but our internal users will see these CSV files show up in Google Drive.

src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
karthikbtf: Got it 👍

PR #1015, opened 2023-03-03
Rename reports photo upload response to 'id' from 'fileId'

karthikbtf:
- similar to [withdrawal photos response](https://github.com/terraware/terraware-server/blob/c9ebbd6e738b1e9d5ed447223d10577a0c33ee36/src/main/kotlin/com/terraformation/backend/nursery/api/WithdrawalsController.kt#L261)
- this is mostly for a reusable photo upload function in the FE codebase [see](https://github.com/terraware/terraware-web/commit/43f989c7bb02e05a8b51b816bd5c2b9ade9e9649#r102961693)


sgrimm (approved):
Sure, consistency is good!

PR #1024, opened 2023-03-15
SW-3203 Limit concurrent thumbnail generation

sgrimm:
Loading too many photos into memory at once to generate thumbnails exhausts the
server's heap space. Add a semaphore to limit the number of photos we can resize
at the same time on a given server instance.

This initial implementation is very naive and doesn't take into account the fact
that memory usage varies depending on the photo's dimensions, but should be
conservative enough to protect us from out-of-memory errors if a client requests
dozens of thumbnails of newly-uploaded images in rapid succession.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/file/ThumbnailStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/file/ThumbnailStore.kt
@@ -15,6 +15,8 @@ import java.net.URI
 import java.nio.file.FileAlreadyExistsException
 import java.nio.file.NoSuchFileException
 import java.time.Clock
+import java.util.concurrent.Semaphore
+import java.util.concurrent.TimeUnit
 import javax.imageio.ImageIO
 import javax.inject.Named
 import kotlin.io.path.Path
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1024** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1024" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1024?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/file/ThumbnailStore.kt
```diff
@@ -122,49 +131,58 @@ class ThumbnailStore(
     val filesRow = filesDao.fetchOneById(fileId) ?: throw FileNotFoundException(fileId)
     val photoUrl = filesRow.storageUrl!!
 
-    val resizedImage = scalePhoto(photoUrl, maxWidth, maxHeight)
-    val buffer = encodeAsJpeg(resizedImage)
-    val size = buffer.size
+    try {
+      if (!semaphore.tryAcquire(0, TimeUnit.SECONDS)) {
+        log.debug("Maximum number of thumbnails already being generated; waiting for one to finish")
+        semaphore.acquire()
```
karthikbtf: does this timeout? just curious..

karthikbtf (approved):
good thing to have!!

src/main/kotlin/com/terraformation/backend/file/ThumbnailStore.kt
sgrimm: It doesn't, but that's a good point; I should make it time out.

karthikbtf (approved):
Thanks, looks good to me!!

PR #1025, opened 2023-03-16
Remove custom OpenAPI sorting

sgrimm:
The OpenAPI library now has support for sorting schema elements automatically,
so use that instead of explicitly sorting everything ourselves.

Add tags to a couple controllers that were missing them so that they get sorted
into the proper positions in the Swagger UI, and add a test to catch any future
cases where we forget to add tags.


sgrimm:
* main
  * **PR #1025** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1025" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1025?utm_source=stack-comment).

karthikbtf (approved):
Nice!

PR #1026, opened 2023-03-16
Remove redundant @RequestParam arguments

sgrimm:
Spring uses the controller argument names as the names of query string parameters
by default, so there's no need to specify them explicitly. Spring also infers
whether or not a query string parameter is required based on the nullability of
the argument's type, so there's no need to explicitly specify that in the
annotation either.

There's no change in behavior or the OpenAPI schema here.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -319,8 +319,8 @@ class AdminController(
 
   @PostMapping("/createFacility")
   fun createFacility(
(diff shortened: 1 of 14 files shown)


sgrimm:
* main
  * **PR #1026** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1026" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1026?utm_source=stack-comment).

karthikbtf (approved):
Much nicer!!

PR #1027, opened 2023-03-16
SW-3205 Add "delete report" button to admin UI

sgrimm:
Allow super-admins to delete reports, including all their files and photos.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
@@ -26,6 +26,7 @@ import com.terraformation.backend.file.model.FileMetadata
 import com.terraformation.backend.log.perClassLogger
 import com.terraformation.backend.report.ReportService
 import com.terraformation.backend.report.event.ReportCreatedEvent
+import com.terraformation.backend.report.event.ReportDeletionStartedEvent
 import com.terraformation.backend.report.event.ReportSubmittedEvent
 import com.terraformation.backend.report.model.ReportBodyModel
 import com.terraformation.backend.report.model.ReportFileModel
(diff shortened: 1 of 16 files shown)


sgrimm:
* main
  * **PR #1027** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1027" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1027?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -220,6 +221,25 @@ class ReportStore(
...
+    // resources and then update the database to remove the references to them; if that happened
+    // inside an enclosing transaction, then a listener throwing an exception could cause the system
+    // to roll back the updates that recorded the successful removal of external resources by an
+    // earlier one.
+    //
+    // There's an unavoidable tradeoff here: if a listener fails, the report data will end up
+    // partially deleted.
+    eventPublisher.publishEvent(ReportDeletionStartedEvent(reportId))
+
+    reportsDao.deleteById(reportId)
```
charlie-tf, line 240: Is there a possible race condition here, in which the event listener attempts to `fetchPhotosByReportId` and `fetchFilesByReportId` but the report has already been deleted from the database at this line?

sgrimm: Event handlers are called synchronously, so that won't be a problem. `eventPublisher.publishEvent()` won't return until all the handlers have finished.

charlie-tf: 👍 Thanks!

charlie-tf (approved):

PR #1029, opened 2023-03-17
Move English enum display names to source code

sgrimm:
When we return enum values that are going to be shown to the user, we look up
localized display names from a properties file.

Previously, the properties files with the English display names were generated as
part of the build process. But that approach makes it difficult to include
additional notes to help translators pick the right translations.

Move the generated files into `src` and remove the logic that generates them from
the reference tables. Add a test to ensure that we don't forget to keep the files
up to date as the data model evolves.


sgrimm:
* main
  * **PR #1029** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1029" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1029?utm_source=stack-comment).

karthikbtf (approved):

PR #1030, opened 2023-03-17

terraformation-inc:


src/main/resources/i18n/Messages_es.properties
karthikbtf, line 1: placeholder for now?

karthikbtf (approved):
👍

PR #1038, opened 2023-03-23
Combine CSV boolean values into single strings

sgrimm:
We accept a number of synonyms for "true" and "false" in boolean columns in
uploaded CSV files. Previously, each synonym was a separate translatable
string, but that's going to be confusing for translators, especially if the
number of words needs to vary between languages.

Combine the word lists into two strings and add a description to each one to tell
translators what to do.

Previously, we used the first word in the list as the canonical localized word
that was accepted by, and returned by, the search API. Add separate strings for
that purpose so we don't have to tell translators to be careful about the order
of the list of synomyms.

The "one synonym per line" format uncovered a bug in gibberish string conversion:
it wasn't preserving line breaks on multiline strings, which included things like
our plaintext email bodies. Update the gibberish logic to preserve line breaks.

### DIFF
--- a/buildSrc/src/main/kotlin/com/terraformation/gradle/RenderGibberishTask.kt
+++ b/buildSrc/src/main/kotlin/com/terraformation/gradle/RenderGibberishTask.kt
@@ -66,7 +66,11 @@ abstract class RenderGibberishTask : DefaultTask() {
 
     englishProperties.forEach { name, english ->
       gibberishProperties[name] =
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #1038** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1038" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1038?utm_source=stack-comment).

karthikbtf (approved):

PR #1039, opened 2023-03-23
SW-3229 Fix Locked by User Name in Report Payload

charlie-tf:
The report payload `lockedByName` field was not being populated. It turned
out to have been trying to look up the user by the `modifiedBy` user id.
Change it to look up by `lockedBy` id.


sgrimm (approved):
Oops!

PR #1040, opened 2023-03-24
SW-3245 Facility Count Only Selected

charlie-tf:
The seedbank/nursery/planting site counts rendered in the report html
document should only include those facilities that were selected for
inclusion in the report, rather than the total counts.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
@@ -38,12 +38,12 @@ class ReportRenderer(
     val organization = organizationStore.fetchOneById(report.metadata.organizationId)
     val context = Context()
 
+    val reportBodyLatestVersion = report.body.toLatestVersion()
     val numFacilitiesSelected =
         mapOf(
-            "seedBanks" to report.body.toLatestVersion().seedBanks.filter { it.selected }.size,
-            "nurseries" to report.body.toLatestVersion().nurseries.filter { it.selected }.size,
-            "plantingSites" to
-                report.body.toLatestVersion().plantingSites.filter { it.selected }.size,
+            "seedBanks" to reportBodyLatestVersion.seedBanks.count { it.selected },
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
```diff
@@ -38,7 +38,16 @@ class ReportRenderer(
     val organization = organizationStore.fetchOneById(report.metadata.organizationId)
     val context = Context()
 
+    val numFacilitiesSelected =
+        mapOf(
+            "seedBanks" to report.body.toLatestVersion().seedBanks.filter { it.selected }.size,
```
karthikbtf: extract out `report.body.toLatestVersion()`?

karthikbtf (approved):

sgrimm (approved):
This seems fine, aside from Karthik's comment which I agree with.

The current behavior is what it is because I was following the designs, which show a seed bank count of 2 with only one seed bank selected. I mention this just so you're on the lookout for any other places where I was taking the designs too literally.

src/main/kotlin/com/terraformation/backend/report/render/ReportRenderer.kt
```diff
@@ -38,7 +38,16 @@ class ReportRenderer(
...
+            "nurseries" to report.body.toLatestVersion().nurseries.filter { it.selected }.size,
```
sgrimm: Could do `count { it.selected }` instead of `filter { it.selected }.size`.

charlie-tf:
> This seems fine, aside from Karthik's comment which I agree with.
> 
> The current behavior is what it is because I was following the designs, which show a seed bank count of 2 with only one seed bank selected. I mention this just so you're on the lookout for any other places where I was taking the designs too literally.

I think what you implemented was fine, the design being what it is. The PRD says only,
```
- # of nurseries
- # seed banks
- # of planting sites
```
Easy enough change.

PR #1041, opened 2023-03-24
SW-3183 Include Withdrawn Quantity in Seed Count

charlie-tf:
We want the total seed count for a seedbank to include the seed quantity
withdrawn from that seedbank as well as the seeds currently stored. This
results in the total number of seeds stored in the seedbank for all time
which is the value that is required for reports. We had to add the total
withdrawn seeds to the seedbank summary statistics to get this value.

(sorry I accidentally pushed this to main but reverted it)


sgrimm (approved):
Seems reasonable! Nice test case.

PR #1053, opened 2023-03-31
SW-3360 Disallow Lock/Unlock When Submitted

charlie-tf:
We want to make sure the data in the database is consistent with its
submission status. Therefore, add a database migration which makes sure that
existing reports are in the submitted state when there is a submitted_by
user. It also adds a constraint that ensures that when there is such a user
the report must be submitted. And when the it is submitted there must be
a submitted_by user.

In the lock and unlock methods of ReportService, add a condition that the
report is not submitted and throw an exception otherwise.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
@@ -113,15 +113,7 @@ class ReportStore(
             .execute()
 
     if (rowsUpdated != 1) {
-      val reportSubmitted =
-          dslContext
-              .selectOne()
-              .from(REPORTS)
-              .where(REPORTS.ID.eq(reportId))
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -111,6 +113,18 @@ class ReportStore(
...
 
     if (rowsUpdated != 1) {
+      val reportSubmitted =
+          dslContext
+              .selectOne()
+              .from(REPORTS)
+              .where(REPORTS.ID.eq(reportId))
+              .and(REPORTS.STATUS_ID.eq(ReportStatus.Submitted))
+              .fetch()
+              .isNotEmpty
```
sgrimm: Maybe a reasonable candidate for a helper function.

sgrimm (approved):
Looks reasonable.

sgrimm (approved):

PR #1054, opened 2023-03-31
SW-3352 MortalityRateInNursery Not Required

charlie-tf:
A change in requirements means that for species in a planting site, the
mortality rate in nursery is no longer required. And the input for this
has been removed on the frontend. To allow a report to be validated,
remove the requirement. We also go through and remove the field from
the model and payloads.


sgrimm (approved):
Seems safe enough. If the input is going away, should the field go away entirely, as opposed to just the validation?

charlie-tf:
> Seems safe enough. If the input is going away, should the field go away entirely, as opposed to just the validation?

Good point, I will remove it.

karthikbtf (approved):
Any related schema changes?

charlie-tf:
> Any related schema changes?

No, if you mean SQL schema changes. The report data is all serialized as json before going into/out of the DB

PR #1064, opened 2023-04-05
Diff OpenAPI YAML in CI builds

sgrimm:
Make it easy to see what changes a given PR would cause to the OpenAPI schema
by including a diff between the locally-generated schema and the staging one
in the GitHub Actions output.

Failure to fetch the schema from staging is logged but not treated as a build
failure (we don't want CI builds to fail if staging is down).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,8 +77,8 @@ jobs:
 
     - name: Diff OpenAPI docs against staging
       run: |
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1064** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1064" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1064?utm_source=stack-comment).

karthikbtf (approved):

PR #1066, opened 2023-04-05
Git ignore IntelliJ project files

karthikbtf:


charlie-tf (approved):

sgrimm (approved):
Sure, seems fine. FYI, this directory only gets used if you check the "Store as project file" checkbox in the run configuration dialog. But we don't share run configurations with each other so ignoring it seems fine.

PR #1080, opened 2023-04-13
SW-3433 Save SeedBank and Nursery Info on Submit

charlie-tf:
When a report is submitted, we want to save build start, build completion, operation start, and capacity for seed banks and nurseries if they are entered in a report. They should be saved after report validation, and only facilities that are selected should have those values saved.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
@@ -18,6 +18,7 @@ import com.terraformation.backend.db.default_schema.ReportId
 import com.terraformation.backend.db.default_schema.ReportStatus
 import com.terraformation.backend.db.default_schema.tables.daos.ReportsDao
 import com.terraformation.backend.db.default_schema.tables.pojos.ReportsRow
+import com.terraformation.backend.db.default_schema.tables.references.FACILITIES
 import com.terraformation.backend.db.default_schema.tables.references.FILES
 import com.terraformation.backend.db.default_schema.tables.references.ORGANIZATION_INTERNAL_TAGS
 import com.terraformation.backend.db.default_schema.tables.references.REPORTS
@@ -228,6 +229,9 @@ class ReportStore(
           .where(REPORTS.ID.eq(reportId))
           .execute()
 
+      saveSeedBankInfo(body)
+      saveNurseryInfo(body)
+
       eventPublisher.publishEvent(ReportSubmittedEvent(reportId, body))
     }
   }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -380,6 +384,39 @@ class ReportStore(
...
+  private fun saveSeedBankInfo(body: ReportBodyModel) {
+    val reportBody = body.toLatestVersion()
+    reportBody.seedBanks
+        .filter { it.selected }
+        .forEach {
+          dslContext
+              .update(FACILITIES)
+              .set(FACILITIES.BUILD_STARTED_DATE, it.buildStartedDate)
+              .set(FACILITIES.BUILD_COMPLETED_DATE, it.buildCompletedDate)
+              .set(FACILITIES.OPERATION_STARTED_DATE, it.operationStartedDate)
```
karthikbtf: Add a where clause checking facilities type is SeedBank? I understand the id check may be sufficient. Same for the nursery info below.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -380,6 +384,39 @@ class ReportStore(
...
+
+  /** Save the nursery buildStartDate, buildCompletedDate, operationStartDate, and capacity */
+  private fun saveNurseryInfo(body: ReportBodyModel) {
+    val reportBody = body.toLatestVersion()
+    reportBody.nurseries
+        .filter { it.selected }
+        .forEach {
+          dslContext
+              .update(FACILITIES)
+              .set(FACILITIES.BUILD_STARTED_DATE, it.buildStartedDate)
```
karthikbtf: Pardon my ignorance.. if there's a nursery in the report body, does that imply the dates have changed? Or can we check the values to optimize/skip-saving, etc.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -378,6 +379,14 @@ abstract class DatabaseTest {
     }
   }
 
+  protected fun getFacilityById(facilityId: FacilityId): FacilitiesRecord {
+    return dslContext
```
karthikbtf: in case it helps, there's a facilitiesDao 
```
facilitiesDao.fetchOneById(facilityId) ?: throw not found exception
```
might be good to throw an exception in any case (remove !! assumption)

karthikbtf (approved):
Looks good, a few thoughts.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
charlie-tf: Yeah, I wondered about this. Seems like I should query for the existing facility and do a comparison check against the existing values to see if anything changed before updating

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -380,6 +387,62 @@ class ReportStore(
     }
   }
 
+  /** Save the seed bank buildStartDate, buildCompletedDate, and operationStartDate */
+  private fun saveSeedBankInfo(body: ReportBodyModel) {
+    val reportBody = body.toLatestVersion()
+    reportBody.seedBanks
+        .filter { it.selected }
+        .forEach {
+          val seedBank = facilitiesDao.fetchOneById(it.id)
```
karthikbtf, line 396: Question: does #L213 imply this is a transaction as well?
If we don't find the seedBank here do we want to throw an exception `?: throw FacilityNotFoundException` or let it pass (we won't end up updating). The current implementation does seem better than throwing an exception (otherwise end-user will have to go through hoops to clean up the report). Also, a very unlikely scenario.

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -380,6 +387,62 @@ class ReportStore(
...
+          dslContext
+              .update(FACILITIES)
+              .set(FACILITIES.BUILD_STARTED_DATE, it.buildStartedDate ?: seedBank?.buildStartedDate)
+              .set(
+                  FACILITIES.BUILD_COMPLETED_DATE,
+                  it.buildCompletedDate ?: seedBank?.buildCompletedDate)
+              .set(
+                  FACILITIES.OPERATION_STARTED_DATE,
+                  it.operationStartedDate ?: seedBank?.operationStartedDate)
+              .where(FACILITIES.TYPE_ID.eq(FacilityType.SeedBank))
```
karthikbtf, line 411: thanks

karthikbtf (approved):
Nice 🎉

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
charlie-tf: Yes, it is in a transaction—I hadn't realized that at first. I think not throwing an exception is probably best like you say.

PR #1083, opened 2023-04-17
SW-3394 Deploy using Tailscale instead of bastion

sgrimm:
When deploying to the AWS hosts, connect to the VPC network using Tailscale and
ssh directly to the hosts rather than bouncing through a bastion host.

### DIFF
--- a/.github/scripts/deploy.sh
+++ b/.github/scripts/deploy.sh
@@ -17,5 +17,6 @@ aws ec2 describe-instances --filters "Name=tag:Application,Values=terraware" \
       echo
       echo "Deploying to $_host"
       echo
-      ssh -n $_host "/usr/local/bin/update.sh terraware-server $COMMIT_SHA"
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #1083** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1083" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1083?utm_source=stack-comment).

sgrimm:
https://github.com/terraware/terraware-server/actions/runs/5260420438/jobs/9507630468 has a test run where I modified the deploy script to run `ls -l /` instead of actually copying code, to prove that it could successfully ssh to the staging hosts.

sgrimm:
Once this lands and we use it successfully, I'll change the remaining Terraware repos and then we should be able to get rid of the bastion hosts.

charlie-tf (approved):
Neat!

PR #1093, opened 2023-04-24
SW-3493 Fix duplicate single quotes in notifications

sgrimm:
The `MessageFormat` class, which is used to format localized messages that contain
placeholders, requires single quote characters to be escaped. By default, Spring
only uses `MessageFormat` if there are actually placeholders. Unfortunately, that
means the escaping rule for single quote characters depends on whether or not the
string happens to have placeholders.

This was causing the added-to-organization notification to have an English title
of, `You''ve been added to a new organization!` (note the two single quote
characters).

Make the escaping behavior consistent by configuring the resource bundle to always
use `MessageFormat` even if there are no placeholders.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
+++ b/src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
@@ -29,6 +29,9 @@ data class NotificationMessage(val title: String, val body: String)
 class Messages {
   private val messageSource =
       ResourceBundleMessageSource().apply {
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1093** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1093" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1093?utm_source=stack-comment).

karthikbtf (approved):

PR #1094, opened 2023-04-25
SW-3504 Omit plots from planting site payloads

sgrimm:
For a real planting site, the list of monitoring plots will be too large to
reasonably include in its entirety when a client fetches a site's information.
Remove them from the planting site API.

Monitoring plots may still be queried using the search API.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -182,7 +182,7 @@ class AdminController(
 
   @GetMapping("/plantingSite/{plantingSiteId}")
   fun getPlantingSite(@PathVariable plantingSiteId: PlantingSiteId, model: Model): String {
-    val plantingSite = plantingSiteStore.fetchSiteById(plantingSiteId)
+    val plantingSite = plantingSiteStore.fetchSiteById(plantingSiteId, true)
     val organization = organizationStore.fetchOneById(plantingSite.organizationId)
 
     val allOrganizations =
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -182,7 +182,7 @@ class AdminController(
 
   @GetMapping("/plantingSite/{plantingSiteId}")
   fun getPlantingSite(@PathVariable plantingSiteId: PlantingSiteId, model: Model): String {
-    val plantingSite = plantingSiteStore.fetchSiteById(plantingSiteId)
+    val plantingSite = plantingSiteStore.fetchSiteById(plantingSiteId, true)
```
karthikbtf, line 185: `false`? Or is this the API to fetch all monitoring plots?

sgrimm: This is for the admin UI, where we want to have the full data set (and where it doesn't matter if that ends up being huge).

sgrimm:
Updated the PR description to clarify that monitoring plots aren't being removed from the search API. But yes, there will be additional APIs to fetch the relevant plots (where "relevant" means taking into account observations, fixed vs. temporary plots, etc.)

karthikbtf:
> Updated the PR description to clarify that monitoring plots aren't being removed from the search API. But yes, there will be additional APIs to fetch the relevant plots (where "relevant" means taking into account observations, fixed vs. temporary plots, etc.)

Thanks 👍

PR #1096, opened 2023-04-25
Add GeoJSON download links to admin UI

sgrimm:
Currently, the admin UI's planting site details page shows a nested list that
includes a line item for each monitoring plot. A real planting site will have tens
of thousands of monitoring plots, making this display uselessly large.

Replace the list of individual monitoring plots with a link to download map data
for a planting zone.

The map data is in GeoJSON format and can be viewed using mapping tools such as
geojson.io. This is mostly for development purposes; the user-facing web app won't
have a view that shows thousands of monitoring plots all at once.

Downloads are per-zone rather than per-site just because the file size would be
unmanageably huge for a real planting site.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -30,6 +30,7 @@ import com.terraformation.backend.db.default_schema.tables.pojos.DeviceTemplates
 import com.terraformation.backend.db.default_schema.tables.pojos.DevicesRow
 import com.terraformation.backend.db.seedbank.StorageLocationId
 import com.terraformation.backend.db.tracking.PlantingSiteId
+import com.terraformation.backend.db.tracking.PlantingZoneId
 import com.terraformation.backend.device.DeviceManagerService
 import com.terraformation.backend.device.DeviceService
 import com.terraformation.backend.device.db.DeviceManagerStore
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1096** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1096" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1096?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -210,6 +214,54 @@ class AdminController(
...
+  ): Map<String, Any> {
+    response.addHeader(
+        HttpHeaders.CONTENT_DISPOSITION, """attachment; filename="zone-${plantingZoneId}.json"""")
+
+    val zone =
+        plantingSiteStore.fetchSiteById(plantingSiteId).plantingZones.first {
+          it.id == plantingZoneId
+        }
+
+    // Return a GeoJSON document that can be viewed using tools such as geojson.io.
```
sgrimm, line 234: If we decide to support additional GeoJSON downloads, it'll make sense to refactor this into something more general, but it didn't seem worth doing to me for this one-off use case.

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -210,6 +214,54 @@ class AdminController(
...
+    return mapOf(
+        "type" to "FeatureCollection",
+        "features" to
+            listOf(
+                mapOf(
+                    "type" to "Feature",
+                    "properties" to emptyMap<String, Any>(),
```
karthikbtf, line 241: In case it helps with debug use-cases, we can add the name of the zones and subzones. Of course we can't do that for monitoring plots, so maybe this partial suggestion is not so useful.

karthikbtf (approved):
Looks good to me 👍

sgrimm:
Going to abandon this in favor of embedding a basic Mapbox viewer in the admin UI.

PR #1097, opened 2023-04-26
Use WGS84 coordinates in tracking tables

sgrimm:
Our API always uses WGS84 (degrees latitude and longitude) when it returns
geometry values to the client, since GeoJSON uses that coordinate system.  But
we've been transforming coordinates to "web Mercator" in the database at write
time, meaning we have to do on-the-fly coordinate system transformation every
time we read a geometry value.

The cost of those transformations is negligible for small amounts of data, but
becomes significant in the face of large planting sites with tens of thousands
of monitoring plots.

Change the code to store WGS84 coordinates in the database instead. Remove
the read-time transformations since they're no longer needed.


sgrimm:
* main
  * **PR #1097** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1097" target="_blank">Graphite</a>  👈
    * **PR #1098** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1098" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1097?utm_source=stack-comment).

karthikbtf (approved):
Nice!

PR #1098, opened 2023-04-26
SW-3491 Support tracking v2 shapefile import

sgrimm:
Update the shapefile importer to support the tracking v2 hierarchy with subzones
and monitoring plots. Import of v1 shapefiles is no longer supported.

The top three levels of the hierarchy (site, zones, and subzones) are loaded from
shapefiles. Monitoring plots are generated programmatically.

The planting site page in the admin UI now shows a map view of the planting site.
It no longer shows a full list of all monitoring plots, just a count of them for
each subzone.

This also introduces a change to how the names of monitoring plots and subzones
are sorted. The names are arbitrary text values and are alphanumeric in the
initial shapefiles we'll be importing, but per the requirements doc, they're
supposed to be integers. We want to sort them in numeric order if the shapefile
was compliant with the spec, but still allow non-numeric names so we can import
the initial set of shapefiles. So we use a collation that detects numbers in
strings and sorts them numerically, while still sorting the non-numeric parts
alphabetically.

For now, we don't have any anonymized example shapefiles to use in automated
tests, so most of the validation tests are removed.


sgrimm:
* main
  * **PR #1097** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1097" target="_blank">Graphite</a> 
    * **PR #1098** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1098" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1098?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -40,11 +43,12 @@ class PlantingSiteStore(
 
   fun fetchSiteById(
       plantingSiteId: PlantingSiteId,
+      includeSubzones: Boolean = false,
```
karthikbtf, line 46: should we set this to true in the api controller or support an option?

karthikbtf (approved):
Looks great! I have one question on exposing subzones in the api.

src/main/resources/templates/admin/organization.html
```diff
@@ -75,7 +75,7 @@ <h4>Create Planting Site</h4>
     <br/>
     <label for="zipfile">Zipfile</label>
     <input type="file" name="zipfile" id="zipfile" required/>
-    Must contain three shapefiles and their associated supplementary datafiles.
+    Must contain four shapefiles and their associated supplementary datafiles.
```
charlie-tf: Should this actually be three?

charlie-tf (approved):
Looks good. Just a quick question

src/main/resources/templates/admin/organization.html
sgrimm: Good catch, thanks. You can tell my initial stab at this was to load monitoring plots from the shapefiles!

charlie-tf (approved):

PR #1109, opened 2023-05-02
SW-3521 Store monitoring plot boundary as Polygon

sgrimm:
When we thought we were going to be loading monitoring plot boundaries from
shapefiles, it made sense to allow complex geometries and normalize them to
MultiPolygon. But now that we're generating them programmatically, we can
guarantee they will never have multiple polygons; storing them that way adds
needless overhead.

Change the geometry type to Polygon and convert any existing monitoring plots
by extracting the first (and only) Polygon from the plot's MultiPolygon.


sgrimm:
* main
  * **PR #1109** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1109" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1109?utm_source=stack-comment).

PR #1110, opened 2023-05-02
Organize database migrations into subdirectories

sgrimm:
The database migrations directory is getting harder to navigate over time.
Group the migrations into sets of 50 in subdirectories.

Flyway automatically scans subdirectories when it's searching for migrations to
run. It only stores migration filenames and checksums, not full paths, so moving
the files to subdirectories doesn't cause them to be flagged as modified.

PR #1111, opened 2023-05-02
SW-3521 Generate permanent monitoring plot clusters

sgrimm:
Group monitoring plots into 2x2 clusters. If all four plots in a cluster fall
within a single subzone, that cluster is eligible to be chosen as a set of
permanent monitoring plots.

An eligible cluster is identified by the presence of a permanent monitoring plot
cluster number. All four plots in a cluster will have the same cluster number.
The cluster numbers are randomized at import time.

The admin UI is updated to demonstrate/test the selection process; currently it
is hardwired to simulate 6 clusters of permanent plots per planting zone and
highlight them on the map.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
@@ -393,16 +393,32 @@ class PlantingSiteImporter(
    * given cluster number in a given planting zone.
    *
    * The cluster numbers allow permanent monitoring plots to be allocated using a SQL query. If we
-   * want 5 clusters of permanent plots from planting zone 4321, we can do a query along the lines
+   * want 3 clusters of permanent plots from planting zone 4321, we can do a query along the lines
    * of
    *
-   *     SELECT mp.id
+   *     SELECT mp.id, mp.permanent_cluster, mp.permanent_cluster_subplot
    *     FROM tracking.monitoring_plots mp
    *     JOIN tracking.planting_subzones ps ON mp.planting_subzone_id = ps.id
    *     WHERE ps.planting_zone_id = 4321
-   *     AND mp.permanent_cluster <= 5;
+   *     AND mp.permanent_cluster <= 3
+   *     ORDER BY mp.permanent_cluster, mp.permanent_cluster_subplot;
    *
-   * and get back a list of all 20 monitoring plots in 5 randomly-chosen 2x2 clusters.
+   * and get back a result like
+   *
+   * | id   | cluster | subplot |
+   * |------|---------|---------|
+   * | 1756 | 1       | 1       |
+   * | 1757 | 1       | 2       |
+   * | 1758 | 1       | 3       |
+   * | 1759 | 1       | 4       |
+   * | 781  | 2       | 1       |
+   * | 782  | 2       | 2       |
+   * | 783  | 2       | 3       |
+   * | 784  | 2       | 4       |
+   * | 360  | 3       | 1       |
+   * | 361  | 3       | 2       |
+   * | 362  | 3       | 3       |
+   * | 363  | 3       | 4       |
    */
   private fun assignPlots(
       allClusters: Collection<Cluster>,
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
@@ -122,7 +122,7 @@ class PlantingSiteImporter(
     val siteFeature = getSiteBoundary(siteFile, validationOptions, problems)
     val zonesByName = getZones(siteFeature, zonesFile, validationOptions, problems)
     val subzonesByZone = getSubzonesByZone(zonesByName, subzonesFile, validationOptions, problems)
-    val plotsBySubzone = generatePlotsBySubzone(siteFeature, subzonesFile)
+    val plotBoundaries = generatePlotBoundaries(siteFeature)
 
     if (problems.isNotEmpty()) {
       throw PlantingSiteUploadProblemsException(problems)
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #1110** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1110" target="_blank">Graphite</a> 
    * **PR #1111** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1111" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1111?utm_source=stack-comment).

sgrimm:
Going to revise this a little bit based on some Slack discussion about plot numbering. Changes should be small.

sgrimm:
Revisions are done.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
```diff
@@ -386,83 +376,167 @@ class PlantingSiteImporter(
...
+    // each plot is in as well as its plot number within that subzone. Plots are still grouped in
+    // 2x2 clusters at this point (though a given cluster might have fewer than 4 plots if some of
+    // them didn't fall within the subzone).
+    val clustersInSubzones: List<List<MonitoringPlotsRow>> =
+        subzoneRows.flatMap { subzoneRow ->
+          var plotNumber = 0
+
+          val subzoneClusters =
+              allPlots
+                  .map { cluster -> cluster.filter { it.coveredBy(subzoneRow.boundary!!) } }
```
karthikbtf: do we want to remove clusters that were already considered for a subzone row - from allPlots? (for runtime perf, only if it makes sense. . i understand there's a cost to the list manipulation... and these are one-time imports.)

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
```diff
@@ -386,83 +376,167 @@ class PlantingSiteImporter(
...
+   * 2x2 groups. Only plots that fall within the planting site are included.
    *
-   * This effectively divides the site into a grid of 25m squares and returns a polygon for each
-   * square.
+   * This effectively divides the site into a grid of 50m squares, divides each square into four 25m
+   * squares, and returns a polygon for each of the smaller squares.
    */
-  private fun generateAllPlots(siteFeature: ShapefileFeature): List<Polygon> {
-    val plots = mutableListOf<Polygon>()
+  private fun generatePlotBoundaries(siteFeature: ShapefileFeature): List<List<Polygon>> {
```
karthikbtf: just for discussion, any benefit to introducing clusters as a data construct here.. ?return a list of clusters vs list of list of polygons.

karthikbtf (approved):
This looks interesting and a fun problem to solve.
I understand how it works and looks good to me.
Some theoretical comments mostly for my understanding.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
sgrimm: That's definitely a reasonable question! It would make the data flow a bit more complex, but I'll try it locally and see how it looks.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
sgrimm: Yeah, that might make this easier to follow. Good suggestion.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
```diff
@@ -386,83 +376,167 @@ class PlantingSiteImporter(
...
+   *
+   * The cluster numbers allow permanent monitoring plots to be allocated using a SQL query. If we
+   * want 5 clusters of permanent plots from planting zone 4321, we can do a query along the lines
+   * of
+   *
+   *     SELECT mp.id
+   *     FROM tracking.monitoring_plots mp
+   *     JOIN tracking.planting_subzones ps ON mp.planting_subzone_id = ps.id
+   *     WHERE ps.planting_zone_id = 4321
+   *     AND mp.permanent_cluster <= 5;
```
charlie-tf: Out of curiosity, will the monitoring plots within a cluster all have contiguous ids? Thus not requiring a GROUP BY here to make sure they aren't mixed up?

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
sgrimm: I tried doing this and it added too much complexity for my taste. The performance win wasn't really noticeable since inserting the database rows is what takes the lion's share of the time.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
sgrimm: They will indeed have contiguous IDs, but I don't think that would matter in the context of this query, unless I'm not understanding the question. Do you mean `ORDER BY` (so the IDs aren't returned in an undefined order)?

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
karthikbtf: Thanks for giving this a try. Makes sense that the db operations take most of the time.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
```diff
@@ -605,4 +606,14 @@ class PlantingSiteImporter(
     ZonesDoNotOverlap("Zones do not overlap"),
     ZonesHaveSubzones("Zones have at least one subzone each"),
   }
+
+  /**
+   * A cluster of up to four monitoring plots. This is a simple wrapper around a list; it's purely
+   * for purposes of code clarity.
+   */
+  private class Cluster(private val plots: List<MonitoringPlotsRow>) :
```
karthikbtf, line 630: Thanks, looks good to me 🙏

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
charlie-tf: I meant that, if the ids weren't contiguous, then one would have to add `mp.permanent_cluster` to the select, I would think, in order to distinguish what cluster each plot belonged to. If contiguous then you could work it out from the ids alone. Not meaning to nitpick, but just trying to understand the example query here

sgrimm: Oh, I see. I'll clarify that, thanks.

sgrimm: Updated the example; does it make more sense now?

charlie-tf: Ah yeah, the example result really helps! Thanks

PR #1112, opened 2023-05-03
SW-3525 Add new growth forms to the supported list

karthikbtf:
- Add Fungi, Lichen and Moss to supported growth forms enum
- Updated csv to reflect these forms
- Added new growth forms to the english enum properties file


src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -76,7 +76,10 @@ VALUES (1, 'Tree'),
        (2, 'Shrub'),
        (3, 'Forb'),
        (4, 'Graminoid'),
-       (5, 'Fern')
+       (5, 'Fern'),
+       (6, 'Fungi'),
```
sgrimm: Are we sure we want "Fungi" instead of "Fungus?" All the other options are singular nouns.

sgrimm (approved):
One nitpicky English grammar question but if we don't care, this seems good to go.

src/main/resources/db/migration/R__TypeCodes.sql
karthikbtf: Good question, I will check with Allison. Thanks!!

sgrimm (approved):
🍄

PR #1114, opened 2023-05-04
SW-3535 Add permissions for planting zones

sgrimm:
In preparation for being able to configure the number of monitoring plots on a
per-planting-zone basis, add read and update permission logic for planting zones.
There are no create or delete permissions since planting zones are only created
via shapefile import, and are never deleted independently of the planting site.

PR #1115, opened 2023-05-04
SW-3535 Support per-planting-zone plot counts

sgrimm:
For each planting zone, keep track of how many permanent and temporary monitoring
plots will be needed for the next observation.

Ordinarily, these numbers will be calculated based on statistical information
about the planting zone, expressed as variance and error margin along with a value
that represents the desired confidence level (Student's t). The confidence level
is expected to stay the same across all planting sites, so it is initialized to
the correct value when a shapefile is imported. The other two values will be
supplied by the science team prior to the first observation of a planting site.

Those numbers are also stored on a per-planting-zone basis, and can be edited in
the admin UI. The admin UI uses those numbers to calculate the number of permanent
monitoring plots, which is then used to calculate the number of temporary plots.
These calculations are done client-side in the admin UI so as to allow the admin
to manually adjust the plot counts before saving if needed. The plot counts can
also be entered directly without using the formula at all.

The map in the admin UI shows the correct number of permanent clusters for each
planting zone rather than a hardwired number of them.


sgrimm:
* main
  * **PR #1114** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1114" target="_blank">Graphite</a> 
    * **PR #1115** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1115" target="_blank">Graphite</a>  👈
      * **PR #1116** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1116" target="_blank">Graphite</a> 
        * **PR #1117** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1117" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1115?utm_source=stack-comment).

PR #1120, opened 2023-05-08
SW-3527 Allow Dashes in Scientific Names During Import

charlie-tf:
Update the import validator to allow dashes in scientific names. We allow for
all three types of dashes that might be encountered, hyphen, en-, and em-.
Also update the tests to account for this change.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/model/Validation.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/model/Validation.kt
@@ -1,6 +1,6 @@
 package com.terraformation.backend.species.model
 
-private val invalidScientificNameChars = Regex("[^A-Za-z. —–-]")
+private val invalidScientificNameChars = Regex("[^A-Za-z. -]")
 
 /**
  * Validates the syntax of a scientific name. This does not check for higher-level conditions such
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/terraformation/backend/species/model/Validation.kt
```diff
@@ -1,6 +1,6 @@
 package com.terraformation.backend.species.model
 
-private val invalidScientificNameChars = Regex("[^A-Za-z. ]")
+private val invalidScientificNameChars = Regex("[^A-Za-z. —–-]")
```
charlie-tf: Note: github doesn't render the en- and em- dashes differently. IntelliJ does, though, I was happily surprised to see

sgrimm (changes requested):
I think if we're going to allow multiple kinds of dashes, we need to normalize them to hyphens.

Otherwise if someone imports a species with an em-dash, they will be baffled when it doesn't show up in searches because they typed a hyphen in the search box.

Also, you'd be able to import "duplicate" species names by using different dashes, and it might not be obvious why a species is showing up twice in the species list.

charlie-tf:
> I think if we're going to allow multiple kinds of dashes, we need to normalize them to hyphens.
> 
> Otherwise if someone imports a species with an em-dash, they will be baffled when it doesn't show up in searches because they typed a hyphen in the search box.
> 
> Also, you'd be able to import "duplicate" species names by using different dashes, and it might not be obvious why a species is showing up twice in the species list.

Those are great points. I'll figure out how to do that 👍

charlie-tf:
I noticed that species added through the frontend client, instead of the importer, are not validated at all. That should be fixed, probably, and I'll create a ticket to capture. EDIT: ticket is [here](https://terraformation.atlassian.net/browse/SW-3554)

sgrimm (approved):
Great, that should cause less confusion.

PR #1121, opened 2023-05-08
SW-3548 Calculate planting site areas

sgrimm:
Planting density is calculated as number of plants per unit of area, so we need to
know the areas of planting sites and their subdivisions. This can be calculated on
the fly from the boundaries, but it never changes, so calculate it at import time.

### DIFF
--- a/src/main/resources/db/migration/0151/V188__PlantingSiteAreas.sql
+++ b/src/main/resources/db/migration/0151/V188__PlantingSiteAreas.sql
@@ -1,0 +1,23 @@
+ALTER TABLE tracking.planting_sites ADD COLUMN area_ha NUMERIC;
+ALTER TABLE tracking.planting_zones ADD COLUMN area_ha NUMERIC;
+ALTER TABLE tracking.planting_subzones ADD COLUMN area_ha NUMERIC;
+
+-- Area must be calculated during shapefile import, so we don't want it to have a default,
+-- but sites in existing test databases need to be made valid.
+UPDATE tracking.planting_zones
+SET area_ha = 1
+WHERE area_ha IS NULL;
+
+UPDATE tracking.planting_subzones
+SET area_ha = 1
+WHERE area_ha IS NULL;
+
+ALTER TABLE tracking.planting_zones ALTER COLUMN area_ha SET NOT NULL;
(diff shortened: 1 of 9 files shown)


sgrimm:
* main
  * **PR #1121** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1121" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1121?utm_source=stack-comment).

src/main/resources/db/migration/0151/V188__PlantingSiteAreas.sql
```diff
@@ -0,0 +1,23 @@
...
+SET area_ha = 1
+WHERE area_ha IS NULL;
+
+ALTER TABLE tracking.planting_zones ALTER COLUMN area_ha SET NOT NULL;
+ALTER TABLE tracking.planting_zones ALTER COLUMN boundary SET NOT NULL;
+
+ALTER TABLE tracking.planting_subzones ALTER COLUMN area_ha SET NOT NULL;
+ALTER TABLE tracking.planting_subzones ALTER COLUMN boundary SET NOT NULL;
+
+ALTER TABLE tracking.planting_sites ADD CONSTRAINT area_positive CHECK (area_ha > 0);
```
karthikbtf, line 21: if area is null, will this check be ok?

karthikbtf (approved):
lgtm, one question

karthikbtf (approved):
lgtm, one question

src/main/resources/db/migration/0151/V188__PlantingSiteAreas.sql
sgrimm: Yes, it works thanks to SQL's null semantics. `area_ha > 0` evaluates to null rather than true or false if `area_ha` is null, and the constraint only fails if the condition evaluates to false.
```
terraware=# create temporary table foo (x integer, check (x > 0));
CREATE TABLE
terraware=# insert into foo values (1);
INSERT 0 1
terraware=# insert into foo values (0);
ERROR:  new row for relation "foo" violates check constraint "foo_x_check"
DETAIL:  Failing row contains (0).
terraware=# insert into foo values (null);
INSERT 0 1
```

PR #1123, opened 2023-05-09
Use separate map layer for permanent plots

sgrimm:
When showing the map of a planting site in the admin UI, we highlight permanent
monitoring plots. Previously, this was done by selecting the line color, but that
meant that on the grid of plots, only the left and bottom borders were colored
because the right and top ones were overwritten by other grid cells.

Add a separate layer for permanent plots so they show up more distinctly on the
map. Restructure the way the plot types are labeled in the GeoJSON data so that
it will be possible to do the same for temporary plots later.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -278,14 +278,21 @@ class AdminController(
                 val numPermanent = zone.numPermanentClusters ?: 0
 
                 zone.plantingSubzones.flatMap { subzone ->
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1123** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1123" target="_blank">Graphite</a>  👈
    * **PR #1124** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1124" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1123?utm_source=stack-comment).

PR #1126, opened 2023-05-09
SW-3509 Handle duplicate accession photo filenames

sgrimm:
Initially, all the metadata about accession photos was stored in a single table
that had a unique constraint to enforce that a given accession could only have
one photo with a given filename. When the table was split up to add support for
nursery withdrawal photos, we lost that unique constraint and thus the server
started accepting uploads of accession photos with duplicate filenames.

When a client attempted to read one of those photos, the request would fail
because the server still assumed filenames were unique and a database query
that was expected to return a single row instead returned multiple rows.

Change the upload API so that uploading a photo overwrites any existing photo 
with the same filename on the same accession. To handle the duplicates that
already exist in the database, update the database queries to only return the
newest file with a given filename.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/PhotosController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/PhotosController.kt
@@ -4,13 +4,11 @@ import com.fasterxml.jackson.annotation.JsonInclude
 import com.terraformation.backend.api.ApiResponse200Photo
 import com.terraformation.backend.api.ApiResponse404
 import com.terraformation.backend.api.ApiResponseSimpleSuccess
-import com.terraformation.backend.api.DuplicateNameException
 import com.terraformation.backend.api.PHOTO_MAXHEIGHT_DESCRIPTION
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #1126** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1126" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1126?utm_source=stack-comment).

karthikbtf (approved):
👍

charlie-tf (approved):
Looks good. Nice test

sgrimm:
Updated this to have "overwrite existing file" semantics as we discussed in Slack. It's different enough to warrant a fresh review.

karthikbtf (approved):
Looks good to me 👍 
Do we need to clean up dupes in the db?

sgrimm:
Existing duplicates shouldn't hurt anything. The code needs to be able to deal with duplicates at read time no matter what, since there will be a short period after the second copy of a file is uploaded and before the first copy is deleted where both of them will exist.

For context, there are 4 duplicate filenames on prod and 9 on staging.

PR #1128, opened 2023-05-09
SW-3219 Send a 'create account' email to new users when added to an organization

karthikbtf:
- we currently send an email with a link to terraware/org-id in params (upon org invitation), even if the user is brand-new to terraware
- feature request is to send a different email that links to the terraware registration page for brand-new users
- also update content to reflect it being a new account that needs to be created, Allison has ok'd the copy
- PR to pre-fill registration page with email https://github.com/terraware/keycloak/pull/35

### original email
original-email

### new email (signup, setup have been renamed to create account)
DEV  You've been added to a new organization in Terraware! - karthik balach@gmail com - Gmail 2023-05-10 11-05-15

### sign up form from the link
Sign Up 2023-05-10 11-04-29

### sample registration link generated by `WebAppUrls`
https://auth.staging.terraware.io/realms/terraware/protocol/openid-connect/registrations?client_id=dev-terraware-server&email=karthik%2Bmay10%40terraformation.com&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fhome?organizationId%3D1&response_type=code&scope=openid


src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -129,6 +130,30 @@ class AppNotificationService(
         organization.id)
   }
 
+  @EventListener
+  fun on(event: UserAddedToTerrawareEvent) {
```
karthikbtf, line 118: we could also re-route this through the listener for UserAddedToOrganization event (or extract some common version out)

src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
```diff
@@ -39,12 +40,34 @@ class OrganizationService(
...
+            false
+          }
+
       val user = userStore.fetchOrCreateByEmail(email)
 
       organizationStore.addUser(organizationId, user.userId, role)
 
-      publisher.publishEvent(
-          UserAddedToOrganizationEvent(user.userId, organizationId, currentUser().userId))
+      if (isNewUser) {
```
karthikbtf, line 54: i debated adding this boolean into the UserAddedToOrganizationEvent but decided to create a new event for separation

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -159,6 +161,25 @@ class EmailNotificationService(
...
+            ?: throw IllegalArgumentException("Admin user must be an individual user")
+    val user =
+        userStore.fetchOneById(event.userId) as? IndividualUser
+            ?: throw IllegalArgumentException("User must be an individual user")
+    val organization = organizationStore.fetchOneById(event.organizationId)
+
+    val terrawareRegistrationUrl =
+        webAppUrls.terrawareRegistrationUrl(event.organizationId).toString()
+
+    emailService.sendUserNotification(
```
karthikbtf, line 177: should this be a keycloak delegation instead?

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -18,19 +19,34 @@ import javax.ws.rs.core.UriBuilder
...
+    val orgHome = fullOrganizationHome(organizationId)
+    val realmBaseUrl =
+        URI(
+            "${keycloakInfo.realmBaseUrl.toString().trimEnd('/')}/protocol/openid-connect/registrations")
+    return UriBuilder.fromUri(realmBaseUrl)
+        .queryParam("client_id", keycloakInfo.clientId)
+        .queryParam("redirect_uri", orgHome)
+        .queryParam("response_type", "code")
+        .queryParam("scope", "openid")
+        .build()
```
karthikbtf: this isn't ideal

sgrimm:
Doesn't seem like there's a super clean way to send someone directly to the registration page with the form already partially filled in. Googling around a bit, it seems like one way people do it is to add JavaScript to the registration page to copy the email address from a query parameter in the URL. Ugly but might be the simplest approach.

karthikbtf:
> Doesn't seem like there's a super clean way to send someone directly to the registration page with the form already partially filled in. Googling around a bit, it seems like one way people do it is to add JavaScript to the registration page to copy the email address from a query parameter in the URL. Ugly but might be the simplest approach.

Thanks, Steve! I will give that a try in the keycloak repo. Guessing we have to sanitize the data.
Will clean this PR a bit as well.

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -18,19 +20,35 @@ import javax.ws.rs.core.UriBuilder
...
 
+  fun terrawareRegistrationUrl(organizationId: OrganizationId, email: String): URI {
+    val orgHome = fullOrganizationHome(organizationId)
+    val realmBaseUrl =
+        URI(
+            "${keycloakInfo.realmBaseUrl.toString().trimEnd('/')}/protocol/openid-connect/registrations")
+    return UriBuilder.fromUri(realmBaseUrl)
+        .queryParam("client_id", keycloakInfo.clientId)
+        .queryParam("email", URLEncoder.encode(email, "UTF-8"))
+        .queryParam("redirect_uri", orgHome)
```
karthikbtf, line 46: I don't believe this does anything but is required by the protocol.
Upon registration, keycloak redirects user to a page saying 'please verify your email'.

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -18,19 +20,35 @@ import javax.ws.rs.core.UriBuilder
...
-  /** Generates a relative path of organization home within the web app */
   fun organizationHome(organizationId: OrganizationId): URI {
     return UriBuilder.fromPath("/home").queryParam("organizationId", organizationId).build()
   }
 
+  fun terrawareRegistrationUrl(organizationId: OrganizationId, email: String): URI {
+    val orgHome = fullOrganizationHome(organizationId)
+    val realmBaseUrl =
+        URI(
+            "${keycloakInfo.realmBaseUrl.toString().trimEnd('/')}/protocol/openid-connect/registrations")
```
sgrimm: Can we construct the path using UriBuilder?

src/test/kotlin/com/terraformation/backend/customer/OrganizationServiceTest.kt
```diff
@@ -232,4 +237,57 @@ internal class OrganizationServiceTest : DatabaseTest(), RunsAsUser {
...
+
+    val otherUser = userStore.fetchByEmail(otherUserEmail)
+    assertNotNull(otherUser, "No new user")
+
+    insertOrganization(organizationId)
+
+    every { user.canAddOrganizationUser(organizationId) } returns true
+    every { user.canSetOrganizationUserRole(organizationId, Role.Contributor) } returns true
+    // mock missing user on initial check
+    every { userStore.fetchByEmail(otherUserEmail) } returnsMany listOf(null, otherUser)
```
sgrimm: Would it work to remove the `insertUser()` for the other user, such that the user genuinely wouldn't exist at first?

sgrimm (approved):
A couple minor comments, but it seems good in general.

src/test/kotlin/com/terraformation/backend/customer/OrganizationServiceTest.kt
karthikbtf: I started out that way but was running into db errors (insert was failing on duplicate users_pkey, which I thought was strange - since that's a db generated id). Will revisit this, more alert now than when I last worked on this test.

karthikbtf: Was able to improve this test. 
Figured out what was happening.
The mock currentUser has a user id 2. When the test adds a new user to an org, the user is inserted into the db without an id. The db auto-increments the primary key (last value of primary key as part of auto-incremented index was 1 for system user insertion). The next incremented id is 2, this conflicts with the mock currentUser's id... hence the error. 
I forced the mock currentUser to have an id of 200 for the Org service tests.

sgrimm (approved):
Changes seem good. Thanks.

PR #1131, opened 2023-05-11
Suppress load balancer cookie warning

sgrimm:
The Keycloak adapter uses an out-of-date version of the Apache HTTP client that
doesn't know how to deal with the expiration times on the cookies that are added
by the load balancer that sits in front of our Keycloak instance. It spits out
a warning log message every time the adapter talks to Keycloak, which happens
often enough to add significant log clutter.

The eventual fix will be to stop using the Keycloak adapter (which is deprecated)
but for now, configure the logger to filter out warnings from the class in
question.


sgrimm:
* main
  * **PR #1131** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1131" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1131?utm_source=stack-comment).

karthikbtf (approved):
👍  
i recall seeing those errors in my intellij console

PR #1133, opened 2023-05-12
SW-3600 Make monitoring plot config mandatory

sgrimm:
To avoid situations where nobody has entered the variance and error margin for a
planting zone and thus the system has no idea how many monitoring plots to assign,
make those values (and the monitoring plot counts derived from them) mandatory
with defaults rather than optional.

In addition, don't allow either the permanent or the temporary plot count to be
zero; we always want at least one permanent cluster and one temporary plot per
observation.


sgrimm:
* main
  * **PR #1133** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1133" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1133?utm_source=stack-comment).

charlie-tf (approved):
👍

PR #1134, opened 2023-05-12
Don't put temporary plots in unplanted subzones

sgrimm:
Since `PlantingZoneModel.chooseTemporaryPlots()` needs to know which subzones have
plants, we can make it filter out the ones that don't, rather than requiring the
caller to do so.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/model/Models.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/model/Models.kt
@@ -84,7 +84,13 @@ data class PlantingZoneModel(
    *   a multiple of the number of subzones) the remaining ones must be placed in the subzones that
    *   have the fewest number of permanent plots. If multiple subzones have the same number of
    *   permanent plots (including 0 of them), temporary plots should be placed in subzones that have
-   *   been planted before being placed in ones that haven't.
+   *   been planted before being "placed" in ones that haven't (but see the next point).
+   * - Only subzones that have been planted should have temporary plots. However, they should be
+   *   excluded only after all the preceding rules have been followed. That is, we want to choose
+   *   plots based on the rules above, and then filter out plots in unplanted subzones, as opposed
+   *   to only spreading plots across planted subzones. Otherwise, for a new project with a small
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1134** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1134" target="_blank">Graphite</a>  👈
    * **PR #1135** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1135" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1134?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/tracking/model/Models.kt
```diff
@@ -84,7 +84,13 @@ data class PlantingZoneModel(
...
    *   have the fewest number of permanent plots. If multiple subzones have the same number of
    *   permanent plots (including 0 of them), temporary plots should be placed in subzones that have
-   *   been planted before being placed in ones that haven't.
+   *   been planted before being "placed" in ones that haven't (but see the next point).
+   * - Only subzones that have been planted should have temporary plots. However, they should be
+   *   excluded only after all the preceding rules have been followed. That is, we want to choose
+   *   plots based on the rules above, and then filter out plots in unplanted subzones, as opposed
+   *   to only spreading plots across planted subzones. Otherwise, for a new project with a small
+   *   number of planted subzones, we would end up piling the entire planting zone's worth of
+   *   temporary plots into a handful of subzones.
```
karthikbtf, line 93: Just for my understanding of the workflow - since we end up with unassigned temporary plots - will this be run at various times later to get more coverage, as the number of subzones with plantings increase?

karthikbtf (approved):
Looks good to me based on comments/description 👍

src/main/kotlin/com/terraformation/backend/tracking/model/Models.kt
sgrimm: Yes, there will be an observation every 6 or 12 months, each of which will have a fresh set of temporary plots.

PR #1153, opened 2023-05-19
SW-3348 Replace Keycloak admin client

sgrimm:
The Keycloak admin client library comes with a bunch of transitive dependencies
that are going to cause problems when we upgrade to Spring Boot 3. We only use a
small portion of its functionality, so replace it with our own implementation.


sgrimm:
* main
  * **PR #1153** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1153" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1153?utm_source=stack-comment).

PR #1169, opened 2023-05-23
Accept canonical enum values in API payloads

sgrimm:
JSON deserialization of our reference table enums was expecting English
display names as values, rather than the non-localized names from the
`name` columns of the reference tables.

Historically, those two were the same since we didn't have any concept of
localizing an enum name. But now that we support localization, there is
a distinction between the enum string values in the API (which uses the
names from the database) and the human-readable values that get included
in exported CSV files and the like.

We were correctly using the names from the database in responses, and listing
those names as the acceptable values in the OpenAPI schema. But the actual
JSON deserialization was expecting the localized English names.

This caused the server to reject the value `CantTell` of the
`RecordedSpeciesCertainty` enum because its localized English name is "Can't
Tell". (Currently, that's the only enum value whose string representation
in the API differs from the localized English name.)

### DIFF
--- a/jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
+++ b/jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
@@ -133,6 +133,7 @@ class TerrawareGenerator : KotlinGenerator() {
               private val displayNames = ConcurrentHashMap<Locale, Map<$enumName, String>>()
               private val byLocalizedName = ConcurrentHashMap<Locale, Map<String, $enumName>>()
               private val byId = values().associateBy { it.id }
+              private val byApiName = values().associateBy { it.displayName }
               
               fun forDisplayName(name: String, locale: Locale): $enumName {
                 val valuesForLocale = byLocalizedName.getOrPut(locale) {
@@ -145,7 +146,10 @@ class TerrawareGenerator : KotlinGenerator() {
               
               @JsonCreator
               @JvmStatic
-              fun forDisplayName(name: String) = forDisplayName(name, Locale.ENGLISH)
+              fun forApiName(name: String): $enumName {
+                return byApiName[name]
+                    ?: throw IllegalArgumentException("Unrecognized value: ${dollarSign}name")
+              }
               
               fun forId(id: Int) = byId[id]
           }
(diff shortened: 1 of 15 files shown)


sgrimm:
* main
  * **PR #1169** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1169" target="_blank">Graphite</a>  👈
    * **PR #1170** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1170" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1169?utm_source=stack-comment).

charlie-tf (approved):
So if I understand correctly, this instructs the code generator to use the `forApiName` method to generate API names, which uses display names instead of localized names?

sgrimm:
There are three places where the mapping between JSON strings and enum values comes into play.

1. Serialization: When we're converting an enum value to JSON, the `@JsonValue` annotation on the `displayName` property (which, now that I'm writing this, I realize I should probably rename to `apiName` for clarity) causes Jackson to use the value of that property as the string representation. That property has the value from the `name` column in the corresponding database table.
2. Deserialization: When we're converting a JSON string to an enum value, the `@JsonCreator` annotation on the method that is now called `forApiName` causes Jackson to call that method to convert the string to an enum value. That method, as of this PR, looks up the enum value whose `displayName` is equal to the string. The bug was that it used to look up the enum value whose localized English name was equal to the string.
3. OpenAPI schema: The OpenAPI schema generator knows that we're using Jackson, and it knows about the `@JsonValue` annotation. So it lists all the `displayName` values as the allowed values for the enum. Didn't need to change anything here, since this was already correct.

charlie-tf:
👍 Cool, thanks for the additional explanation.

sgrimm:
Updated this to rename `displayName` to `jsonValue` since `displayName` no longer has anything to do with what that property is actually used for. The PR now touches a lot more code but hopefully it'll be a bit more self-evident what's going on with the various names.

sgrimm:
@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1169).

sgrimm:
@sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1169).

PR #1175, opened 2023-05-24
Track per-species observed plant totals

sgrimm:
We're going to be presenting various views of past observations that include
aggregated plant counts. Since a single monitoring plot can contain hundreds
of plants, and a single observation can contain dozens of monitoring plots,
doing the aggregation on the fly would be costly, especially for views that
show the results from many observations at once.

Add three tables to help reduce the aggregation cost. They record the number
of plants of each recording state (live, dead, existing), as well as total
plants and mortality rate (both of the latter excluding existing plants),
on a per-species, per-observation basis. Each table aggregates the results
at a different level of granularity: by monitoring plot, by planting zone,
and by planting site. They're updated (transactionally) when an observation
of a monitoring plot is completed.

For UI views that include per-species plant counts and mortality rates, we'll use
these new tables directly. For views that include total plant counts and mortality
rates across all species, we'll do on-the-fly aggregation using these tables; the
number of species in an observation will be small enough that this won't be
unacceptably costly.


sgrimm:
* main
  * **PR #1175** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1175" target="_blank">Graphite</a>  👈
    * **PR #1178** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1178" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1175?utm_source=stack-comment).

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -304,6 +304,15 @@ COMMENT ON COLUMN tracking.observations.completed_time IS 'Server-generated date
 COMMENT ON COLUMN tracking.observations.end_date IS 'Last day of the observation. This is typically the last day of the same month as `start_date`.';
 COMMENT ON COLUMN tracking.observations.start_date IS 'First day of the observation. This is either the first day of the month following the end of the planting season, or 6 months after that day.';
 
+COMMENT ON TABLE tracking.observed_plot_species_totals IS 'Aggregated per-monitoring-plot, per-species totals of plants recorded during observations.';
+COMMENT ON COLUMN tracking.observed_plot_species_totals.total_plants IS 'Total number of plants observed, not counting existing plants.';
```
karthikbtf: per-monitoring-plot?

src/main/resources/db/migration/0151/V193__ObservedSpeciesTotals.sql
```diff
@@ -0,0 +1,50 @@
...
+    total_existing INTEGER NOT NULL DEFAULT 0,
+    total_plants INTEGER NOT NULL DEFAULT 0,
+    mortality_rate INTEGER NOT NULL DEFAULT 0,
+
+    PRIMARY KEY (observation_id, monitoring_plot_id, species_id)
+);
+
+CREATE INDEX ON tracking.observed_plot_species_totals (observation_id);
+CREATE INDEX ON tracking.observed_plot_species_totals (monitoring_plot_id);
+CREATE INDEX ON tracking.observed_plot_species_totals (species_id);
```
karthikbtf, line 16: Just curious if this could be a function with parameters, or would that be more complex and less readable? Not requesting changes.

karthikbtf (approved):
Looks good to me 👍

src/main/resources/db/migration/0151/V193__ObservedSpeciesTotals.sql
sgrimm: It could have been, but like you said, it'd be more complex, and didn't seem worth it for a one-off migration like this. (Maybe if we were creating a dozen of these tables instead of three of them, the situation would be different.)

The main benefit would be to eliminate the chance that the tables would be defined inconsistently, but the tests will catch that and we'd still want the tests even if we used a function here.

karthikbtf: Thanks for the clarification, yes makes sense 👍

sgrimm:
@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1175).

sgrimm:
@sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1175).

PR #1176, opened 2023-05-25
SW-1031 Run tests that depend on external services

sgrimm:
By default, we skip tests like `S3FileStoreTest` in CI because we don't want our
CI builds to fail if the AWS API is having a glitch that's unrelated to the
changes we're making to our code. But that means we don't detect problems in
the code under test, or, more commonly, problems caused by upgrades of the client
libraries we use to talk to those services.

Add a build step that runs those tests (which are now required to have class names
ending in `ExternalTest` but doesn't cause the whole workflow to fail if they
don't pass. The failures will still be visible in the build log, so we can decide
on a case-by-case basis whether we care about them.

### DIFF
--- a/.github/scripts/set-environment.sh
+++ b/.github/scripts/set-environment.sh
@@ -5,15 +5,13 @@ set -euo pipefail
 commit_sha="${GITHUB_SHA:0:12}"
 docker_image='terraware/terraware-server'
 
+TIER=STAGING
+
 # Define tier based on branch ref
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #1176** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1176" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1176?utm_source=stack-comment).

charlie-tf (approved):
Nice, all makes sense

PR #1177, opened 2023-05-25
SW-3555 Restrict admin UI to super admins

sgrimm:
The time has long since passed when it was necessary to use the admin UI for
organization-level administrative operations like adding users, but the admin UI
remained accessible to organization admins anyway. Permissions were still enforced
correctly, so there wasn't a security concern, but the UI is ugly and there's no
reason we should be exposing it to end users.

Change the security configuration so it is only accessible by super-admin users.

There are no changes to the admin UI itself here, but this change will let us
simplify some of the logic since we'll no longer need to, e.g., conditionally show
parts of the UI depending on whether or not the user is a super admin.


sgrimm:
* main
  * **PR #1177** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1177" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1177?utm_source=stack-comment).

charlie-tf (approved):
Looks good, thanks for the simplification

PR #1183, opened 2023-05-26
SW-3730 Observation results API

sgrimm:
Add GET endpoints to fetch observation results.

* `/api/v1/tracking/observations/results?organizationId={id}` gets the results 
  of all observations across all an organization's planting sites.
* `/api/v1/tracking/observations/results?plantingSiteId={id}` gets the results 
  of all observations of a single planting site.
* `/api/v1/tracking/observations/{id}/results` gets tthe results of a single
  observation.
  
The summary calculations are tested against the numbers from a spreadsheet we're
using as part of the requirements process.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
@@ -217,9 +217,10 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
               val subzones = record[plantingSubzoneMultiset]
 
               val isCompleted =
-                  subzones.all { subzone ->
-                    subzone.monitoringPlots.all { it.completedTime != null }
-                  }
+                  subzones.isNotEmpty() &&
+                      subzones.all { subzone ->
+                        subzone.monitoringPlots.all { it.completedTime != null }
+                      }
               val completedTime =
                   if (isCompleted) {
                     subzones.maxOf { subzone ->
@@ -284,7 +285,7 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
         .fetch { record ->
           val zones = record[plantingZoneMultiset]
 
-          val isCompleted = zones.all { it.completedTime != null }
+          val isCompleted = zones.isNotEmpty() && zones.all { it.completedTime != null }
           val completedTime =
               if (isCompleted) {
                 zones.maxOf { it.completedTime!! }
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
@@ -1,0 +1,346 @@
+package com.terraformation.backend.tracking.db
+
+import com.terraformation.backend.customer.model.IndividualUser
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.asNonNullable
+import com.terraformation.backend.db.default_schema.OrganizationId
+import com.terraformation.backend.db.default_schema.tables.references.USERS
+import com.terraformation.backend.db.forMultiset
+import com.terraformation.backend.db.tracking.ObservationId
+import com.terraformation.backend.db.tracking.PlantingSiteId
+import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOTS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVATIONS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PHOTOS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PLOTS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVED_PLOT_SPECIES_TOTALS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVED_ZONE_SPECIES_TOTALS
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONES
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_ZONES
+import com.terraformation.backend.tracking.model.MONITORING_PLOTS_PER_HECTARE
+import com.terraformation.backend.tracking.model.ObservationMonitoringPlotPhotoPayload
+import com.terraformation.backend.tracking.model.ObservationMonitoringPlotResultsPayload
+import com.terraformation.backend.tracking.model.ObservationMonitoringPlotStatus
+import com.terraformation.backend.tracking.model.ObservationPlantingSubzoneResultsPayload
+import com.terraformation.backend.tracking.model.ObservationPlantingZoneResultsPayload
+import com.terraformation.backend.tracking.model.ObservationResultsPayload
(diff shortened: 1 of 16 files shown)


sgrimm:
* main
  * **PR #1193** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1193" target="_blank">Graphite</a> 
    * **PR #1183** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1183" target="_blank">Graphite</a>  👈
      * **PR #1194** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1194" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1183?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -0,0 +1,346 @@
...
+          val totalPlants = totalIdentifiedPlants
+
+          ObservationResultsPayload(
+              completedTime = completedTime,
+              mortalityRate = mortalityRate,
+              observationId = record[OBSERVATIONS.ID.asNonNullable()],
+              plantingSiteId = record[OBSERVATIONS.PLANTING_SITE_ID.asNonNullable()],
+              plantingZones = zones,
+              startDate = record[OBSERVATIONS.START_DATE.asNonNullable()],
+              state = record[OBSERVATIONS.STATE_ID.asNonNullable()],
```
karthikbtf, line 370: There's a new comment in figma about rolling up the state (showing the "least" progress). https://www.figma.com/file/ZO2Y8oGdXQaInnKFRUpDLq?node-id=2004:235545#442294441

karthikbtf (commented):
Looks good so far. Curious what the abridged results will add to this.
A bit uncertain about completed times versus planting site time zones. Do they need to be transformed client-side or use as-is?

I am using this branch to create FE services/reducers, etc. will have more feedback soon.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
karthikbtf: After some review and cross referencing designs - and recollecting a discussion from the roundtable, here's my thought..
FE will be loading the full observations results for the org - and use internal redux selectors to provide the various views internally.
This is probably planned but just calling out that we need an endpoint for loading the monitoring plot photos.
`/api/v1/tracking/observations/{observationId}/plots/{plotId}/photos/{photoId}`?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -0,0 +1,346 @@
...
+            OBSERVATIONS.START_DATE,
+            OBSERVATIONS.STATE_ID,
+            plantingZoneMultiset)
+        .from(OBSERVATIONS)
+        .where(condition)
+        .orderBy(OBSERVATIONS.ID)
+        .fetch { record ->
+          val zones = record[plantingZoneMultiset]
+
+          val isCompleted = zones.all { it.completedTime != null }
```
karthikbtf: For an upcoming observation, isCompleted is evaluating to true (zones is empty), do we need an isEmpty() check as well?  Ends up throwing an exception in the zones.maxOf (guarded by isCompleted).

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
sgrimm: > There's a new comment in figma about rolling up the state (showing the "least" progress).

Thanks. I believe the observation-level state the server keeps track of should be consistent with that discussion.

sgrimm: > This is probably planned but just calling out that we need an endpoint for loading the monitoring plot photos.

Yep, that's definitely planned. Jira task is SW-3699 (let's see if GitHub turns that into a link!)

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
sgrimm: Good catch, thanks. This does indeed need to check for zones being non-empty.

sgrimm:
> Curious what the abridged results will add to this.

I was thinking about that in the context of the "list of all past observations" page in the UI, where you wouldn't want to fetch, say, the monitoring plot boundaries for every monitoring plot in every observation when you're not going to use that data for anything.

However, we can leave that as a future optimization; it will be impossible for the list of observations to be large any time soon given how scheduling works.

> A bit uncertain about completed times versus planting site time zones. Do they need to be transformed client-side or use as-is?

As things stand, they'd need to be transformed client-side, but we could do the transformation server-side instead.

karthikbtf:
> > Curious what the abridged results will add to this.
> 
> I was thinking about that in the context of the "list of all past observations" page in the UI, where you wouldn't want to fetch, say, the monitoring plot boundaries for every monitoring plot in every observation when you're not going to use that data for anything.
> 
> However, we can leave that as a future optimization; it will be impossible for the list of observations to be large any time soon given how scheduling works.
> 
> > A bit uncertain about completed times versus planting site time zones. Do they need to be transformed client-side or use as-is?
> 
> As things stand, they'd need to be transformed client-side, but we could do the transformation server-side instead.

Thanks Steve. Thought about this some more, I think we can do this client side - the display values are only month/year - client has to do some processing anyways. cc @charlie-tf I was incorrect in assuming that completed times were `yyyy-mm-dd` format (for the map slider). Here's an example completed time ` "completedTime": "2023-06-01T21:47:21.303929Z",`. We can have the client redux/reducer convert the dates to month/year format with time zone applied.

karthikbtf:
@sgrimm would it be a bunch of non-trivial work to return the corresponding names where we have ids (planting site, species, zone, subzone, monitoring plot)?
Trying to merge this client-side but the mixed state of redux/non-redux is complicating things a bit - still doable just trying to see which approach is cleaner.
cc @charlie-tf

sgrimm:
Yes, I can include names. I guess I'll need to include zone and site boundaries as well for the map views? (I'd been assuming the client would get those from the site data.)

karthikbtf:
> Yes, I can include names. I guess I'll need to include zone and site boundaries as well for the map views? (I'd been assuming the client would get those from the site data.)

Good question. Yes we have the site data. I guess it would be redundant to have multiple sources for site related data. Maybe the client should just merge the data for its various purposes (application of redux). I will handle the names and polygons client side. Cancel my request.

karthikbtf (approved):
Thanks so much 👍 
This was useful to cross-reference and confirm FE behavior/observation.

sgrimm:
@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1183).

sgrimm:
[Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1183) rebased this pull request as part of a merge.

sgrimm:
@sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1183).

PR #1187, opened 2023-06-05
Track "Other" and "Can't Tell" plant totals

sgrimm:
Previously, we were only calculating total observations of plants of known
species, but we need to be able to show totals for "Can't Tell" plants as well as
break the "Other" plant totals down by user-supplied species names.

Change the totals tables to include both species ID and name columns whose
nullability depends on the recorded species certainty.

Since there aren't any observations in production yet, this is done by dropping
and recreating the tables rather than altering them.


sgrimm:
* main
  * **PR #1187** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1187" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1187?utm_source=stack-comment).

PR #1191, opened 2023-06-06
EnumsTest shouldn't fail on extra values

sgrimm:
If `Enums_en.properties` has a key that doesn't match any enum values, but does
have keys for all the values of all the enums, `EnumsTest` should pass.

This was causing CI failures after we deleted an enum value; the old English
display name was still in Phrase's string list and Phrase added it to back to
the properties file on export.


sgrimm:
* main
  * **PR #1191** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1191" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1191?utm_source=stack-comment).

karthikbtf (approved):
👍 
noticed resurrection of deleted strings in web as well

PR #1192, opened 2023-06-07
SW-3730 Drop total plants from observed species totals

sgrimm:
The total number of observed plants can be calculated from the live, dead, and
existing totals.

The total was included as an explicit column in order to support exposing it
via the search API, but we've decided against using the search API to fetch
observation results, so it's no longer useful.


sgrimm:
* main
  * **PR #1192** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1192" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1192?utm_source=stack-comment).

charlie-tf (approved):

PR #1198, opened 2023-06-08
Start migration subdirectories at multiples of 50

sgrimm:
Make the database migration subdirectories start at multiples of 50 rather than at
multiples of 50 plus one. It's a little more intuitive if, e.g., migration 100 is
in the same directory as migration 101.

No changes to the contents of any migrations here, just renaming the directories
and moving the multiple-of-50-numbered migrations to their new homes.

PR #1197, opened 2023-06-08
SW-3731 Track plantings since last observation

sgrimm:
Replace the `planting_site_populations` and `planting_subzone_populations` views
with tables that are updated when there are outplanting nursery withdrawals, and
add an analogous table for zone-level totals.

In addition to the all-time totals, these tables also have a new column that
tracks the number of plants added since the last observation. This number
is also adjusted when there are outplanting nursery withdrawals, but is reset to
zero when an observation is completed.

Tracking these numbers explicitly will reduce the cost of the queries to populate
the plants dashboard, especially once we add the "plants planted since last
observation" stat.

Once this code is deployed, a followup migration will backfill the values for
existing planting sites. The backfill needs to happen separately to avoid the case
where an outplanting withdrawal happens while this code is in the middle of being
deployed, and happens to be handled by a previous release which would fail to
update the numbers.

PR #1199, opened 2023-06-08
SW-3731 Backfill planting site populations

sgrimm:
Backfill the plant totals for planting sites, zones, and subzones. This overwrites
any existing values with the correct totals across all plantings.

This should be deployed as its own release so that it only runs when all the
server instances are already aware of these tables.

### DIFF
--- a/src/main/resources/db/migration/0200/V200__BackfillPopulations.sql
+++ b/src/main/resources/db/migration/0200/V200__BackfillPopulations.sql
@@ -1,0 +1,50 @@
+INSERT INTO tracking.planting_subzone_populations (
+    planting_subzone_id,
+    species_id,
+    total_plants,
+    plants_since_last_observation)
+SELECT plantings.planting_subzone_id,
+       plantings.species_id,
+       SUM(plantings.num_plants),
+       SUM(plantings.num_plants)
+FROM tracking.plantings
+WHERE plantings.planting_subzone_id IS NOT NULL
+GROUP BY plantings.planting_site_id, plantings.planting_subzone_id, plantings.species_id
+ON CONFLICT (planting_subzone_id, species_id)
+    DO UPDATE
+    SET total_plants = excluded.total_plants,
+        plants_since_last_observation = excluded.plants_since_last_observation;
+
+INSERT INTO tracking.planting_zone_populations (
+    planting_zone_id,
+    species_id,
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1197** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1197" target="_blank">Graphite</a> 
    * **PR #1198** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1198" target="_blank">Graphite</a> 
      * **PR #1199** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1199" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1199?utm_source=stack-comment).

src/main/resources/db/migration/0200/V200__BackfillPopulations.sql
```diff
@@ -0,0 +1,50 @@
...
+    total_plants,
+    plants_since_last_observation)
+SELECT plantings.planting_subzone_id,
+       plantings.species_id,
+       SUM(plantings.num_plants),
+       SUM(plantings.num_plants)
+FROM tracking.plantings
+WHERE plantings.planting_subzone_id IS NOT NULL
+GROUP BY plantings.planting_site_id, plantings.planting_subzone_id, plantings.species_id
+ON CONFLICT (planting_subzone_id, species_id)
```
karthikbtf, line 13: conflicts likely in the migration or just in case we need to re-run?

karthikbtf (approved):
LGTM 👍

src/main/resources/db/migration/0200/V200__BackfillPopulations.sql
sgrimm: If there was an outplanting withdrawal in between the time the previous code change was released and this one is released, these tables will already have some data in them, but the data will be wrong (because the code assumes the numbers from previous plantings are already reflected in these tables) so we need to overwrite it in the migration.

karthikbtf: Thanks, makes sense. Yes, I forgot we will be running this migration after the BE has been upgraded with the other changes (handling plants since last observation).

PR #1201, opened 2023-06-08
Add limit parameter to observation results API

sgrimm:
To allow the plants dashboard to show statistics from the most recent
observation of a given planting site, allow the client to request a maximum
number of results and change the result order so the most recently completed
observation always comes first. The client can thus pass in a planting site ID
and a limit of 1 to fetch the latest observation results for the site.

The default is still to return results from all observations.


sgrimm:
* main
  * **PR #1201** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1201" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1201?utm_source=stack-comment).

charlie-tf (approved):
Makes sense 👍

PR #1203, opened 2023-06-09
Make observation results test more strict

sgrimm:
The tests in `ObservationResultsStoreTest` that compared calculated results
against a set of example spreadsheets was too lax and failed to catch some
errors in the spreadsheets. Rework them so that instead of verifying that each
spreadsheet row has a corresponding set of numbers in the calculated observation
results, they instead convert the observation results into a spreadsheet which
must be exactly equal to the spreadsheets with the expected values.

The previous version of the test failed to catch that we were returning non-live
species in the results, which violates the specs; fix that bug.

Freshen the spreadsheets with freshly-exported copies that fix a few problems the
original test didn't catch (a species row that shouldn't have been there, two
species rows that should have been there but weren't) and that express mortality
rates as rounded percentages since those are what we're returning in our
calculations.


sgrimm:
* main
  * **PR #1203** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1203" target="_blank">Graphite</a>  👈
    * **PR #1205** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1205" target="_blank">Graphite</a> 
      * **PR #1206** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1206" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1203?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
```diff
@@ -410,8 +374,20 @@ class ObservationResultsStoreTest : DatabaseTest(), RunsAsUser {
...
+  ) {
+    val actualRendered = actual.map { it.joinToString(",") }.sorted().joinToString("\n")
+    val expected =
+        mapCsv(path) { mapCsvRow(it.toList()).joinToString(",") }.sorted().joinToString("\n")
+
+    assertEquals(expected, actualRendered, path)
   }
+
+  private fun getSpeciesNameValue(species: ObservationSpeciesResultsPayload): String =
+      species.speciesName ?: species.speciesId?.let { speciesNames[it] } ?: ""
```
karthikbtf, line 390: had to do this in the FE as well... just thinking aloud.. wish kotlin/java had a union construct (speciesName or speciesId)

karthikbtf (approved):
Looks pretty clear to me 👍

sgrimm:
@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1203).

sgrimm:
@sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1203).

PR #1208, opened 2023-06-13
Multiple observations in ObservationResultsStoreTest

sgrimm:
We're going to be changing some of the observation results calculations to look at
data from previous observations, and the spreadsheets we're using to define the
expected results in the product requirements are being updated to include multiple
observations for people to inspect.

The revised stats CSVs will have an extra header row to indicate which observation
a particular column relates to. (This observation number is purely for human
consumption; the test code assumes that each observation will have a fixed number
of data columns in a fixed order.)

In anticipation of these changes, rework `ObservationResultsStoreTest` to handle
spreadsheets where the rows can contain data from multiple observations.

The actual test data is unchanged here, as is the business logic for the
calculations.


sgrimm:
* main
  * **PR #1208** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1208" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1208?utm_source=stack-comment).

charlie-tf (approved):

PR #1212, opened 2023-06-13
Support strings with multiple links in email messages

sgrimm:
Currently, the helper function that replaces square brackets with `<a>` tags in
localizable strings in email notification bodies only supports one link per
string. Update it to support arbitrary numbers of links in preparation for adding
an email notification that has two links in one sentence.

PR #1211, opened 2023-06-13
Refactor date-based observation queries

sgrimm:
Refactor two query methods in `ObservationStore` that get a list of observations
that's filtered based on what the current date is in the planting sites' time
zones. No change in behavior here; this is in preparation for adding a third query
that does something similar.

PR #1213, opened 2023-06-13
SW-3743 Notify when observation is 1 month away

sgrimm:
Notify all organization members when an observation is due to start the following
month.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -179,6 +179,25 @@ class ObservationStore(
               .where(PLANTINGS.PLANTING_SITE_ID.eq(OBSERVATIONS.PLANTING_SITE_ID))
               .and(PLANTINGS.PLANTING_SUBZONE_ID.isNotNull))
 
+  /**
+   * Returns a list of observations that are starting in 1 month or less and for which we have yet
+   * to send out notifications that they're coming up.
+   */
+  fun fetchNonNotifiedUpcomingObservations(): List<ExistingObservationModel> {
+    val maxStartDate =
+        LocalDate.ofInstant(clock.instant(), ZoneOffset.UTC).plusMonths(1).plusDays(1)
+
+    return fetchWithDateFilter(
+        listOf(
+            OBSERVATIONS.STATE_ID.eq(ObservationState.Upcoming),
+            OBSERVATIONS.START_DATE.le(maxStartDate),
+            OBSERVATIONS.UPCOMING_NOTIFICATION_SENT_TIME.isNull,
+            plantingSiteHasPlantings,
+        )) { todayAtSite, record ->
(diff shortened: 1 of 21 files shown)


sgrimm:
* main
  * **PR #1211** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1211" target="_blank">Graphite</a> 
    * **PR #1212** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1212" target="_blank">Graphite</a> 
      * **PR #1213** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1213" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1213?utm_source=stack-comment).

charlie-tf (approved):
Looks good. Noting that observations whose start date is in the past will get an _upcoming_ observation notification, I think. Unlikely to be a problem in practice, though.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -179,6 +179,25 @@ class ObservationStore(
...
+        LocalDate.ofInstant(clock.instant(), ZoneOffset.UTC).plusMonths(1).plusDays(1)
+
+    return fetchWithDateFilter(
+        listOf(
+            OBSERVATIONS.STATE_ID.eq(ObservationState.Upcoming),
+            OBSERVATIONS.START_DATE.le(maxStartDate),
+            OBSERVATIONS.UPCOMING_NOTIFICATION_SENT_TIME.isNull,
+            plantingSiteHasPlantings,
+        )) { todayAtSite, record ->
+          record[OBSERVATIONS.START_DATE]!! <= todayAtSite.plusMonths(1)
```
karthikbtf, line 197: Just for my understanding.. we wouldn't need to filter out observations with start dates that are back by more than a month - because they would be tagged as Overdue and not covered by this search, yes?

karthikbtf (approved):
Looks good to me.

sgrimm:
Observations are tagged as overdue when their end dates pass. Usually the end date should be the end of the same month as the start date, but in theory it's possible to create an observation whose start date is, say, 2 months in the past and whose end date is in the future. In that case, ordinarily the "start observations when they're scheduled to start" logic would flip the state from Upcoming to InProgress and this code would ignore it.

For the same reason, we should never generate a notification for an observation whose start date is in the past, because the "start observations whose start times have arrived" operation happens before this.

sgrimm:
@sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1213).

sgrimm:
[Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1213) rebased this pull request as part of a merge.

sgrimm:
@sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1213).

PR #1214, opened 2023-06-14
SW-3348 Automatically log out incompatible sessions

sgrimm:
When we make a non-backward-compatible change to one of the classes that is
stored as part of the user's login session data, treat the session as
nonexistent instead of bombing out with an internal error.

In practice, this will often be invisible to the user: they will be treated
as unauthenticated, so they'll be redirected to Keycloak. If Keycloak still
considers them logged in, they'll be issued a new access token and redirected
straight back to the web app which will then create a fresh session for them.

This is in preparation for moving away from the Keycloak adapter library; when
we switch away from it, sessions that include serialized objects from that
library will no longer be deserializable.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/SessionConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/SessionConfig.kt
@@ -1,0 +1,52 @@
+package com.terraformation.backend.auth
+
+import com.terraformation.backend.log.perClassLogger
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.core.convert.converter.Converter
+import org.springframework.core.convert.support.GenericConversionService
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1214** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1214" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1214?utm_source=stack-comment).

sgrimm:
This class is copied pretty much verbatim from the funder portal.

karthikbtf (approved):
Looks good 👍 
Do we remove this later?

sgrimm:
My plan is to keep it around since the formats of the serialized objects could change due to Spring Security upgrades or changes to our application code. If there's no change to the object formats, this will be a harmless no-op.

PR #1216, opened 2023-06-15
SW-3348 Use Spring OAuth2 instead of Keycloak adapter

sgrimm:
The Keycloak adapter for Spring Security we've been using to authenticate users
to terraware-server has been deprecated and won't be updated to support new
versions of Spring going forward. The Keycloak team recommends people switch to
Spring Security's native OAuth2 support. Make that change.

This change should be invisible to clients, except that existing login sessions
will no longer be valid and users will have to get a fresh access token from
Keycloak (which may involve logging in again, or may happen transparently,
depending on whether or not Keycloak recognizes the user as already logged in).

The Keycloak credentials and server address are now part of the Spring Security
configuration. Dev environment configurations will need to be updated accordingly;
the example in `KEYCLOAK.md` shows what the configuration should look like.


sgrimm:
* main
  * **PR #1216** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1216" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1216?utm_source=stack-comment).

sgrimm:
I realize this one is pretty tough to review without in-depth knowledge of Spring Security; we'll probably need to lean on manual testing to make sure this hasn't broken anything.

sgrimm:
One thing in particular I'll want to test once this is running on staging, but that's kind of hard to test before then: running a FE dev environment locally that proxies API requests to the staging terraware-server instance. Getting that working with the Keycloak adapter took a couple rounds of experimentation and the same may be true here. That said, I think there's a pretty good chance it'll work out of the box since Spring Security's post-login redirect flow is simpler than the Keycloak adapters (one less redirect hop).

karthikbtf:
> I realize this one is pretty tough to review without in-depth knowledge of Spring Security; we'll probably need to lean on manual testing to make sure this hasn't broken anything.

Sounds good to me. Looks similar to the funder backend now.

PR #1225, opened 2023-06-19
SW-3769 Skip deleted species in summary statistics

karthikbtf:
- extracted out the species query separately
- i don't know if this is the best way to accomplish what we want, open to suggestions
- added a test

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -15,6 +15,7 @@ import com.terraformation.backend.db.default_schema.FacilityType
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.UserType
 import com.terraformation.backend.db.default_schema.tables.references.FILES
+import com.terraformation.backend.db.default_schema.tables.references.SPECIES
 import com.terraformation.backend.db.default_schema.tables.references.USERS
 import com.terraformation.backend.db.seedbank.AccessionId
 import com.terraformation.backend.db.seedbank.AccessionQuantityHistoryType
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -867,11 +868,21 @@ class AccessionStore(
...
+    val speciesQuery =
+        dslContext
+            .select(DSL.countDistinct(ACCESSIONS.SPECIES_ID))
+            .from(ACCESSIONS)
+            .innerJoin(SPECIES)
+            .on(ACCESSIONS.SPECIES_ID.eq(SPECIES.ID))
+            .and(SPECIES.DELETED_TIME.isNull)
+            .and(ACCESSIONS.ID.`in`(subquery))
+
+    val speciesCount = speciesQuery.fetchOne()?.value1()
```
sgrimm: May as well inline `speciesQuery` here. The reason the existing query in this function is stored in a variable is so we can log it and then execute it, but we're not logging this query.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -867,11 +868,21 @@ class AccessionStore(
                     1)
                 .else_(0))
 
+    val speciesQuery =
+        dslContext
+            .select(DSL.countDistinct(ACCESSIONS.SPECIES_ID))
+            .from(ACCESSIONS)
+            .innerJoin(SPECIES)
+            .on(ACCESSIONS.SPECIES_ID.eq(SPECIES.ID))
+            .and(SPECIES.DELETED_TIME.isNull)
```
sgrimm: The first `.and` should be a `.where` here. The end result will be the same in this case because it's a simple inner join, but semantically, the accession ID and species deletion status aren't determining how to join `accessions` and `species` together.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -867,11 +868,21 @@ class AccessionStore(
                     1)
                 .else_(0))
 
+    val speciesQuery =
+        dslContext
+            .select(DSL.countDistinct(ACCESSIONS.SPECIES_ID))
+            .from(ACCESSIONS)
+            .innerJoin(SPECIES)
```
sgrimm: Maybe replace `innerJoin` with `join` since that's the style we use elsewhere in the code (inner join is the default join type).

sgrimm (changes requested):
Seems reasonable at a high level, just a couple implementation details to comment on.

sgrimm (approved):
Looks great!

PR #1235, opened 2023-06-21
Separate observation result payloads from models

sgrimm:
Some of the observation result statistics require input values that don't need
to be returned to clients, and there are about to be more such values. The
distance between the internal representation of results and the client-facing
representation will become large enough that it no longer makes sense to directly
use the payload classes in the calculations.

The client-visible behavior difference here is that the observation result API
payloads no longer include a few values (e.g., `totalExisting`) that weren't
specified in the product requirements.


charlie-tf (approved):
Looks good. Nice clean refactor

PR #1236, opened 2023-06-21
SW-3788 Revised observation calculations

sgrimm:
Update the observation result calculations to reflect the latest round of
requirements changes. The major differences:

* Mortality rate now only considers plants in permanent monitoring plots, and
  includes dead plants from previous observations, not just from the current
  observation. This requires keeping a cumulative count of dead plants across
  observations, and tracking the number of live plants in permanent plots
  separately from the number of plants across all plots; the database schema
  is updated accordingly.

* Related to the above, when the number of permanent monitoring plots goes
  down, the mortality rate calculation needs to exclude the plants from plots
  that were considered permanent in earlier observations but are no longer
  included in the current observation.

* Existing plants are included in the "total plants" value but are not
  included in the mortality rate or planting density calculations.

* Estimated total plants are calculated per planting zone, not just for the
  entire planting site.

An updated set of spreadsheets is included here to test that the calculations
return the expected values. These spreadsheets have already been reviewed for
correctness.


src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -352,52 +405,55 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
...
           if (zones.isNotEmpty() && zones.all { it.plantingDensity != null }) {
-            val totalArea = zones.sumOf { it.areaHa }
+            plantingDensity =
+                zones
+                    .flatMap { zone ->
+                      zone.plantingSubzones.flatMap { subzone ->
+                        subzone.monitoringPlots.map { plot -> plot.plantingDensity }
+                      }
+                    }
+                    .average()
```
charlie-tf, line 422: Just thinking aloud. So this should be a good estimate provided that 1) the plots are equally sized (they are) and 2) they are uniformly spatially distributed across the site (true also at large scale)

charlie-tf (approved):
Makes sense. I think I understand what you've done. Thanks for all the comments

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
sgrimm: Yes, that's the idea. If we ever support variable-size plots, which isn't too likely but could theoretically happen, we'd use each plot's area when calculating the planting density rather than using a constant.

PR #1242, opened 2023-06-22
SW-2710 Declare annotation property aliases

sgrimm:
We have custom OpenAPI annotations to document that controller methods can
return certain kinds of responses. The annotations have parameters that supply
the values of some of the parameters of the underlying OpenAPI annotations.

Future versions of Spring are going to stop automatically using parameters on
annotations like these as the values of the underlying annotations, and will
require applications to explicitly declare that they want the parameters to be
treated as aliases for the parameters of other annotations. Spring Boot 3 logs
warnings about the upcoming behavior change.

Add the appropriate declarations so our annotations won't break in future Spring
versions.


sgrimm:
* main
  * **PR #1242** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1242" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1242?utm_source=stack-comment).

charlie-tf (approved):
👍

PR #1246, opened 2023-06-22
Download GBIF data instead of uploading it

sgrimm:
The admin UI allows importing GBIF species data from a zipfile. Previously, the
zipfile was uploaded by the browser using a file input. It is a huge file and
requires special treatment to avoid running into request size limits. And the
way file uploads are handled is different in Spring Boot 3, so we'd need to
implement a brand-new mechanism for dealing with the file.

Change the admin UI so it accepts a URL as input and pulls the file from there.
The common case will be in dev environments with a `file://` URL pointing to the
local filesystem. It's not quite as convenient, but GBIF import is a pretty rare
operation so the added burden should be miniscule.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -60,6 +60,7 @@ import com.terraformation.backend.tracking.model.PlantingSiteModel
 import com.terraformation.backend.tracking.model.Shapefile
 import io.swagger.v3.oas.annotations.Hidden
 import java.math.BigDecimal
+import java.net.URI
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1246** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1246" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1246?utm_source=stack-comment).

charlie-tf (approved):
Would this be a `file://` url then?

sgrimm:
> Would this be a `file://` url then?

Yep. Updated the PR description to clarify. (The admin UI also has a message suggesting a `file://` URL.)

PR #1251, opened 2023-06-23
SW-2710 Upgrade to Spring Boot 3

sgrimm:
Spring Boot 3 (and Spring Framework 6, which it uses) has some API differences
from previous versions that require code changes on our part as well as upgrades
of dependencies that are Spring-aware. The changes of note:

* Many of the packages that used to live under `javax` were moved to `jakarta`
  because the Java EE standard is now maintained by the Eclipse Foundation
  instead of being considered part of Java proper. Spring Boot 3 uses the new
  package names. This accounts for the lion's share of the edits since there
  were imports of `javax` classes in a large percentage of the source files.

* The behavior of a couple of annotations has changed. Most notably, we no
  longer need to annotate `TerrawareServerConfig` to tell Spring that it should
  use the constructors to set the config values; that's now the default.

* The serialized format of login sessions has changed; existing login sessions
  will be invalidated when this change is deployed (commit 5b070af means the
  impact on users should be minimal) and the frontend Cypress tests will need a
  new canned session object.


sgrimm:
* main
  * **PR #1251** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1251" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1251?utm_source=stack-comment).

sgrimm:
@charlie-tf This PR has two commits, one of which is just the mechanical edits to `import` statements. The idea was that you can skim the imports-only one quickly and then focus on the other one (which is small by comparison).

sgrimm:
I overlooked the `generateOpenApiDocs` task, which in theory (and in the current `main` code) doesn't need to connect to a database. Looks like avoiding the database connection isn't as straightforward in Spring Boot 3. This has been an ongoing source of headaches even prior to Spring Boot 3, so I think I will change it so it launches an ephemeral database in a container.

src/main/kotlin/com/terraformation/backend/api/ControllerExceptionHandler.kt
```diff
@@ -113,9 +114,10 @@ class ControllerExceptionHandler : ResponseEntityExceptionHandler() {
         .warn("Generic response exception thrown on $description use ClientFacingException", ex)
 
     @Suppress("USELESS_ELVIS") // ex.message can be null despite being annotated otherwise
-    val message = ex.message ?: ex.status.reasonPhrase
+    val message =
+        ex.message ?: (ex.statusCode as? HttpStatus)?.reasonPhrase ?: ex.statusCode.toString()
```
charlie-tf: If `ex.statusCode` is potentially null, then `ex.statusCode.toString()` would also fail, correct?

src/main/resources/application-default.yaml
```diff
@@ -34,6 +34,7 @@ logging:
   level:
     com.terraformation: DEBUG
     org.springdoc: DEBUG
+    org.springframework.boot.devtools.restart.ChangeableUrls: WARN
```
charlie-tf, line 40: Sorry I'm missing context here. What does this change?

src/main/resources/application-apidoc.yaml
```diff
@@ -8,8 +8,11 @@
 # server is running.
 
 spring:
+  datasource:
+    # Use an in-memory database so Spring can autowire beans that require a data source.
+    url: jdbc:h2:mem:testdb
```
charlie-tf, line 13: 👍

charlie-tf (commented):
Just a couple of questions, but looks good

sgrimm:
Updated the PR to hopefully answer both of the questions; let me know if it's still unclear.

charlie-tf (approved):
Okay, that's clearer. Thanks!

PR #1252, opened 2023-06-23
SW-3815 Read Google credentials from environment

sgrimm:
The credentials for the Google API, which we use to export report files to Google
Drive, have to be in JSON form. By default, the Google client library expects the
credentials to be in a file, but it also allows callers to directly supply a JSON
object.

Add a Terraware config value that lets us pass the JSON object in an environment
variable so we don't have to write it to a file on the server.


sgrimm:
* main
  * **PR #1252** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1252" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1252?utm_source=stack-comment).

charlie-tf (approved):
Looks good to me!

PR #1255, opened 2023-06-26
Allow enums to use string IDs

sgrimm:
IUCN conservation categories are identified with two-letter codes. We'll want to
use those codes directly in our database schema rather than defining a synthetic
numeric ID like we do for our other enumerated types. We'll also want to use the
codes in the API, as opposed to our usual approach of using enum display names.

Update the enum code generator to support both integers and strings as enum IDs,
and to optionally allow the ID, rather than the name, to be used as the JSON
value.

Existing enums should continue to work as before.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/EnumFromReferenceTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/EnumFromReferenceTable.kt
@@ -3,8 +3,8 @@ package com.terraformation.backend.db
 import java.util.Locale
 import org.springframework.context.support.ResourceBundleMessageSource
 
-interface EnumFromReferenceTable<T : Enum<T>> {
-  val id: Int
+interface EnumFromReferenceTable<ID : Any, T : Enum<T>> {
+  val id: ID
   /** JSON string representation of this enum value. */
   val jsonValue: String
   val tableName: String
@@ -21,7 +21,7 @@ interface EnumFromReferenceTable<T : Enum<T>> {
           setDefaultEncoding("UTF-8")
         }
 
-    fun <T : EnumFromReferenceTable<T>> loadLocalizedDisplayNames(
+    fun <T : EnumFromReferenceTable<*, T>> loadLocalizedDisplayNames(
         locale: Locale,
         values: Array<T>
     ): Map<T, String> {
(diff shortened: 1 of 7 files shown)


sgrimm:
* main
  * **PR #1255** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1255" target="_blank">Graphite</a>  👈
    * **PR #1256** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1256" target="_blank">Graphite</a> 
      * **PR #1257** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1257" target="_blank">Graphite</a> 
        * **PR #1258** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1258" target="_blank">Graphite</a> 
          * **PR #1259** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1259" target="_blank">Graphite</a> 
            * **PR #1260** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1260" target="_blank">Graphite</a> 
              * **PR #1261** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1261" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1255?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/db/EnumFromReferenceTable.kt
```diff
@@ -21,7 +21,7 @@ interface EnumFromReferenceTable<T : Enum<T>> {
           setDefaultEncoding("UTF-8")
         }
 
-    fun <T : EnumFromReferenceTable<T>> loadLocalizedDisplayNames(
+    fun <T : EnumFromReferenceTable<*, T>> loadLocalizedDisplayNames(
```
charlie-tf, line 24: Is the `*` kinda like an `Any` type?

charlie-tf (approved):
Minor question about Kotlin syntax I didn't quite understand. Besides that everything makes sense

src/main/kotlin/com/terraformation/backend/db/EnumFromReferenceTable.kt
sgrimm: > Is the * kinda like an Any type?

Kind of, but `List<*>` says "use type inference to figure out the list type" instead of "the list can contain a mix of any kinds of objects" which is what `List<Any>` means.

I started to type a longer explanation but this explains it better than I would have: https://stackoverflow.com/a/65534086

PR #1267, opened 2023-06-28

terraware-phrase:


src/main/resources/i18n/Enums_es.properties
```diff
@@ -82,6 +82,9 @@ public.Role.Contributor=Colaborador
 public.Role.Manager=Gerente
 public.Role.Owner=Propietario
 public.SeedStorageBehavior.Intermediate=Intermediario
+public.SeedStorageBehavior.LikelyIntermediate=Probablemente intermedio
+public.SeedStorageBehavior.LikelyOrthodox=Probablemente ortodoxo
+public.SeedStorageBehavior.LikelyRecalcitrant=Probable recalcitrante
```
sgrimm: @constanzauanini Should this be "Probablemente" like the other two?

constanzauanini: Yes, "probablemente" would be the correct way to say it

PR #1268, opened 2023-06-28
Make AccessionModel.state non-nullable

sgrimm:
Accession states are never null (they are populated at creation time and never
cleared) so there's no reason for `AccessionModel` to be able to represent an
accession with a null state.

No behavior change here, just marginally cleaner code.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -194,7 +194,6 @@ class AccessionStore(
         parentStore.getOrganizationId(facilityId) ?: throw FacilityNotFoundException(facilityId)
     val state =
         when {
-          accession.state == null -> AccessionState.AwaitingCheckIn
           accession.state == AccessionState.UsedUp ->
               throw IllegalArgumentException("Accessions cannot be set to Used Up at creation time")
           accession.state.isV2Compatible -> accession.state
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #1268** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1268" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1268?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -327,7 +326,7 @@ class AccessionStore(
     val organizationId =
         parentStore.getOrganizationId(facilityId) ?: throw FacilityNotFoundException(facilityId)
 
-    if (updated.state?.isV2Compatible != true) {
```
charlie-tf, line 330: Ugh, I hate the `== true` or `!= true` construction that nullability sometimes forces us into.

charlie-tf (approved):
Nice improvement! Thanks!

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: Yeah, I'm not nuts about that either, but I like it _slightly_ better than `?: false`.

PR #1269, opened 2023-06-29
SW-3647 Accession species only editable if no deliveries

sgrimm:
In preparation for an upcoming change that causes changes to an accession's
species to be propagated to seedling batches that come from that accession,
add logic to prevent editing an accession's species if the accession's seeds
have already been sent to a nursery, and seedlings have already been withdrawn
from the nursery and delivered to a planting site.

This prevents accession species changes from causing inconsistencies between
nursery withdrawals and the species populations at planting sites. In the
future we may remove this restriction and make the species changes propagate to
planting site data too, but doing so would introduce a number of edge cases
that would require more detailed requirements definition.

The fact that seedlings from an accession's seeds have been delivered to planting
sites is exposed as a new `hasDeliveries` field on the accession payload. Clients
can use this to inform users that they can't edit the accession's species.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -5,6 +5,7 @@ import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.model.IndividualUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.AccessionNotFoundException
+import com.terraformation.backend.db.AccessionSpeciesHasDeliveriesException
 import com.terraformation.backend.db.FacilityNotFoundException
 import com.terraformation.backend.db.FacilityTypeMismatchException
 import com.terraformation.backend.db.IdentifierGenerator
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #1269** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1269" target="_blank">Graphite</a>  👈
    * **PR #1270** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1270" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1269?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -335,9 +351,11 @@ class AccessionStore(
       throw FacilityNotFoundException(facilityId)
     }
 
+    val speciesId = if (existing.hasDeliveries) existing.speciesId else updated.speciesId
+
     requirePermissions {
       updateAccession(accessionId)
-      updated.speciesId?.let { readSpecies(it) }
+      speciesId?.let { readSpecies(it) }
```
charlie-tf: Should we return an error code in the case where we don't actually update the species to the one in the payload (i.e. `existing.hasDeliveries` is true)?

charlie-tf (approved):
Looks great!

PR #1270, opened 2023-06-29
SW-3647 Apply accession species changes to batches

sgrimm:
It can be difficult to distinguish some species of seeds from others; sometimes
a definitive species ID can only happen after the seeds have sprouted and the
seedlings can be examined.

If it turns out that the initial guess at the species was wrong, we want to
update the accession to have the correct species. Currently, though, the species
of the seedling batches don't change, and there's no way to edit them manually;
they remain listed under the old incorrect species.

Update the system so that changing the species of an accession will automatically
update the species of all the seedling batches that came from that accession.

It is not currently permitted to change the species of an accession if there are
already outplanting withdrawals of its seedlings; we thus don't need to reflect
changes of accession species in the planting histories of planting sites. The
expectation is that this restriction is no big deal in practice because species
will have already been identified by the time seedlings go out for planting.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/Exceptions.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/Exceptions.kt
@@ -67,6 +67,10 @@ abstract class MismatchedStateException(message: String) : RuntimeException(mess
 class AccessionNotFoundException(val accessionId: AccessionId) :
     EntityNotFoundException("Accession $accessionId not found")
 
+class AccessionSpeciesHasDeliveriesException(val accessionId: AccessionId) :
+    MismatchedStateException(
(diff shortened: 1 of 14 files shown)


sgrimm:
* main
  * **PR #1269** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1269" target="_blank">Graphite</a> 
    * **PR #1270** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1270" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1270?utm_source=stack-comment).

charlie-tf (approved):

PR #1282, opened 2023-07-06
Update build scripts for Gradle 8.2

sgrimm:
- Use property assignment rather than `.set()` (technically still an incubating
  feature but it's enabled by default in 8.2)
- Upgrade a couple plugins to newer versions that don't call newly-deprecated
  Gradle APIs
- Enable "all warnings as errors" compilation mode on build scripts
- Remove a null check that is no longer necessary

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,7 @@ plugins {
   // kotlin("kapt")
 
   id("dev.monosoul.jooq-docker") version "3.0.22"
-  id("com.diffplug.spotless") version "6.4.2"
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #1282** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1282" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1282?utm_source=stack-comment).

karthikbtf (approved):
👍

PR #1283, opened 2023-07-06
Require planting sites to have at least 5 plots

sgrimm:
If you use the "draw site on map" feature of the admin UI to draw a planting site
that's only big enough for one 4-plot cluster, the server creates the planting
site but then can't start any observations because it can't allocate any temporary
plots.

Update the importer code to ensure that there are at least 5 plots available; if
not, site creation fails. Since plots are allocated in groups of 4, technically
"5 plots minimum" is really "8 plots minimum" and thus the planting site must be
at least 100x50 meters.


sgrimm:
* main
  * **PR #1283** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1283" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1283?utm_source=stack-comment).

karthikbtf (approved):
Looks clear to me.

PR #1284, opened 2023-07-06
Ignore irrelevant fields in plant observation payload

sgrimm:
If a recorded plant's species is known, the server is supposed to ignore the
value of the `speciesName` field in the plot observation completion API
payload, but it doesn't; instead the request fails due to a database constraint
violation.

Update the code to conditionally ignore species ID and name like the API
documentation says it does.


sgrimm:
* main
  * **PR #1284** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1284" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1284?utm_source=stack-comment).

joshbuchea (approved):
I'm not familiar with the syntax here, but the change looks pretty straight forward.

PR #1288, opened 2023-07-10
SW-3868 Add country code to users, locales

sgrimm:
To support variation of number formatting rules in different Spanish-speaking
countries, keep track of users' countries. If the user has selected a country,
their locale will include a country code as well as a language.

The server will automatically include the user's country code in their locale
when they update their profile information, to account for the fact that clients
have language selectors that don't include country codes. That is, if the user
sets their locale to `es` and their country to `MX`, the server will save the
locale as `es-MX`.

Number and date formatting will automatically use country-specific rules once
the user has set their country.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -248,7 +248,7 @@ class UserStore(
             ?: throw IllegalStateException("Current user not found in users table")
 
     // If the country code is set, include it in the locale.
-    val newLocale = model.locale?.let { Locale.of(it.language, model.countryCode ?: it.country) }
+    val newLocale = model.locale?.let { Locale.of(it.language, model.countryCode ?: "") }
 
     dslContext.transaction { _ ->
       usersDao.update(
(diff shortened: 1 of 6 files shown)


sgrimm:
* main
  * **PR #1288** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1288" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1288?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -247,13 +247,17 @@ class UserStore(
         usersDao.fetchOneById(model.userId)
             ?: throw IllegalStateException("Current user not found in users table")
 
+    // If the country code is set, include it in the locale.
+    val newLocale = model.locale?.let { Locale.of(it.language, model.countryCode ?: it.country) }
```
karthikbtf: just curious what happens when the country doesn't apply to the locale.. like es-ZH for example.., guessing it's a no-op and we default to `es` where applicable?

karthikbtf (approved):
Looks good to me 👍

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
karthikbtf: unrelated, do we want to be able to clear the country?

sgrimm: If there are no country-specific rules for a language+country pair, it will fall back on the default rules for the language.

And good point on clearing the country. I had that there for testing and forgot to remove it.

karthikbtf (approved):
Looks good to me!!

PR #1317, opened 2023-08-01
SW-3977 Admin UI to create minimal planting site

sgrimm:
For testing observations in the mobile app, we want to be able to create the
smallest possible planting site, with one permanent cluster and one temporary
plot. Previously, this was impossible to do in the admin UI because the required
precision of the boundary lines was greater than the precision of the map.

Add a button to place a minimal-size planting site on the map. The user can move
it around on the map and change where the temporary plot is located relative to
the permanent cluster.

This requires adding some logic to the planting site creation code to account for
the fact that the map UI does coordinate transformations differently than the
server does, meaning that the site boundary submitted by the client will always
be off by a tiny amount from the server's point of view.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
@@ -988,6 +988,7 @@ class AdminController(
       @RequestParam organizationId: OrganizationId,
       @RequestParam siteName: String,
       @RequestParam boundary: String,
+      @RequestParam fitSiteToPlots: Boolean,
       redirectAttributes: RedirectAttributes,
   ): String {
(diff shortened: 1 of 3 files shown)


sgrimm:
* main
  * **PR #1317** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1317" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1317?utm_source=stack-comment).

sgrimm:
Screenshot 2023-08-01 at 11 37 39 AM

karthikbtf (approved):
Nice! Looks good to me.

PR #1319, opened 2023-08-02
SW-3814 Non-sensitive config for prod and staging

sgrimm:
Some of the configuration for the prod and staging environments is safe to expose
to the world; add it to config files to reduce the amount of configuration that
needs to be supplied in environment variables at deployment time.

### DIFF
--- a/src/main/resources/application-prod.yaml
+++ b/src/main/resources/application-prod.yaml
@@ -37,5 +37,5 @@ terraware:
   report:
     export-enabled: true
   request-log:
-    email-regex: ".*@terraformation\.com"
+    email-regex: ".*@terraformation\\.com"
   web-app-url: "https://terraware.io/"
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1319** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1319" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1319?utm_source=stack-comment).

karthikbtf (approved):
Nice!

sgrimm:
I haven't forgotten about this, just holding off on it because my "move Terraware to ECS" task is paused for the moment.

sgrimm:
I've thought better of this approach; having environment-specific config files in the open-source code base, while harmless from a security point of view, doesn't feel like the right solution.

PR #1320, opened 2023-08-03
SW-4007 Use nursery time zone to check withdrawn date

sgrimm:
If a nursery withdrawal is dated earlier than the time a batch's remaining
quantity was most recently manually observed, we don't override the
manually-entered quantity.

Unfortunately, we were using UTC to figure out which date the last observation
fell on, meaning the system could erroneously think that a withdrawal happened
before the last observation instead of on the same day. This caused the system
to not update the quantities of batches when it should have.

A followup change will backfill the missing quantity updates.


sgrimm:
* main
  * **PR #1320** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1320" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1320?utm_source=stack-comment).

charlie-tf (approved):
Changes make sense. Thanks!

PR #1322, opened 2023-08-08
SW-4041 Return permanent plot live/dead counts

sgrimm:
The mortality rate graphs in the web app need to show live and dead plant totals
that only consider plants in permanent plots. We were already calculating those
totals on the server side since they are the inputs to the mortality rate
calculation; expose them in the observation results API payloads.


sgrimm:
* main
  * **PR #1322** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1322" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1322?utm_source=stack-comment).

karthikbtf (approved):
Thanks!

PR #1324, opened 2023-08-09
SW-4028 Rotate thumbnail image source

karthikbtf:
- Rotate image source (if needed) when creating thumbnails
- This is a copy of funder-backend PR https://github.com/terraware/funder-backend/pull/110
- Tested locally

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -92,6 +92,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.session:spring-session-jdbc")
 
+  implementation("com.drewnoakes:metadata-extractor:2.18.0")
   implementation("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
   implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jacksonVersion")
   implementation("com.google.api-client:google-api-client:2.2.0")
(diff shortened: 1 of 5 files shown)


karthikbtf:
For funder portal, we had fewer images in staging and went ahead and deleted the thumbnail data, in order to recreate correctly rotated thumbnails.
In staging and perhaps prod - given the number of existing thumbnails, we could just let the the thumbnails be and expect rotation to work going forward?

sgrimm:
Not sure if we tried this the last time around, but I notice there's a Thumbnailator option related to this: https://coobird.github.io/thumbnailator/javadoc/0.4.20/net/coobird/thumbnailator/Thumbnails.Builder.html#useExifOrientation(boolean) -- did that not do the right thing?

karthikbtf:
> Not sure if we tried this the last time around, but I notice there's a Thumbnailator option related to this: https://coobird.github.io/thumbnailator/javadoc/0.4.20/net/coobird/thumbnailator/Thumbnails.Builder.html#useExifOrientation(boolean) -- did that not do the right thing?

I had explored that first, it only works if the image source to Thumbnailator has exif orientation data preserved.
We lose that information in buffered (edit) streams. Only works if we use a File or Url as image source to Thumbnailator.

karthikbtf:
> > Not sure if we tried this the last time around, but I notice there's a Thumbnailator option related to this: https://coobird.github.io/thumbnailator/javadoc/0.4.20/net/coobird/thumbnailator/Thumbnails.Builder.html#useExifOrientation(boolean) -- did that not do the right thing?
> 
> I had explored that first, it only works if the image source to Thumbnailator has exif orientation data preserved. We lose that information in io streams. Only works if we use a File or Url as image source to Thumbnailator.

Found the [PR](https://github.com/terraware/funder-backend/pull/108) from that attempt, for context. I believe thumbnailator's default setting is to preserve orientation.

karthikbtf:
> Ah, right, now I remember discussing that. In that case, this seems like the right solution.
> 
> Leaving existing thumbnails as-is seems fine for now; we can revisit if it turns out to be troublesome for users.

Great, thanks Steve.

PR #1331, opened 2023-08-14
SW-4071 Thread safety for reading Geometry from DB

sgrimm:
The JTS `WKBReader` class is not thread-safe, but we were sharing an instance of
it between threads in the code that reads geometry values from the database.

Create a fresh instance each time; it's a cheap class to instantiate. The
instances share a `GeometryFactory`, which is documented as thread-safe.


sgrimm:
* main
  * **PR #1331** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1331" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1331?utm_source=stack-comment).

PR #1338, opened 2023-08-18
SW-4084 Allow Used Up accessions to be imported

sgrimm:
We were explicitly disallowing accessions from being created in Used Up state
because it doesn't make much sense for a typical interactive workflow, but it
does make sense to be able to import historical accession data that includes
accessions that have already been fully withdrawn.


sgrimm:
* main
  * **PR #1338** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1338" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1338?utm_source=stack-comment).

charlie-tf (approved):
Looks good!

PR #1346, opened 2023-08-21
Gradle: Use lazy property to get build directory

sgrimm:
Gradle plugins are supposed to use lazy evaluation when fetching the build
directory since it can change at runtime, but we were using the non-lazy
getter. It is deprecated starting in Gradle 8.3; switch to the appropriate
lazy property.


sgrimm:
* main
  * **PR #1346** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1346" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1346?utm_source=stack-comment).

charlie-tf (approved):
Neat 👍

PR #1354, opened 2023-08-29
Local Docker improvements

nickgraz:
- Added some sample files so the Keycloak variables are easier to implement for local development
-- [Sample `.env`](https://github.com/terraware/terraware-server/pull/1354/files#diff-f478ca7c0a624094d72e39755682e9ac21d821e076b09f44f76be6f90d438762) for running with Docker 
-- [Sample `application-dev.yaml`](https://github.com/terraware/terraware-server/pull/1354/files#diff-bc7da0862d6ebbed553d3bb138d645685062a7cf8f8ddd6374920a961f14e4d9) for running with IntelliJ
- Added a `make` command for running just the Postgres container ([`make run-postgres`](https://github.com/terraware/terraware-server/pull/1354/files#diff-2d5cf8144e7448f74d50e9f30614362fbba55750d498273b487f52e836a9721aR24))
- [Added a check](https://github.com/terraware/terraware-server/pull/1354/files#diff-2d5cf8144e7448f74d50e9f30614362fbba55750d498273b487f52e836a9721aR18) for Docker `.env` file to use that when running the server via `make`
- Updated readme with info related to changes


docker/Makefile
```diff
@@ -14,7 +14,15 @@ image: prepare
 	docker build -f Dockerfile -t $(IMAGE_NAME) $(BUILD_DIR)
 
 run: image
-	docker-compose -f docker-compose.yml up
+    # This will not work on Windows
```
nickgraz, line 17: Is this OK?

README.md
```diff
@@ -42,6 +42,19 @@ Fetch the API schema to confirm the server is up and running. (This doesn't requ
 
     curl http://localhost:8080/v3/api-docs.yaml
 
+### Running the server via `docker`
+
+If you wish to run the Terraware server and/or Postgres database via docker, check out the `./docker/Makefile`.
+
+#### Run both the server and database
+This is useful if you just want the server running locally for frontend development
```
sgrimm, line 50: FYI, the frontend repo also has a docker-compose file for running the server locally, and that's the one the frontend developers typically use. Nothing wrong with also keeping the one in this repo up to date, though.

README.md
```diff
@@ -42,6 +42,19 @@ Fetch the API schema to confirm the server is up and running. (This doesn't requ
...
+1. Copy `./docker/.env.sample` to `./docker/.env` and fill in the necessary Keycloak environment variables
+2. Run `cd docker && make run`
+
+### Run just the database
+This is useful if you are doing backend development and running the server in IntelliJ, and you don't want to install Postgres natively
+1. Run `cd docker && make run-postgres`
```
sgrimm: `make -C docker run-postgres` would also work (main difference is that the shell will stay pointed at the repo root directory afterwards).

.gitignore
```diff
@@ -3,6 +3,7 @@ __pycache__
 .gradle
 .venv
 application-dev*.yaml
+!application-dev*.yaml.sample
```
sgrimm, line 6: Oh, interesting! I would have thought the `.sample` suffix would have caused the rule on the previous line to not match.

docker/Makefile
```diff
@@ -14,7 +14,15 @@ image: prepare
...
+	@if [ -f .env ]; then\
+		docker-compose -f docker-compose.yml --env-file .env up;\
+	else\
+		docker-compose -f docker-compose.yml up;\
+	fi
+
+run-postgres:
+	docker-compose -f docker-compose.yml up postgres
```
sgrimm, line 25: Would it make sense to add `-d` here so it runs in the background?

docker/docker-compose.yml
```diff
@@ -16,6 +17,8 @@ version: "3.8"
 services:
   postgres:
     image: "postgis/postgis:13-3.0"
+    ports:
+      - "5432:5432"
```
sgrimm, line 21: This is the only part of the PR I'm not sure about, because it means if you already have a local Postgres server running (either in Docker or natively), then `docker-compose up` will fail to start this service because the host port is in use.

I do see the benefit of being able to connect to this database using local native tools on the host, e.g., IntelliJ's database browser.

If people are going to be using this as the main database instance in their dev environments, such that connecting to it with local tools is a regular desire, then it may be worth making this use a persistent volume for the database so the data doesn't get wiped on container restart.

docker/Makefile
sgrimm: Yeah, we don't officially support Windows dev environments, so this is totally fine.

sgrimm (changes requested):
I'm marking this as "request changes" but it's mostly good. The database port/volume thing is the only part of this that gives me pause and needs some further discussion.

docker/Makefile
nickgraz: Personally I like to have it running in the foreground so I can easily see the logs and shut it down. I can add something like `make background=true run-postgres` if you'd like.

docker/docker-compose.yml
nickgraz: I added a persistent volume so the data doesn't get wiped. I also added a new `make` target `run-server` so that a user can _only_ run the server via Docker compose in case they're running Postgres natively or in Docker elsewhere

docker/Makefile
sgrimm: Nah, no need for that. This is fine.

docker/docker-compose.yml
sgrimm: What do you think about using a Docker volume rather than a bind mount for this? Docker volumes are significantly faster than bind mounts on the Mac because they operate entirely inside the Linux VM rather than having to proxy filesystem operations across the host/VM boundary.

nickgraz: Now that VirtioFS is supported in Mac and Docker (not experimental anymore), it technically isn't a bind mount anymore, the filesystem is accessed directly by the container. I think this is better because now there's no management of the volumes needed (no need to prune, the dir can just be deleted if you want), but it is _assuming_ that users are using this new-fangled tech in their Mac and Docker setup. 

All that being said, if you still want me to make this an OG volume, let me know!

sgrimm (approved):
Oops, I totally dropped the ball on this! The updates seem good, and you're right that my objection about bind mounts is outdated.

PR #1369, opened 2023-09-19
SW-4223 Add Parent ID to admin facility devices form

nickgraz:
- Add the field to the admin devices form
- Plumb the field through the form controller to it saves in the DB


sgrimm:
`./gradlew spotlessApply` will fix the formatting failure here.

src/main/kotlin/com/terraformation/backend/customer/api/AdminController.kt
```diff
@@ -829,6 +830,7 @@ class AdminController(
       @RequestParam model: String,
       @RequestParam name: String?,
       @RequestParam protocol: String?,
+      @RequestParam parentId: DeviceId?,
```
sgrimm, line 832: Definitely a bit of a nitpick, but move this up a line so the request parameters stay in alphabetical order. Same with the new `DevicesRow` constructor argument below this.

sgrimm (changes requested):

sgrimm (approved):

PR #1371, opened 2023-09-20
SW-4238 Update permissions for "Terraformation Contact" role

karthikbtf:
- Add 'Terraformation Contact' to `isAdminOrHigher` / `isManagerOrHigher` checks, since this role will have the same permissions as an Admin
- Disallow adding a user with 'Terraformation Contact' role
- Disallow updating a user's role as 'Terraformation Contact'
- Disallow modifying a 'Terraformation Contact' user with another role
- Leveraged existing permissions layers instead of special-casing these semantics in OrganizationStore or OrganizationService. Advantage of this is we can use the SystemUser (in admin ui) to create a TF contact org user (until we have more clarity on super-admin privileges here). Downside to this approach is being unable to mock the role. Added an `IndividualUserTest` to test for the logic instead.
- Open to doing this some other way if there's a preference.


karthikbtf:
Redoing this a bit differently

PR #1374, opened 2023-09-20
SW-4242 Admin UI feature to assign Terraformation Contact (nice to have feature)

karthikbtf:
- introduced `assignTerraformationContact` function in OrganizationService which handles removing existing contact and assigning new contact
- added some tests
- updated AdminController and Admin UI to show/reassign Terraformation Contact (made some assumptions here as this is an interim solution)
- caveat, super admin needs to be part of the org in order to assign a Terraformation Contact (for now at least)

##### when there is no assigned TF contact
Terraware Server Placeholder Admin UI 2023-09-21 01-52-02

##### with assigned TF contact
Terraware Server Placeholder Admin UI 2023-09-21 01-51-39

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
@@ -78,7 +78,17 @@ class OrganizationService(
     }
   }
 
-  fun assignTerraformationContact(organizationId: OrganizationId, email: String): UserId {
+  /**
+   * Assigns a Terraformation Contact in an organization. Removes existing Terraformation Contact
+   * from the organization, if one exists. If email of user to assign as Terraformation Contact,
+   * already exists as an organization user, the role is simply updated. Otherwise, a new user is
+   * created and added as the Terraformation Contact.
+   *
+   * @param email, email of user to assign as Terraformation Contact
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/terraformation/backend/customer/OrganizationService.kt
```diff
@@ -77,6 +78,47 @@ class OrganizationService(
...
+          if (existingUser != null) {
+            dslContext
+                .selectOne()
+                .from(ORGANIZATION_USERS)
+                .where(ORGANIZATION_USERS.ORGANIZATION_ID.eq(organizationId))
+                .and(ORGANIZATION_USERS.USER_ID.eq(existingUser.userId))
+                .fetch()
+                .isNotEmpty
+          } else {
+            false
```
sgrimm: I think it'd be better to throw an exception if the user doesn't exist. Can't see a good use case for wanting to create a new TF employee user as part of the "assign someone as contact" flow; more likely would be a typo in the email address.

karthikbtf: That's a valid point, will update this.

sgrimm (approved):

PR #1375, opened 2023-09-21
SW-4246 Validate that Terraformation Contact's email ends with '@terraformation.com'

karthikbtf:
- Added a regex pattern in Admin UI for automatic form validation
- Added validation/checks at entry points that create or set the Terraformation Contact user
- Added tests to capture these scenarios
- Did not add a config to grab the email pattern for now, we can do that if it is cleaner


src/main/resources/templates/admin/organization.html
```diff
@@ -363,8 +363,8 @@ <h4>Assign New Terraformation Contact (interim solution - until we add TW suppor
     <p>
         The current Terraformation Contact will be removed from the org (if one exists) and the new one will be assigned.
     </p>
-    <label for="email" style="margin-top: 20px">Enter new Terraformation Contact email:</label>
-    <input type="email" id="terraformationContactEmail" name="terraformationContactEmail" style="min-width: 300px; padding: 2px;">
+    <label for="email" style="margin-top: 20px">Enter the Terraformation Contact's <i>(@terraformation.com)</i> email:</label>
+    <input type="email" id="terraformationContactEmail" name="terraformationContactEmail" style="min-width: 300px; padding: 2px;" pattern="^\S+@terraformation\.com$">
```
charlie-tf, line 367: Cool. Does this render the form non-submittable when the pattern isn't matched?

src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
```diff
@@ -594,6 +607,17 @@ class OrganizationStore(
           .and(ORGANIZATION_USERS.USER_ID.eq(userId))
           .fetchOne(ORGANIZATION_USERS.ROLE_ID)
 
+  private fun getUserEmail(userId: UserId): String =
+      dslContext.select(USERS.EMAIL).from(USERS).where(USERS.ID.eq(userId)).fetchOne(USERS.EMAIL)
+          ?: throw UserNotFoundException(userId)
+
+  private fun validateTerraformationContactEmail(userId: UserId) =
+      getUserEmail(userId).let { it ->
+        if (!it.endsWith(suffix = "@terraformation.com", ignoreCase = false)) {
```
charlie-tf: Do we check for malformed emails anywhere? For instance, `foo@bar@terraformation.com` would pass the checks but isn't a valid email.

charlie-tf (approved):
Looks good, just a couple minor questions

src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
sgrimm: That wouldn't end up being a valid user, so it'd fail the user ID lookup later on; I think this is fine.

However, I'd probably do `ignoreCase=true` here because `TERRAFORMATION.COM` is technically a valid spelling of our email domain.

karthikbtf: Oops, thanks for catching that, intended it to be `ignoreCase=true` but kept thinking `caseSensitive` and used false. Will fix this.

src/main/resources/templates/admin/organization.html
karthikbtf: Yes, the form will pop-up a message above the input asking to correct the pattern.

sgrimm:
Can you rebase this? I merged your other PRs and this has conflicts now.

sgrimm (approved):

PR #1381, opened 2023-09-25
SW-4253 / SW-4254 API to schedule and reschedule observations

karthikbtf:
- Added API to schedule/reschedule observations
- Added validation checks on the requests, based on requirements in the PRD
- Added tests for scheduling and rescheduling observations
- Also tested the API with real requests


sgrimm (approved):
Looks like the right way to do all this!

PR #1384, opened 2023-10-02
SW-4257 User notifications to schedule observations

karthikbtf:
- Send email and app notifications to org admin/owner when trigger conditions are met, to shedule an obseration

- Trigger to notify scheduling an observation:
```
2 weeks have elapsed since the previous observation for a planting site and there is no future observation scheduled
  OR
Plants are allocated to a detailed map planting site for the first time
```

- Trigger to notify a reminder to schedule an observation:
```
6 weeks have elapsed since the previous observation for a planting site
OR
It has been 4 weeks since plants were first allocated to a detailed map planting site and there is no future observation scheduled
```
- Added a gating flag in TerrawareServerConfig to avoid sending notifications if a trigger condition is met based on presence of plantings alone. This is only to help synchronize releasing this feature along with the FE's schedule observation workflow, which is also currently hidden in production.
- Added tests
- Tested locally by adjusting the test clock


src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -243,6 +246,19 @@ class TerrawareServerConfig(
...
+   * This configuration is a temporary measure to avoid sending user notifications on a first
+   * planting. We don't want this in production until scheduling observations as a feature is
+   * released.
+   */
+  class ObservationsConfig(
+      /**
+       * Whether to send notifications to schedule on observation when the first planting is
+       * allocated to a planting site sub zone.
+       */
+      @DefaultValue("false") val notifyOnFirstPlanting: Boolean = false,
```
sgrimm, line 259: Perhaps this should be gated on an explicit "observation scheduling enabled" flag on the planting site and/or the organization. Right now we're using the presence of subzones as a proxy for "this is an accelerator participant" because there's no way for non-participant sites to have subzones, and we're assuming that all accelerator participants are eligible for observation scheduling. Which was fine when observations were manually scheduled by TF employees, but may not be fine-grained enough now.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -626,6 +626,32 @@ class ObservationStore(
...
+    return dslContext
+        .select(OBSERVATIONS.asterisk())
+        .from(OBSERVATIONS)
+        .where(OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
+        .and(OBSERVATIONS.STATE_ID.eq(ObservationState.Completed))
+        .and(OBSERVATIONS.COMPLETED_TIME.isNotNull)
+        .and(noUpcomingOrInProgressObservations)
+        .orderBy(OBSERVATIONS.COMPLETED_TIME.desc())
+        .limit(1)
+        .fetchOne { row -> ExistingObservationModel.of(row) }
```
sgrimm: I assume this is the SQL you referred to in your PR description. Maybe something like this?
```kotlin
return dslContext
    .select(OBSERVATIONS.asterisk())
    .from(OBSERVATIONS)
    .where(OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
    .and(OBSERVATIONS.COMPLETED_TIME.isNotNull)
    .andNotExists(
        DSL.selectOne()
            .from(OBSERVATIONS)
            // Need the planting site ID in the NOT EXISTS
            .where(OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
            .and(
                OBSERVATIONS.STATE_ID.`in`(
                    ObservationState.InProgress,
                    // I think you probably need Overdue here too
                    ObservationState.Overdue,
                    ObservationState.Upcoming)))
    .orderBy(OBSERVATIONS.COMPLETED_TIME.desc())
    .limit(1)
    .fetchOne { row -> ExistingObservationModel.of(row) }
```

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -122,6 +123,17 @@ class PlantingSiteStore(
...
+  fun fetchOldestPlantingTime(plantingSiteId: PlantingSiteId): Instant? {
+    return dslContext
+        .select(PLANTINGS.CREATED_TIME)
+        .from(PLANTINGS)
+        .where(PLANTINGS.PLANTING_SITE_ID.eq(plantingSiteId))
+        .and(PLANTINGS.CREATED_TIME.isNotNull)
+        .orderBy(PLANTINGS.CREATED_TIME.asc())
+        .limit(1)
+        .fetchOne { row -> row.value1() }
+  }
```
sgrimm: A bit more concise:
```kotlin
return dslContext
    .select(DSL.min(PLANTINGS.CREATED_TIME))
    .from(PLANTINGS)
    .where(PLANTINGS.PLANTING_SITE_ID.eq(plantingSiteId))
    .fetchOne(DSL.min(PLANTINGS.CREATED_TIME))
```

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -429,6 +441,49 @@ class PlantingSiteStore(
...
 
+  fun fetchNonNotifiedSitesToScheduleObservations(): List<PlantingSiteId> {
+    requirePermissions { manageNotifications() }
+
+    return dslContext
+        .select(PLANTING_SITES.ID)
+        .from(PLANTING_SITES)
+        .where(PLANTING_SITES.SCHEDULE_OBSERVATION_NOTIFICATION_SENT_TIME.isNull)
+        .fetch { row -> row.value1() }
+        .filter { countReportedPlantsInSubzones(it).isNotEmpty() }
```
sgrimm: This is going to have an N+1 problem with the filtering, though admittedly N will probably be small enough to not cause problems. Still, maybe better to do it in one query, something like
```kotlin
return dslContext
    .select(PLANTING_SITES.ID)
    .from(PLANTING_SITES)
    .where(PLANTING_SITES.SCHEDULE_OBSERVATION_NOTIFICATION_SENT_TIME.isNull)
    .andExists(
        DSL.selectOne()
            .from(PLANTINGS)
            .where(PLANTINGS.PLANTING_SITE_ID.eq(PLANTING_SITES.ID))
            .and(PLANTINGS.PLANTING_SUBZONE_ID.isNotNull))
    // This is a bit more efficient than the lambda
    .fetch(PLANTING_SITES.ID.asNonNullable())
```
and similarly for the query below this one.

sgrimm (commented):
I didn't go over this in fine detail; I mostly looked at the areas you mentioned in the PR description as needing feedback.

karthikbtf:
> Marking "request changes" just to make it clear that it's been looked at already

Thanks for the feedback Steve, this is helpful. I will follow up.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
karthikbtf: This is great 👍

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
karthikbtf: I may have missed adding more context here. The UI workflow to schedule/reschedule observations is hidden behind a new opt-in flag and not visible in production. Any site with a detailed map and a planting is eligible for the first 'schedule observation notification', if this is triggered in production before we ungate the UI workflow - user would be directed to a non-working UI workflow (email link). Hence this flag to disable triggering the notification, my thought was we can remove this flag after the FE schedule observation workflow is released in production. We would have to enable the config flag in staging for QA purposes.

karthikbtf:
> > Marking "request changes" just to make it clear that it's been looked at already
> 
> Thanks for the feedback Steve, this is helpful. I will follow up.

Incorporated the feedback which was very helpful, thanks again. Only outstanding item is the suggestion on how to gate, I will respond separately in that thread.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
karthikbtf: The gating mechanism here only enables/disables the notifications/reminders to schedule observations. Does not gate the scheduling of observations. The UI workflow to schedule observations is hidden behind an opt-in flag.
Plan is to enable the terraware flag in dev/staging alone, for QA purposes. When we release scheduling observations (FE) to production, we can remove this terraware config flag so notifications also work. 

If I understand your suggestion correctly, is it to gate the feature of scheduling observations (BE and FE) based on properties in the org/planting-site?

karthikbtf:
> > > Marking "request changes" just to make it clear that it's been looked at already
> > 
> > 
> > Thanks for the feedback Steve, this is helpful. I will follow up.
> 
> Incorporated the feedback which was very helpful, thanks again. Only outstanding item is the suggestion on how to gate, I will respond separately in that thread.

Also, No Rush on this PR, I will be back next Wednesday to follow up 🙏

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
sgrimm: Yeah, I guess my thought here is that it might make sense to treat "this planting site is eligible for observations" as an explicit thing in the system, and if the flag is off, you don't get the UI or the notifications.

That said, though, this flag is fine for the purpose you describe. My idea is definitely more of a product feature.

src/main/kotlin/com/terraformation/backend/daily/ObservationScheduler.kt
```diff
@@ -79,6 +85,34 @@ class ObservationScheduler(
...
+        observationService.fetchNonNotifiedSitesToRemindSchedulingObservations())
+  }
+
+  private fun notifyScheduleObservation(sites: Collection<PlantingSiteId>) {
+    sites.forEach { site ->
+      try {
+        eventPublisher.publishEvent(ScheduleObservationNotificationEvent(site))
+        plantingSiteStore.markScheduleObservationNotificationComplete(site)
+      } catch (e: Exception) {
+        log.error("Unable to mark planting site ${site} schedule observation complete")
```
sgrimm: Don't need the curly braces around the variable name here.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -317,6 +321,44 @@ class EmailNotificationService(
...
+    emailService.sendOrganizationNotification(
+        plantingSite.organizationId,
+        ScheduleObservation(
+            config,
+            plantingSite.organizationId,
+            plantingSite.id,
+            plantingSite.name,
+            webAppUrls.fullObservations(plantingSite.organizationId, plantingSite.id).toString(),
+        ),
+        roles = setOf(Role.Admin, Role.Owner, Role.Manager))
```
sgrimm: I'm surprised we want these to go out to managers, not just admins/owners (not saying the code is wrong, just surprised by the product decision).

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -626,6 +626,36 @@ class ObservationStore(
...
+  }
+
+  fun hasObservations(plantingSiteId: PlantingSiteId): Boolean {
+    return dslContext
+        .selectOne()
+        .from(OBSERVATIONS)
+        .where(OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
+        .fetch()
+        .isNotEmpty
+  }
```
sgrimm: This will return a row for each of the observations for the site; one option would be to add `.limit(1)`.

However, jOOQ has a convenience function for this pattern that takes a table name and a varargs list of conditions. I only just recently noticed it myself, but it's perfect for this:
```kotlin
return dslContext
    .fetchExists(
        OBSERVATIONS,
        OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
```

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -429,6 +431,74 @@ class PlantingSiteStore(
     }
   }
 
+  fun hasSubzonePlantings(plantingSiteId: PlantingSiteId): Boolean {
+    return dslContext
+        .selectOne()
+        .from(PLANTINGS)
+        .where(PLANTINGS.PLANTING_SITE_ID.eq(plantingSiteId))
+        .and(PLANTINGS.PLANTING_SUBZONE_ID.isNotNull)
+        .fetch()
```
sgrimm: See previous comment about `fetchExists()`

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -429,6 +431,74 @@ class PlantingSiteStore(
     }
   }
 
+  fun hasSubzonePlantings(plantingSiteId: PlantingSiteId): Boolean {
+    return dslContext
```
sgrimm: This could have a check for read permission on the planting site.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -210,4 +234,30 @@ class ObservationService(
...
+            elapsedWeeks(it, weeks)
+          }
+              ?: false)
+
+  private fun fetchNonNotifiedSitesForThresholds(
+      completedTimeElapsedWeeks: Long,
+      firstPlantingElapsedWeeks: Long,
+      siteIds: List<PlantingSiteId>
+  ): Collection<PlantingSiteId> {
+    requirePermissions { manageNotifications() }
```
sgrimm: This is harmless, of course, but I usually only put permission checks on public methods (unless the private methods have more restrictive permissions than their callers).

sgrimm (changes requested):
A few minor things but it seems good at a high level.

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
karthikbtf: This is my mistake. Should only go out to Admins and Owners.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
karthikbtf: This is great, will use `fetchExists` for these types of checks.

karthikbtf:
> A few minor things but it seems good at a high level.

Thanks for the feedback 👍 
Followed up on the feedback. Also refactored the code to remove all those identical functions, introduced a `NotificationCriteriaModel` which now defines sealed classes/trigger-params for the various use-cases. The service/stores no longer have those many functions for the various use-cases, leverages the parameters in the criteria function parameter instead. Let me know what you think of this pattern.

src/main/kotlin/com/terraformation/backend/tracking/model/NotificationCriteriaModel.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
+import com.terraformation.backend.tracking.event.ObservationSchedulingNotificationEvent
+import com.terraformation.backend.tracking.event.ScheduleObservationNotificationEvent
+import com.terraformation.backend.tracking.event.ScheduleObservationReminderNotificationEvent
+import java.time.Instant
+import org.jooq.Condition
+import org.jooq.TableField
+import org.jooq.impl.DSL
+
+class NotificationCriteriaModel {
```
sgrimm: Structurally this seems reasonable. Maybe shorten the names:

* `NotificationCriteriaModel` -> `NotificationCriteria` since it's not really a model class, just a namespace.
* `ObservationSchedulingNotifications` -> `ObservationScheduling` since "notification" is already implicit from the outer class.

src/test/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStoreTest.kt
```diff
@@ -810,4 +811,86 @@ internal class PlantingSiteStoreTest : DatabaseTest(), RunsAsUser {
...
+    fun `returns false when there are no plantings in subzones for a site with subzones`() {
+      insertPlantingSite(id = plantingSiteId)
+      insertPlantingZone()
+      insertPlantingSubzone()
+
+      assertFalse(store.hasSubzonePlantings(plantingSiteId))
+    }
+
+    @Test
+    fun `returns true when there are no plantings in subzones`() {
```
sgrimm: Should remove "no" from the name of this test.

sgrimm (changes requested):
Looks good! One test needs to be renamed, and I have another naming thought which you can take or leave as you prefer.

src/test/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStoreTest.kt
karthikbtf: Oops, thanks. Good catch. Copy/pasta

sgrimm (approved):
Looks good!

PR #1386, opened 2023-10-03
SW-4259 'Observation not scheduled' notification emails to Terraformation Contact/Support

karthikbtf:
- Send email notifications to Terraformation Contact (or Terraformation Support) when observations haven't been scheduled in a while
- First notification is sent when 8 weeks have elapsed since the last completed observation (or 6 weeks since the first planting)
- Second notification is sent when 16 weeks have elapsed since the last completed observation (or 14 weeks since the first planting)
- Added a support config in TerrawareServerConfig to configure a support email
- Email is sent to the org's Terraformation Contact user if one exists, otherwise to the support email (if configured)
- Added tests
- Tested locally with the test clock, had to increase test clock html input to 3 (to advance clock to 16 weeks)
- There is no app notification for this use case


src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -359,6 +362,39 @@ class EmailNotificationService(
...
 
+  @EventListener
+  fun on(event: ObservationNotScheduledNotificationEvent) {
+    val plantingSite =
+        plantingSiteStore.fetchSiteById(
+            event.plantingSiteId,
+            PlantingSiteDepth.Site,
+        )
+    val organizationId = parentStore.getOrganizationId(event.plantingSiteId)!!
+    // return if we don't have a TF contact to send email to
```
sgrimm: This comment seems inaccurate

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -359,6 +362,39 @@ class EmailNotificationService(
...
+    val user = getTerraformationContactUser(organizationId)
+    val organization =
+        organizationStore.fetchOneById(
+            plantingSite.organizationId, OrganizationStore.FetchDepth.Organization)
+
+    user?.let {
```
sgrimm: I think `if (user != null)` would probably be a bit easier to read here than the `let` expression.

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -104,6 +104,15 @@ class EmailService(
     }
   }
 
+  /**
+   * Sends an email notification to a support email if configured.
+   *
+   * @param [model] Model object containing values that can be referenced by the template.
+   */
+  fun sendSupportNotification(model: EmailTemplateModel) {
+    config.support.email?.let { Locale.ENGLISH.use { send(model, listOf(it)) } }
```
sgrimm: The use of `it` is slightly unclear here since it's in a nested lambda. Maybe introduce a name for it like `email`?

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -484,6 +500,27 @@ class PlantingSiteStore(
         .execute()
   }
 
+  fun markObservationNotScheduledFirstNotificationComplete(plantingSiteId: PlantingSiteId) {
+    requirePermissions { manageNotifications() }
+
+    dslContext
+        .update(PLANTING_SITES)
+        .set(PLANTING_SITES.OBSERVATION_NOT_SCHEDULED_FIRST_NOTIFICATION_SENT_TIME, clock.instant())
+        .where(PLANTING_SITES.ID.eq(plantingSiteId))
```
sgrimm: There are now enough of these that I wonder if it makes sense to refactor them to call a helper method that takes the timestamp field to update as an argument.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -89,6 +89,20 @@ notification.email.html.footer.2=PO Box 3470, PMB 15777
...
+# {0} is the name of an organization. {1} is the name of a planting site.
+notification.observation.notScheduled.email.subject=Organization {0} has not scheduled an observation of planting site {1}
+# {0} is the name of an organization. {1} is the name of a planting site.
+notification.observation.notScheduledSupport.email.body.1=Organization {0} has not scheduled an observation of planting site {1}
+notification.observation.notScheduledSupport.email.body.2=There is currently no assigned Terraformation primary project contact for this organization. Please assign one and let them know of the unscheduled observation.
+# {0} is the name of a planting site
+notification.observation.notScheduledSupport.email.body.3=Planting site {0} is ready to have it’s next observation scheduled, but the organization has not scheduled one.
+# {0} is the name of an organization. {1} is the name of a planting site.
+notification.observation.notScheduledSupport.email.subject=Organization {0} has not scheduled an observation of planting site {1}
+
```
sgrimm: Blank line will get removed by Phrase

sgrimm (changes requested):
Just some small stuff; generally this looks fine, though I worry that we're now accumulating a lot of nearly-identical notification functions.

karthikbtf:
> Just some small stuff; generally this looks fine, though I worry that we're now accumulating a lot of nearly-identical notification functions.

Thanks again for the feedback, applied the changes. Also modified the parent PR to remove the various almost identical functions, which I agree was not too pretty. Updated this PR to leverage the criteria model in the parent PR.

sgrimm (approved):
Looks good, but of course if you go with my naming suggestions on the parent PR, this one should change too. If not, this is good to go as-is once the parent is merged.

PR #1387, opened 2023-10-05
Stop adding Row suffix to embeddable POJOs

sgrimm:
If we define an embeddable type to represent a compound key, the generated POJO
class shouldn't have a Row suffix on its name because it doesn't represent a
table row.

Update the naming strategy to exclude embeddable types.

Currently, we don't have any places in the code where we're referring to these
classes by name; everywhere we use them, we use type inference. So no changes
to the application code are needed here. But this will be less confusing in
the future if/when we do need to refer to the type of a compound key.


sgrimm:
* main
  * **PR #1387** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1387" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1387?utm_source=stack-comment).

nickgraz (approved):

PR #1393, opened 2023-10-12
SW-4279 Don't send org email notifications to Terraformation Contact users

karthikbtf:
- TF Contact receives very specific email notifications as defined in the PRD
- Skip sending routine org email notifications to TF Contact users
- Refactored a couple notifications to use the `sendOrganizationNotification`
- Not sure how this current approach will ensure correctness for future email notifications
- Another option is to initialize this default roles list in the EmailService.sendOrganizationNotification function argument instead, future email notifications will by default skip TF Contacts unless explicitly passed into the roles


nickgraz (commented):
I think it is fine how it is, but if "Skip sending routine org email notifications to TF Contact users" is part of the PRD, I think it makes sense to go with your idea and bake it in to the `SendEmail.sendOrganizationNotification` args as a default. This will isolate the business logic to one place instead of having to sprinkle it in several places.

sgrimm (changes requested):
Agreed with Nick; your other option seems safer.

sgrimm (approved):

PR #1394, opened 2023-10-12
SW-4301 Refine permissions for schedule/reschedule observations

karthikbtf:
- We want admins/owners/TFContact/superadmins alone to be able to schedule/reschedule observations
- The UI handles this gating currently
- ObservationService.scheduleObservation/rescheduleObservation functions were piggy-backing onto create/update observations permissions earlier (update permissions is a bit more relaxed)
- Updated the functions them to use new permissions checks for schedule/reschedule observations
- Updated the tests

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
@@ -537,6 +537,13 @@ class PermissionRequirements(private val user: TerrawareUser) {
     }
   }
 
+  fun rescheduleObservation(observationId: ObservationId) {
+    if (!user.canRescheduleObservation(observationId)) {
+      updateObservation(observationId)
+      throw AccessDeniedException("No permission to reschedule observation $observationId")
+    }
+  }
+
   fun sendAlert(facilityId: FacilityId) {
     if (!user.canSendAlert(facilityId)) {
       readFacility(facilityId)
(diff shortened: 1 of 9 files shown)


src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
```diff
@@ -593,6 +592,21 @@ internal class PermissionRequirementsTest : RunsAsUser {
             canUpdateStorageLocation(storageLocationId)
           }
 
+  @Test
+  fun updateTerraformationContact() {
```
karthikbtf, line 596: this is an old test, reordered in alphabetical order

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -537,6 +537,13 @@ class PermissionRequirements(private val user: TerrawareUser) {
     }
   }
 
+  fun rescheduleObservation(observationId: ObservationId) {
+    if (!user.canRescheduleObservation(observationId)) {
+      updateObservation(observationId)
```
sgrimm: Is this the right fallback permission check? If the user doesn't have reschedule permission or update permission, then this will throw an error message of "No permission to update observation." I think this should be `readObservation` instead so it'll either throw "No permission to reschedule" or "Observation not found."

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -573,6 +580,14 @@ class PermissionRequirements(private val user: TerrawareUser) {
     }
   }
 
+  fun scheduleObservation(plantingSiteId: PlantingSiteId) {
+    if (!user.canScheduleObservation(plantingSiteId)) {
+      createObservation(plantingSiteId)
```
sgrimm: Similar to the above comment -- if the user lacks both of these permissions but does have read permission on the planting site, the error message will be misleading here.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
karthikbtf: Makes sense. Had started out that way, not sure why I overthought this. Will fix this and the other case.

sgrimm (approved):

PR #1404, opened 2023-10-16
Use Java 21 runtime

sgrimm:
Previously, we were targeting Java 19 with the build (because that was the newest
version supported by the Kotlin compiler at one point) and using Java 20 as the
runtime JVM.

Update the runtime to use Java 21, and the build to target Java 20.

No functional change (we aren't using any of Java 21's new features) but it
should give us some minor performance improvements.

### DIFF
--- a/jooq/build.gradle.kts
+++ b/jooq/build.gradle.kts
@@ -37,11 +37,11 @@ gradlePlugin {
   }
 }
 
-java { targetCompatibility = JavaVersion.VERSION_19 }
+java { targetCompatibility = JavaVersion.VERSION_20 }
 
 tasks.withType<KotlinCompile> {
   compilerOptions {
-    jvmTarget.set(JvmTarget.JVM_19)
+    jvmTarget.set(JvmTarget.JVM_20)
     allWarningsAsErrors.set(true)
   }
 }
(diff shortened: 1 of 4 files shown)


sgrimm:
* main
  * **PR #1404** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1404" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1404?utm_source=stack-comment).

jooq/build.gradle.kts
```diff
@@ -37,11 +37,11 @@ gradlePlugin {
   }
 }
 
-java { targetCompatibility = JavaVersion.VERSION_19 }
+java { targetCompatibility = JavaVersion.VERSION_20 }
```
nickgraz, line 40: Does this mean that we always support 1 version behind the current version?

nickgraz (approved):

jooq/build.gradle.kts
sgrimm: > Does this mean that we always support 1 version behind the current version?

Not always, but often. This is really the target compatibility of the Kotlin compiler version we're using, and the Kotlin complier tends to take a little time to add support for newly-released JVM versions. So when we upgrade Kotlin to a version that supports generating Java 21 bytecode, we can bump this up to 21.

It doesn't matter too much on a practical level since the JVM has rock-solid backward compatibility.

PR #1408, opened 2023-10-18
SW-4311 Implement monitoring plot replacement

sgrimm:
Implement the business logic for replacing a monitoring plot in an observation.

The behavior varies depending on a combination of factors: whether the given
plot is permanent or temporary in the observation, whether or not there are
already completed plots at the site from the current or a prior observation,
whether the replacement request is temporary or long-term, and whether other
subzones are finished planting or not. Various permutations, as outlined in
the requirements document, are covered by test cases.

This feature introduces the concept of a monitoring plot existing at a site,
possibly having observation results from previous observations, but being
unavailable for inclusion in future observations. Add a flag to the monitoring
plots table to keep track of availability.


sgrimm:
* main
  * **PR #1408** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1408" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1408?utm_source=stack-comment).

sgrimm:
I recommend reading the requirements doc before reviewing this.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -232,20 +236,117 @@ class ObservationService(
...
+          plantingSiteStore.fetchSiteById(observation.plantingSiteId, PlantingSiteDepth.Plot)
+      val plantingZone =
+          plantingSite.plantingZones.first { zone ->
+            zone.plantingSubzones.any { subzone ->
+              subzone.monitoringPlots.any { it.id == monitoringPlotId }
+            }
+          }
+      val plantingSubzone =
+          plantingZone.plantingSubzones.first { subzone ->
+            subzone.monitoringPlots.any { it.id == monitoringPlotId }
```
karthikbtf: extract out locally for reuse?

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -232,20 +236,117 @@ class ObservationService(
...
-    }
+      log.info("Replacing monitoring plot $monitoringPlotId in observation $observationId")
 
-    eventPublisher.publishEvent(
-        ObservationPlotReplacedEvent(duration, justification, observation, monitoringPlotId))
+      if (observationPlot.model.isPermanent) {
+        val isFirstObservation =
+            observationStore.fetchObservationsByPlantingSite(observation.plantingSiteId).none {
+              it.state == ObservationState.Completed
+            }
```
karthikbtf, line 265: Do we need to worry about multiple overdue observations for a site - with potential for reassignment of permanent plot from same cluster across those observations? Perhaps not a real world scenario.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -232,20 +236,117 @@ class ObservationService(
...
+                log.info(
+                    "Long-term replacement or no other permanent cluster available; removing " +
+                        "plot from this observation and making the other plots in its cluster " +
+                        "only available as temporary plots in future observations.")
+                plantingSiteStore.makePlotUnavailable(monitoringPlotId)
+              } else {
+                swappedPlots
+              }
+
+          // Only add the new permanent cluster if its plots are all in planted subzones; otherwise
```
karthikbtf: Just curious why the replacement cluster is the max cluster in the zone instead of the subzone, I am likely missing some nuance.

karthikbtf (approved):
This was nicely documented with tests, easy to follow. Did take me a while to wrap my head around the various cases, tests helped me understand that better.
Vaguely recall a comment in the PRD about revisiting the calculations, no changes needed there?

karthikbtf:
Thanks for moving that migration file I created into the 0200 directory, not sure how I missed that in my PR.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
sgrimm: Expanded on this comment to hopefully better explain the reasoning; does it make sense now?

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
sgrimm: In theory there's nothing in the data model to prevent that, but if it happens, there has been a major breakdown on the business side and it won't matter how we handle plot replacement. Probably the right solution is to forbid having multiple in-progress observations (and enforce that rule in the code).

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
karthikbtf: Thanks for the clarified comments, makes sense 👍

karthikbtf (approved):
Nice!

PR #1409, opened 2023-10-18
Log when scheduled tasks run

karthikbtf:
- was mostly looking for a heartbeat in the datadog logs
- if these should be debug statements I can make the change


nickgraz (approved):
Info level seems fine here

sgrimm:
I don't object to this change per se, but I'm unclear on the justification for it (the heartbeat in Datadog). Are the log messages from the load balancer health checks not usable as heartbeats?

karthikbtf:
> I don't object to this change per se, but I'm unclear on the justification for it (the heartbeat in Datadog). Are the log messages from the load balancer health checks not usable as heartbeats?

Was looking for a sign that the scheduled jobs were running, I agree this isn't something we need to spam the logs with. Not something we debug or look at often anyways. Was debugging the observation notifications and everything is good now.

PR #1423, opened 2023-10-23
SW-4336 Return monitoring plot names in the replacement result

karthikbtf:
- client needs to display the plot names from the result ids
- client can fetch the names using a search, or, the replace plot API response can include the names
- wanted to give the API response approach a try in case it is simpler than the client having to search
- if we prefer to return ids alone I can revisit the client approach

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
@@ -37,6 +37,8 @@ import com.terraformation.backend.tracking.model.NotificationCriteria
 import com.terraformation.backend.tracking.model.PlantingSiteDepth
 import com.terraformation.backend.tracking.model.ReplacementDuration
 import com.terraformation.backend.tracking.model.ReplacementResult
+import com.terraformation.backend.tracking.model.ReplacementResultIds
+import com.terraformation.backend.tracking.model.ReplacementResultPlot
 import jakarta.inject.Named
 import java.io.InputStream
 import java.time.Instant
@@ -273,7 +275,7 @@ class ObservationService(
                         "swap this permanent cluster with a higher-numbered one.")
                 plantingSiteStore.swapWithLastPermanentCluster(monitoringPlotId)
               } else {
-                ReplacementResult(emptySet(), emptySet())
+                ReplacementResultIds(emptySet(), emptySet())
               }
 
           val removalResult =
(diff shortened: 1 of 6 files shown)


sgrimm (changes requested):
The goal here makes sense, but I have a few implementation thoughts.

Moving the plot name lookup into the controller class seems to me like it'd be a slightly better separation of concerns than putting it in the service class. Efficiency should be identical since it's the same lookup either way.

If you do that, then I'd undo the rename of `ReplacementResults` and instead change `ReplacementResultPlot` to `ReplacementResultPayload` (and move it to the controller's source file) just to make it clear it's a client-facing class.

Would it make sense to return `AssignedPlotPayload` instead, so the client has everything it needs to update the table view on the fly? Not a huge difference but might save an additional round-trip to the server to refresh the plots list. If that makes sense, then you could change `ObservationStore.fetchObservationPlotDetails` to take an optional `Collection<MonitoringPlotId>?` parameter to limit the results, and then map it to `AssignedPlotPayload` the same way the controller currently does.

karthikbtf:
> The goal here makes sense, but I have a few implementation thoughts.
> 
> Moving the plot name lookup into the controller class seems to me like it'd be a slightly better separation of concerns than putting it in the service class. Efficiency should be identical since it's the same lookup either way.
> 
> If you do that, then I'd undo the rename of `ReplacementResults` and instead change `ReplacementResultPlot` to `ReplacementResultPayload` (and move it to the controller's source file) just to make it clear it's a client-facing class.
> 
> Would it make sense to return `AssignedPlotPayload` instead, so the client has everything it needs to update the table view on the fly? Not a huge difference but might save an additional round-trip to the server to refresh the plots list. If that makes sense, then you could change `ObservationStore.fetchObservationPlotDetails` to take an optional `Collection<MonitoringPlotId>?` parameter to limit the results, and then map it to `AssignedPlotPayload` the same way the controller currently does.

Moving this to the controller does sound like the right approach, thanks.

Would returning observation plot details only return the replaced plot information, and lose the removed plots info?

On the fly update in client sounds like an efficient approach, we would have to apply the incremental changes to the redux store. I can revisit this later, something we could do for a bunch of other API as well - could be a bunch of refactor.

Let me try the client side approach of searching monitoring plots by id, seems like this should be the right/simpler approach for now.

sgrimm:
> Would returning observation plot details only return the replaced plot information, and lose the removed plots info?

Yeah, maybe makes sense to only return the IDs for removed plots since the other details would be irrelevant.

karthikbtf:
Searching for monitoring plot names client-side seems to be working well. Will close this.

PR #1429, opened 2023-10-25
SW-4361 Add internal tag for accelerator participants

sgrimm:
We're going to be adding system behavior that only applies to participants in
the Terraformation Accelerator. Add an internal tag so super-admins can indicate
which organizations are in the accelerator.

Nothing uses this new tag yet, but it will appear in the admin UI.


sgrimm:
* main
  * **PR #1429** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1429" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1429?utm_source=stack-comment).

charlie-tf (approved):

PR #1438, opened 2023-10-30
SW-4365 Carry accession ID across nursery transfers

sgrimm:
If a batch was originally created via a nursery transfer from a seed bank, it will
have an accession ID indicating where it came from. If plants from the batch are
then transferred to a different nursery, the newly-created batch should have the
same accession ID so it can be traced back to its original seeds.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
@@ -655,6 +655,7 @@ class BatchStore(
         val newBatch =
             create(
                 NewBatchModel(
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1438** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1438" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1438?utm_source=stack-comment).

nickgraz (approved):

PR #1440, opened 2023-10-31
SW-4376 Add API to change batch statuses

sgrimm:
Calculating germination and loss rates will require that we distinguish between
the user overwriting the quantities of a batch and the user moving seedlings from
one status to another.

Add a new API endpoint `POST /api/v1/nursery/batches/{id}/changeStatuses` to move
seedlings between statuses. It causes a new type of quantity history entry to be
recorded, which will allow the germination and loss rate calculations to treat
status changes correctly.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
@@ -328,6 +328,34 @@ class BatchStore(
     }
   }
 
+  fun changeStatuses(
+      batchId: BatchId,
+      germinatingQuantityToChange: Int,
+      notReadyQuantityToChange: Int
+  ) {
+    requirePermissions { updateBatch(batchId) }
+
+    if (germinatingQuantityToChange == 0 && notReadyQuantityToChange == 0) {
+      return
+    }
+
+    retryVersionedBatchUpdate(batchId) { batch ->
+      if (batch.germinatingQuantity < germinatingQuantityToChange ||
+          batch.notReadyQuantity < notReadyQuantityToChange) {
+        throw BatchInventoryInsufficientException(batchId)
+      }
+
+      updateQuantities(
+          batchId,
+          batch.version,
+          batch.germinatingQuantity - germinatingQuantityToChange,
+          batch.notReadyQuantity - notReadyQuantityToChange + germinatingQuantityToChange,
+          batch.readyQuantity + notReadyQuantityToChange,
+          BatchQuantityHistoryType.StatusChanged,
+      )
+    }
+  }
+
   fun delete(batchId: BatchId) {
     requirePermissions { deleteBatch(batchId) }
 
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #1440** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1440" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1440?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -505,83 +533,56 @@ class BatchStore(
...
+                  // recently than the withdrawal, just record the withdrawal without subtracting
+                  // it from inventory.
+                  //
+                  // We figure this out inside the retry loop because the latest observed time
+                  // might have been updated by whatever operation caused our version number to
+                  // be out of date.
+                  val latestObservedDate =
+                      LocalDate.ofInstant(batch.latestObservedTime, nurseryTimeZone)
+                  val withdrawalIsNewerThanObservation =
+                      withdrawal.withdrawnDate >= latestObservedDate
```
karthikbtf, line 549: should this be just '>' to be considered newer?

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -328,6 +328,34 @@ class BatchStore(
...
+
+    if (germinatingQuantityToChange == 0 && notReadyQuantityToChange == 0) {
+      return
+    }
+
+    retryVersionedBatchUpdate(batchId) { batch ->
+      if (batch.germinatingQuantity < germinatingQuantityToChange ||
+          batch.notReadyQuantity < notReadyQuantityToChange) {
+        throw BatchInventoryInsufficientException(batchId)
+      }
```
karthikbtf, line 346: is this within the retry so we can capture issues from concurrent edits?

karthikbtf (approved):
Looks good, some comments.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: That won't work in the case where you create a batch on the same day you withdraw from it: the withdrawal won't be treated as newer than the initial observation and the quantities won't be deducted.

Fundamentally the problem is that we're prompting the user for a date and we have no way of telling what time of day the withdrawal happened, so we have to break the tie somewhat arbitrarily.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
sgrimm: Yes,  exactly -- someone else might have withdrawn or changed statuses such that the current status change can no longer succeed.

PR #1442, opened 2023-10-31
SW-4408 Change criteria for observations that can be rescheduled

karthikbtf:
- Upcoming observations can be rescheduled
- Overdue or InProgress observations can be rescheduled only if they have no observed plots
- Completed observations cannot be rescheduled

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
@@ -649,7 +649,7 @@ class ObservationServiceTest : DatabaseTest(), RunsAsUser {
       insertMonitoringPlot()
 
       val observationId = insertObservation(state = ObservationState.valueOf(stateName))
-      insertObservationPlot(completedTime = Instant.EPOCH)
+      insertObservationPlot(claimedBy = user.userId, completedBy = user.userId)
 
       val startDate = LocalDate.EPOCH
       val endDate = startDate.plusDays(1)
--- a/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
@@ -68,6 +68,8 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
 import org.springframework.http.MediaType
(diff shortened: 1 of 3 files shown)


src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
```diff
@@ -619,8 +621,35 @@ class ObservationServiceTest : DatabaseTest(), RunsAsUser {
...
+    fun `throws exception rescheduling an In-Progress or Overdue observation if there are observed plots`(
+        stateName: String
+    ) {
+      insertPlantingSite()
+      insertPlantingZone(numPermanentClusters = 1, numTemporaryPlots = 1)
+      insertPlantingSubzone()
+      insertMonitoringPlot()
+
+      val observationId = insertObservation(state = ObservationState.valueOf(stateName))
+      insertObservationPlot(completedTime = Instant.EPOCH)
```
sgrimm: This will probably fail; you need to set both the claimed and completed fields. If you do `insertObservationPlot(claimedBy = user.userId, completedBy = user.userId)` it will fill in the timestamps for you.

That said, maybe it makes sense to automatically set claimed by/time in `insertObservationPlot()` if completed by/time are set; the way it is now is extra verbosity for no real benefit.

karthikbtf: Sounds good, will make that change. This test passed btw, but the change does make sense.. will update.

src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
```diff
@@ -619,8 +621,35 @@ class ObservationServiceTest : DatabaseTest(), RunsAsUser {
...
+    fun `throws exception rescheduling an In-Progress or Overdue observation if there are observed plots`(
+        stateName: String
+    ) {
+      insertPlantingSite()
+      insertPlantingZone(numPermanentClusters = 1, numTemporaryPlots = 1)
+      insertPlantingSubzone()
+      insertMonitoringPlot()
+
+      val observationId = insertObservation(state = ObservationState.valueOf(stateName))
+      insertObservationPlot(claimedBy = user.userId, completedBy = user.userId)
```
karthikbtf, line 652: Setting both claimedBy and completedBy here.
Was having trouble setting claimedBy to the value of completedBy in insertObservationPlot without breaking the alphabetical order of function arguments.

Screen Shot 2023-10-31 at 4 23 42 PM

sgrimm: Yeah, you'd have to change the order to make this work. What you're doing here is fine for now; we can tweak the insert function in a separate PR since we'd also want to update all the call sites which would be out of scope for this PR.

sgrimm (approved):

PR #1443, opened 2023-10-31
SW-4407 Fix adding user to org from admin UI

sgrimm:
Commit 7c9a28a introduced the ability for a super-admin to add a user to any
organization without the super-admin having to be a member of the organization.
Unfortunately, this would fail when the system tried to generate in-app and
email notifications to tell the user they'd been added.

For now, switch to the system user when generating these notifications. We will be
revamping the permission system for super-admins in the near future, at which
point we'll replace this with something more robust.


sgrimm:
* main
  * **PR #1443** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1443" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1443?utm_source=stack-comment).

karthikbtf (approved):
Fair enough.

PR #1451, opened 2023-11-01
Return scheduled 'endDate' of observation in API response results payload 

karthikbtf:
- Populated 'endDate' in ObservationResultsModel
- This will be [shown](https://terraformation.atlassian.net/browse/SW-3999) in the observation results table


karthikbtf:
Client can also fetch it from the observation payload, if we prefer to not add this in Observation results responses. We do send the start date with the results.

karthikbtf:
Will use the endDate in observation response payloads.

nickgraz (approved):

PR #1459, opened 2023-11-06
SW-4418 Aggregate germination and loss rates

sgrimm:
Return the aggregate germination and loss rates in the summary data for species
and for nursery facilities. This replaces the existing aggregate loss rate
calculation, which was less sophisticated than the current one.

To support calculating the aggregate numbers without having to walk through the
quantity history of every batch, the batches table now stores the numerators and
denominators of the batch-level rate calculations, such that the SQL query that
reads the aggregate stats can simply add them up and divide to get the aggregate
rates.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
@@ -267,14 +267,20 @@ class BatchStore(
             .filter { it.value1() == WithdrawalPurpose.Dead }
             .sumOf { it.value2().toLong() }
 
-    val totalIncludingWithdrawn = totalWithdrawn + (inventory?.totalQuantity ?: 0)
-    val lossRate =
-        if (totalIncludingWithdrawn > 0) {
-          // Round to nearest integer percentage.
-          (totalDead * 100 + totalIncludingWithdrawn / 2) / totalIncludingWithdrawn
-        } else {
-          0
-        }
+    val (germinationRate: Double?, lossRate: Double?) =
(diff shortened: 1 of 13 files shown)


sgrimm:
* main
  * **PR #1459** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1459" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1459?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -949,13 +956,24 @@ class BatchStore(
...
-                DSL.sum(BATCHES.READY_QUANTITY))
+                DSL.sum(BATCHES.READY_QUANTITY),
+                DSL.sum(BATCHES.TOTAL_GERMINATED)
+                    .times(100.0)
+                    .div(DSL.sum(BATCHES.TOTAL_GERMINATION_CANDIDATES))
+                    .cast(Double::class.java),
+                DSL.sum(BATCHES.TOTAL_LOST)
+                    .times(100.0)
+                    .div(DSL.sum(BATCHES.TOTAL_LOSS_CANDIDATES))
+                    .cast(Double::class.java),
```
karthikbtf: extract out for reuse (if it makes sense to do so)?

karthikbtf (approved):
Looks good to me 👍

PR #1461, opened 2023-11-07
SW-4428 Force send email notification to TF contact on plot replacement request

karthikbtf:
- disable the opt in flag option
- also added an email title (matching the subject) similar to other observation notification emails (i can remove this if needed, this is mostly for visual appeal)
- wrapped freemarker tags with `<!-- htmlmin:ignore>`

DEV  Organization my org has requested an observation plot change - karthik balach@gmail com - Gmail 2023-11-07 11-36-31

### DIFF
--- a/src/main/resources/templates/email/observation/plotReplaced/body.ftlh.mjml
+++ b/src/main/resources/templates/email/observation/plotReplaced/body.ftlh.mjml
@@ -8,19 +8,22 @@
 
         <mj-section padding="0px">
             <mj-column mj-class="body-wrapper">
-                <#if !hasPrimaryContact>
+                <mj-text mj-class="text-headline06-bold">
+                ${strings("notification.observation.monitoringPlotReplaced.email.body.1", organizationName)}
+                </mj-text>
+                <!-- htmlmin:ignore --><#if !hasPrimaryContact><!-- htmlmin:ignore -->
                     <mj-text mj-class="text-body03">
-                        ${strings("notification.observation.monitoringPlotReplaced.email.body.1")}
+                        ${strings("notification.observation.monitoringPlotReplaced.email.body.2")}
                     </mj-text>
-                </#if>
+                <!-- htmlmin:ignore --></#if><!-- htmlmin:ignore -->
                 <mj-text mj-class="text-body03">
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/email/observation/plotReplaced/body.ftlh.mjml
```diff
@@ -8,19 +8,22 @@
 
         <mj-section padding="0px">
             <mj-column mj-class="body-wrapper">
-                <#if !hasPrimaryContact>
+                <mj-text mj-class="text-headline06-bold">
+                ${strings("notification.observation.monitoringPlotReplaced.email.body.1", organizationName)}
+                </mj-text>
+                <!-- htmlmin:ignore --><#if !hasPrimaryContact><!-- htmlmin:ignore -->
```
karthikbtf: Couldn't get this to work, taking a look in the debugger. Since we will be removing this content maybe it's not so urgent.

src/main/resources/templates/email/observation/plotReplaced/body.ftlh.mjml
```diff
@@ -8,19 +8,22 @@
...
-                <#if !hasPrimaryContact>
-                    <mj-text mj-class="text-body03">
-                        ${strings("notification.observation.monitoringPlotReplaced.email.body.1")}
-                    </mj-text>
-                </#if>
+                <mj-text mj-class="text-headline06-bold">
+                ${strings("notification.observation.monitoringPlotReplaced.email.body.1", organizationName)}
+                </mj-text>
+                <mj-text mj-class="text-body03">
+                  <!-- htmlmin:ignore --><#if !hasPrimaryContact><!-- htmlmin:ignore -->
```
karthikbtf, line 15: Got this to work by moving it inside the mjml tag, looked at device/unknownAutomation as a guide.
When the ignore tags were outside mjml, the freemarker tags were not being preserved (looked in build/resource/..templates/email/..plotReplaced/body.ftlh).

sgrimm: That makes sense now that I see it. The MJML tool doesn't know anything about those comments, just treats them as content, so if they appeared where it didn't expect content, it would just treat them as if they were actual comments and ignore them.

PR #1463, opened 2023-11-08
SW-4432 Make batch history payload OpenAPI schema TS-friendly

sgrimm:
The code generator we use to generate TypeScript types from the OpenAPI schema
doesn't make use of discriminator mappings on polymorphic payloads, so it was
failing to generate usable types for the batch history payloads.

Change the payload type hierarchy and annotations to produce a schema that is
still correct but has what the code generator needs:

- `BatchHistoryPayload` no longer has any properties, which allows the code
  generator to turn it into a simple union type. The common properties have
  been moved to a new `BatchHistoryPayloadCommonProps` class, which each
  subclass includes in its "allOf" schema property.
- The `type` field is on each subclass is now a string rather than an enum.
  It now has a a `@Schema` annotation on it to specify that it must have a
  specific value. This allows the code generator to restrict each subclass's
  `type` value to a specific string so it can be used as a discriminator.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesHistoryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesHistoryController.kt
@@ -146,17 +146,14 @@ class BatchesHistoryController(
   }
 }
 
-enum class BatchHistoryPayloadType {
-  DetailsEdited,
-  IncomingWithdrawal,
-  OutgoingWithdrawal,
-  PhotoCreated,
-  PhotoDeleted,
-  QuantityEdited,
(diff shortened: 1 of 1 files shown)


sgrimm:
* main
  * **PR #1463** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1463" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1463?utm_source=stack-comment).

charlie-tf (approved):
Cool. Looks good!

PR #1464, opened 2023-11-09
SW-4439 Throw better error for unsupported image type

sgrimm:
Currently, the content type validation in the image upload endpoints just looks at
the client-supplied MIME type and believes it. It's thus possible for a client to
say it's uploading a JPEG image but actually upload some other kind of file, and
the server will store it in the file store.

But when we then try to generate a thumbnail for that image, it will fail because
the Java ImageIO class doesn't know how to load images of unsupported types.

Previously, this resulted in a NullPointerException with no additional detail.
Add logic to try to detect the file type from its content and throw a more useful
exception so we can diagnose why thumbnail generation is failing.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/util/ImageUtils.kt
+++ b/src/main/kotlin/com/terraformation/backend/util/ImageUtils.kt
@@ -9,8 +9,10 @@ import java.net.URI
 import javax.imageio.ImageIO
 import net.coobird.thumbnailator.filters.Flip
 import net.coobird.thumbnailator.filters.Rotation
+import org.apache.tika.Tika
+import org.springframework.web.reactive.function.UnsupportedMediaTypeException
 
-/** Utility class for image maniuplation */
+/** Utility class for image manipulation */
 @Named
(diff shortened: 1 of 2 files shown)


sgrimm:
* main
  * **PR #1464** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1464?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1465** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1465?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1466** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1464?utm_source=stack-comment).

charlie-tf (approved):
Nice!

PR #1470, opened 2023-11-10
SW-4429 Make "has no contact" a separate notification

sgrimm:
Some notifications go to the organization's primary project contact, and if there
isn't a contact, they're instead sent to the support mailing list.

Previously, if we had to fall back on the support list, we would add a paragraph
to the text of the notification email about the organization not having a contact.
This meant that each individual notification would have to have conditional logic
to either include that paragraph or not.

Replace that approach with a separate "this org has no contact" notification.

Since the new notification is only ever sent to the support mailing list, its text
is included directly in the template files rather than coming from the localizable
strings list.


sgrimm:
* main
  * **PR #1470** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1470?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1470?utm_source=stack-comment).

src/main/resources/templates/email/organization/missingContact/body.ftlh.mjml
```diff
@@ -9,13 +8,11 @@
...
-                ${strings("notification.observation.notScheduledSupport.email.body.1", organizationName, plantingSiteName)}
+                    Organization ${organizationName} has no contact!
                 </mj-text>
                 <mj-text mj-class="text-body03">
-                ${strings("notification.observation.notScheduledSupport.email.body.2")}
-                </mj-text>
-                <mj-text mj-class="text-body03">
-                ${strings("notification.observation.notScheduledSupport.email.body.3", plantingSiteName)}
+                    There is currently no assigned Terraformation primary project contact for
+                    organization ${organizationName}. Please assign one.
```
karthikbtf: Should we include the org id since org names aren't unique?

karthikbtf (approved):
Thanks, changes look good to me, clean straightforward. One suggestion but feel free to ignore if it won't help.

karthikbtf (approved):
LGTM 👍

src/main/resources/templates/email/organization/missingContact/body.ftlh.mjml
```diff
@@ -9,13 +8,11 @@
         <mj-section padding="0px">
             <mj-column mj-class="body-wrapper">
                 <mj-text mj-class="text-headline06-bold">
-                ${strings("notification.observation.notScheduledSupport.email.body.1", organizationName, plantingSiteName)}
+                    Organization ${organizationName} has no contact!
```
nickgraz, line 11: Is the assumption that the "support mailing list" recipients will always speak English? Is someone at Terraformation in the support mailing list?

nickgraz (approved):
Seems legit assuming the support mailing list is guaranteed to have English reading recipients.

src/main/resources/templates/email/organization/missingContact/body.ftlh.mjml
sgrimm: Yes, the support mailing list is an internal thing and it will always be English-speaking TF employees reading it.

PR #1474, opened 2023-11-14
SW-4411 Rework planting site notification schema

sgrimm:
We're about to add several more reminder notifications for planting sites, and
it will start to get a little unwieldy to have a separate column in the planting
sites table to keep track of each one.

Create a new table `planting_site_notifications` that tracks how many of each
type of notification we've sent for a planting site. This models the reminder
notifications a bit differently than before: they are considered to have the
same notification type but a different "notification number" where the first
notification about something is number 1, the first reminder is number 2, etc.


sgrimm:
* main
  * **PR #1474** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1474?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1474?utm_source=stack-comment).

src/main/resources/db/migration/0200/V225__PlantingSiteNotifications.sql
```diff
@@ -0,0 +1,55 @@
+CREATE TABLE tracking.planting_site_notifications (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    planting_site_id BIGINT NOT NULL REFERENCES tracking.planting_sites ON DELETE CASCADE,
+    notification_type_id INTEGER NOT NULL REFERENCES notification_types,
```
karthikbtf, line 4: should we validate that notification type is one of  schedule observation, observation not scheduled (support)?

src/main/kotlin/com/terraformation/backend/tracking/model/NotificationCriteria.kt
```diff
@@ -1,74 +1,89 @@
...
 
-  object ScheduleObservations : ObservationScheduling {
+  data object ScheduleObservations : ObservationScheduling {
     override val completedTimeElapsedWeeks: Long = 2
     override val firstPlantingElapsedWeeks: Long = 0
-    override val notificationNotCompletedCondition: Condition =
-        DSL.condition(PLANTING_SITES.SCHEDULE_OBSERVATION_NOTIFICATION_SENT_TIME.isNull)
-    override val notificationCompletedField: TableField<PlantingSitesRecord, Instant?> =
-        PLANTING_SITES.SCHEDULE_OBSERVATION_NOTIFICATION_SENT_TIME
+    override val notificationType: NotificationType = NotificationType.ScheduleObservation
```
karthikbtf, line 53: should be `ScheduleObservationReminder`? same for number=2

karthikbtf (commented):
This is a very nice approach. Thanks for making the reminders extensible without adding more property columns.
A few comments/questions.

src/main/kotlin/com/terraformation/backend/tracking/model/NotificationCriteria.kt
karthikbtf: My mistake, ScheduleObservation is correct.

karthikbtf (approved):

PR #1477, opened 2023-11-15
SW-4457 Schema for organization type and additional properties

karthikbtf:
- Plan is to capture 3 new categories of information 
  a) managed facility types - 0 or more of (seed bank, nursery, planting site) 
  b) organization type (predetermined list of types), for 'Other' we want to capture additional information 'type details' (100 char limit) 
  c) website - text with no formatting restrictions
- All of these are optional in the db (no defaulting to any values for existing orgs)
- Client will require new orgs to pick an organization type (b), existing orgs can populate them in the org details page at any time
- The managed facilities property is not re-rendered in the client today, we *may* choose to do this later. For now it's just ingested.
- Wondering if this property should just be a comma joined string of facility types (only consumed internally by us), client would have to handle i18n display and map it back to english values for persistence. For now it is a bigger production just to be consistent with how we store multiple values of any type.


src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
```diff
@@ -0,0 +1,19 @@
...
+
+CREATE TABLE organization_types (
+    id INTEGER NOT NULL PRIMARY KEY,
+    name TEXT NOT NULL
+);
+
+CREATE TABLE organization_managed_facility_types (
+  organization_id BIGINT NOT NULL REFERENCES organizations,
+  managed_facility_type_id INTEGER NOT NULL REFERENCES managed_facility_types,
+  UNIQUE (organization_id, managed_facility_type_id)
```
charlie-tf: Would it be better to make this `PRIMARY KEY (organization_id, managed_facility_type_id)`?

karthikbtf: I can do that

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -157,8 +160,10 @@ val ID_WRAPPERS =
                     "GbifTaxonId",
                     listOf("gbif_taxa\\.id", "gbif_.*\\.taxon_id", "gbif_.*\\..*_usage_id")),
                 IdWrapper("InternalTagId", listOf("internal_tags\\.id", ".*\\.internal_tag_id")),
+                IdWrapper("ManagedFacilityTypeId", listOf(".*\\.managed_facility_type_id")),
```
sgrimm: This should be included in the column list for the enum type rather than defined as an ID wrapper.

sgrimm (changes requested):

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
```diff
@@ -0,0 +1,22 @@
...
+  managed_facility_type_id INTEGER NOT NULL REFERENCES managed_facility_types,
+  PRIMARY KEY (organization_id, managed_facility_type_id)
+);
+
+ALTER TABLE organizations ADD COLUMN type_id INTEGER REFERENCES organization_types;
+ALTER TABLE organizations ADD COLUMN type_details VARCHAR(100);
+ALTER TABLE organizations ADD COLUMN website TEXT;
+
+/* Allow type details to be populated only if type id is 'Other' (6) */
+ALTER TABLE organizations ADD CONSTRAINT other_type_details CHECK (type_details IS NULL OR (type_details <> '' AND type_id = 6));
```
charlie-tf: Does this check fail if `type_details = ''`, regardless of the value for `type_id`?

charlie-tf: Should it instead be `CHECK (type_details IS NULL OR (type_details IS NOT NULL AND type_id = 6));`?

charlie-tf (commented):
Question about the constraint

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
charlie-tf: Oh, maybe it should be `CHECK (type_details IS NULL OR type_id = 6);`

karthikbtf: we want a non-empty value for type details if type id is 'Other'

charlie-tf: Oh, I see. It looks like the constraint as written should allow for `type_details` to be NULL and `type_id = 6` at the same time.

karthikbtf: I will update the comment.

karthikbtf: > Oh, I see. It looks like the constraint as written should allow for `type_details` to be NULL and `type_id = 6` at the same time.

Ugh, yes, good catch - will fix that. Thanks!!

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
```diff
@@ -0,0 +1,23 @@
...
+  PRIMARY KEY (organization_id, managed_facility_type_id)
+);
+
+ALTER TABLE organizations ADD COLUMN organization_type_id INTEGER REFERENCES organization_types;
+ALTER TABLE organizations ADD COLUMN organization_type_details VARCHAR(100);
+ALTER TABLE organizations ADD COLUMN website TEXT;
+
+/* Require non-empty type details if type id is 'Other' (6), type details should be null otherwise. */
+ALTER TABLE organizations ADD CONSTRAINT other_type_details
+  CHECK ((organization_type_details IS NULL AND organization_type_id != 6) OR (organization_type_details <> '' AND organization_type_id = 6));
```
charlie-tf: Sorry to nitpick! But I think this still allows `organization_type_details IS NULL` at the same time as `organization_type_id = 6`

charlie-tf: Should the second clause be `(organization_type_details IS NOT NULL AND organization_type_details <> '' AND organization_type_id = 6)`?

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -159,6 +162,7 @@ val ID_WRAPPERS =
                 IdWrapper("InternalTagId", listOf("internal_tags\\.id", ".*\\.internal_tag_id")),
                 IdWrapper("NotificationId", listOf("notifications\\.id", ".*\\.notification_id")),
                 IdWrapper("OrganizationId", listOf("organizations\\.id", ".*\\.organization_id")),
+                IdWrapper("OrganizationTypeId", listOf("organizations\\.type_id")),
```
sgrimm: This could also be on the enum table's column list.

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
karthikbtf: No worries. I believe `<> ''` catches not null and not empty.

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
karthikbtf: Oops, I added it to the enum table's list but missed removing it from here.

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
karthikbtf: Fwiw, I am adding tests in an upcoming PR.

charlie-tf: >No worries. I believe <> '' catches not null and not empty.

Okay! I didn't realize that. On googling a little bit, you seem to be correct :laughing:

karthikbtf: > > No worries. I believe <> '' catches not null and not empty.
> 
> Okay! I didn't realize that. On googling a little bit, you seem to be correct 😆

I added the check anyways to avoid confusion :)

karthikbtf: Ok looks like the `<> ''` only may work in sql but not postgresql. Testing locally in psql I needed both the 'is not null' and '<> ''' conditions.

charlie-tf: Oh! Thanks so much for testing that out ❤️

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -41,6 +41,8 @@ val ENUM_TABLES =
                 EnumTable("facility_connection_states", "facilities\\.connection_state_id"),
                 EnumTable("facility_types", "facilities\\.type_id"),
                 EnumTable("growth_forms", listOf("growth_forms\\.id", ".*\\.growth_form_id")),
+                EnumTable("managed_facility_types", listOf("managed_facility_types\\.id",
```
sgrimm: Is "managed facility type" an engineering invention, or is it what the ops team uses for this? I find it a bit confusing in that we already have a concept of a "facility type" in the system that has nothing to do with this, and also there are types here (planting sites) that aren't facilities in the sense we use that word elsewhere in the system.

karthikbtf: This is what I am trying to capture with the 'managed facility type'
```
“Does your organization manage any of the following? Check all that apply.”

Options are checkboxes:
- Seed Bank/Seed Collection
- Nursery
- Planting Site
```
The value is only captured for downstream analytics, will not be re-rendered in the client (as of current plans).
I could also make it a comma separated value of 'TEXT' type.

sgrimm: The word "facility" is my main concern here because it has a specific meaning in our software. The structure you have here with an enum table is totally fine.

How about "managed location types" or "managed operational areas" or something like that?

karthikbtf: > The word "facility" is my main concern here because it has a specific meaning in our software. The structure you have here with an enum table is totally fine.
> 
> How about "managed location types" or "managed operational areas" or something like that?

Location types sounds good to me. Will use that, thanks so much!

karthikbtf: > > The word "facility" is my main concern here because it has a specific meaning in our software. The structure you have here with an enum table is totally fine.
> > How about "managed location types" or "managed operational areas" or something like that?
> 
> Location types sounds good to me. Will use that, thanks so much!

Saw your comment in the JIRA, thanks.  Will go with 'location' for now and update later if there's any feedback in the JIRA.

src/main/resources/db/migration/0200/V226__OrganizationTypes.sql
```diff
@@ -0,0 +1,27 @@
...
+ALTER TABLE organizations ADD COLUMN organization_type_id INTEGER REFERENCES organization_types;
+ALTER TABLE organizations ADD COLUMN organization_type_details VARCHAR(100);
+ALTER TABLE organizations ADD COLUMN website TEXT;
+
+/* Require non-empty type details if type id is 'Other' (6), type details should be null otherwise. */
+ALTER TABLE organizations ADD CONSTRAINT other_type_details
+  CHECK (
+    (organization_type_details IS NULL AND organization_type_id != 6)
+      OR
+    (organization_type_details IS NOT NULL AND organization_type_details <> '' AND organization_type_id = 6)
```
sgrimm: Another option here is to split the empty-string check out into its own constraint, like
```
CHECK (organization_type_details IS NULL OR organization_type_details != ''),
CHECK ((organization_type_details IS NULL AND organization_type_id != 6)
    OR (organization_type_details IS NOT NULL AND organization_type_id = 6))
```

src/main/resources/i18n/Enums_en.properties
```diff
@@ -83,6 +86,12 @@ public.NotificationType.ScheduleObservationReminder=Reminder To Schedule Observa
 public.NotificationType.SensorOutOfBounds=Sensor Out Of Bounds
 public.NotificationType.UnknownAutomationTriggered=Unknown Automation Triggered
 public.NotificationType.UserAddedToOrganization=User Added to Organization
+public.OrganizationType.Government=Governmental Agency
+public.OrganizationType.NGO=Non-Governmental Organizations(NGO)
```
sgrimm: Maybe a space before the `(` here.

sgrimm (approved):

sgrimm (approved):
Changes look good.

karthikbtf:
> Changes look good.

Thanks for the review. I will follow up on website validation once we hear back on JIRA.

PR #1478, opened 2023-11-15
SW-4458 API updates to populate organization type and additional properties

karthikbtf:
- updated payloads in create/update organization to support new properties
- updated store to account for these properties
- added tests
- tested REST API using a local client

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
@@ -17,7 +17,9 @@ import com.terraformation.backend.db.CannotRemoveLastOwnerException
 import com.terraformation.backend.db.OrganizationHasOtherUsersException
 import com.terraformation.backend.db.OrganizationNotFoundException
 import com.terraformation.backend.db.UserNotFoundException
+import com.terraformation.backend.db.default_schema.ManagedFacilityType
 import com.terraformation.backend.db.default_schema.OrganizationId
+import com.terraformation.backend.db.default_schema.OrganizationType
 import com.terraformation.backend.db.default_schema.Role
 import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.db.default_schema.tables.pojos.OrganizationsRow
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
```diff
@@ -148,25 +155,40 @@ class OrganizationStore(
...
             modifiedTime = clock.instant(),
             name = row.name,
+            organizationTypeId = row.organizationTypeId,
+            organizationTypeDetails = row.organizationTypeDetails,
             timeZone = row.timeZone,
+            website = row.website,
         )
 
-    organizationsDao.insert(fullRow)
+    return dslContext.transactionResult { _ ->
```
karthikbtf, line 164: If the missing transaction scope was intentional, I can remove this.

sgrimm: Not intentional; this is a good change.

src/main/kotlin/com/terraformation/backend/customer/db/OrganizationStore.kt
```diff
@@ -148,25 +155,40 @@ class OrganizationStore(
...
+            .execute()
+      }
 
-    return fullRow.toModel(totalUsers = 1)
+      managedFacilityTypes.forEach {
+        with(ORGANIZATION_MANAGED_FACILITY_TYPES) {
+          dslContext
+              .insertInto(ORGANIZATION_MANAGED_FACILITY_TYPES)
+              .set(ORGANIZATION_ID, fullRow.id)
+              .set(MANAGED_FACILITY_TYPE_ID, it)
```
sgrimm: Kotlin style: only use the default `it` lambda function parameter for single-line function bodies. Or in other words, `forEach` should specify a variable name.

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -252,16 +256,29 @@ data class CreateOrganizationRequestPayload(
         maxLength = 6)
     val countrySubdivisionCode: String?,
     val description: String?,
+    val managedLocationTypes: List<ManagedLocationType>?,
```
sgrimm: Minor nit: Could make this a `Set` instead of a `List` so the OpenAPI schema will indicate that there shouldn't be duplicate values.

sgrimm (approved):

sgrimm (approved):

PR #1484, opened 2023-11-17
SW-4466 Add shapefile area sanity check

sgrimm:
Detect when someone tries to upload a shapefile that spans a very large area,
rather than attempting to generate a covering grid of monitoring plots; the
grid generation would eventually succeed, but would take long enough for the
shapefile upload to time out.

The size limit here is somewhat arbitrary, but is significantly larger than
any of the planting sites we've already created or any of the large dummy
sites we've used for testing.


sgrimm:
* main
  * **PR #1484** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1484?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1484?utm_source=stack-comment).

nickgraz (approved):

PR #1503, opened 2023-11-22
SW-4474 Add user API to delete a planting site, with restrictions

karthikbtf:
- planting site can be deleted if it has no plantings
- introduced user-facing deletePlantingSite in PlantingSiteService to differentiate from default implementation in PlantingSiteStore
- added tests

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -476,6 +476,15 @@ class PlantingSiteStore(
     )
   }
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -476,6 +476,15 @@ class PlantingSiteStore(
     )
   }
 
+  fun hasPlantings(plantingSiteId: PlantingSiteId): Boolean {
```
karthikbtf, line 479: There is a `hasSubzonePlantings` as well, not sure if it would be better to merge the two and expose additional arguments.

nickgraz (approved):

PR #1516, opened 2023-11-28
Improve TestEventPublisher

sgrimm:
Add the ability to clear the event list to support asserting that no additional
events were published after a certain point.

Add an assertion method to check that no events were published.

Allow callers to pass custom messages to the assertion methods.

PR #1517, opened 2023-11-28
SW-4412 Mark planting seasons as active/inactive

sgrimm:
Add a scheduled job that transitions planting seasons into and out of active state
as their start and end times arrive. Each transition generates an application
event which will be used to trigger notifications.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -35,8 +35,10 @@ import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONE
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_ZONES
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_ZONE_POPULATIONS
 import com.terraformation.backend.log.perClassLogger
+import com.terraformation.backend.tracking.event.PlantingSeasonEndedEvent
 import com.terraformation.backend.tracking.event.PlantingSeasonRescheduledEvent
 import com.terraformation.backend.tracking.event.PlantingSeasonScheduledEvent
+import com.terraformation.backend.tracking.event.PlantingSeasonStartedEvent
 import com.terraformation.backend.tracking.event.PlantingSiteDeletionStartedEvent
 import com.terraformation.backend.tracking.model.CannotUpdatePastPlantingSeasonException
 import com.terraformation.backend.tracking.model.ExistingPlantingSeasonModel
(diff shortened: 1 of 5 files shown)


sgrimm:
* main
  * **PR #1516** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1516?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1517** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1517?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #1518** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1517?utm_source=stack-comment).

karthikbtf (approved):
Looks straightforward to me 👍

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -979,4 +986,77 @@ class PlantingSiteStore(
...
+  private fun endPlantingSeasons() {
+    dslContext
+        .select(
+            PLANTING_SEASONS.PLANTING_SITE_ID.asNonNullable(), PLANTING_SEASONS.ID.asNonNullable())
+        .from(PLANTING_SEASONS)
+        .where(PLANTING_SEASONS.END_TIME.le(clock.instant()))
+        .and(PLANTING_SEASONS.IS_ACTIVE.isTrue)
+        .fetch()
+        .forEach { (plantingSiteId, plantingSeasonId) ->
+          endPlantingSeason(plantingSiteId, plantingSeasonId)
```
nickgraz, line 1059: Does this produce another database call in a loop, or is there some `dslContext` ORM functionality going on that is combining the queries?

nickgraz (approved):
A small question about `dslContext`, but looks great otherwise.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
sgrimm: We don't want to combine calls here because we want each transition to be in its own database transaction. Otherwise a failure would cause successful transitions from the same run to be rolled back.

nickgraz: Ah yes, I am now recalling that you have already given me this exact answer to this same question before. Sorry for the repetition!

PR #1521, opened 2023-11-28
SW-4515 Add API functionality to return species that are used in the org

karthikbtf:
- Enhanced SpeciesStore findAllSpecies to return in-use species controlled by an input arg
- Similar enhancement in SpeciesController list all species API
- Added tests
- I had originally separated out this functionality, seemed wasteful to have extra API/functions when we could extend the list all species API

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
@@ -62,25 +62,13 @@ class SpeciesStore(
           }
 
   private val usedInAccessions: Condition =
-      DSL.exists(
-          DSL.selectOne()
-              .from(ACCESSIONS)
-              .where(ACCESSIONS.SPECIES_ID.eq(SPECIES.ID))
-              .and(ACCESSIONS.facilities.ORGANIZATION_ID.eq(SPECIES.ORGANIZATION_ID)))
+      DSL.exists(DSL.selectOne().from(ACCESSIONS).where(ACCESSIONS.SPECIES_ID.eq(SPECIES.ID)))
 
   private val usedInBatches: Condition =
-      DSL.exists(
-          DSL.selectOne()
-              .from(BATCHES)
-              .where(BATCHES.SPECIES_ID.eq(SPECIES.ID))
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -58,6 +61,27 @@ class SpeciesStore(
                 .toSet()
           }
 
+  private val usedInAccessions: Condition =
+      DSL.exists(
+          DSL.selectOne()
+              .from(ACCESSIONS)
+              .where(ACCESSIONS.SPECIES_ID.eq(SPECIES.ID))
+              .and(ACCESSIONS.facilities.ORGANIZATION_ID.eq(SPECIES.ORGANIZATION_ID)))
```
sgrimm: You don't need the organization ID filter here or in the other conditions. Species are per-organization so if the species ID matches, the species must be in the right organization.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
sgrimm: Ah, did you do this because there's an index on facility ID but not on species ID on the accessions table?

Probably makes sense to add an index on species ID on the tables in question.

karthikbtf: I didn't account for the species ids being unique across orgs, didn't think about indices to be honest. I can add an index on species ID for those tables.

sgrimm: Some of them might already have indexes (I only looked at accessions).

karthikbtf: Yes, batches has an index on species_id. Accessions and plantings don't, yet.

karthikbtf: Should I be adding these to `R__Indexes.sql` vs a new migration file?

karthikbtf: Plantings has an index as well, just accessions.

sgrimm (approved):

PR #1522, opened 2023-11-28
SW-4480 Disallow deletion of in-use species using customer facing API

karthikbtf:
- introduced a deleteSpecies call in SpeciesService that checks if species is in-use
- updated controller to use service function, controller already had a 409 response annotation (previous behavior?)
- added tests
- cleaned up PlantingSiteInUseException to use ResourceInUseException


src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -151,6 +151,22 @@ class SpeciesStore(
         .fetch { ExistingSpeciesModel.of(it, speciesEcosystemTypesMultiset) }
   }
 
+  fun isInUse(speciesId: SpeciesId, organizationId: OrganizationId): Boolean {
+    requirePermissions { readOrganization(organizationId) }
```
sgrimm: This could be `readSpecies(speciesId)` and then it wouldn't need the org ID.

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
```diff
@@ -151,6 +151,22 @@ class SpeciesStore(
...
+
+    return dslContext.fetchExists(
+        DSL.selectOne()
+            .from(SPECIES)
+            .where(SPECIES.ID.eq(speciesId))
+            .and(SPECIES.ORGANIZATION_ID.eq(organizationId))
+            .and(SPECIES.DELETED_TIME.isNull)
```
sgrimm: Is filtering out deleted species the right thing here? A deleted species from before this change can be in use and in that case I'd think you'd want this method to return true.

charlie-tf (approved):
Looks good!

src/main/kotlin/com/terraformation/backend/tracking/PlantingSiteService.kt
```diff
@@ -19,14 +19,14 @@ class PlantingSiteService(
...
-   * Deletes a planting site if it has no plantings. throws [PlantingSiteInUseException] if planting
+   * Deletes a planting site if it has no plantings. throws [ResourceInUseException] if planting
    * site has plantings
    */
   fun deletePlantingSite(plantingSiteId: PlantingSiteId) {
     requirePermissions { deletePlantingSite(plantingSiteId) }
 
     if (plantingSiteStore.hasPlantings(plantingSiteId)) {
-      throw PlantingSiteInUseException(plantingSiteId)
+      throw ResourceInUseException("Planting site $plantingSiteId is currently in use.")
```
sgrimm: Why remove the more specific exception here?

src/main/kotlin/com/terraformation/backend/species/db/SpeciesStore.kt
karthikbtf: Makes sense, will fix this.

src/main/kotlin/com/terraformation/backend/species/api/SpeciesController.kt
```diff
@@ -143,7 +143,7 @@ class SpeciesController(
       organizationId: OrganizationId,
```
karthikbtf: Ok to remove this org id from before? Looks unused.

sgrimm: Yeah, that seems fine to remove.

sgrimm (approved):

PR #1527, opened 2023-11-29
Use compound primary key for subzone populations

sgrimm:
Commit c9ae259d replaced `plot_populations` with `planting_subzone_populations`.
There was an embedded type defined for `plot_populations` to represent its
compound primary key, but it wasn't replaced with one for the new table.

Update the embedded type and use it in the subzone populations search table;
the incorrect primary key on that table could have resulted in incorrect search
results, though luckily the web app isn't currently doing any searches that
would trigger the problem.


sgrimm:
* main
  * **PR #1527** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1527?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1527?utm_source=stack-comment).

nickgraz (approved):
Nice catch

PR #1528, opened 2023-11-29
SW-4482 Support notes for accession quantity edits in customer facing API

karthikbtf:
- update schema to support notes in accession quantity history (nullable)
- add notes to accession payloads
- store notes in history
- retrieve notes from history
- update tests

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
@@ -85,11 +85,15 @@ data class SeedQuantityPayload(
                 "the \"units\" field is \"Seeds\". This is always calculated on the server side " +
                 "and is ignored on input.",
         readOnly = true)
-    val grams: BigDecimal? = null
+    val grams: BigDecimal? = null,
+    @Schema(description = "Notes associated with this quantity when being changed by end user.")
+    val notes: String? = null,
 ) {
-  constructor(model: SeedQuantityModel) : this(model.quantity, model.units, model.grams)
+  constructor(
+      model: SeedQuantityModel
+  ) : this(model.quantity, model.units, model.grams, model.notes)
 
-  fun toModel() = SeedQuantityModel(quantity, units)
+  fun toModel() = SeedQuantityModel(quantity, units, notes)
 }
 
 fun SeedQuantityModel.toPayload() = SeedQuantityPayload(this)
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
@@ -85,15 +85,11 @@ data class SeedQuantityPayload(
                 "the \"units\" field is \"Seeds\". This is always calculated on the server side " +
                 "and is ignored on input.",
         readOnly = true)
-    val grams: BigDecimal? = null,
-    @Schema(description = "Notes associated with this quantity when being changed by end user.")
-    val notes: String? = null,
+    val grams: BigDecimal? = null
 ) {
-  constructor(
-      model: SeedQuantityModel
-  ) : this(model.quantity, model.units, model.grams, model.notes)
+  constructor(model: SeedQuantityModel) : this(model.quantity, model.units, model.grams)
 
-  fun toModel() = SeedQuantityModel(quantity, units, notes)
+  fun toModel() = SeedQuantityModel(quantity, units)
 }
 
 fun SeedQuantityModel.toPayload() = SeedQuantityPayload(this)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
```diff
@@ -85,11 +85,15 @@ data class SeedQuantityPayload(
...
         readOnly = true)
-    val grams: BigDecimal? = null
+    val grams: BigDecimal? = null,
+    @Schema(description = "Notes associated with this quantity when being changed by end user.")
+    val notes: String? = null,
 ) {
-  constructor(model: SeedQuantityModel) : this(model.quantity, model.units, model.grams)
+  constructor(
+      model: SeedQuantityModel
+  ) : this(model.quantity, model.units, model.grams, model.notes)
```
karthikbtf: This codepath to set notes is unused, will remove it.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
```diff
@@ -85,11 +85,15 @@ data class SeedQuantityPayload(
                 "the \"units\" field is \"Seeds\". This is always calculated on the server side " +
                 "and is ignored on input.",
         readOnly = true)
-    val grams: BigDecimal? = null
+    val grams: BigDecimal? = null,
+    @Schema(description = "Notes associated with this quantity when being changed by end user.")
+    val notes: String? = null,
```
karthikbtf: Added notes to the remaining quantity payload, even though it's only plumbed into updating the history. Notes are not stored as part of the accession payload (no client requirement so far).
Would it be better to extract this property out of the remaining quantity payload? That would require more explicit plumbing of notes elsewhere.

nickgraz (approved):
Looks good to me

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
sgrimm: Yeah, this feels like the wrong place to me since there are several quantities on an accession and notes only apply to the remaining quantity, not, e.g., the subset quantity.

Maybe `UpdateAccessionRequestPayloadV2` should have a field for remaining quantity notes.

src/main/kotlin/com/terraformation/backend/seedbank/model/SeedQuantity.kt
```diff
@@ -42,8 +42,11 @@ fun SeedQuantityUnits.fromGrams(quantity: BigDecimal): BigDecimal {
   return value.stripTrailingZeros()
 }
 
-data class SeedQuantityModel(val quantity: BigDecimal, val units: SeedQuantityUnits) :
-    Comparable<SeedQuantityModel> {
+data class SeedQuantityModel(
+    val quantity: BigDecimal,
+    val units: SeedQuantityUnits,
+    val notes: String? = null,
```
sgrimm: This also feels like the wrong place for notes; a quantity is more like a numeric value and it doesn't seem to me like there could be reasonable semantics for notes when, e.g., you subtract one quantity from another.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
```diff
@@ -374,6 +374,8 @@ data class UpdateAccessionRequestPayloadV2(
                 "existing value, it is considered a new observation, and the new value will " +
                 "override any previously-calculated remaining quantities.")
     val remainingQuantity: SeedQuantityPayload? = null,
+    @Schema(description = "Notes associated with remaining quantity updates if any.")
+    val remainingQuantityNotes: String? = null,
```
karthikbtf, line 386: This is definitely cleaner.

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
```diff
@@ -91,6 +91,7 @@ data class AccessionModel(
     val projectId: ProjectId? = null,
     val receivedDate: LocalDate? = null,
     val remaining: SeedQuantityModel? = null,
+    val remainingNotes: String? = null,
```
karthikbtf: Is this ok or would it be better to pass the notes into AccessionStore.update as a separate argument instead? Here the notes are only used as input during update, it won't be returned back in the model at any point.

sgrimm: I don't love either option but I think it's the inevitable consequence of our "PUT the whole accession to change anything, and we'll infer the desired changes by diffing with the current state" API design.

On balance I think I'd probably prefer to make it a separate argument, since this model class represents a snapshot of an accession's data and the notes are really an input to the update operation rather than part of the desired top-level state of the accession.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -353,7 +354,7 @@ class AccessionStore(
     throw RuntimeException("BUG! Inserting accession failed but error was not caught.")
   }
 
-  fun update(updated: AccessionModel) {
+  fun update(updated: AccessionModel, updateContext: AccessionUpdateContext? = null) {
```
sgrimm, line 357: Nice, this is definitely a more extensible way to do it than just passing notes in directly.

sgrimm (approved):

PR #1545, opened 2023-12-04
Add config option to keep invalid photos

sgrimm:
For debugging photo format issues in the mobile app, it'll be helpful to be
able to examine files that failed the format validation check. Add a config
option to skip deleting invalid files.

We'll enable the option in staging but keep the default behavior in prod.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
@@ -62,6 +62,12 @@ class TerrawareServerConfig(
      */
     val useTestClock: Boolean = false,
 
+    /**
+     * Retain uploaded files that fail file validation. This may be set in test environments so we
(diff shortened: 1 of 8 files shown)


sgrimm:
* main
  * **PR #1545** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1545?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1545?utm_source=stack-comment).

sgrimm:
> Out of curiosity - are images sometimes failing validation because it is the wrong format? Or issues with the bytes uploaded? Something else?

The problem we're looking at now seems to be that they're in the wrong format, but we won't know for sure until after this change is live and we can actually look at the files to see what's wrong.

PR #1546, opened 2023-12-05
SW-3936 Report settings API

sgrimm:
Add endpoints to read and modify the report settings for an organization and
its projects.


sgrimm:
* main
  * **PR #1546** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1546?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1546?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/report/api/ReportsSettingsController.kt
```diff
@@ -0,0 +1,101 @@
...
+@CustomerEndpoint
+@RequestMapping("/api/v1/reports/settings")
+@RestController
+class ReportsSettingsController(
+    private val reportStore: ReportStore,
+) {
+  @GetMapping
+  @Operation(summary = "Gets the report settings for an organization.")
+  fun getReportSettings(
+      @RequestParam organizationId: OrganizationId
```
karthikbtf, line 29: Not requesting a change, just curious why request param over a path param.

src/test/kotlin/com/terraformation/backend/report/db/ReportStoreTest.kt
```diff
@@ -581,4 +595,202 @@ class ReportStoreTest : DatabaseTest(), RunsAsUser {
...
+
+  @Nested
+  inner class UpdateSettings {
+    @Test
+    fun `supports enabling organization-level reports`() {
+      val settings =
+          ReportSettingsModel(
+              isConfigured = true,
+              organizationId = organizationId,
+              organizationEnabled = false,
```
karthikbtf: set to true?

src/test/kotlin/com/terraformation/backend/report/db/ReportStoreTest.kt
```diff
@@ -581,4 +595,202 @@ class ReportStoreTest : DatabaseTest(), RunsAsUser {
...
+              projects =
+                  listOf(
+                      ReportProjectSettingsModel(
+                          projectId = unconfiguredProjectId,
+                          isConfigured = false,
+                          isEnabled = true),
+                  )),
+          store.fetchSettingsByOrganization(organizationId),
+          "Projects list should be empty if no projects are enabled")
```
karthikbtf: The string description mentions empty projects list but the assert expects a non-empty projects list.

karthikbtf (approved):
Looks good to me, a few comments.

src/test/kotlin/com/terraformation/backend/report/db/ReportStoreTest.kt
sgrimm: Good catch. I was going back and forth on what the default return value should be. And in fact, this assertion isn't even useful at all because the default behavior is already covered by its own test.

src/main/kotlin/com/terraformation/backend/report/api/ReportsSettingsController.kt
sgrimm: > Not requesting a change, just curious why request param over a path param.

Definitely a subjective call, but my general approach is to use path parameters when the meaning of the parameter is declared by the parent. So for example, in `/api/v1/organizations/123` it is clear that 123 is an organization ID. But in `/api/v1/reports/settings/123` what does 123 refer to?

sgrimm: I did briefly consider using `/api/v1/organizations/123/reports/settings` instead and maybe that'd be better, but I decided it'd be better to keep this under `/api/v1/reports` so it's grouped with the other reports-related endpoints.

karthikbtf: Thanks, yes this makes sense as a better approach 👍

PR #1550, opened 2023-12-06
Add 'where' to project update query

nickgraz:
Fixes a bug which occurred during project update. Since there is no `where` it was attempting to insert a new row, which conflicted with the pre-existing row due to a constraint on the project name.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
@@ -210,7 +210,7 @@ class ProjectStoreTest : DatabaseTest(), RunsAsUser {
       store.update(projectId) {
         ExistingProjectModel(
             description = "New description",
-            id = projectId,
+            id = ProjectId(-1),
             name = "New name",
             organizationId = OrganizationId(-1),
         )


nickgraz:
* main
  * **PR #1550** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1550?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1550?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -84,6 +84,7 @@ class ProjectStore(
           .set(PROJECTS.MODIFIED_BY, currentUser().userId)
           .set(PROJECTS.MODIFIED_TIME, clock.instant())
           .set(PROJECTS.NAME, updated.name)
+          .where(PROJECTS.ID.eq(updated.id))
```
sgrimm: This should use `projectId`; otherwise the update function can return a model with a different project ID and we'll update the wrong one.

src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
```diff
@@ -210,7 +210,7 @@ class ProjectStoreTest : DatabaseTest(), RunsAsUser {
       store.update(projectId) {
         ExistingProjectModel(
             description = "New description",
-            id = ProjectId(-1),
+            id = projectId,
```
sgrimm: Can probably remove this change -- the intent of the original test was to make sure any changes to the project ID in the returned model would be ignored.

sgrimm (approved):

PR #1566, opened 2023-12-12
Add accessionNumber to the batch store fetch function

nickgraz:


nickgraz:
* main
  * **PR #1566** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1566?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1566?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
```diff
@@ -124,7 +126,10 @@ class BatchStore(
     val subLocationIds = fetchSubLocationIds(batchId)
     val totalWithdrawn = getTotalWithdrawn(batchId)
 
-    return ExistingBatchModel(batchesRow, subLocationIds, totalWithdrawn)
+    val accessionId = batchesRow.accessionId
+    val accessionNumber = if (accessionId != null) fetchAccessionNumber(accessionId) else null
```
sgrimm: This is totally fine as is, but FYI, another way to write it would be
```kotlin
val accessionNumber = batchesRow.accessionId?.let { fetchAccessionNumber(it) }
```

src/main/kotlin/com/terraformation/backend/nursery/model/BatchModel.kt
```diff
@@ -92,8 +93,10 @@ data class ExistingBatchModel(
       row: BatchesRow,
       subLocationIds: Set<SubLocationId> = emptySet(),
       totalWithdrawn: Int,
+      accessionNumber: String? = null
```
sgrimm: A trailing comma would be good here so this line doesn't need to change if we add another argument.

sgrimm (approved):

sgrimm:
We'll have to think about how this should work if/when we add fine-grained permissions. Like, if the accession and the batch are in different projects and someone only has permission to see the batch's project, should they be able to see the accession number? Not even remotely a topic for this PR, but a future product discussion we'll need to have.

src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
nickgraz: Ah yes, much better

nickgraz:
> We'll have to think about how this should work if/when we add fine-grained permissions. Like, if the accession and the batch are in different projects and someone only has permission to see the batch's project, should they be able to see the accession number? Not even remotely a topic for this PR, but a future product discussion we'll need to have.

Good point. Maybe we bring it up at the next weekly sync?

PR #1570, opened 2023-12-14
Add the project name to the AccessionModel

nickgraz:
- Pull the field into the model within the AccessionStore when using `fetchOneById`
- Add the field to the "GET accession" response body


nickgraz:
* `main`
  * **PR #1570** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1570?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1570?utm_source=stack-comment).

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreCreateTest.kt
```diff
@@ -220,14 +220,18 @@ internal class AccessionStoreCreateTest : AccessionStoreTest() {
       assertNotNull(prop.get(accession), "Field ${prop.name} is null in example object")
     }
 
-    val stored = store.create(accession.toModel(clock))
+    val accessionModel = accession.toModel(clock)
+    val stored = store.create(accessionModel)
```
sgrimm: Why split this?

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreCreateTest.kt
```diff
@@ -220,14 +220,18 @@ internal class AccessionStoreCreateTest : AccessionStoreTest() {
...
 
     accessionModelProperties
         .filter { (payloadFieldNames[it.name] ?: it.name) in propertyNames }
         .forEach { prop ->
           assertNotNull(prop.get(stored), "Field ${prop.name} is null in stored object")
         }
 
+    val storedAccession = stored.id?.let { store.fetchOneById(it) }
+    assertEquals(storedAccession?.projectName, "Project Foo")
```
sgrimm: This could use `stored` directly rather than fetching it again.

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreCreateTest.kt
nickgraz: Ah, good catch. I was originally using `accessionModel.id?.let` below, but that wasn't correct. This isn't being used anymore and the below code should be refactored too

nickgraz:
> The implementation seems straightforward enough, but I'm puzzled why we need this. Doesn't the client already have a list of projects it can use to look up the names, like it does with facilities? (We return the facility ID in the accession payload but we don't include the name.)

Hmm... I will check on that. Generally the data is requested within the view, as opposed to pre-loading all the data whether or not it is used. But I have seen some non-async things to get facilities and other things so I could be wrong.

nickgraz:
It looks like the facilities functions I mentioned are reading from the `Organization`, so that data is already a child of that specific entity and can be accessed without asking the server. We could potentially do the same thing if we return all the projects within the "GET /api/v1/organizations" call, which is what populates the organization data into the app context

nickgraz:
Never mind, I found in the main app component that we are preloading all projects. I can plumb it down to the accession view. This is somewhat of an anti pattern in my opinion, but the data is be loaded into redux so it is possible.

nickgraz:
Going to ignore this and pull the data from the projects that are loaded in at the app level

PR #1575, opened 2023-12-19
Fix jOOQ deprecation warning

sgrimm:
We use a custom class for PostGIS geometry columns, and we have a converter that
plugs into jOOQ so it uses our class when reading and writing those columns.

When we were constructing a custom jOOQ `Field` object, we were telling jOOQ to
look up the converter for our class in its global list of converters. That pattern
is being deprecated per https://github.com/jOOQ/jOOQ/issues/15286 because it can
cause undefined behavior. The existing code still works, but causes jOOQ to log
warning messages.

Update our custom field construction to explicitly specify the converter it needs.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/GeometryBinding.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/GeometryBinding.kt
@@ -11,6 +11,7 @@ import org.jooq.BindingSetSQLOutputContext
 import org.jooq.BindingSetStatementContext
 import org.jooq.Converter
 import org.jooq.impl.DSL
+import org.jooq.impl.SQLDataType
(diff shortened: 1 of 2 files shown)


sgrimm:
* `main`
  * **PR #1575** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1575?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1575?utm_source=stack-comment).

charlie-tf (approved):

PR #1576, opened 2023-12-20
SW-4586 Add BatchId to Facility Inventories Search Table

charlie-tf:
The table to show nursery batches by nursery searches the FacilityInventories
search table. In order to correctly route to the withdrawal flow the batchIds
for each returned search row is needed. These are added as a comma-
separated text aggregation field on the database view, and returned in the
search table.

### DIFF
--- a/src/main/resources/db/migration/0200/V232__FacilityInventoryBatchId.sql
+++ b/src/main/resources/db/migration/0200/V232__FacilityInventoryBatchId.sql
@@ -1,0 +1,14 @@
+DROP VIEW nursery.facility_inventories;
+
+CREATE OR REPLACE VIEW nursery.facility_inventories AS
+SELECT organization_id,
+       species_id,
+       facility_id,
+       id                                       AS batch_id,
+       SUM(ready_quantity)                      AS ready_quantity,
+       SUM(not_ready_quantity)                  AS not_ready_quantity,
+       SUM(germinating_quantity)                AS germinating_quantity,
+       SUM(ready_quantity + not_ready_quantity) AS total_quantity
+FROM nursery.batches
+GROUP BY organization_id, species_id, facility_id, batch_id
+HAVING SUM(ready_quantity + not_ready_quantity + germinating_quantity) > 0;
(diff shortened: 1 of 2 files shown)


sgrimm (changes requested):
This doesn't seem right to me. That view / search table isn't per-batch; it gives you the total inventories per species per facility across all batches. If you add batch ID to the group-by clause, then it effectively becomes a copy of the batches table with no aggregation.

Maybe you want a multi-value sublist that joins on the facility and species IDs.

charlie-tf:
> This doesn't seem right to me. That view / search table isn't per-batch; it gives you the total inventories per species per facility across all batches. If you add batch ID to the group-by clause, then it effectively becomes a copy of the batches table with no aggregation.
> 
> Maybe you want a multi-value sublist that joins on the facility and species IDs.

Kay, thanks. I'll look into that 👍 Could be that what I need is to hit a different search table on the FE.

src/main/resources/db/migration/0200/V232__FacilityInventoryBatchId.sql
```diff
@@ -4,11 +4,11 @@ CREATE OR REPLACE VIEW nursery.facility_inventories AS
 SELECT organization_id,
        species_id,
        facility_id,
-       id                                       AS batch_id,
+       STRING_AGG(CAST(id AS TEXT), ',')        AS batch_ids,
```
charlie-tf, line 7: @sgrimm This is more what I was actually looking for, and doesn't disaggregate the view like what I had before

sgrimm: This seems inconsistent with the way we handle lists of associated objects everywhere else in the search API. If you export this view as a CSV, you will get a comma-delimited value whereas other multi-value fields get exported with newline delimiters.

Probably clearest to explain what I mean by example; I will throw together a quick PR to demonstrate.

charlie-tf:
Oops, failed a test. Fixing..........

src/main/resources/db/migration/0200/V232__FacilityInventoryBatchId.sql
sgrimm: #1582 has the alternate approach to this.

charlie-tf: Thanks so much

PR #1577, opened 2023-12-20
Return user ID from DatabaseTest.insertUser()

sgrimm:
Allow tests to insert users more succinctly when they need the user IDs
afterwards. Update existing tests to get the user IDs from the insert function
rather than declaring them separately.

`insertUser()` now allows passing in a null user ID to let the database
allocate a new one, but existing tests don't use that feature yet in order to
avoid potential problems with ID collisions. The default is still to use the
user ID from the active security context.


sgrimm:
* `main`
  * **PR #1577** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1577?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1577?utm_source=stack-comment).

nickgraz (approved):
Nice refactor

PR #1585, opened 2023-12-21
Fix global role migration

sgrimm:
Migration 233 assumed that the global roles table was already populated, but that
won't be the case if it is deployed at the same time as migration 232 since the
roles are inserted in `R__TypeCodes.sql` after the numbered migrations.

### DIFF
--- a/src/main/resources/db/migration/0200/V233__DropSuperAdminType.sql
+++ b/src/main/resources/db/migration/0200/V233__DropSuperAdminType.sql
@@ -1,3 +1,7 @@
+INSERT INTO global_roles (id, name)
+VALUES (1, 'Super-Admin')
+ON CONFLICT DO NOTHING;
+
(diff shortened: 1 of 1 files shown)


sgrimm:
* `main`
  * **PR #1585** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1585?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1585?utm_source=stack-comment).

sgrimm:
Editing migrations after the fact is usually not allowed because the checksums won't match and Flyway will bomb out at start time, but in this case, the migration was never applied successfully anyway.

nickgraz (approved):
Makes sense

PR #1589, opened 2023-12-26
SW-4595 Fix incorrect apostrophes

sgrimm:
ASCII apostrophe characters are special in properties files and have to be doubled
to escape them. But some text editors "helpfully" detect pairs of apostrophes and
turn the second one into a closing single quote mark. Then Phrase, or other
software, notices the remaining apostrophe and doubles it again, such that we end
up with strings that have an ASCII apostophe and a single quote together.

We also had a couple sentences that should have used the word "its" but instead
used "it’s."

Remove the unnecessary single quote characters and add a test to catch the problem
in the future.

### DIFF
--- a/src/main/resources/i18n/Messages_en.properties
+++ b/src/main/resources/i18n/Messages_en.properties
@@ -100,14 +100,14 @@ notification.observation.monitoringPlotReplaced.email.subject=Organization {0} h
 # {0} is the name of an organization. {1} is the name of a planting site.
 notification.observation.notScheduled.email.body.1=Organization {0} has not scheduled an observation of planting site {1}
 # {0} is the name of a planting site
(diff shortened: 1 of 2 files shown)


sgrimm:
* `main`
  * **PR #1589** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1589?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1589?utm_source=stack-comment).

nickgraz (approved):
If I had a nickel for every time the "helpful" apostrophes caused issues...

PR #1632, opened 2024-01-16
SW-4649 Add batch ID to withdrawal when a nursery transfer is created

nickgraz:


nickgraz:
* `main`
  * **PR #1632** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1632?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1632?utm_source=stack-comment).

src/main/resources/db/migration/0200/V239__SeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE seedbank.withdrawals ADD COLUMN batch_id BIGINT NOT NULL REFERENCES nursery.batches;
```
sgrimm: Why `NOT NULL`? Won't this break for withdrawals that aren't nursery transfers, or am I misunderstanding?

sgrimm: Also, when you add a `NOT NULL` column, you need to specify a default value or the `ALTER` statement will fail if there are existing rows.

One way to do it is to add the column as nullable, backfill the values for existing rows, then alter the column to be non-nullable. That works if the table isn't very active (such that there won't be new rows inserted while the backfill is running). If the table is active, then it needs to happen in multiple phases which I can talk about in more detail if it's useful.

But my hunch is that this column ought to be nullable.

nickgraz: Hmm... I think probably a misunderstanding on my part. I thought all withdrawals would be associated to a batch. I will change it up

PR #1633, opened 2024-01-16
SW-4649 Backfill batch ID into seedbank withdrawals

nickgraz:
Migration to backfill the `batch_id` into the accession withdrawals table hinges on the `created_time` check, which I have truncated to seconds and given it a 2 second window since they are not created at the exact same time. 

@sgrimm let me know what you think, I think 2 seconds might be too high but there is a chance that the withdrawal was created at the end of the previous second and the batch at the beginning of the next second. Hopefully the other checks are sufficient to make the chance that an incorrect batch ID is applied impossible.

### DIFF
--- a/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
+++ b/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
@@ -1,6 +1,6 @@
 UPDATE seedbank.withdrawals withdrawals
 SET batch_id = (
-    SELECT batchHistory.batch_id
+    SELECT MIN(batchHistory.batch_id)
     FROM nursery.batch_quantity_history batchHistory
     INNER JOIN nursery.batches batches
         ON batches.id = batchHistory.batch_id
--- a/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
+++ b/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
@@ -1,0 +1,14 @@
+UPDATE seedbank.withdrawals
+SET
+    batch_id = joined.batch_id
+FROM (
+    SELECT withdrawals.id withdrawalId, batches.batch_id
+     FROM nursery.batch_quantity_history batches
+        INNER JOIN seedbank.withdrawals withdrawals
+            ON batches.germinating_quantity = withdrawals.withdrawn_quantity
+            AND batches.created_by = withdrawals.created_by
+            AND date_trunc('second', withdrawals.created_time)
+                BETWEEN (date_trunc('second', batches.created_time) - '1 second'::interval)
+                AND (date_trunc('second', batches.created_time) + '1 second'::interval)
+) as joined
+WHERE joined.withdrawalId = seedbank.withdrawals.id;
--- a/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
+++ b/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
@@ -11,7 +11,7 @@ SET batch_id = (
         AND batchHistory.created_by = withdrawals.created_by
         AND batchHistory.created_time
             BETWEEN withdrawals.created_time
-            AND (withdrawals.created_time::date + '1 second'::interval)
+            AND (withdrawals.created_time + '1 second'::interval)
 )
 WHERE withdrawals.batch_id IS NULL
     AND withdrawals.purpose_id = 9;
--- a/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
+++ b/src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
@@ -1,14 +1,17 @@
-UPDATE seedbank.withdrawals
-SET
-    batch_id = joined.batch_id
-FROM (
-    SELECT withdrawals.id withdrawalId, batches.batch_id
-     FROM nursery.batch_quantity_history batches
-        INNER JOIN seedbank.withdrawals withdrawals
-            ON batches.germinating_quantity = withdrawals.withdrawn_quantity
-            AND batches.created_by = withdrawals.created_by
-            AND date_trunc('second', withdrawals.created_time)
-                BETWEEN (date_trunc('second', batches.created_time) - '1 second'::interval)
-                AND (date_trunc('second', batches.created_time) + '1 second'::interval)
-) as joined
-WHERE joined.withdrawalId = seedbank.withdrawals.id;
+UPDATE seedbank.withdrawals withdrawals
+SET batch_id = (
+    SELECT batchHistory.batch_id
+    FROM nursery.batch_quantity_history batchHistory
+    INNER JOIN nursery.batches batches
+        ON batches.id = batchHistory.batch_id
+    INNER JOIN seedbank.accessions accessions
+        ON accessions.id = withdrawals.accession_id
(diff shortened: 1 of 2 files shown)


nickgraz:
* `main`
  * **PR #1633** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1634** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1634?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1633?utm_source=stack-comment).

src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,14 @@
+UPDATE seedbank.withdrawals
+SET
+    batch_id = joined.batch_id
+FROM (
+    SELECT withdrawals.id withdrawalId, batches.batch_id
+     FROM nursery.batch_quantity_history batches
+        INNER JOIN seedbank.withdrawals withdrawals
+            ON batches.germinating_quantity = withdrawals.withdrawn_quantity
```
sgrimm: Maybe filter on species ID and purpose too? If the purpose is anything other than a nursery transfer, it isn't a candidate.

nickgraz:
Now that I think about it, I think what I have here is actually a 3 second window if the `between` is inclusive of the entire second.

src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,14 @@
...
+            AND batches.created_by = withdrawals.created_by
+            AND date_trunc('second', withdrawals.created_time)
+                BETWEEN (date_trunc('second', batches.created_time) - '1 second'::interval)
+                AND (date_trunc('second', batches.created_time) + '1 second'::interval)
```
sgrimm: Doing this with a time window seems fine, but I don't think truncating to seconds is necessary; you can add/subtract 1 second to the full timestamp and get the same filtering behavior.

You could take advantage of the fact that the original code was always creating the withdrawal before the batch, so the batch can't have an earlier timestamp.

src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,14 @@
...
+) as joined
+WHERE joined.withdrawalId = seedbank.withdrawals.id;
```
sgrimm: Maybe add `AND seedbank.withdrawals.batch_id IS NULL` to avoid overwriting any values that were populated by #1632

src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,14 @@
+UPDATE seedbank.withdrawals
+SET
+    batch_id = joined.batch_id
+FROM (
+    SELECT withdrawals.id withdrawalId, batches.batch_id
+     FROM nursery.batch_quantity_history batches
+        INNER JOIN seedbank.withdrawals withdrawals
```
sgrimm: `UPDATE ... FROM` already implicitly joins with the target table.

I think this might be better structured as something like
```sql
UPDATE seedbank.withdrawals
SET batch_id = (
  SELECT batches.batch_id
  FROM nursery.batch_quantity_history batches
  WHERE batches.created_by = withdrawals.created_by
  AND [...]
)
WHERE batch_id IS NULL;
```

sgrimm (changes requested):

src/main/resources/db/migration/0200/V240__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,17 @@
...
+    INNER JOIN nursery.batches batches
+        ON batches.id = batchHistory.batch_id
+    INNER JOIN seedbank.accessions accessions
+        ON accessions.id = withdrawals.accession_id
+    WHERE batches.species_id = accessions.species_id
+        AND batchHistory.germinating_quantity = withdrawals.withdrawn_quantity
+        AND batchHistory.created_by = withdrawals.created_by
+        AND batchHistory.created_time
+            BETWEEN withdrawals.created_time
+            AND (withdrawals.created_time::date + '1 second'::interval)
```
sgrimm: Won't casting to a date cause this to evaluate to 1 second past midnight, which will almost always be earlier than the creation time? Does `withdrawals.created_time + '1 second'::interval` not work?

How have you been testing this? I'm surprised this `BETWEEN` clause works for anything but midnight creation times.

nickgraz: I've been testing with my local database, clearing batch IDs and running the query. I thought it was working as expected because postgres was responding with `UPDATED 41` (41 being the number of rows I expected to update). After looking a bit closer, you are right, it was not working.

sgrimm (changes requested):
This is looking better, but there's one minor gotcha: if there are two matching batch history entries in the same 1-second span, the subquery will return both IDs which will cause the update to fail.

I checked and there's a case of this in the production database: two nursery withdrawals (IDs 508 and 509) from the same accession with created times within a tenth of a second of each other. How that happened, I don't know; maybe our double-submit prevention was broken at one point?

I think it should work to make the subquery return the minimum matching batch ID (that is, wrap the ID field in `MIN()`).

sgrimm (approved):
I think this should do the trick!

sgrimm:
Oh, one thing that I should have noticed about the previous PR and that you can either add to this one or do separately: the new batch ID column could use an index. Otherwise the database has to do a full scan of the table to search for the withdrawal for a particular batch.

nickgraz:
I thought about that, but we aren't actually searching for that anywhere are we? Or are you saying that specifically to help out with this migration?

sgrimm:
I'm assuming one of the next steps in this series is going to be "get the withdrawal that created batch X" which would need to do a full scan without an index. But maybe I'm wrong about that being a thing we're going to need to do.

An index would also make it more efficient for the database to enforce the foreign key constraint if we ever delete batches.

nickgraz:
For now we just want to return the batch ID with the withdrawal entry, so we can link to the batch page. I need to add that in the next PR.

Do we delete batches? I can add the index now if you want, but I am not sure we will use it yet.

sgrimm:
Yes, [batches can be deleted](https://github.com/terraware/terraware-server/blob/be4ce75aff529944d4f1d0fe2fecade47fe29317/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt#L479). Which I guess is another thing I missed in my initial review.

I'm guessing the behavior we want if a batch is deleted is for the batch ID to be removed from the withdrawal but the withdrawal to otherwise remain untouched. If so, easiest is to make the foreign-key constraint do it automatically.

So something like this, which again, could be part of this PR or a separate one:
```sql
CREATE INDEX ON seedbank.withdrawals (batch_id);

ALTER TABLE seedbank.withdrawals
    DROP CONSTRAINT withdrawals_batch_id_fkey;

ALTER TABLE seedbank.withdrawals
    ADD FOREIGN KEY (batch_id)
    REFERENCES nursery.batches
    ON DELETE SET NULL;
```
The last line is the key one.

nickgraz:
Awesome, thanks @sgrimm

sgrimm (approved):
Thanks!

PR #1634, opened 2024-01-17
SW-4649 Add batch ID in the accession history response

nickgraz:


nickgraz:
* `main`
  * **PR #1633** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1634** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1634?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1634?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/WithdrawalStore.kt
```diff
@@ -109,6 +110,7 @@ class WithdrawalStore(
                 notes = record[NOTES],
                 type = type,
                 userId = record[WITHDRAWN_BY],
+                batchId = record[BATCH_ID],
```
sgrimm: Move this up so the argument list stays alphabetical.

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsController.kt
```diff
@@ -117,10 +118,11 @@ data class AccessionHistoryEntryPayload(
     @Schema(description = "User-entered notes about the event, if any.") //
     val notes: String?,
     val type: AccessionHistoryType,
+    val batchId: BatchId?,
```
sgrimm: Move this up so the field list stays alphabetical. It could also use a description.

src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
```diff
@@ -41,6 +42,11 @@ data class AccessionHistoryModel(
     val type: AccessionHistoryType,
     /** If the event was attributed to a Terraware user, that user's ID. */
     val userId: UserId?,
+    /**
+     * The associated batch ID, if applicable. Batches that are created during nursery transfers are
+     * associated to the history entry.
+     */
+    val batchId: BatchId? = null,
```
sgrimm: Move this up so the fields stay alphabetical.

sgrimm (changes requested):
A few minor ordering comments but the big thing is that `WithdrawalStoreTest` should be updated to test that the batch ID is being fetched properly.

src/main/kotlin/com/terraformation/backend/seedbank/model/AccessionHistoryModel.kt
nickgraz: Should this still be moved up, even though it is an optional parameter?

nickgraz: Never mind, I see we can use named arguments

docker/docker-compose.yml
```diff
@@ -16,7 +16,7 @@ version: "3.8"
 
 services:
   postgres:
-    image: "postgis/postgis:13-3.0"
+    image: "postgis/postgis:15-3.3"
```
sgrimm: This change seems unrelated, and also, it's set to 13 currently because we use 13 in production.

sgrimm (changes requested):

docker/docker-compose.yml
nickgraz: Ugh sorry, I added this because I wanted it compatible with my local DB that runs in `terraware-web`, which is using `15-3.3`. Regardless, I didn't mean to commit it, sorry about that!

sgrimm (approved):

PR #1636, opened 2024-01-18
Use local volume for postgres database within docker setup

nickgraz:


nickgraz:
* `main`
  * **PR #1636** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1636?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1636?utm_source=stack-comment).

nickgraz:
Apologies, the gt create command only added the file that was deleted, I didn't commit the other changes, other files incoming.

.gitignore
```diff
@@ -12,5 +12,3 @@ openapi.yaml
 .env
 .DS_Store
 .run
```
sgrimm, line 14: Does it make sense to leave `docker/volumes` in here so that people who ran the Docker setup before this change won't accidentally check in their database files?

docker/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
     ports:
       - "5432:5432"
     volumes:
-      - "./volumes/database:/var/lib/postgresql"
+      - "$HOME/docker/volumes/postgres:/var/lib/postgresql"
```
sgrimm: How about something like `"${POSTGRES_VOLUME_DIR:-$HOME/docker/volumes/postgres}:/var/lib/postgresql"` so it's possible for someone to use a different directory if they want?

sgrimm (approved):
Ah, this makes more sense. A couple optional suggestions inline but I think this is fine as-is too.

.gitignore
nickgraz: I thought I was the only one that used this 😅 ? Good call though

docker/docker-compose.yml
nickgraz: Good call

sgrimm (approved):

PR #1643, opened 2024-01-23
SW-4774 Add user's global roles to the API response for GET /me

nickgraz:
Return the global roles for the user in the GET /me API call response, the front end will use these roles to determine what things they can / can't see with regard to the accelerator admin area of the site.

### DIFF
--- a/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
+++ b/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
@@ -42,7 +42,7 @@ val ENUM_TABLES =
                 EnumTable("ecosystem_types", listOf(".*\\.ecosystem_type_id")),
                 EnumTable("facility_connection_states", listOf("facilities\\.connection_state_id")),
                 EnumTable("facility_types", listOf("facilities\\.type_id")),
-                EnumTable("global_roles", listOf(".*\\.global_role_id"), isLocalizable = false),
+                EnumTable("global_roles", listOf(".*\\.global_role_id"), isLocalizable = false, useIdAsJsonValue = true),
                 EnumTable("growth_forms", listOf("growth_forms\\.id", ".*\\.growth_form_id")),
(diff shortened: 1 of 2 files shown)


nickgraz:
* `main`
  * **PR #1643** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1643?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1643?utm_source=stack-comment).

sgrimm (approved):
This is fine. One thing you might consider is setting `useIdAsJsonValue = true` on `GlobalRole`'s entry in `Config.kt` so the values in the payloads don't have whitespace or hyphens. That is, you'll get back `SuperAdmin` instead of `Super-Admin`.

nickgraz:
Oh, that does seem nicer, thanks.

nickgraz:
That seemed to just give me the IDs `("Super-Admin" | "Accelerator Admin")[];` -> `("1" | "2")[];`

nickgraz:
Perhaps I need to implement a `GlobalRoleConverter` class for that functionality?

sgrimm:
Could be that I forgot to make that flag affect the OpenAPI schema generation. What values do you get back when you actually call the endpoint with that flag set?

nickgraz:
The API is also returning the ID values

nickgraz:
I committed just change just now so you can see exactly what I did

sgrimm:
Oh, ugh, sorry, I gave you the wrong flag. I meant `isLocalizable = false`, not `useIdAsJsonValue = true`.

sgrimm:
Actually, scratch that. I am thinking of the funder portal's enum setup. What you had originally is the right way to do it for terraware-server.

nickgraz:
Sounds good

PR #1647, opened 2024-01-27
SW-4791 Allow searching for null IDs in sublists

sgrimm:
`IdWrapperField` wasn't handling null search values like the other field classes
do; it needs to ad an `IS NULL` condition if the list of exact search values
includes a null. This was causing us to be unable to search for batches that
aren't associated with projects.


sgrimm:
* `main`
  * **PR #1647** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1647?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1647?utm_source=stack-comment).

nickgraz (approved):
Nice find, thanks for the quick turnaround!

PR #1653, opened 2024-01-29
Use correct Renovate keyword for grouping

sgrimm:
The Renovate config file uses the keyword `matchFileNames` for rules that match
based on the name of the package file. We were using `matchPaths` which caused
our Node and Python package updates to be generated individually rather than
grouped together as intended.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,12 +12,12 @@
         },
         {
             "description": "Group Node dependencies",
-            "matchPaths": ["package.json"],
+            "matchFileNames": ["package.json"],
             "groupName": "Node dependencies"
         },
         {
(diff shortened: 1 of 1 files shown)


sgrimm:
* `main`
  * **PR #1653** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1653?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1653?utm_source=stack-comment).

nickgraz (approved):

PR #1656, opened 2024-02-02
Fix facility creation race condition

sgrimm:
In end-to-end tests, we were sometimes sending two facility creation requests at
the exact same time from different browsers. This exposed a race in the facility
creation code where the same facility number would be used for both facilities,
causing one of the creations to fail with a unique key violation.

Fix it by serializing facility creation using a row lock on the organization.


sgrimm:
* `main`
  * **PR #1656** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1656?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1656?utm_source=stack-comment).

karthikbtf (approved):
Good find!

PR #1669, opened 2024-02-09
SW-4770 Fix totalSeedsStoredForProject stat calculation

nickgraz:
- Scope the subquery for seedbank project stats to the facility so that `totalSeedsStoredForProject` is scoped to that seedbank

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -936,13 +936,13 @@ class AccessionStore(
             .and(ACCESSIONS.STATE_ID.`in`(AccessionState.activeValues)))
   }
 
-  fun getSummaryStatistics(projectId: ProjectId): AccessionSummaryStatistics {
+  fun getSummaryStatistics(facilityId: FacilityId, projectId: ProjectId): AccessionSummaryStatistics {
     requirePermissions { readProject(projectId) }
 
     return getSummaryStatistics(
         DSL.select(ACCESSIONS.ID)
             .from(ACCESSIONS)
-            .where(ACCESSIONS.PROJECT_ID.eq(projectId))
+            .where(ACCESSIONS.PROJECT_ID.eq(projectId).and(ACCESSIONS.FACILITY_ID.eq(facilityId)))
             .and(ACCESSIONS.STATE_ID.`in`(AccessionState.activeValues)))
   }
 
--- a/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
@@ -936,7 +936,10 @@ class AccessionStore(
             .and(ACCESSIONS.STATE_ID.`in`(AccessionState.activeValues)))
   }
 
-  fun getSummaryStatistics(facilityId: FacilityId, projectId: ProjectId): AccessionSummaryStatistics {
+  fun getSummaryStatistics(
+      facilityId: FacilityId,
+      projectId: ProjectId
+  ): AccessionSummaryStatistics {
     requirePermissions { readProject(projectId) }
 
     return getSummaryStatistics(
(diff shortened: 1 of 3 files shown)


nickgraz:
* `main`
  * **PR #1669** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1669?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1669?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -936,13 +936,13 @@ class AccessionStore(
             .and(ACCESSIONS.STATE_ID.`in`(AccessionState.activeValues)))
   }
 
-  fun getSummaryStatistics(projectId: ProjectId): AccessionSummaryStatistics {
+  fun getSummaryStatistics(facilityId: FacilityId, projectId: ProjectId): AccessionSummaryStatistics {
```
nickgraz: Since this is currently only used in one place, I figured it was OK to modify it directly instead of making another function that takes both `facilityId` and `projectId`

src/test/kotlin/com/terraformation/backend/report/ReportServiceTest.kt
```diff
@@ -337,6 +337,14 @@ class ReportServiceTest : DatabaseTest(), RunsAsUser {
               remainingUnitsId = SeedQuantityUnits.Seeds,
           ),
       )
+      insertAccession(
+          AccessionsRow(
+              facilityId = nonProjectSeedBankId,
```
karthikbtf, line 342: should this be `otherProjectSeedbankId`? I see L#334 as well in the older code.. maybe this is fine.

karthikbtf (approved):
Good find/fix, one question.

src/test/kotlin/com/terraformation/backend/report/ReportServiceTest.kt
nickgraz: Since seed banks aren't directly tied to projects, I am pretty sure this is functionally the same

nickgraz: But I'll wait for @sgrimm to confirm since I'm not certain

karthikbtf: Test is certainly functionally correct.
Comment was mostly from parsing the const name, assuming it was earmarked for testing without projects. All good! Nothing to follow up.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -936,13 +936,16 @@ class AccessionStore(
...
+      facilityId: FacilityId,
+      projectId: ProjectId
+  ): AccessionSummaryStatistics {
     requirePermissions { readProject(projectId) }
 
     return getSummaryStatistics(
         DSL.select(ACCESSIONS.ID)
             .from(ACCESSIONS)
-            .where(ACCESSIONS.PROJECT_ID.eq(projectId))
+            .where(ACCESSIONS.PROJECT_ID.eq(projectId).and(ACCESSIONS.FACILITY_ID.eq(facilityId)))
```
sgrimm: It'd be more consistent to move the `.and` out one level so it's invoked on the return value of `where()` rather than being part of `where()`'s argument. Or to be more succinct: insert a right parenthesis before `.and` and remove one from the end of the line, which will cause this new `and` to line up with the existing one on the next line.

sgrimm (approved):
One small thing to tweak in the code, but this seems simple enough.

Can you maybe use a shorter commit title and move some of the detail into the rest of the commit message? Generally terraware-server doesn't use commit titles longer than 80 characters or so. That way commands like `git shortlog` won't need to do lots of line wrapping in a 100-column terminal.

PR #1676, opened 2024-02-12
SW-4802 Move findUnusedSquare to helper class

sgrimm:
For user-defined planting zone boundaries, we'll want to validate that the
boundaries are large enough to hold monitoring plots before inserting the zone
data into the database.

Ideally, we wanto to do that using the same "find a plot location" algorithm
that's used in real observations. That code is currently in `PlantingZoneModel`
which can only represent planting zones that exist in the database.

Move the zone search logic into a helper class `UnusedSquareFinder`.

This is mostly a straight copy-paste of the existing implementation, with the
following changes:

- The logic for excluding existing monitoring plots now lives in
  `findUnusedSquares` since it was only ever used via that function anyway.
  This requires changing the unit test for that functionality.
- The various functions that used to be defined inside `findUnusedSquare` are
  now just regular methods in the new class.
- Likewise, some of the local variables from `findUnusedSquare` are now
  instance variables in the new class.
- It's no longer possible to specify a grid interval thats different than the
  square size. We weren't ever doing that anyway.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/model/PlantingZoneModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/model/PlantingZoneModel.kt
@@ -3,16 +3,9 @@ package com.terraformation.backend.tracking.model
 import com.terraformation.backend.db.tracking.MonitoringPlotId
 import com.terraformation.backend.db.tracking.PlantingSubzoneId
 import com.terraformation.backend.db.tracking.PlantingZoneId
-import com.terraformation.backend.util.Turtle
-import com.terraformation.backend.util.createRectangle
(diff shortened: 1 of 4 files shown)


sgrimm:
* `main`
  * **PR #1676** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1676?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1676?utm_source=stack-comment).

nickgraz (approved):
Nice work, this is the kinda stuff that makes our open source offering even more desirable!

PR #1678, opened 2024-02-13
SW-4706 SW-4708 Add enum values for accelerator global roles

nickgraz:
- Add the new "TF Expert" and "Read Only" roles into the `global_roles` enum table

PR #1677, opened 2024-02-13
SW-4710 Add cohorts table, add cohorts_phase table

nickgraz:
- Add a new table to hold cohort data
- Add a new enum table to hold cohort phase enums
- Add relationship between cohort and cohort phase within cohort table
- Add the phases from the PRD as enums into the `cohort_phases` table

PR #1679, opened 2024-02-13
SW-4710 Add permissions for cohort management

nickgraz:
- Add "TF Expert" and "Read Only" global role capabilities to permissions checks 
- Add permission checks for CRUD on Cohort model according to this table:


nickgraz:
* `main`
  * **PR #1677** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1677?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1678** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1679** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1679?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
        * **PR #1680** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
          * **PR #1681** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1679?utm_source=stack-comment).

nickgraz:
I added those tests for the new roles (and the accelerator admin role). One question this brings up: should these new roles be superseding to the org roles? I structured these tests so the new global roles are only applied to the permissions from the new accelerator models (cohort and participant), but the individual user is still set up with org-level permissions, which dictate their access to "other stuff" (observations, planting sites, etc...). Is that correct?

For example, right now a "TF expert" user can't modify an org's planting site unless they also have the "admin" role for that org (set in `PermissionTest.kt:1331` for example).

Should a TF Expert be able to modify _any_ org's planting sites if it is associated to a project that is associated to a participant? It seems the way it is set up now is better, IE global roles only apply to new accelerator-specific models and a user must be granted access to individual orgs still. But I am not sure if that is what the PRD is representing. Also feel free to tell me this is a product question if you don't have an opinion or aren't sure.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -1269,6 +1269,170 @@ internal class PermissionTest : DatabaseTest() {
...
+        setTestClock = false,
+        updateAppVersions = false,
+        updateCohort = false,
+        updateDeviceTemplates = false,
+        updateGlobalRoles = false,
+        updateParticipant = false,
+    )
+  }
+
+  fun `ready only user has correct privileges`() {
```
sgrimm: Typo: "ready" should be "read"

sgrimm (approved):
One minor typo but this looks good otherwise.

On the topic of org permissions, yes, we will need these roles to grant some level of permission to read org data, but I'd rather do that as a followup to keep this PR more focused. It's not going to be a trivial "is one of these roles ==> access granted" situation because in most cases we'll only want them to have access to data in accelerator-tagged organizations and/or in projects that are actually associated with participants.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
nickgraz: 🤦‍♂️... thanks

PR #1680, opened 2024-02-13
SW-4710 Add cohort store and model

nickgraz:


nickgraz:
* `main`
  * **PR #1677** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1677?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1678** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1679** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
        * **PR #1680** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1680?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
          * **PR #1681** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1680?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
```diff
@@ -0,0 +1,121 @@
...
+  private fun fetch(condition: Condition?): List<ExistingCohortModel> {
+    val user = currentUser()
+
+    val participantIdsMultiset =
+        DSL.multiset(
+                DSL.select(PARTICIPANTS.ID)
+                    .from(PARTICIPANTS)
+                    .where(PARTICIPANTS.COHORT_ID.eq(COHORTS.ID))
+                    .orderBy(PARTICIPANTS.ID))
+            .convertFrom { result -> result.map { it[PARTICIPANTS.ID.asNonNullable()] } }
```
sgrimm: Granted the data sets are likely to be pretty small so maybe it doesn't matter much, but I wonder if it's right to always fetch the list of participants. Will we actually use the participant IDs most of the time when we're working with cohorts, or will we only need them in specific contexts?

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.terraformation.backend.db.default_schema.tables.pojos.CohortsRow
+import com.terraformation.backend.db.default_schema.tables.references.COHORTS
+import org.jooq.Field
+import org.jooq.Record
+
+data class CohortModel<ID : CohortId?>(
+    val id: ID,
+    val name: String,
+    val participantIds: List<ParticipantId>,
+    val phaseId: CohortPhase?
```
sgrimm: The naming convention I've been using in model classes is that enums don't have the `Id` suffix on their names. The suffix is there on the `Row` classes just because they're generated from the column names of the underlying tables, but in the context of the model, the phase is a value of type `CohortPhase`, not an ID.

Also, should this be `CohortPhase` instead of `CohortPhase?`?

sgrimm (commented):
Seems like the right direction so far.

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
nickgraz: Yes, it needs to be updated now that phase is required in the database.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
nickgraz: So far I haven't seen any view that shows cohorts without participants. It seems they are always combined. Come to think of it, the only place cohorts are visible, you also see all the participants, projects, and managers associated. Do you think that should be implemented in another method (like `fetchWithFullParticipantData` or similar) or should it be baked into this? I am not sure that this method would be used on its own yet unless we were also going to merge the "base cohort data" (this fetch) and the "full participant / project data" together on the frontend.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
```diff
@@ -0,0 +1,122 @@
...
+    cohortsDao.insert(row)
+
+    return row.toModel()
+  }
+
+  fun delete(cohortId: CohortId) {
+    requirePermissions { deleteCohort(cohortId) }
+
+    try {
+      dslContext.transaction { _ ->
```
sgrimm, line 67: This doesn't need to be in a transaction since it's only one SQL statement.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
```diff
@@ -0,0 +1,122 @@
...
+        val rowsDeleted = dslContext.deleteFrom(COHORTS).where(COHORTS.ID.eq(cohortId)).execute()
```
sgrimm, line 68: This is 100% fine, but FYI in case you didn't know it existed, the DAO classes have a `deleteById` method.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
sgrimm: I was thinking things like drop-down menus of cohorts when editing participants. But if it's always needed, then doing it here is fine. If we find it's unnecessary we can make this work like `PlantingSiteStore` where the caller can say whether to return the child objects.

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.terraformation.backend.db.default_schema.ParticipantId
+import com.terraformation.backend.db.default_schema.tables.pojos.CohortsRow
+import com.terraformation.backend.db.default_schema.tables.references.COHORTS
+import org.jooq.Field
+import org.jooq.Record
+
+data class CohortModel<ID : CohortId?>(
+    val id: ID,
+    val name: String,
+    val participantIds: List<ParticipantId>,
```
sgrimm: This could maybe be a `Set` instead of a `List` to codify the fact that duplicate IDs aren't allowed.

Also, I wonder if this should be a second type parameter like `ID` so that we can use the type system to forbid setting a list of participant IDs on a `NewCohortModel`. Though at that point it might be cleaner to make `ExistingCohortModel` and `NewCohortModel` entirely separate classes (and leave the ID and participants off the latter) since half the properties would be different between the two.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1712,6 +1717,7 @@ abstract class DatabaseTest {
       name: String = "Participant ${nextParticipantNumber++}",
       createdBy: Any = currentUser().userId,
       createdTime: Instant = Instant.EPOCH,
+      cohortId: Any? = null
```
sgrimm: Toss a trailing comma on here.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1721,17 +1727,42 @@ abstract class DatabaseTest {
             modifiedBy = createdBy.toIdWrapper { UserId(it) },
             modifiedTime = createdTime,
             name = name,
-        )
+            cohortId = cohortId?.toIdWrapper { CohortId(it) })
```
sgrimm: This could be above `createdBy` to keep the arguments in alphabetical order.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1721,17 +1727,42 @@ abstract class DatabaseTest {
...
+  ): CohortId {
+    val row =
+        CohortsRow(
+            createdBy = createdBy.toIdWrapper { UserId(it) },
+            createdTime = createdTime,
+            id = id?.toIdWrapper { CohortId(it) },
+            modifiedBy = createdBy.toIdWrapper { UserId(it) },
+            modifiedTime = createdTime,
+            name = name,
+            phaseId = phase)
```
sgrimm: Nit: Maybe throw a trailing comma on here just so future diffs that add more columns to this table don't have to change this line.

sgrimm (approved):
This looks good! I have a couple minor nitpicks that should be fixed before merge, and a couple things you might want to consider changing if you want.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
nickgraz: I copied this from `ParticipantStore`. It seems like this route was chosen because the DAO `deleteById` method does not return the rows deleted, so we wouldn't be able to throw `CohortNotFoundException` if a cohort that doesn't exist was requested to be deleted. I am going to leave it as is, but if you think I should use the DAO let me know.

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
nickgraz: Going with the second type parameter, let me know what you think, I could go either way re: leave as is or entirely separate classes.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
nickgraz: I think the transaction is needed, without it, I get `org.jooq.exception.DataAccessException` thrown in the catch block where we are checking for participants (`val hasParticipants = dslContext.fetchExists(PARTICIPANTS, PARTICIPANTS.COHORT_ID.eq(cohortId))`). Or do I need to reset the context somehow so it knows that the previous error is OK to ignore? 

```
Caused by: org.jooq.exception.DataAccessException: SQL [select exists (select 1 as "one" from "participants" where "participants"."cohort_id" = ?)]; ERROR: current transaction is aborted, commands ignored until end of transaction block
```

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
nickgraz: I ended up implementing the `CohortDepth` in the same way as `PlantingSiteDepth`.

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
nickgraz: Scratch that, seeing your comment in #1681, I decided to make it non-nullable and default to an empty set. This allowed me to remove the second type parameter.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
sgrimm: Ah, good point. Then this is definitely the right way to do it.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
sgrimm: You're right; I didn't think about the error case leaving the (implicit) open transaction in a dirty state.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
```diff
@@ -0,0 +1,136 @@
...
+
+  private fun participantIdsMultiset(): Field<Set<ParticipantId>> =
+      DSL.multiset(
+              DSL.select(PARTICIPANTS.ID)
+                  .from(PARTICIPANTS)
+                  .where(PARTICIPANTS.COHORT_ID.eq(COHORTS.ID))
+                  .orderBy(PARTICIPANTS.ID))
+          .convertFrom { result -> result.map { it[PARTICIPANTS.ID.asNonNullable()] }.toSet() }
+
+  private fun fetch(condition: Condition?, depth: CohortDepth?): List<ExistingCohortModel> {
```
sgrimm: Could `depth` be non-nullable here?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
nickgraz: Yes, it should be. I conflated nullability with default value in the API and plumbed it through as nullable. I switched it back here and in the API to be non-nullable.

nickgraz:
### Merge activity

* **Feb 15, 1:16 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1680).
* **Feb 15, 1:16 PM EST**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1680).

PR #1682, opened 2024-02-14
SW-4865 Store batch ID on new accession withdrawal

sgrimm:
When seeds are withdrawn to nurseries, we want to store the batch ID that's
created by the withdrawal. The ID was being passed to `WithdrawalStore` but it
wasn't actually saving the value to the database.

Fix the bug and rerun the migration that backfills batch IDs in any nursery
transfers that happened since the initial run.


sgrimm:
* `main`
  * **PR #1682** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1682?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1682?utm_source=stack-comment).

nickgraz (approved):
Nice find

src/main/resources/db/migration/0200/V243__BackfillSeedbankWithdrawalsBatchId.sql
```diff
@@ -0,0 +1,17 @@
+UPDATE seedbank.withdrawals withdrawals
```
nickgraz: I just merged in my cohort table PR, this will need to be updated to `V244`, sorry!

PR #1683, opened 2024-02-14

terraware-phrase:

### DIFF
--- a/src/main/resources/i18n/Enums_es.properties
+++ b/src/main/resources/i18n/Enums_es.properties
@@ -8,6 +8,10 @@ nursery.WithdrawalPurpose.Dead=Muerto
 nursery.WithdrawalPurpose.NurseryTransfer=Transferencia de viveros
 nursery.WithdrawalPurpose.Other=Otros
 nursery.WithdrawalPurpose.OutPlant=Planta exterior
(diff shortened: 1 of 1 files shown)


src/main/resources/i18n/Enums_es.properties
```diff
@@ -8,6 +8,10 @@ nursery.WithdrawalPurpose.Dead=Muerto
 nursery.WithdrawalPurpose.NurseryTransfer=Transferencia de viveros
 nursery.WithdrawalPurpose.Other=Otros
 nursery.WithdrawalPurpose.OutPlant=Planta exterior
+public.CohortPhase.Phase0DueDiligence=Fase 0\: Debida diligencia
```
nickgraz, line 11: @sgrimm these translations are from [these changes](https://github.com/terraware/terraware-server/pull/1677/files#diff-eeb8b11bdf1dd7ebe5c6280994439d59094103ce578b58f93029affebc04a44fR11-R14),  is the escaped `:` OK? Should I change the underlying English values to something other than `-`?

sgrimm: The escaped `:` is fine here. No surprise that different languages would prefer different punctuation.

PR #1685, opened 2024-02-15
Accept Polygon planting site boundaries

sgrimm:
Allow clients to supply planting site boundaries as either MultiPolygon or Polygon
geometries. This will save clients the trouble of converting single-polygon site
boundaries to MultiPolygon solely because it's required by the API.

This requires updating our OpenAPI schema customization code to handle properties
that are references to payload classes but have explicit annotations to list which
types are allowed.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
@@ -21,13 +21,16 @@ import com.terraformation.backend.tracking.model.PlantingSiteReportedPlantTotals
 import com.terraformation.backend.tracking.model.PlantingSubzoneModel
 import com.terraformation.backend.tracking.model.PlantingZoneModel
 import com.terraformation.backend.tracking.model.UpdatedPlantingSeasonModel
+import com.terraformation.backend.util.toMultiPolygon
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Schema
 import java.math.BigDecimal
 import java.time.Instant
 import java.time.LocalDate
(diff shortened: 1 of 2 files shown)


sgrimm:
* `main`
  * **PR #1685** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1685?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #1686** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1685?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
```diff
@@ -282,14 +287,21 @@ data class UpdatedPlantingSeasonPayload(
 }
 
 data class CreatePlantingSiteRequestPayload(
-    val boundary: MultiPolygon? = null,
+    @Schema(oneOf = [MultiPolygon::class, Polygon::class]) //
```
karthikbtf, line 290: trailing `//`?

karthikbtf (approved):
Nice!

src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
sgrimm: It's to prevent ktfmt from putting the `val boundary` on the same line as the annotation.

PR #1689, opened 2024-02-21
SW-4835 Create rest api for getting global roles and updating a user's global roles

nickgraz:
- Add GET and POST API for interacting with global roles. 
- Permission check for the POST API is within the `userStore` call

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
@@ -12,39 +12,37 @@ import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.db.default_schema.UserId
 import io.swagger.v3.oas.annotations.Operation
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
-import org.springframework.web.servlet.mvc.support.RedirectAttributes
 
 @AcceleratorEndpoint
-@RequestMapping("/api/v1/accelerator/acl")
+@RequestMapping("/api/v1/accelerator")
 @RestController
 class GlobalRolesController(
     private val userStore: UserStore,
 ) {
   @ApiResponse200
-  @GetMapping("/globalRoles")
-  @Operation(summary = "Gets the list of global roles and users with their roles.")
-  fun listGlobalRoles(): GlobalRoleListResponsePayload {
+  @GetMapping("/users")
+  @Operation(summary = "Gets the list of users that have global roles.")
+  fun listGlobalRoles(): GlobalRoleUsersListResponsePayload {
     requirePermissions { readGlobalRoles() }
 
-    return GlobalRoleListResponsePayload(
-        GlobalRole.entries.sortedBy { it.jsonValue }.map { GlobalRolePayload(it) },
-        userStore.fetchWithGlobalRoles().map { UserWithRolesPayload(it) })
+    return GlobalRoleUsersListResponsePayload(
+        userStore.fetchWithGlobalRoles().map { UserWithGlobalRolesPayload(it) })
   }
 
   @ApiResponse200
   @ApiResponse404
-  @PostMapping("/globalRoles")
-  @Operation(summary = "")
+  @PostMapping("/globalRoles/{userId}")
+  @Operation(summary = "Apply the supplied global roles to the user.")
   fun updateGlobalRoles(
-      @RequestParam userId: UserId,
-      @RequestParam roles: List<String>,
-      redirectAttributes: RedirectAttributes,
+      @PathVariable("userId") userId: UserId,
+      @RequestBody payload: UpdateGlobalRolesRequestPayload,
   ): SuccessResponsePayload {
-    val roleEnums = roles.map { GlobalRole.forJsonValue(it) }.toSet()
+    val roleEnums = payload.globalRoles.map { GlobalRole.forJsonValue(it) }.toSet()
 
     userStore.updateGlobalRoles(userId, roleEnums)
 
@@ -52,19 +50,7 @@ class GlobalRolesController(
   }
 }
 
-data class GlobalRolePayload(
-    val id: Int,
-    val name: String,
-) {
-  constructor(
-      globalRole: GlobalRole
-  ) : this(
-      id = globalRole.id,
-      name = globalRole.name,
-  )
-}
-
-data class UserWithRolesPayload(
+data class UserWithGlobalRolesPayload(
     val id: UserId,
     val email: String,
     val firstName: String?,
@@ -81,7 +67,9 @@ data class UserWithRolesPayload(
       lastName = user.lastName)
 }
 
-data class GlobalRoleListResponsePayload(
-    val globalRoles: List<GlobalRolePayload>,
-    val users: List<UserWithRolesPayload>
-) : SuccessResponsePayload
+data class GlobalRoleUsersListResponsePayload(val users: List<UserWithGlobalRolesPayload>) :
+    SuccessResponsePayload
+
+data class UpdateGlobalRolesRequestPayload(
+    val globalRoles: List<String>,
+)
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
@@ -1,0 +1,70 @@
+package com.terraformation.backend.accelerator.api
+
+import com.terraformation.backend.api.AcceleratorEndpoint
+import com.terraformation.backend.api.ApiResponse200
+import com.terraformation.backend.api.ApiResponse404
+import com.terraformation.backend.api.SimpleSuccessResponsePayload
+import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.model.IndividualUser
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.default_schema.GlobalRole
+import com.terraformation.backend.db.default_schema.UserId
+import io.swagger.v3.oas.annotations.Operation
+import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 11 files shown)


nickgraz:
* **#1689** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1689?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1690** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1690?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,70 @@
...
+import io.swagger.v3.oas.annotations.Operation
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.servlet.mvc.support.RedirectAttributes
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/acl")
```
karthikbtf: `/api/v1/accelerator/globalroles` and drop the `/globalRoles` from get/post mapping

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,70 @@
...
+        GlobalRole.entries.sortedBy { it.jsonValue }.map { GlobalRolePayload(it) },
+        userStore.fetchWithGlobalRoles())
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @PostMapping("/globalRoles")
+  @Operation(summary = "")
+  fun updateGlobalRoles(
+      @RequestParam userId: UserId,
```
karthikbtf: ideally a path param

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,70 @@
...
+  constructor(
+      globalRole: GlobalRole
+  ) : this(
+      id = globalRole.id,
+      name = globalRole.name,
+  )
+}
+
+data class GlobalRoleListResponsePayload(
+    val globalRoles: List<GlobalRolePayload>,
```
karthikbtf: Curious what the use-case is for this data. Is it to populate the dropdown when assigning global roles? If so, we could just hardcode that in the FE.

karthikbtf (approved):
Looks good, I have a few suggestions.

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
nickgraz: Yea, I am seeing the designs now for the first time and going to re-work this. Originally all I had was a few lines in the PRD. What it seems like we need is a "list users with global roles" endpoint (which might already exist somewhere) and an "update global roles for user" endpoint. It can definitely be hard coded in the UI since there are only 4 values and they likely won't change for a while.

nickgraz: @karthikbtf can you take a look now? I couldn't find a "list users" endpoint, so I made the "list" endpoint here for users only. Specifically, users with global roles (as opposed to all users). Also removed the global roles since we can just hardcode them on the frontend. The two APIs now are `GET /accelerator/users` and `POST /accelerator/users/{userId}`.

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,75 @@
...
+import io.swagger.v3.oas.annotations.Operation
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/globalRoles")
```
karthikbtf: ~~`globalroles`? api is case in-sensitive anyways but most for reading~~ please ignore, not relevant.

karthikbtf (approved):
Looks good to me 🔥

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
sgrimm: Slight nit with the placement here: this isn't really accelerator-specific functionality so IMO it belongs outside of `/api/v1/accelerator`. Maybe under `/api/v1/users` instead? Or maybe give it its own top-level name `/api/v1/globalRoles`? And the class can go in the `com.terraformation.backend.customer.api` package.

To Karthik's point about case sensitivity, just as an FYI, you can include multiple values in this annotation, e.g., `@RequestMapping("/api/v1/globalRoles", "/api/v1/globalroles")`. Not saying it's necessary in this case, just pointing out the feature for future reference.

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,75 @@
...
+@RestController
+class GlobalRolesController(
+    private val userStore: UserStore,
+) {
+  @ApiResponse200
+  @GetMapping("/users")
+  @Operation(summary = "Gets the list of users that have global roles.")
+  fun listGlobalRoles(): GlobalRoleUsersListResponsePayload {
```
sgrimm: Would it make sense to have this take an optional list of roles? ("No, keep it simple for now" is a totally fine answer!)

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,75 @@
...
+    requirePermissions { readGlobalRoles() }
+
+    return GlobalRoleUsersListResponsePayload(
+        userStore.fetchWithGlobalRoles().map { UserWithGlobalRolesPayload(it) })
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @PostMapping("/users/{userId}")
```
sgrimm: If this controller moves under `/api/v1/users`, then this path could turn into `/api/v1/users/{userId}/globalRoles` which, to me, feels more consistent with how we do paths elsewhere in the API.

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,75 @@
...
+      firstName = user.firstName,
+      globalRoles = user.globalRoles,
+      lastName = user.lastName)
+}
+
+data class GlobalRoleUsersListResponsePayload(val users: List<UserWithGlobalRolesPayload>) :
+    SuccessResponsePayload
+
+data class UpdateGlobalRolesRequestPayload(
+    val globalRoles: List<String>,
```
sgrimm: Did you try making this `Set<GlobalRole>` instead? I believe that should work, but maybe you tried it and it didn't. That way you don't need to do the string-to-enum conversion in application code, and as an added bonus, the OpenAPI schema should include a list of allowed values.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
nickgraz: There seems to be more functionality in the user management for the accelerator, and as a result it is getting its own PRD. I think that would be a great feature to add at that time, but for now I am going to keep it simple.

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
nickgraz: Since it is users with global roles, it seems like perhaps it should be `/api/v1/globalRoles/users` for the list and `/api/v1/users/{userId}/globalRoles`. Is it OK for them to have two different top level URL segments in the same file? I could also go with `/api/v1/users/globalRoles` for the list but, to me, semantically it means "the global roles that belong to users".

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
nickgraz: I changed it to `Set<GlobalRole>`

src/main/kotlin/com/terraformation/backend/accelerator/api/GlobalRolesController.kt
sgrimm: Having two separate top-level paths is fine, I think.

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
```diff
@@ -0,0 +1,73 @@
...
+@RequestMapping("/api/v1")
+@RestController
+class GlobalRolesController(
+    private val userStore: UserStore,
+) {
+  @ApiResponse200
+  @GetMapping("/globalRoles/users")
+  @Operation(summary = "Gets the list of users that have global roles.")
+  fun listGlobalRoles(): GlobalRoleUsersListResponsePayload {
+    requirePermissions { readGlobalRoles() }
```
sgrimm: Would it make sense for this permission check to live in `UserStore.fetchWithGlobalRoles()`?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
nickgraz: Yes I think that makes more sense, updated.

nickgraz:
### Merge activity

* **Mar 6, 5:46 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1689).
* **Mar 6, 5:46 PM EST**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1689) rebased this pull request as part of a merge.
* **Mar 6, 5:47 PM EST**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1689).

PR #1690, opened 2024-02-21
SW-4835 Add permissions for ability to set specific global roles

nickgraz:
- Add a permission check for `canUpdateSpecificGlobalRoles` which takes a set of roles. Since the permissions are based on the roles that are being updated, we check them individually against the current user's authorization.
- I left the `canUpdateGlobalRoles` permission as a "super admin fallback", in the event no roles are passed (for deleting all roles). It is also used in the admin UI for checking whether or not to show the global role UI. I think this will eventually get removed and we can decide if this permission is still needed then.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -337,9 +337,10 @@ class UserStore(
   }
 
   fun updateGlobalRoles(userId: UserId, roles: Set<GlobalRole>) {
-    requirePermissions { updateSpecificGlobalRoles(roles) }
     if (roles.isEmpty()) {
       requirePermissions { updateGlobalRoles() }
+    } else {
+      requirePermissions { updateSpecificGlobalRoles(roles) }
     }
 
     val user = fetchOneById(userId)
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -337,7 +337,10 @@ class UserStore(
   }
 
   fun updateGlobalRoles(userId: UserId, roles: Set<GlobalRole>) {
-    requirePermissions { updateGlobalRoles() }
+    requirePermissions { updateSpecificGlobalRoles(roles) }
+    if (roles.isEmpty()) {
+      requirePermissions { updateGlobalRoles() }
+    }
 
     val user = fetchOneById(userId)
     if (user !is IndividualUser || !user.email.endsWith("@terraformation.com", ignoreCase = true)) {
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -459,10 +459,10 @@ data class IndividualUser(
   override fun canUpdateSpecificGlobalRoles(globalRoles: Set<GlobalRole>): Boolean =
       globalRoles.all {
         when (it) {
-          GlobalRole.SuperAdmin -> isSuperAdmin()
           GlobalRole.AcceleratorAdmin -> isAcceleratorAdmin()
-          GlobalRole.TFExpert -> isAcceleratorAdmin()
           GlobalRole.ReadOnly -> isAcceleratorAdmin()
+          GlobalRole.SuperAdmin -> isSuperAdmin()
+          GlobalRole.TFExpert -> isAcceleratorAdmin()
         }
       }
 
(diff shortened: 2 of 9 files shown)


nickgraz:
* **#1689** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1689?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1690** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -451,6 +451,16 @@ data class IndividualUser(
 
   override fun canUpdateGlobalRoles(): Boolean = isSuperAdmin()
 
+  override fun canUpdateSpecificGlobalRoles(globalRoles: Set<GlobalRole>): Boolean =
+      globalRoles.all {
+        when (it) {
+          GlobalRole.SuperAdmin -> isSuperAdmin()
```
karthikbtf: nice

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -337,7 +337,11 @@ class UserStore(
   }
 
   fun updateGlobalRoles(userId: UserId, roles: Set<GlobalRole>) {
-    requirePermissions { updateGlobalRoles() }
+    if (roles.isEmpty()) {
```
karthikbtf, line 340: Can a TF-expert remove a 'super-admin' role from someone?

karthikbtf (commented):
Looking good, one question on nuking all roles.

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
nickgraz: No they should not be able to, if the roles are empty (remove all roles for user), the user must be able to `updateGlobalRoles` which is a blanket permission only given to "super admin".

karthikbtf: Got it.

karthikbtf (approved):

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -451,6 +451,16 @@ data class IndividualUser(
...
+          GlobalRole.AcceleratorAdmin -> isAcceleratorAdmin()
+          GlobalRole.TFExpert -> isAcceleratorAdmin()
+          GlobalRole.ReadOnly -> isAcceleratorAdmin()
```
sgrimm: Maybe alphabetize this just to make it clearer that these don't need to be in descending order of power?

src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
```diff
@@ -708,6 +709,31 @@ internal class PermissionRequirementsTest : RunsAsUser {
...
+    grant { user.canUpdateSpecificGlobalRoles(setOf(GlobalRole.SuperAdmin)) }
+    requirements.updateSpecificGlobalRoles(setOf(GlobalRole.SuperAdmin))
+
+    grant {
+      user.canUpdateSpecificGlobalRoles(
+          setOf(GlobalRole.AcceleratorAdmin, GlobalRole.TFExpert, GlobalRole.ReadOnly))
+    }
+    requirements.updateSpecificGlobalRoles(
+        setOf(GlobalRole.AcceleratorAdmin, GlobalRole.TFExpert, GlobalRole.ReadOnly))
+  }
```
sgrimm: I wonder if some of this exercising of the logic for matching against what's in the set of roles belongs in `PermissionTest` instead since it's really testing the logic in `IndividualUser` rather than the logic in `PermissionRequirements`.

nickgraz: I moved the tests over to `PermissionTest`

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -337,7 +337,11 @@ class UserStore(
...
+      requirePermissions { updateGlobalRoles() }
+    } else {
+      requirePermissions { updateSpecificGlobalRoles(roles) }
```
sgrimm, line 343: This check should probably look at the difference between the current and desired roles, rather than at the final list of roles.

If I want to add a new role to someone, it shouldn't fail because I don't have permission to add their existing roles. And likewise, if I want to remove a role, we should check for permission on the role I'm removing, not on the roles I'm leaving in place.

nickgraz:
### Merge activity

* **Mar 6, 5:46 PM EST**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1690).
* **Mar 6, 5:46 PM EST**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1690).

PR #1691, opened 2024-02-22
Add accelerator api to packages that need to be scanned for api docs

karthikbtf:
- added accelerator api package to yaml


sgrimm (approved):
Good catch!

PR #1700, opened 2024-02-28
SW-4934 Define read endpoints for deliverables

sgrimm:
To allow clients to start coding against the API, define the endpoints for listing
and fetching deliverable data.

Currently, these endpoints all return dummy responses.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
@@ -1,0 +1,206 @@
+package com.terraformation.backend.accelerator.api
+
+import com.terraformation.backend.api.AcceleratorEndpoint
+import com.terraformation.backend.api.ApiResponse200
+import com.terraformation.backend.api.ApiResponse404
+import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.db.accelerator.DeliverableCategory
+import com.terraformation.backend.db.accelerator.DeliverableId
+import com.terraformation.backend.db.accelerator.DeliverableType
+import com.terraformation.backend.db.accelerator.DocumentStore
+import com.terraformation.backend.db.accelerator.ParticipantId
+import com.terraformation.backend.db.accelerator.SubmissionDocumentId
+import com.terraformation.backend.db.accelerator.SubmissionStatus
+import com.terraformation.backend.db.default_schema.OrganizationId
+import com.terraformation.backend.db.default_schema.ProjectId
+import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.Parameter
+import io.swagger.v3.oas.annotations.headers.Header
+import io.swagger.v3.oas.annotations.media.Schema
+import io.swagger.v3.oas.annotations.responses.ApiResponse
+import java.net.URI
+import java.time.Instant
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/deliverables")
+@RestController
+class DeliverablesController {
+  @ApiResponse200
+  @GetMapping
+  @Operation(
+      summary = "Lists the deliverables for accelerator projects",
+      description =
+          "The list may optionally be filtered based on certain criteria as specified in the " +
+              "query string. If no filter parameters are supplied, lists all the deliverables " +
+              "in all the participants and projects that are visible to the user. For users with " +
+              "accelerator admin privileges, this will be the full list of all deliverables for " +
+              "all accelerator projects.")
+  fun listDeliverables(
+      @Parameter(
+          description =
+              "List deliverables for projects belonging to this organization. Ignored if " +
+                  "participantId or projectId is specified.")
+      organizationId: OrganizationId? = null,
+      @Parameter(
+          description =
+              "List deliverables for all projects in this participant. Ignored if projectId is " +
+                  "specified.")
+      participantId: ParticipantId? = null,
+      @Parameter(description = "List deliverables for this project only.")
+      projectId: ProjectId? = null,
+  ): ListDeliverablesResponsePayload {
+    return ListDeliverablesResponsePayload(
+        listOf(
+            ListDeliverablesElement(
+                category = DeliverableCategory.LegalEligibility,
+                descriptionHtml = "<p>A description</p>",
+                id = DeliverableId(1),
+                name = "Incorporation Documents",
+                numDocuments = 2,
+                organizationId = OrganizationId(1),
+                organizationName = "Test Org",
+                participantId = ParticipantId(2),
+                participantName = "Random Participant",
+                projectId = ProjectId(3),
+                projectName = "Omega Project",
+                status = SubmissionStatus.Rejected,
+                type = DeliverableType.Document,
+            ),
+            ListDeliverablesElement(
+                category = DeliverableCategory.FinancialViability,
+                descriptionHtml = "<p>All about money!</p>",
+                id = DeliverableId(1),
+                name = "Budget",
+                numDocuments = 0,
+                organizationId = OrganizationId(1),
+                organizationName = "Test Org",
+                participantId = ParticipantId(2),
+                participantName = "Random Participant",
+                projectId = ProjectId(3),
+                projectName = "Omega Project",
+                status = SubmissionStatus.NotSubmitted,
+                type = DeliverableType.Document,
+            ),
+        ))
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{deliverableId}")
+  @Operation(
+      summary = "Gets the details of a single deliverable and its submission documents, if any.")
+  fun getDeliverable(@PathVariable deliverableId: DeliverableId): GetDeliverableResponsePayload {
+    return GetDeliverableResponsePayload(
+        DeliverablePayload(
+            category = DeliverableCategory.LegalEligibility,
+            descriptionHtml = "<p>A description</p>",
+            documents =
+                listOf(
+                    SubmissionDocumentPayload(
+                        Instant.now(),
+                        "Project's articles of incorporation",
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#1700** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1700?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,207 @@
...
+    val descriptionHtml: String?,
+    val id: DeliverableId,
+    val name: String,
+    @Schema(
+        description =
+            "Number of documents submitted for this deliverable. Only valid for deliverables of " +
+                "type Document.")
+    val numDocuments: Int?,
+    val organizationId: OrganizationId,
+    val organizationName: String,
```
sgrimm, line 159: I included both IDs and names so this won't be dependent on the client having already fetched stuff like the list of projects and so that it'll work in the admin view where the IDs can point to organizations the user isn't a member of.

The downside is that the responses will be bigger; open to other suggestions if that seems like the wrong tradeoff.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,207 @@
...
+    val documents: List<SubmissionDocumentPayload>,
+    @Schema(
+        description =
+            "If the deliverable has been reviewed, the user-visible feedback from the review.")
+    val feedback: String?,
+    val id: DeliverableId,
+    @Schema(
+        description =
+            "Internal-only comment on the submission. Only present if the current user has accelerator admin privileges.")
+    val internalComment: String?,
```
sgrimm, line 189: Another approach would be for the admin view to use a separate API endpoint to fetch deliverables, but I thought keeping the API as small as possible was the better tradeoff. Please give feedback if that's the wrong call!

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,207 @@
...
+    val type: DeliverableType,
+)
+
+data class SubmissionDocumentPayload(
+    val createdTime: Instant,
+    val description: String,
+    val documentStore: DocumentStore,
+    val id: SubmissionDocumentId,
+    val name: String,
+)
```
sgrimm, line 174: This doesn't include a download URL because, at least on Dropbox, it requires an API call to generate each URL which would add latency and another possible point of failure. The link should point to the document download endpoint (defined in the controller class above) which the client can generate.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
karthikbtf: I like this approach, simplifies client side code and fewer requests. Client has a way to fetch project name from id if needed (Nick set something up), but we don't have something similar for organizations.
Guessing the response payload sizes should be ok, we aren't dealing with large number of deliverables (yet?).

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
karthikbtf: This works for the client, implementation so far has been assuming a single API endpoint to list deliverables.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,206 @@
...
+  @GetMapping
+  @Operation(
+      summary = "Lists the deliverables for accelerator projects",
+      description =
+          "The list may optionally be filtered based on certain criteria as specified in the " +
+              "query string. If no filter parameters are supplied, lists all the deliverables " +
+              "in all the participants and projects that are visible to the user. For users with " +
+              "accelerator admin privileges, this will be the full list of all deliverables for " +
+              "all accelerator projects.")
+  fun listDeliverables(
```
nickgraz, line 44: We will also need optional `category` and `status` query params. There is also a search bar in the FE table... originally built assuming usage of the Search API and doing a fuzzy search on `name` and `project_name` (I'll double check on `project_name`, both were added based on an assumption but the designs don't mention what the search bar should filter on). Do you think that should be added here or should we filter the results on the FE?

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,206 @@
...
+            documents =
+                listOf(
+                    SubmissionDocumentPayload(
+                        Instant.now(),
+                        "Project's articles of incorporation",
+                        DocumentStore.Google,
+                        SubmissionDocumentId(13974),
+                        "Incorporation Documents_2024-02-28_Omega_Projects articles of incorporation.doc",
+                    )),
+            feedback = "Is this a joke? This is a bunch of cat photos, not a legal document.",
```
nickgraz, line 112: 💯

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -0,0 +1,206 @@
...
+            participantName = "Random Participant",
+            projectId = ProjectId(3),
+            projectName = "Omega Project",
+            status = SubmissionStatus.Rejected,
+            type = DeliverableType.Document,
+        ))
+  }
+
+  @ApiResponse(
+      responseCode = "307",
```
sgrimm, line 128: @karthikbtf @nickgraz Would it be better for this to return HTTP 200 with a response body that includes the target URL, rather than returning a redirect?

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
nickgraz: I agree, this is nice.

nickgraz (commented):
I think this is useful enough to merge before doing the write endpoints. Gives us a good amount of work using these new types and refactoring some of the mock data out. I think the `category` and `status` query params should be added to the `listDeliverables` API though, the `search` param can maybe wait until I get confirmation of what the search in the table does from design.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
karthikbtf: Redirect sounds good. Ideally client would generate an anchor tag with href pointing to `/api/v1/accelerator/deliverables/{deliverableId}/documents/{documentId}` and target = '_blank'. With a 200 / body with target url, it would be extra steps on the client side to generate the href. Nick, thoughts?

nickgraz: I think this is good as-is, on the frontend we will just have this URL open in a new window so the browser will follow the redirect no problem.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
nickgraz: I just added a question in the PRD about this field, the UI seems to indicate that this is _multiple_ comments. I think this is good as-is for now, and hopefully for MVP we can just allow multiple comments within the individual field (which it looks like the underlying column will support).

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
sgrimm: My assumption was that for MVP the client would be grabbing the whole list with just the filters that were implied by the top nav and then doing the other filtering locally. I think that's how we do it in some other places in the UI. But it's easy to add those filter criteria here. Should I also add a way to change the sort order, or will we be doing client-side sorting?

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
nickgraz: You beat me to it!

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
karthikbtf: I believe this is a single text blurb.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
sgrimm: Does it need to support multiple values for category and/or status?

nickgraz: That works for me, we can do all the filtering client side for now. I think we have generally always tried to leverage the API to do the filtering but in this case, for MVP, I think it is fine to do it client side. The sorting we can absolutely do client side, even after MVP. I think we should only consider server side sorting if we ever decide to implement pagination server side as well, otherwise we would be making requests that have no need for the server to compute the result of.

nickgraz: > Does it need to support multiple values for category and/or status?

Yea the frontend uses multi select filters. But I think you've convinced me to just do it all in the frontend for now so you can disregard.

nickgraz (approved):

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
karthikbtf: We have been relying on server side sorting for i18n purposes. Especially if there are numbers involved, the display vs sort semantics. Guessing the number of documents won't be in the 1000s and we should be ok.
We do have some unrelated client side sorting in the codebase, this required having both 'number' and 'number(raw)' in the payload.
Maybe here if we have both localized and raw numbers, we should be ok client side.

sgrimm:
Merging this so it's available for code generation on the client side, but obviously we can tweak it as we go if it turns out to not be quite right.

PR #1704, opened 2024-02-29
Add templateUrl to the DeliverablePayload

nickgraz:


nickgraz:
* **#1704** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1704?utm_source=stack-comment).

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -134,6 +134,7 @@ class DeliverablesController {
             projectId = ProjectId(3),
             projectName = "Omega Project",
             status = SubmissionStatus.Rejected,
+            templateUrl = "https://docs.google.com/document/d/1NkNoho843CE-6NM8rmmHfdAT7HJS6vWhfzezcvkj3yM/edit?usp=sharing",
```
sgrimm: Remember that this is an open-source project (including the PR history) -- is that a document we want the world to know about?

sgrimm (approved):
Change is fine, but the example might not be.

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
sgrimm: If it's just a template, the answer might be "yes" -- just making sure.

nickgraz: It is a (near) empty Google Doc so I think it is alright.

nickgraz: I replaced it with a public URL. @karthikbtf you may want to delete that test doc, you could potentially get random share requests, sorry about that.

karthikbtf: Thanks for calling this out. It is an empty doc I created just for this purpose, was aware of the open source situation. I will delete it.

PR #1716, opened 2024-03-04
SW-4962 Fix document upload endpoint API to accept projectId as a path param

karthikbtf:
- the projectId was defined as a multipart/form-data request part but it's type was ProjectId, led to unsupported MediaType errors
- fix is to either set the schema for projectId to be "int64" and convert to ProjectId inside the function
- instead moved projectId to the path as a PathParam as this is our common practice


nickgraz (approved):
👍

PR #1717, opened 2024-03-04
SW-4655: Added exportable check to SearchField.

tommylau523:
Added an exportable variable the defaults to `true`. If a search field is converted to its raw form, it will set exportable to `false`.

If a raw search field is requested for a csv export, it will throw an Exception, that results in a 400 Bad Request error.

Due to the CSV text format, it may look like the following instead of a JSON:
```
"status","message"
"error","The search field is not exportable"
```

Tests not added yet. Will set up test suite for the entire Search API in a separate PR.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
@@ -45,16 +45,15 @@ private constructor(
   override val nullable: Boolean = original.nullable || targetPath.sublists.any { !it.isRequired }
 
   /**
-   *  True if the target field can be exported to a CSV file. Delegate to original field for
-   *  exportable check.
+   * True if the target field can be exported to a CSV file. Delegate to original field for
+   * exportable check.
    */
   override val exportable: Boolean = original.exportable
 
   override fun raw(): SearchField? {
     return if (localize) {
       val rawOriginal = original.raw() ?: return null
-      AliasField(
-          rawFieldName(), SearchFieldPath(targetPath.prefix, rawOriginal), rawOriginal)
+      AliasField(rawFieldName(), SearchFieldPath(targetPath.prefix, rawOriginal), rawOriginal)
     } else {
       null
     }
--- a/src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
@@ -44,6 +44,12 @@ private constructor(
    */
   override val nullable: Boolean = original.nullable || targetPath.sublists.any { !it.isRequired }
 
+  /**
+   *  True if the target field can be exported to a CSV file. Delegate to original field for
+   *  exportable check.
+   */
+  override val exportable: Boolean = original.exportable
+
   override fun raw(): SearchField? {
     return if (localize) {
       val rawOriginal = original.raw() ?: return null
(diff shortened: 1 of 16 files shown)


src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
```diff
@@ -24,7 +24,13 @@ private constructor(
...
      * this.
      */
-    val original: SearchField
+    val original: SearchField,
+
+    /**
+     * Whether this field can be exported to a CSV file. Only default (localized) fields may be
+     * requested for CSV export, since CSV files need to match what’s shown onscreen.
+     */
+    override val exportable: Boolean = true,
```
sgrimm: Maybe there's a case here that I'm overlooking, but for alias fields, I think it might make sense to always delegate to the original field here (that is, get rid of this argument entirely and let the `by original` take care of passing the call through). My thinking is that exportability is really a property of the rendering format, which is always going to be determined by the underlying field.

src/main/kotlin/com/terraformation/backend/search/SerachExceptions.kt
```diff
@@ -0,0 +1,8 @@
+package com.terraformation.backend.search
```
sgrimm: File naming nit: "Search" is misspelled. But luckily it doesn't matter: our convention for source files with multiple exception classes is to call them `Exceptions.kt`.

sgrimm: Also, maybe this exception belongs in the PR that adds the code that checks for exportability -- as far as I can tell, it's unused at this point.

sgrimm (changes requested):
Seems like the right approach. I have a couple specific bits of feedback (so requesting changes) but nothing major.

src/main/kotlin/com/terraformation/backend/search/SerachExceptions.kt
sgrimm: Forget my last comment. I overlooked the change that throws this exception.

src/main/kotlin/com/terraformation/backend/search/SerachExceptions.kt
```diff
@@ -0,0 +1,8 @@
...
+
+import jakarta.ws.rs.ClientErrorException
+import jakarta.ws.rs.core.Response
+
+/** An exception thrown when a non-exportable search field is requested for export to CSV. */
+class SearchFieldNotExportableException(message: String = "The search field is not exportable") :
```
sgrimm: Maybe include the field name in the message here?

src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
tommylau523: I had added an override to take the exportable value from `original.exportable`. Could you please look it over again? Appreciate the review and the feedback 😃

src/main/kotlin/com/terraformation/backend/search/field/AliasField.kt
```diff
@@ -44,6 +44,12 @@ private constructor(
    */
   override val nullable: Boolean = original.nullable || targetPath.sublists.any { !it.isRequired }
 
+  /**
+   * True if the target field can be exported to a CSV file. Delegate to original field for
+   * exportable check.
+   */
+  override val exportable: Boolean = original.exportable
```
sgrimm: The class already delegates to the original (the `by` on line 27) for all properties/functions that aren't explicitly overridden. So you could remove this entirely and it'd do the right thing.

sgrimm (approved):
Nice!

PR #1712, opened 2024-03-02
WIP SW-4952 Implement document upload endpoint

sgrimm:
DO NOT MERGE

This is an initial implementation of the document upload endpoint. Currently it
only writes files to Google Drive, not to Dropbox, and the Google Drive semantics
aren't properly abstracted yet. But the endpoint works and writes files to Google
Drive with the name collision behavior described in the requirements.

PR #1711, opened 2024-03-01
SW-4958 Add per-project document storage settings

sgrimm:
When documents are uploaded for accelerator projects, they need to be stored in
project-specific locations and have a project-specific element added to their
filenames. Add a table to store those settings, and allow them to be edited from
the list of a participant's projects in the admin UI.

PR #1719, opened 2024-03-06
Add fetchOneById method to SubmissionStore

nickgraz:
- Add a `fetchOneById` function to the `submissionStore` which fetches the submission and any associated submission documents.
- I had to add `insertSubmission`, `insertDeliverable`, and `insertModule` test database functions in order to test out the `fetchByOneId` function.
- Add `canReadSubmission` permission, available to all global roles and members of the submission's project's organization

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
@@ -16,7 +16,7 @@ data class SubmissionModel<ID : SubmissionId?>(
     val internalComment: String? = null,
     val projectId: ProjectId,
     val submissionDocumentIds: Set<SubmissionDocumentId>,
-    val submissionStatusId: SubmissionStatus,
+    val submissionStatus: SubmissionStatus,
 ) {
   companion object {
     fun of(
@@ -30,7 +30,7 @@ data class SubmissionModel<ID : SubmissionId?>(
           internalComment = record[SUBMISSIONS.INTERNAL_COMMENT]!!,
           projectId = record[SUBMISSIONS.PROJECT_ID]!!,
           submissionDocumentIds = submissionDocumentIds?.let { record[it] } ?: emptySet(),
-          submissionStatusId = record[SUBMISSIONS.SUBMISSION_STATUS_ID]!!,
+          submissionStatus = record[SUBMISSIONS.SUBMISSION_STATUS_ID]!!,
       )
     }
   }
--- a/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
@@ -26,8 +26,8 @@ data class SubmissionModel<ID : SubmissionId?>(
       return ExistingSubmissionModel(
           id = record[SUBMISSIONS.ID]!!,
           deliverableId = record[SUBMISSIONS.DELIVERABLE_ID]!!,
-          feedback = record[SUBMISSIONS.FEEDBACK]!!,
-          internalComment = record[SUBMISSIONS.INTERNAL_COMMENT]!!,
+          feedback = record[SUBMISSIONS.FEEDBACK],
+          internalComment = record[SUBMISSIONS.INTERNAL_COMMENT],
           projectId = record[SUBMISSIONS.PROJECT_ID]!!,
           submissionDocumentIds = submissionDocumentIds?.let { record[it] } ?: emptySet(),
           submissionStatus = record[SUBMISSIONS.SUBMISSION_STATUS_ID]!!,
--- a/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
@@ -1,0 +1,39 @@
+package com.terraformation.backend.accelerator.model
+
+import com.terraformation.backend.db.accelerator.DeliverableId
+import com.terraformation.backend.db.accelerator.SubmissionDocumentId
+import com.terraformation.backend.db.accelerator.SubmissionId
+import com.terraformation.backend.db.accelerator.SubmissionStatus
+import com.terraformation.backend.db.accelerator.tables.references.SUBMISSIONS
+import com.terraformation.backend.db.default_schema.ProjectId
+import org.jooq.Field
+import org.jooq.Record
+
+data class SubmissionModel<ID : SubmissionId?>(
+    val id: ID,
+    val deliverableId: DeliverableId,
+    val feedback: String? = null,
+    val internalComment: String? = null,
+    val projectId: ProjectId,
+    val submissionDocumentIds: Set<SubmissionDocumentId>,
+    val submissionStatusId: SubmissionStatus,
+) {
+  companion object {
+    fun of(
+        record: Record,
+        submissionDocumentIds: Field<Set<SubmissionDocumentId>>? = null
+    ): ExistingSubmissionModel {
+      return ExistingSubmissionModel(
+          id = record[SUBMISSIONS.ID]!!,
+          deliverableId = record[SUBMISSIONS.DELIVERABLE_ID]!!,
+          feedback = record[SUBMISSIONS.FEEDBACK]!!,
+          internalComment = record[SUBMISSIONS.INTERNAL_COMMENT]!!,
+          projectId = record[SUBMISSIONS.PROJECT_ID]!!,
+          submissionDocumentIds = submissionDocumentIds?.let { record[it] } ?: emptySet(),
+          submissionStatusId = record[SUBMISSIONS.SUBMISSION_STATUS_ID]!!,
+      )
+    }
+  }
+}
+
+typealias ExistingSubmissionModel = SubmissionModel<SubmissionId>
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
@@ -4,6 +4,8 @@ import com.terraformation.backend.customer.model.IndividualUser
 import com.terraformation.backend.db.AccessionNotFoundException
 import com.terraformation.backend.db.DeviceNotFoundException
 import com.terraformation.backend.db.FacilityNotFoundException
+import com.terraformation.backend.db.accelerator.SubmissionId
+import com.terraformation.backend.db.accelerator.tables.references.SUBMISSIONS
 import com.terraformation.backend.db.default_schema.AutomationId
 import com.terraformation.backend.db.default_schema.DeviceId
 import com.terraformation.backend.db.default_schema.DeviceManagerId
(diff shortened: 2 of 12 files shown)


nickgraz:
* **#1719** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1719?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1712** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1712?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1724](https://github.com/terraware/terraware-server/pull/1724) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1724?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1711** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1711?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/model/SubmissionModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.jooq.Record
+
+data class SubmissionModel<ID : SubmissionId?>(
+    val id: ID,
+    val deliverableId: DeliverableId,
+    val feedback: String? = null,
+    val internalComment: String? = null,
+    val projectId: ProjectId,
+    val submissionDocumentIds: Set<SubmissionDocumentId>,
+    val submissionStatusId: SubmissionStatus,
```
sgrimm: Maybe remove the `Id` suffix from this field name since from the model's point of view, this is just an enum value rather than a numeric ID.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -526,6 +543,45 @@ abstract class DatabaseTest {
     return row.id!!.also { inserted.projectIds.add(it) }
   }
 
+  private var nextDeliverableNumber: Int = 1
+
+  protected fun insertDeliverable(
+      createdBy: UserId = currentUser().userId,
+      createdTime: Instant = Instant.EPOCH,
+      deliverableCategoryId: DeliverableCategory? = DeliverableCategory.FinancialViability,
+      deliverableTypeId: DeliverableType? = DeliverableType.Document,
```
sgrimm: These arguments could be non-nullable since the columns are non-nullable in the database.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -526,6 +543,45 @@ abstract class DatabaseTest {
...
+      descriptionHtml: String? = "",
+      id: Any? = null,
+      moduleId: Any? = null,
```
sgrimm: Module ID is required in the database so maybe default this to `inserted.moduleId`?

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1787,62 +1899,98 @@ abstract class DatabaseTest {
...
     val speciesIds = mutableListOf<SpeciesId>()
     val subLocationIds = mutableListOf<SubLocationId>()
+    val submissionIds = mutableListOf<SubmissionId>()
     val uploadIds = mutableListOf<UploadId>()
     val userIds = mutableListOf<UserId>()
     val withdrawalIds = mutableListOf<WithdrawalId>()
 
     val accessionId
       get() = accessionIds.last()
+
```
sgrimm: Did ktfmt insert these blank lines? Interesting that it suddenly decided to do that when it hadn't previously.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -641,6 +697,35 @@ abstract class DatabaseTest {
     return actualSpeciesId.also { inserted.speciesIds.add(it) }
   }
 
+  fun insertSubmission(
+      createdBy: Any = currentUser().userId,
+      createdTime: Instant = Instant.EPOCH,
+      deliverableId: Any? = null,
```
sgrimm: Could default to `inserted.deliverableId`.

sgrimm (commented):
Not sure if you wanted feedback on this yet, but here's some feedback!

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
nickgraz: Yea it seems like it did. I recently updated the IntelliJ plugin to version 1.1.0.47. Would you like me to revert it?

sgrimm: Looks like the ktfmt version in our Gradle config is out of date and Renovate hasn't been catching it. Best bet is maybe to upgrade it in Gradle and run it on the code base all at once to avoid getting piecemeal diffs like this; I'll go do that now.

nickgraz:
@sgrimm I have some pending changes to address all this feedback but I can't push until the `upload` branch is updated with `main`. Is it OK if I do that and fix the merge conflicts? Or will it mess up your graphite locally if I do that?

sgrimm:
I was going to upgrade the branch right after sending out the ktfmt upgrade PR, so hang on a few minutes and I'll take care of it.

nickgraz:
@sgrimm I have some pending changes to address all this feedback but I can't push until the `upload` branch is updated with `main`. Is it OK if I do that and fix the merge conflicts? Or will it mess up your graphite locally if I do that?

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -1889,6 +1905,7 @@ internal class PermissionTest : DatabaseTest() {
         readCohort: Boolean = false,
         readInternalTags: Boolean = false,
         readParticipant: Boolean = false,
+        readSubmission: Boolean = false,
```
nickgraz: @sgrimm is there a better place for this? The permission for this is technically allowed for global roles _and_ org members. So it is both globally scoped and organizationally scoped.

sgrimm: It might be cleaner to define a new `expect()` method that takes a list of submission IDs. The user with the global role can read all of them, but the org user can only read the ones from their org.

src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
```diff
@@ -137,13 +139,21 @@ class ParentStore(private val dslContext: DSLContext) {
   fun getOrganizationId(speciesId: SpeciesId): OrganizationId? =
       fetchFieldById(speciesId, SPECIES.ID, SPECIES.ORGANIZATION_ID)
 
+  fun getOrganizationId(submissionId: SubmissionId): OrganizationId? {
+    val projectId = getProjectId(submissionId)
+    return projectId?.let { getOrganizationId(it) }
+  }
```
sgrimm: Rather than two separate fetches, this could take advantage of jOOQ's implicit join feature. We do this in other places in this file such as `getFacilityConnectionState()`. It'd look something like
```kotlin
fetchFieldById(submissionId, SUBMISSIONS.ID, SUBMISSIONS.projects.ORGANIZATION_ID)
```

src/test/kotlin/com/terraformation/backend/accelerator/db/SubmissionStoreTest.kt
```diff
@@ -0,0 +1,62 @@
...
+      val submissionId = insertSubmission()
+
+      assertEquals(
+          ExistingSubmissionModel(
+              id = submissionId,
+              feedback = "",
+              internalComment = "",
+              projectId = projectId,
+              deliverableId = deliverableId,
+              submissionDocumentIds = emptySet(),
```
sgrimm: Maybe worth inserting a few document IDs to make sure the multiset works.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -481,9 +503,7 @@ internal class PermissionTest : DatabaseTest() {
         updateProject = true,
     )
 
-    permissions.expect(
-        deleteSelf = true,
-    )
+    permissions.expect(deleteSelf = true, readSubmission = true)
```
sgrimm: Trailing comma would be good here just to keep future diffs cleaner (but as noted below, this should probably be a separate `expect()` call anyway.)

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -526,6 +543,48 @@ abstract class DatabaseTest {
...
 
+  private var nextDeliverableNumber: Int = 1
+
+  protected fun insertDeliverable(
+      createdBy: UserId = currentUser().userId,
+      createdTime: Instant = Instant.EPOCH,
+      deliverableCategoryId: DeliverableCategory = DeliverableCategory.FinancialViability,
+      deliverableTypeId: DeliverableType = DeliverableType.Document,
+      descriptionHtml: String? = "",
+      id: DeliverableId = nextDeliverableNumber.toIdWrapper { DeliverableId(it) },
```
sgrimm: A couple things here. The usual pattern is to make the ID nullable and let the database allocate an ID if the caller doesn't pass one in. And also, the ID arguments are usually of type `Any?` such that a caller that wants specific IDs can just use integer arguments like `id = 1` for brevity.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -526,6 +543,48 @@ abstract class DatabaseTest {
...
+      isSensitive: Boolean = false,
+      isRequired: Boolean = false,
+      moduleId: ModuleId = inserted.moduleId,
+      name: String = "Deliverable $nextDeliverableNumber",
+      position: Int = 0,
+      subtitle: String? = "",
```
sgrimm: This could default to null. (And we should probably remove subtitle from the data model anyway since it got removed from the requirements, but that's not a topic for this PR.)

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -526,6 +543,48 @@ abstract class DatabaseTest {
...
+      createdTime: Instant = Instant.EPOCH,
+      deliverableCategoryId: DeliverableCategory = DeliverableCategory.FinancialViability,
+      deliverableTypeId: DeliverableType = DeliverableType.Document,
+      descriptionHtml: String? = "",
+      id: DeliverableId = nextDeliverableNumber.toIdWrapper { DeliverableId(it) },
+      isSensitive: Boolean = false,
+      isRequired: Boolean = false,
+      moduleId: ModuleId = inserted.moduleId,
+      name: String = "Deliverable $nextDeliverableNumber",
+      position: Int = 0,
```
sgrimm: Positions should start with 1 rather than 0 since `DeliverablesImporter` flips the signs on the positions to detect when there are existing deliverables missing from the imported file.

Using `nextDeliverableNumber` as the default would probably make sense here to avoid position collisions.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -641,6 +700,39 @@ abstract class DatabaseTest {
     return actualSpeciesId.also { inserted.speciesIds.add(it) }
   }
 
+  private var nextSubmissionNumber = 1
+
+  fun insertSubmission(
+      createdBy: UserId = currentUser().userId,
+      createdTime: Instant = Instant.EPOCH,
+      deliverableId: DeliverableId = inserted.deliverableId,
+      feedback: String? = "",
```
sgrimm: This and `internalComment` should default to null, I'd say.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
nickgraz: Will remove in a subsequent PR

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
nickgraz: I broke it out into its own `expect`.

sgrimm (approved):
This looks good! Maybe remove the to-do item from the PR description before merging.

PR #1726, opened 2024-03-08
SW-4992 Created tables for project voting, and added ProjectVoteStore.

tommylau523:
SQL Tables added:
vote_options (ENUM for YES/NO/CONDITIONAL)
project_voters 
project_vote_records

Permissions added:
readProjectVotes
updateProjectVotes

Store operations (and permissions needed):
fetchesAllVotesByProject (readProjectVotes)
upsert (updateProjectVotes)
delete (updateProjectVotes)


src/main/resources/db/migration/0200/V249__ProjectVoting.sql
```diff
@@ -0,0 +1,27 @@
...
+    user_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    PRIMARY KEY (user_id, project_id)
+);
+
+-- Add voting records
+CREATE TABLE accelerator.project_vote_records (
+    user_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    vote_option_id BIGINT NOT NULL REFERENCES accelerator.vote_options,
```
sgrimm: A thought: if this were nullable, we wouldn't need a separate `project_voters` table. We'd populate this table with the roster of voters we would have inserted in that table.

src/main/resources/db/migration/0200/V249__ProjectVoting.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE accelerator.project_voters (
+    user_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    PRIMARY KEY (user_id, project_id)
+);
+
+-- Add voting records
+CREATE TABLE accelerator.project_vote_records (
+    user_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
```
sgrimm: I know you're discussing this with the product folks, but I think this is going to need a phase ID too and the primary key will be (user, project, phase). I believe there's a vote at each phase transition, so a project will be voted on more than once over its lifetime.

src/main/resources/db/migration/0200/V249__ProjectVoting.sql
```diff
@@ -0,0 +1,27 @@
+-- Add vote options enum table (YES, CONDITIONAL, NO)
+CREATE TABLE accelerator.vote_options (
+    id INTEGER PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    value TEXT NOT NULL,
```
sgrimm: Call this column `name` -- our code generation configuration expects enum tables to have `id` and `name` columns.

sgrimm (commented):
This seems like the right general direction to me so far.

src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
```diff
@@ -31,3 +33,7 @@ class SubmissionDocumentNotFoundException(id: SubmissionDocumentId) :
 
 class SubmissionNotFoundException(id: SubmissionId) :
     EntityNotFoundException("Submission $id not found")
+
+class ProjectVoteNotFoundException(projectId: ProjectId, phase: CohortPhase?, userId: UserId?) :
```
sgrimm: Nit: This could move up a couple positions to keep the classes in alphabetical order (or at least to keep the "project" exceptions together).

sgrimm: Also, maybe add `null` default values for the nullable arguments?

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,101 @@
...
+        ProjectVotesRow(
+            userId = model.userId,
+            projectId = model.projectId,
+            phaseId = model.phase,
+            voteOptionId = model.voteOption,
+            conditionalInfo = model.conditionalInfo,
+            createdBy = currentUserId,
+            createdTime = now,
+            modifiedBy = currentUserId,
+            modifiedTime = now)
```
sgrimm: Nit: Maybe put a trailing comma here just to keep future diffs tidier.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    projectVotesDao.insert(row)
+    return row.toModel()
+  }
+
+  fun delete(projectId: ProjectId, phase: CohortPhase? = null, userId: UserId? = null) {
+    dslContext.transaction { _ ->
+      val conditions = mutableListOf(PROJECT_VOTES.PROJECT_ID.eq(projectId))
+      if (phase != null) conditions.addLast(PROJECT_VOTES.PHASE_ID.eq(phase))
+      if (userId != null) conditions.addLast(PROJECT_VOTES.USER_ID.eq(userId))
```
sgrimm: The way you have it here is totally fine, just pointing out a possible alternate idiom:
```kotlin
val conditions = listOfNotNull(
    PROJECT_VOTES.PROJECT_ID.eq(projectId),
    phase?.let { PROJECT_VOTES.PHASE_ID.eq(it) },
    userId?.let { PROJECT_VOTES.USER_ID.eq(it) },
)
```

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,101 @@
...
+              .set(MODIFIED_BY, currentUser().userId)
+              .set(MODIFIED_TIME, clock.instant())
+              .set(VOTE_OPTION_ID, voteOption)
+              .set(CONDITIONAL_INFO, conditionalInfo)
+              .where(
+                  listOf(
+                      PROJECT_VOTES.PROJECT_ID.eq(projectId),
+                      PROJECT_VOTES.PHASE_ID.eq(phase),
+                      PROJECT_VOTES.USER_ID.eq(userId),
+                  ))
```
sgrimm: This works, but just to keep it more syntactically similar to the raw SQL, we usually structure static lists of conditions like
```kotlin
.where(PROJECT_VOTES.PROJECT_ID.eq(project_id))
.and(PROJECT_VOTES.PHASE_ID.eq(phase))
.and(PROJECT_VOTES.USER_ID.eq(userId))
```

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  fun updateVoteOption(
+      projectId: ProjectId,
+      phase: CohortPhase,
+      userId: UserId,
+      voteOption: VoteOption? = null,
+      conditionalInfo: String? = null
+  ) {
```
sgrimm: Would it make sense to combine this and `create()` so the caller just cares about setting the vote and doesn't need to know whether or not there was a previous vote?

src/main/kotlin/com/terraformation/backend/accelerator/model/VoteModel.kt
```diff
@@ -0,0 +1,56 @@
...
+          conditionalInfo = record[PROJECT_VOTES.CONDITIONAL_INFO],
+      )
+    }
+
+    fun create(
+        projectId: ProjectId,
+        phase: CohortPhase,
+        userId: UserId,
+        voteOption: VoteOption?,
+        conditionalInfo: String?
```
sgrimm: What's the advantage of calling this vs. directly using the constructor?

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -474,4 +474,6 @@ COMMENT ON TABLE accelerator.submission_statuses IS '(Enum) Statuses of submissi
 
 COMMENT ON TABLE accelerator.submissions IS 'Information about the current states of the information supplied by specific projects in response to deliverables.';
 
+COMMENT ON TABLE accelerator.vote_options IS '(Enum) Available vote options.';
+COMMENT ON TABLE accelerator.project_votes IS 'Vote options (nullable) selected by voters.';
```
sgrimm: This comment should go up above line 468 to keep the tables in alphabetical order within the `accelerator` schema.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
```diff
@@ -0,0 +1,592 @@
...
+
+  @Nested
+  inner class Create {
+    @Test
+    fun `creates blank vote for self`() {
+      val projectId = insertProject()
+
+      clock.instant = Instant.EPOCH.plusSeconds(500)
+      val phase: CohortPhase = CohortPhase.Phase0DueDiligence
+      store.create(VoteModel.create(projectId, phase, user.userId, null, null))
```
sgrimm: Maybe include `null` defaults for the optional arguments here? (In the constructor, of course, if you decide to replace the `create` method with constructor calls.)

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
```diff
@@ -0,0 +1,592 @@
...
+
+      assertEquals(
+          listOf(
+              ProjectVotesRow(
+                  projectId = projectId,
+                  phaseId = phase,
+                  userId = user.userId,
+                  voteOptionId = null,
+                  conditionalInfo = null,
```
sgrimm: Totally okay to include these for explicitness in the test, but FYI, in case you didn't notice, the `Row` classes all have null defaults for all their properties.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
```diff
@@ -0,0 +1,592 @@
...
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+class VoteStoreTest : DatabaseTest(), RunsAsUser {
+  override val user = mockUser()
+
+  private val clock = TestClock()
+  private val store: VoteStore by lazy { VoteStore(clock, dslContext, projectVotesDao) }
+
+  data class VoteKey(val userId: UserId, val projectId: ProjectId, val phase: CohortPhase)
```
sgrimm, line 27: Might be worth adding this to `EMBEDDABLES` at the bottom of `Config.kt` so jOOQ will generate a class for it and include it as a field in the various table-related classes. (The fields for the individual columns will still be there too.)

sgrimm (changes requested):
This is shaping up nicely! Some comments inline, and of course it needs permissions as you note in the PR description, but it mostly looks solid. I especially appreciate the comprehensive test coverage.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
tommylau523: I may be missing something but the `VoteKey` here is only used as a key to the MutableMap for holding testing variables (to avoid naming by variables `vote100project1phase1` etc. ).

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
tommylau523: Done!

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
tommylau523: Combined into an `upsert` function

src/main/kotlin/com/terraformation/backend/accelerator/model/VoteModel.kt
tommylau523: Removed.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
sgrimm: Yeah, fair point that this is only for the test's internal use. I definitely feel the pain of figuring out naming for test fixtures like this!

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
tommylau523: I saw in in your PR! I applied your suggestion. It does look a lot cleaner (and maybe a slight performance boost over mutableList I am guessing)

tommylau523:
Hi @sgrimm, ready for review again! Apologies for the long PR. I didn't expect the PR to add up like this. I have a good idea how I can break it up into smaller pieces for future PRs.

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -474,4 +476,5 @@ COMMENT ON TABLE accelerator.submission_statuses IS '(Enum) Statuses of submissi
 
 COMMENT ON TABLE accelerator.submissions IS 'Information about the current states of the information supplied by specific projects in response to deliverables.';
 
+COMMENT ON TABLE accelerator.vote_options IS '(Enum) Available vote options.';
```
sgrimm, line 479: Nit: Blank line after this so the trailing comment is more visually distinct.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1946,6 +1950,32 @@ abstract class DatabaseTest {
         )
 
     cohortModulesDao.insert(row)
+
+  fun insertVote(
+      projectId: ProjectId,
```
sgrimm: A couple things on these insert helper functions:

* ID arguments that tests are likely to pass in can be of type `Any`. Then when they're assigned to properties in the `Row` class, the `.toIdWrapper` extension method turns them into the correct type. The idea is that in tests, we can do, e.g., `insertVote(projectId = 1)` for conciseness.
* For dependencies like this, default to the most recently inserted row of the type in question, which gets tracked   in the `Inserted` class, an instance of which is created as the `inserted` property. Insert functions that create new IDs record the IDs there. The idea is that callers can omit the ID argument entirely if they're setting up some test data.

So this would be `projectId: Any = inserted.projectId` and then in the row creation, `projectId = projectId.toIdWrapper { ProjectId(it) }`.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1946,6 +1950,32 @@ abstract class DatabaseTest {
...
+      phase: CohortPhase = CohortPhase.Phase0DueDiligence,
+      user: UserId = currentUser().userId,
+      voteOption: VoteOption? = null,
+      conditionalInfo: String? = null,
+      createdBy: UserId = currentUser().userId,
+      createdTime: Instant = Instant.EPOCH,
+  ): ProjectVotesRow {
+    val row =
+        ProjectVotesRow(
+            createdBy = createdBy.toIdWrapper { UserId(it) },
```
sgrimm: Since the user IDs here are already of type `UserId`, there's no need to call `toIdWrapper`; this can just be `createdBy = createdBy`. (And it's fine for `createdBy` to not be of type `Any` here since tests are unlikely to want to explicitly pass in specific user ID values.)

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,81 @@
...
+          .from(PROJECT_VOTES)
+          .where(PROJECT_ID.eq(projectId))
+          .orderBy(listOf(PROJECT_ID, PHASE_ID, USER_ID))
+          .fetch { VoteModel.of(it) }
+    }
+  }
+
+  fun delete(projectId: ProjectId, phase: CohortPhase? = null, userId: UserId? = null) {
+    requirePermissions { updateProjectVotes(projectId) }
+    dslContext.transaction { _ ->
```
sgrimm: Does this need to be in a transaction?

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,81 @@
...
+    private val projectVotesDao: ProjectVotesDao,
+) {
+  fun fetchAllVotesByProject(projectId: ProjectId): List<VoteModel> {
+    requirePermissions { updateProjectVotes(projectId) }
+    return with(PROJECT_VOTES) {
+      dslContext
+          .select(PROJECT_VOTES.asterisk())
+          .from(PROJECT_VOTES)
+          .where(PROJECT_ID.eq(projectId))
+          .orderBy(listOf(PROJECT_ID, PHASE_ID, USER_ID))
```
sgrimm: Don't need the `listOf` here -- `orderBy` takes a vararg list of fields.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.jooq.DSLContext
+
+@Named
+class VoteStore(
+    private val clock: InstantSource,
+    private val dslContext: DSLContext,
+    private val projectVotesDao: ProjectVotesDao,
+) {
+  fun fetchAllVotesByProject(projectId: ProjectId): List<VoteModel> {
+    requirePermissions { updateProjectVotes(projectId) }
```
sgrimm: I believe users with the "Read Only" global role are supposed to be able to view votes, so this might need a separate `readProjectVotes` permission.

sgrimm (changes requested):
A few comments but this is getting close.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
tommylau523: No! Good catch!

tommylau523:
Appreciate the reviews @sgrimm! Ready again. I added a JOIN FROM USERS to retrieve user info since we will likely need to return those in the API. Please take a look!

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -0,0 +1,82 @@
...
+      dslContext
+          .select(PROJECT_VOTES.asterisk(), USERS.EMAIL, USERS.FIRST_NAME, USERS.LAST_NAME)
+          .from(PROJECT_VOTES)
+          .join(USERS)
+          .on(USERS.ID.eq(PROJECT_VOTES.USER_ID))
+          .where(PROJECT_ID.eq(projectId))
+          .orderBy(PROJECT_ID, PHASE_ID, USER_ID)
+          .fetch { VoteModel.of(it) }
+    }
+  }
```
sgrimm: I wonder if we should do this via the search API instead of a dedicated "fetch the fields we anticipate the UI will need" implementation so callers can specify what exact information they want about users and projects as well as what filters they want. We do use the "pull in data from other tables based on what the web app will need" approach in some places already, and I'm definitely not saying this is wrong, just that it may be a sign we want more flexible query capability.

Where I'm coming from here is thinking about how `VoteModel` is going to be used for internal business logic like calculating the vote totals. We won't care about email addresses in that kind of context but we'll still have to fetch them to populate the field in the model class.

sgrimm (approved):
This is good to merge. I do have one thought about the read side of this but we can think about it after this is in.

sgrimm:
It'd be nice to update the PR description before merging, both to remove the TODO items and to reflect the fact that this no longer just creates tables but also does other things.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
tommylau523: I am happy to revert this last commit for now. Instead I will stash this commit and we can think about how the front-end can get relevant information for rendering voter names or emails etc.

PR #1732, opened 2024-03-08
WIP - Create SubmissionService tests for uploading submission files

nickgraz:
- Implement `ProjectDocumentSettingsStore` with `fetchOneById`
  - And `ExistingProjectDocumentSettingsModel`
- Implement `SubmissionDocumentStore` with `fetchOneById`
  - And `ExistingSubmissionDocumentModel`
- Add a new method for `SubmissionService.receiveDocument` that takes the `deliverable`, `projectDocumentSettings`, and `receiver` as additional arguments so we can easily mock the `receiver` in tests.
- Add tests for `SubmissionService.receiveDocument`
  - One that verifies the upload works
  - One that verifies name collisions are handled correctly

TODO
[ ] `canReadProjectDocumentSetting` permission
[ ] `canReadDeliverable` permission (right now using `canReadAllDeliverables`)


nickgraz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1732?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1732** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1742** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1739** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1739?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1738** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1738?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1737** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1737?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/accelerator/SubmissionServiceTest.kt
```diff
@@ -0,0 +1,221 @@
...
+      assertEquals(
+          listOf(submissionDocument1, submissionDocument2),
+          listOf(
+              ExistingSubmissionDocumentModel(
+                  id = submissionDocumentId1,
+                  documentStore = DocumentStore.Google,
+                  description = mockDocumentDescription,
+                  location = mockStoredFileLocation,
+                  name = expectedFileName1,
+                  submissionId = SubmissionId(2)),
```
nickgraz: This part is a bit iffy. I tried making the `setUp` function only run once with `BeforeAll` and `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`, but ran into user auth issues. I think ideally the "create project, insert project document settings" only happens once for this entire test class. But I was not able to make it work, so the second test has _the next_ submission ID.

Perhaps there's something I am missing? Or maybe I need to delete the entries after each test? Or maybe I am overthinking it?

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionService.kt
```diff
@@ -50,24 +52,17 @@ class SubmissionService(
   private val illegalFilenameCharacters = Regex("[<>:\"/\\\\|?*]")
 
   fun receiveDocument(
+      contentType: String,
+      deliverable: ExistingDeliverableModel,
+      description: String,
```
sgrimm, line 57: Alphabetical order, yay!

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -134,12 +134,12 @@ class DeliverablesController(
...
-            ProjectId(projectId),
-            deliverableId,
-            description,
-            file.contentType ?: MediaType.APPLICATION_OCTET_STREAM_VALUE)
+            inputStream = file.inputStream,
+            originalName = file.getFilename(),
+            projectId = ProjectId(projectId),
+            deliverableId = deliverableId,
+            description = description,
+            contentType = file.contentType ?: MediaType.APPLICATION_OCTET_STREAM_VALUE)
```
sgrimm: May as well alphabetize here too to match the actual argument order on the method.

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
```diff
@@ -14,15 +14,15 @@ import com.terraformation.backend.db.accelerator.tables.references.SUBMISSION_DO
...
+import java.time.Clock
 import org.jooq.Condition
 import org.jooq.DSLContext
 import org.jooq.impl.DSL
 import org.springframework.context.ApplicationEventPublisher
 
 @Named
 class SubmissionStore(
-    private val clock: InstantSource,
+    private val clock: Clock,
```
sgrimm: I've been using `InstantSource` rather than `Clock` just to keep the interfaces minimal. `Clock` inherits from `InstantSource` and adds a bunch of time-zone-related functionality we usually don't need.

src/test/kotlin/com/terraformation/backend/accelerator/SubmissionServiceTest.kt
```diff
@@ -0,0 +1,220 @@
...
+        googleDriveReceiver.upload(
+            inputStream = inputStream,
+            fileName = expectedFileName,
+            contentType = contentType,
+        )
+      }
+
+      val submissionDocument = submissionDocumentStore.fetchOneById(submissionDocumentId)
+      assertEquals(
+          submissionDocument,
```
sgrimm: Expected value comes first.

src/test/kotlin/com/terraformation/backend/accelerator/SubmissionServiceTest.kt
sgrimm: Here's how to reset a table's ID sequence for each test method in a `DatabaseTest` subclass:
```kotlin
override val tablesToResetSequences = listOf(SUBMISSIONS)
```

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -552,6 +553,25 @@ abstract class DatabaseTest {
     return row.id!!.also { inserted.projectIds.add(it) }
   }
 
+  protected fun insertProjectDocumentSettings(
+      dropboxFolderPath: String,
+      fileNaming: String,
+      googleFolderUrl: URI,
```
sgrimm: These could have defaults.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -552,6 +553,25 @@ abstract class DatabaseTest {
...
+      projectId: Any? = this.inserted.projectId,
```
sgrimm: No need for the `this.` prefix here.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -1972,6 +1992,7 @@ abstract class DatabaseTest {
     val plantingSubzoneIds = mutableListOf<PlantingSubzoneId>()
     val plantingZoneIds = mutableListOf<PlantingZoneId>()
     val projectIds = mutableListOf<ProjectId>()
+    val projectDocumentSettingsProjectIds = mutableListOf<ProjectId>()
```
sgrimm: Maybe this isn't needed? The IDs here ought to always be in `projectIds`.

sgrimm (changes requested):
I have some specific comments inline but high level, I think this is probably a bigger change than is required just for testing SubmissionService.

In particular, I'm not seeing why adding the new store classes for project document settings and for submissions is needed for the test. We may want those classes for other reasons, of course; it just seems like having them here will force us to also do things like add permissions for them that aren't really related to testing SubmissionService.

We might be able to get away with simpler ways to stub out the Dropbox/Google interactions without needing to expose another `receiveDocument` method that's only needed for tests. Some alternate approaches:

* Mock `GoogleDriveWriter` and `DropboxWriter` and let `SubmissionService` create real `Reciever` objects using those mock writers. Probably the least invasive approach but the mock configuration will be a little fiddly.
* Move the "which receiver to create for a `DocumentStore` value" logic into a separate `@Named` class that `SubmissionService` depends on. Then in the test, replace that with a stub that returns a mock `Receiver`.
* Variant on the previous approach is to move the `Receiver` creation logic to a protected method in `SubmissionService` and then, in the test, subclass `SubmissionService` to override that method.

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionService.kt
nickgraz: Finally got one without you asking!

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
nickgraz: Good call

nickgraz:
> I have some specific comments inline but high level, I think this is probably a bigger change than is required just for testing SubmissionService.
> 
> In particular, I'm not seeing why adding the new store classes for project document settings and for submissions is needed for the test. We may want those classes for other reasons, of course; it just seems like having them here will force us to also do things like add permissions for them that aren't really related to testing SubmissionService.
> 
> We might be able to get away with simpler ways to stub out the Dropbox/Google interactions without needing to expose another `receiveDocument` method that's only needed for tests. Some alternate approaches:
> 
> * Mock `GoogleDriveWriter` and `DropboxWriter` and let `SubmissionService` create real `Reciever` objects using those mock writers. Probably the least invasive approach but the mock configuration will be a little fiddly.
> * Move the "which receiver to create for a `DocumentStore` value" logic into a separate `@Named` class that `SubmissionService` depends on. Then in the test, replace that with a stub that returns a mock `Receiver`.
> * Variant on the previous approach is to move the `Receiver` creation logic to a protected method in `SubmissionService` and then, in the test, subclass `SubmissionService` to override that method.

@sgrimm For some reason this comment is not showing up in the Graphite UI, sorry for addressing all the other inline comments and ignoring this one. Unless you think I should finish this first, I am switching over to the scoring and voting work and will address this later in the week.

sgrimm:
Switching over to the voting and scoring work sounds good to me. We can revisit this later.

sgrimm:
@nickgraz I think we should probably close this -- seems unlikely we'll come back to it any time soon and if you agree with my feedback, it might be easier to start from scratch rather than updating this. If you agree, I'll also close the other PR this one depends on.

PR #1745, opened 2024-03-11
SW-5008 Created scaffolding for Votes API

tommylau523:


src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/voting/")
```
sgrimm: No need for a trailing slash here since there are leading slashes on the method paths.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,91 @@
...
+class ProjectVotingController() {
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @GetMapping("/{projectId}")
+  @Operation(summary = "Gets vote selections for a single project.")
+  fun getProjectVotes(
+      @PathVariable("projectId") projectId: ProjectId,
+  ): ProjectVotesResponsePayload {
+    return ProjectVotesResponsePayload(ProjectVotesPayload(projectId, listOf()))
```
sgrimm: Could do `emptyList()` instead of `listOf()` but either works.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/voting/")
+@RestController
+class ProjectVotingController() {
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @GetMapping("/{projectId}")
```
sgrimm: Should we have another layer in the path here? Like `/projects/{projectId}` or maybe flip it around so it's `/api/v1/accelerator/projects/{projectId}/voting`?

I'm thinking that at some point we'll have a voting-related endpoint that doesn't target a single project and we'd want something other than a project ID at this path position.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,91 @@
...
+  fun getProjectVotes(
+      @PathVariable("projectId") projectId: ProjectId,
+  ): ProjectVotesResponsePayload {
+    return ProjectVotesResponsePayload(ProjectVotesPayload(projectId, listOf()))
+  }
+
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @PostMapping("/{projectId}")
```
sgrimm: We generally use PUT rather than POST for "overwrite any existing value" operations.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,91 @@
...
+  fun deleteProjectVotes(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: DeleteProjectVotesRequestPayload,
+  ): ProjectVotesResponsePayload {
+    return ProjectVotesResponsePayload(ProjectVotesPayload(projectId, listOf()))
+  }
+}
+
+data class VoteSelection(
+    val user: UserId,
```
sgrimm: This would be more consistent with the existing API style as `userId` rather than `user`.

src/main/kotlin/com/terraformation/backend/api/Annotations.kt
```diff
@@ -74,6 +74,14 @@ annotation class ApiResponse400(
     val description: String = "The request was not valid."
 )
 
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+@ApiResponseSimpleError(responseCode = "403")
+annotation class ApiResponse403(
```
sgrimm, line 80: Good call. We've been bad about including this in the schema in the API in general.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
tommylau523: I think this makes more sense to me. I think from the design mocks, it does point towards project being a center piece and functionalities built on top of it

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/projects/{projectId}/voting")
+@RestController
+class ProjectVotingController() {
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @GetMapping("/")
```
sgrimm: It might be better to do `@GetMapping` (without the path) here so that `GET /api/v1/accelerator/projects/1234/voting` works. I believe with the slash here, you'd need to do `GET /api/v1/accelerator/projects/1234/voting/`.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,99 @@
...
+
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @DeleteMapping("/")
+  @Operation(summary = "Delete voters that match the query options from the project.")
+  fun deleteProjectVotes(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: DeleteProjectVotesRequestPayload,
+  ): ProjectVotesResponsePayload {
```
sgrimm: Should this return a simple success acknowledgement (`SimpleSuccessResponsePayload`)? It's unclear to me what the content of the response payload would be here otherwise.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,99 @@
...
+  @PutMapping("/")
+  @Operation(
+      summary = "Upserts vote selections for a single project.",
+      description =
+          "Update the user's vote for the project phase. If the (user, project, phase) does not, " +
+              "exist, a new entry is created. If `voteOption` is `null`, the user has not voted. ")
+  fun upsertProjectVotes(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: UpsertProjectVotesRequestPayload,
+  ): ProjectVotesResponsePayload {
```
sgrimm: Will this return the full set of votes for the project, or just the ones that were updated?

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v1/accelerator/projects/{projectId}/voting")
```
sgrimm: Maybe `votes` instead of `voting`? Our path elements are usually nouns if they refer to a specific kind of entity.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
tommylau523: Good call!

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
tommylau523: I am leaning towards returning the updated votes.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @PutMapping
+  @Operation(
+      summary = "Upserts vote selections for a single project.",
+      description =
+          "Update the user's vote for the project phase. If the (user, project, phase) does not, " +
```
sgrimm: There's a stray comma at the end of this string.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,114 @@
...
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @GetMapping
+  @Operation(
+      summary = "Gets vote selections for a single project.",
+      description =
+          "List every vote selection for this project, organized by phases. Each phase will " +
+              "contain a list of eligible voters and their selections. If a `voteOption` is " +
+              "`null`, the voter has not yet casted a vote. ")
```
sgrimm: Might make sense to put that last piece of information in a schema annotation on `voteOption` itself (either in addition to having it here or instead of).

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,114 @@
...
+  @ApiResponse200
+  @ApiResponse403
+  @ApiResponse404
+  @PutMapping
+  @Operation(
+      summary = "Upserts vote selections for a single project.",
+      description =
+          "Update the user's vote for the project phase. If the (user, project, phase) does not, " +
+              "exist, a new entry is created. If a `voteOption` is set to `null`, the voter has " +
+              "not yet casted a vote. ")
```
sgrimm: Ditto the above comment about also including this as docs on the payload field, but also, maybe it should explicitly say that a null `voteOption` will _remove_ any existing vote.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
```diff
@@ -0,0 +1,114 @@
...
+  @ApiResponse403
+  @ApiResponse404
+  @DeleteMapping
+  @Operation(
+      summary = "Remove one or more voters from the project/phase.",
+      description =
+          "Delete the voters from the project phase, making them ineligible from voting. This is " +
+              "different from undoing a vote (by setting the `voteOption` to `null`). `projectId`" +
+              "must be provided. If neither `phaseId` or `userId` is provided, all voters from " +
+              "every phase are removed. If only `phaseId` is provided, all voters from that " +
```
sgrimm: "Default is to delete all voters" makes me a bit nervous.

Do we need bulk voter deletion in the API at all? If so, maybe as a guard against accidents it'd make sense to add a payload field with an enum that you have to explicitly set to say you want to delete a certain set of things, and then validate that the IDs match what the user says they intended.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotingController.kt
tommylau523: I don't think bulk delete is part of the requirements and it doesn't seem like we will have many use cases for that. I am opting to use a boolean flag for deleting a phase of voters, and not exposing deleting a whole project of voters at all.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+data class VoteSelection(
+    val userId: UserId,
+    @Schema(
+        description =
+            "The vote the user has selected. Can be yes/no/conditional or `null` if " +
+                "a vote is not yet selected.")
+    val voteOption: VoteOption? = null,
```
karthikbtf, line 86: @tommylau523 if the user selects 'Conditional', the designs show a text entry where user expands on decision. We would need to store/return this text data.

tommylau523: I missed a field called conditionalInfo here, I will push in a patch.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
```diff
@@ -0,0 +1,129 @@
...
+    val conditionalInfo: String? = null,
+    val email: String? = null,
```
karthikbtf: Should `email` be always present?

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
```diff
@@ -0,0 +1,129 @@
...
+data class DeleteProjectVotesRequestPayload(
+    val projectId: ProjectId,
+    val phase: CohortPhase,
+    @Schema(description = "If set to `null`, all voters in the phase will be removed. ")
+    val userId: UserId? = null,
+    @Schema(
+        description =
+            "A safeguard flag that must be set to `true` for deleting all voters in " +
+                "a project phase. ")
+    val phaseDelete: Boolean = false,
```
karthikbtf: Do we set this to any specific value for a single user delete? If not, maybe make it optional.
I like this safeguard, is there a bulk delete use-case?

karthikbtf (approved):
Looks good. A few comments.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
tommylau523: I am thinking maybe in a case where a phase of vote is created incorrectly. There is not really a delete use case at all in the requirements.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
tommylau523: Good call. Let me do so.

karthikbtf (approved):
Thanks @tommylau523 !

PR #1749, opened 2024-03-12
SW-5018 Added Project Scores tables

tommylau523:


src/main/resources/db/migration/0250/V251__ProjectScores.sql
```diff
@@ -0,0 +1,24 @@
...
+   name TEXT NOT NULL,
+   UNIQUE (name)
+);
+
+-- Add voting records
+CREATE TABLE accelerator.project_scores (
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    phase_id INTEGER NOT NULL REFERENCES accelerator.cohort_phases,
+    score_category_id INTEGER NOT NULL REFERENCES accelerator.score_categories,
+    score INTEGER CHECK (score >= -2 AND score <= 2),
```
sgrimm, line 13: Nice.

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -465,8 +465,11 @@ COMMENT ON TABLE accelerator.participants IS 'Accelerator participant details.';
 COMMENT ON TABLE accelerator.project_document_settings IS 'Per-project configuration for storage of submitted documents.';
 COMMENT ON COLUMN accelerator.project_document_settings.file_naming IS 'Identifier that is included in generated filenames. This is often, but not necessarily, the same as the project name.';
 
+COMMENT ON TABLE accelerator.project_scores IS 'Score options (nullable) selected by voters.';
```
sgrimm: Blank line between these two just so each table's comment(s) are visually separated from other tables.

src/main/resources/i18n/Enums_en.properties
```diff
@@ -19,6 +19,18 @@ accelerator.DeliverableCategory.SupplementalFiles=Supplemental Files
 accelerator.DeliverableCategory.VerraNonPermanenceRiskToolNPRT=Verra Non-Permanence Risk Tool (NPRT)
 # Noun, as in a legal document
 accelerator.DeliverableType.Document=Document
+accelerator.ScoreCategory.Carbon=Carbon
+accelerator.ScoreCategory.Finance = 'Finance'
```
sgrimm: This should just be `=Finance` rather than ` = 'Finance'` so it doesn't show up as a diff when Phrase uploads translations of this file. It treats the English files as translations too and it generates new copies of them and includes them in the PRs if there are changes.

src/main/resources/db/migration/R__Comments.sql
sgrimm: Also, is the "(nullable)" accurate?

sgrimm (changes requested):
Looks generally good; just a couple minor comments.

src/main/resources/db/migration/R__Comments.sql
tommylau523: Nullable is accurate because a scorer can select a 'Not Assigned Score Yet' state essentially

src/main/resources/i18n/Enums_en.properties
tommylau523: My mistake! I overlook the formatting of those.

src/main/resources/db/migration/R__Comments.sql
tommylau523: I separated the comments to one about the score column itself. Should give a bit of clarity

nickgraz (approved):
Nice work 🎉

PR #1750, opened 2024-03-12
SW-4991 Wait for more documents before notifying

sgrimm:
To avoid flooding accelerator admins with notifications when users upload multiple
documents in rapid succession to satisfy deliverables, defer the "deliverable
ready for review" notification until 5 minutes have passed without a new upload
for the deliverable.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
@@ -1,0 +1,67 @@
+package com.terraformation.backend.accelerator
+
+import com.terraformation.backend.accelerator.db.DeliverableStore
+import com.terraformation.backend.accelerator.event.DeliverableDocumentUploadedEvent
+import com.terraformation.backend.accelerator.event.DeliverableReadyForReviewEvent
+import com.terraformation.backend.customer.model.SystemUser
+import com.terraformation.backend.log.perClassLogger
+import jakarta.inject.Named
+import java.time.Duration
+import java.time.InstantSource
+import org.jobrunr.scheduling.JobScheduler
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.context.annotation.Lazy
+import org.springframework.context.event.EventListener
+
+@Named
+class SubmissionNotifier(
+    private val clock: InstantSource,
+    private val deliverableStore: DeliverableStore,
+    private val eventPublisher: ApplicationEventPublisher,
+    @Lazy private val scheduler: JobScheduler,
+    private val systemUser: SystemUser,
+) {
+  companion object {
+    /**
+     * Wait this long before notifying about new submissions to give the user time to submit
+     * additional items for a deliverable. This is the delay after the _last_ submission, that is,
+     * the countdown resets each time something new is submitted for a deliverable.
+     */
+    private val notificationDelay = Duration.ofMinutes(5)
+
+    private val log = perClassLogger()
+  }
+
+  /** Schedules a "ready for review" notification when a document is uploaded. */
+  @EventListener
+  fun on(event: DeliverableDocumentUploadedEvent) {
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#1750** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1752** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1752?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1751** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

sgrimm:
Test failure is a bug in an existing test; will fix it as a separate PR and stack this on the new PR.

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val deliverableStore: DeliverableStore,
+    private val eventPublisher: ApplicationEventPublisher,
+    @Lazy private val scheduler: JobScheduler,
+    private val systemUser: SystemUser,
+) {
+  companion object {
+    /**
+     * Wait this long before notifying about new submissions to give the user time to submit
+     * additional items for a deliverable. This is the delay after the _last_ submission, that is,
+     * the countdown resets each time something new is submitted for a deliverable.
```
tommylau523: Elegant solution! I may not be understanding this correctly, but from what I can gather, it is not resetting the countdown, but instead each event has a 5-minute delay to check whether it is the latest upload, before notifying, is that correct?

sgrimm: Yeah, I'll change the wording here to clarify that. "The countdown resets" is the user-visible behavior but you're right about how it actually works.

sgrimm:
> I am wondering if it makes sense for the notification to include document changed.

One wrinkle with that would be that (thanks to this change) there might be multiple documents. But maybe we'll get user feedback saying it's something people want to see.

sgrimm:
### Merge activity

* **Mar 12, 3:22 PM EDT**: @sgrimm started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1750).
* **Mar 12, 3:26 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1750) rebased this pull request as part of a merge.
* **Mar 12, 3:27 PM EDT**: @sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1750).

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
```diff
@@ -0,0 +1,68 @@
...
+     */
+    private val notificationDelay = Duration.ofMinutes(5)
+
+    private val log = perClassLogger()
+  }
+
+  /** Schedules a "ready for review" notification when a document is uploaded. */
+  @EventListener
+  fun on(event: DeliverableDocumentUploadedEvent) {
+    scheduler.schedule<SubmissionNotifier>(clock.instant().plus(notificationDelay)) {
```
karthikbtf, line 39: Much simpler approach 👍 
Do we need to worry about the corner case where service is restarted during this wait time?

sgrimm: No worry about that. JobRunr (which is where the `scheduler` object comes from) is a persistent scheduler. This call actually just inserts into the database, and then a background thread periodically scans to see if there are any jobs whose times have come. It also takes care of guaranteeing that only one host in the cluster will run a given job.

karthikbtf: Nice! Thanks for the explanation.

PR #1753, opened 2024-03-12
SW-5021 Created skeleton for Scores API

tommylau523:


src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,74 @@
...
+  ): UpsertProjectScoresResponsePayload {
+    return UpsertProjectScoresResponsePayload(projectId, payload.phase, emptyList())
+  }
+}
+
+data class CategoryScore(
+    val scoreCategory: ScoreCategory,
+    @Schema(
+        description = "Must be between -2 to 2. If set to `null`, a score has not been selected.")
+    val score: Int? = null,
```
nickgraz: Can this change to `value`? Also this will need to allow for strings as well for the qualitative scores.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,74 @@
...
+              "exist, a new entry is created. Setting a `score` to `null` removes the score.")
+  fun upsertProjectScores(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: UpsertProjectScoresRequestPayload,
+  ): UpsertProjectScoresResponsePayload {
+    return UpsertProjectScoresResponsePayload(projectId, payload.phase, emptyList())
+  }
+}
+
+data class CategoryScore(
```
nickgraz: I think this should just be `Score`.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,74 @@
...
+    val scoreCategory: ScoreCategory,
```
nickgraz: Can this just be `category`?

tommylau523:
* **#1753** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1749** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1749?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,74 @@
...
+              "exist, a new entry is created. Setting a `score` to `null` removes the score.")
+  fun upsertProjectScores(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: UpsertProjectScoresRequestPayload,
+  ): UpsertProjectScoresResponsePayload {
+    return UpsertProjectScoresResponsePayload(projectId, payload.phase, emptyList())
+  }
+}
+
+data class Score(
```
nickgraz: I think we will also need something like `inputType`, each `category` contains a number score (dropdown with -2 -> 2) and qualitative (text) input. For example "Carbon Score" and "Carbon Qualitative".

We also will have a (and possibly more in the future) system generated score. I think it would be alongside the other scores, so maybe something like `type: 'user' | 'system'` should be added. Perhaps @sgrimm has an opinion on where the system generated score will exist. We can also just do the math on the frontend, but I suspect folks outside engineering will want this value to be in the database.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,74 @@
...
+    val category: ScoreCategory,
+    @Schema(
+        description = "Must be between -2 to 2. If set to `null`, a score has not been selected.")
+    val value: Int? = null,
+)
+
+data class PhaseScores(val phase: CohortPhase, val scores: List<Score>)
```
nickgraz: This will also need `modifiedTime`, we have to display this on the frontend.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
tommylau523: Qualitative is my oversight. Adding in the next patch. 

As for system generated score, are they going to be tied to score categories? We will need to update our table schema to reflect which category is user-updated and which is system generated. 

Will update API schema to accept int | String for score as well.

tommylau523: I am thinking that we can keep every score with the following fields:

category
value: int?
qualitative: string?
modifiedTime

Each phase should always have all the categories at least according to the current PRD.

nickgraz: I don't have a super strong opinion about how it is implemented, it seems like it should just be considered "another score" for simplicity. In the mock data I use `category: "calculated"` to differentiate, but it would probably be redundant if we additionally have `type: "system"`. 

Perhaps the `category` is nullable and only applies to user supplied scores? Alternatively we ditch `type` altogether and just use the `category` to differentiate between system and user values, IE every category that isn't "system" or "calculated" is a user value.

nickgraz: > I am thinking that we can keep every score with the following fields:
> 
> category value: int? qualitative: string? modifiedTime
> 
> Each phase should always have all the categories at least according to the current PRD.

This looks good to me! The `modifiedTime` is only needed at the scorecard or phase scores level though.

tommylau523: We will have `modifiedTime` per category because we have modifiedTime stored in each of the score row. We can either have the backend returned the latest modifiedTime, or have the frontend take the latestTime of the cards. Whichever one you prefer!

tommylau523: @nickgraz I added a separate return for system scores. For now there may just be one that is hard-coded to be average. I think in the future, if there are more system scores, we will add a separate ENUM table and keep track of those. Will be nicer going forward to make the distinction between user score and system score. Let me know what you think!

nickgraz: Sounds good to me!

nickgraz: One slight snag with this approach, the "system calculated score" shouldn't be generated until all the other scores have a value. This is probably not a big deal, the math can be done on the fly if the conditions are right.

Another thing, which I think you should ask product about perhaps, this seems like the value would not be available in the database, only in the request flow. I _think_ they want it in the database for reporting purposes.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,86 @@
...
+  fun upsertProjectScores(
+      @PathVariable("projectId") projectId: ProjectId,
+      @RequestBody payload: UpsertProjectScoresRequestPayload,
+  ): UpsertProjectScoresResponsePayload {
+    return UpsertProjectScoresResponsePayload(projectId, payload.phase, emptyList())
+  }
+}
+
+data class Score(
+    val category: ScoreCategory? = null,
```
sgrimm: What would it mean to have a null category?

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,86 @@
...
+    @Schema(
+        description = "Must be between -2 to 2. If set to `null`, a score has not been selected.")
+    val value: Int? = null,
+    val qualitative: String? = null,
+    val modifiedTime: Instant = Instant.now(),
```
sgrimm: Would it be better to define separate payload classes for reads and for writes? I don't think we want clients to be able to supply a modified time when writing, if only because who knows if their clocks are accurate. Obviously we can ignore this field but IMO it'd be better to just not have it at all on the write path.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,86 @@
...
+)
+
+data class SystemScore(
+    val name: String? = null,
```
sgrimm: Is there more than one system-calculated score? What would the possible names be here? What would it mean for the name to be null?

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,86 @@
...
+)
+
+data class PhaseScores(
+    val phase: CohortPhase,
+    val scores: List<Score>,
+    val systemScores: List<SystemScore>
+)
+
+data class UpsertProjectScoresRequestPayload(
+    val projectId: ProjectId,
```
sgrimm: The project ID is already in the URL path, I believe.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
```diff
@@ -0,0 +1,86 @@
...
+data class GetProjectScoresResponsePayload(
+    val projectId: ProjectId,
+    val projectName: String,
+    val phases: List<PhaseScores>
+) : SuccessResponsePayload
+
+data class UpsertProjectScoresResponsePayload(
+    val projectId: ProjectId,
+    val phaseId: CohortPhase,
+    val results: List<Score>
```
sgrimm: This could include the system-generated total score too (otherwise the client has to fetch it after every upsert, I expect).

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
tommylau523: We were discussing to have a null category to represent system scores. I think I will revert this change. System calculated scores in general will need more design and iterations

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
tommylau523: I think at this point there is only one. I think this is to protect us from having more than one in the future. I think it is best to take it out for now.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
sgrimm: Granted, the client can calculate averages too, but I think it's probably good to have a single source of truth just in case at some point we decide to make it a weighted average or something.

src/main/kotlin/com/terraformation/backend/accelerator/api/ScoringController.kt
tommylau523: We probably want the logic to be part of ScoreService regardless. We either want to handle updating the calculated score on the database on writes, or have the service compute the score on request. 

Our database looks different depending on the approach but we should be able to expose that for reporting either way.

The first approach gives us easier access to the values (if we ever dump scores directly from the db for example)

The second approach gives us less coupling to worry about in the data perspective. 

I will make sure to check with product about whether we have that calculated score stored in database requirement.

tommylau523: @sgrimm hoping to get your thoughts on this. Whether we should write calculated scores to the database, and keep it as a service output. We can make this decision a bit downstream as well.

sgrimm: My inclination would be to store it in the database, though for a different reason: if we change the score calculation in the future, we may not want the calculated scores for older projects to change retroactively.

sgrimm: Also, in terms of the API representation, the total score could be a single `BigDecimal?` field that's a sibling of the list of per-category scores. (`BigDecimal` rather than `Double` so we can have reliable control of the number of decimal places.)

tommylau523: That sounds good to me. I think the system score changes will likely be part of a different PR. If the api data types look good to @nickgraz and @sgrimm, we can merge this one in!

src/main/resources/i18n/Enums_en.properties
```diff
@@ -19,6 +19,18 @@ accelerator.DeliverableCategory.SupplementalFiles=Supplemental Files
...
+accelerator.ScoreCategory.Forestry=Forestry
+accelerator.ScoreCategory.Legal=Legal
+accelerator.ScoreCategory.Community=Community
+accelerator.ScoreCategory.GIS=GIS
+accelerator.ScoreCategory.ClimateImpact=Climate Impact
+accelerator.ScoreCategory.ExpansionPotential=Expansion Potential
+accelerator.ScoreCategory.ExperienceAndUnderstanding=Experience and Understanding
+accelerator.ScoreCategory.OperationalCapacity=Operational Capacity
+accelerator.ScoreCategory.ResponsivenessAndAttentionToDetail=Responsiveness and Attention to Detail
+accelerator.ScoreCategory.ValuesAlignment=Values Alignment
```
sgrimm: These are all only for internal use, so I think we can include `isLocalizable = false` as an argument to the enum entry in `Config.kt` and then we don't need these strings in the strings table.

src/main/resources/db/migration/0250/V251__ProjectScores.sql
```diff
@@ -21,4 +21,3 @@ CREATE TABLE accelerator.project_scores (
 
 -- Speed up query by projects
 CREATE INDEX ON accelerator.project_scores(project_id);
```
sgrimm: Need to undo the edit to this file; editing migration scripts that have already been applied will cause the server to fail to start.

sgrimm (changes requested):

src/main/resources/i18n/Enums_en.properties
```diff
@@ -19,18 +19,6 @@ accelerator.DeliverableCategory.SupplementalFiles=Supplemental Files
...
-accelerator.ScoreCategory.Forestry=Forestry
-accelerator.ScoreCategory.Legal=Legal
-accelerator.ScoreCategory.Community=Community
-accelerator.ScoreCategory.GIS=GIS
-accelerator.ScoreCategory.ClimateImpact=Climate Impact
-accelerator.ScoreCategory.ExpansionPotential=Expansion Potential
-accelerator.ScoreCategory.ExperienceAndUnderstanding=Experience and Understanding
-accelerator.ScoreCategory.OperationalCapacity=Operational Capacity
-accelerator.ScoreCategory.ResponsivenessAndAttentionToDetail=Responsiveness and Attention to Detail
-accelerator.ScoreCategory.ValuesAlignment=Values Alignment
```
sgrimm, line 33: Phrase will re-add these next time it exports strings as a PR. Removing them here is fine, it'll just also require someone with Phrase admin rights (I'm one of them) to go delete these strings from our strings list in the Phrase UI once this is merged. Their GitHub import feature can only add and update strings, not delete them.

This is just an FYI so you're not puzzled when this deletion doesn't stick.

sgrimm (approved):
Looks good! I may end up tweaking this as I work on the implementation, of course, but this is a good base to build on.

src/main/resources/i18n/Enums_en.properties
sgrimm: I've deleted these from Phrase.

PR #1755, opened 2024-03-13

terraware-phrase:


src/main/resources/i18n/Enums_es.properties
```diff
@@ -19,6 +19,18 @@ accelerator.DeliverableCategory.SupplementalFiles=Archivos suplementarios
...
 # Noun, as in a legal document
 accelerator.DeliverableType.Document=Documento
+accelerator.ScoreCategory.Carbon=Carbono
+accelerator.ScoreCategory.ClimateImpact=Impacto climático
+accelerator.ScoreCategory.Community=Comunidad
+accelerator.ScoreCategory.ExpansionPotential=Capacidad de ampliación
+accelerator.ScoreCategory.ExperienceAndUnderstanding=Experiencia y percepción
+accelerator.ScoreCategory.Finance=Finanzas
+accelerator.ScoreCategory.Forestry=Silvicultura
+accelerator.ScoreCategory.GIS=Sistema de información geográfica (GIS)
```
tommylau523: @alexTerraformation It seems like in other places, GIS should not be translated. What should I do here?

sgrimm: I think this is one of those cases where we should just override the translation. I'll go update this in the Phrase UI and tell it to update the PR.

tommylau523: Thanks! In a separate UI I have also marked these an not localizable. Perhaps we can just close this PR.

sgrimm: Also, the translator got it wrong here -- the Spanish translation for "GIS" is actually "SIG" (it's correct in other strings).

PR #1759, opened 2024-03-13
SW-5042 Added updateProjectVoteDecisions permission, only for SystemUser

tommylau523:

PR #1757, opened 2024-03-13
SW 5040 Created VoteDecisions data tables

tommylau523:

PR #1756, opened 2024-03-13
Added Cohort phase check for Vote upserts and deletes

tommylau523:

PR #1760, opened 2024-03-14
SW-5034 Extended VoteStore to handle updating vote decisions

tommylau523:
Added VoteDecisionModel, implemetned fetch and update VoteDecision in store

Added updating functionalities to VoteStoreTest


tommylau523:
* **#1760** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1760?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1757** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1757?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#1759](https://github.com/terraware/terraware-server/pull/1759) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1759?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#1756** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1756?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -34,13 +35,27 @@ class VoteStore(
     }
   }
 
-  fun delete(projectId: ProjectId, phase: CohortPhase, userId: UserId? = null) {
+  fun fetchAllVoteDecisions(projectId: ProjectId): List<VoteDecisionModel> {
+    requirePermissions { readProjectVotes(projectId) }
+    return with(PROJECT_VOTE_DECISIONS) {
+      dslContext
+          .select(PROJECT_VOTE_DECISIONS.asterisk())
+          .from(PROJECT_VOTE_DECISIONS)
```
sgrimm: This is totally fine as-is, but FYI, for single-table "select all columns" queries you can also do `selectFrom(PROJECT_VOTE_DECISIONS)` which selects all the columns and also causes the `fetch` method to return a `ProjectVoteDecisionsRecord` instead of a generic `Record`.

In this case since you're mapping the record to a custom class anyway, there's no functional difference (and doing it this way is 100% fine), just pointing out the feature so you'll know about it in the future.

For queries where you're selecting columns from more than one table, you have to use the `select(...).from(...)` style.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -101,4 +120,50 @@ class VoteStore(
...
+      phase: CohortPhase,
+      now: Instant = clock.instant()
+  ): VoteDecisionModel {
+    val votes =
+        dslContext
+            .select(PROJECT_VOTES.VOTE_OPTION_ID)
+            .from(PROJECT_VOTES)
+            .where(PROJECT_VOTES.PROJECT_ID.eq(projectId))
+            .and((PROJECT_VOTES.PHASE_ID.eq(phase)))
+            .fetch { it.value1() }
```
sgrimm: `.fetch(PROJECT_VOTES.VOTE_OPTION_ID)` will be slightly more efficient here.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -101,4 +120,50 @@ class VoteStore(
...
+        } else if (votes.any { it == null }) {
+          // If one or more voter has not submitted a vote
+          null
+        } else {
+          val counts = votes.groupingBy { it!! }.eachCount()
+          val highestCount = counts.maxOf { it.value }
+          val majority = counts.filterValues { it == highestCount }.keys
+
+          // Return relative majority only if not a tie
+          if (majority.size == 1) majority.first() else null
```
sgrimm: It occurred to me that this could be `majority.singleOrNull()` but I think they way you have it is probably a bit better because it makes the intent more explicit.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -101,4 +120,50 @@ class VoteStore(
...
+    with(PROJECT_VOTES) {
+      dslContext
+          .insertInto(PROJECT_VOTE_DECISIONS)
+          .set(PROJECT_ID, projectId)
+          .set(PHASE_ID, phase)
+          .set(VOTE_OPTION_ID, decision)
+          .set(MODIFIED_TIME, now)
+          .onDuplicateKeyUpdate()
+          .set(PROJECT_ID, projectId)
+          .set(PHASE_ID, phase)
```
sgrimm: Setting the project and phase would be a no-op here, I think, since they're the primary key columns that caused the duplicate key update to happen.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
```diff
@@ -40,13 +42,13 @@ class VoteStoreTest : DatabaseTest(), RunsAsUser {
...
   @Nested
-  inner class FetchAllVotesByProject {
+  inner class FetchAllVotes {
     @Test
     fun `fetches with no vote`() {
       val projectId = insertProject(participantId = inserted.participantId)
       clock.instant = Instant.EPOCH.plusSeconds(500)
 
-      assertEquals(listOf<VoteModel>(), store.fetchAllVotesByProject(projectId))
+      assertEquals(listOf<VoteModel>(), store.fetchAllVotes(projectId))
```
sgrimm: Could be `emptyListOf<VoteModel>()` to make the intent a little more explicit.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -34,13 +35,27 @@ class VoteStore(
...
 
-  fun delete(projectId: ProjectId, phase: CohortPhase, userId: UserId? = null) {
+  fun fetchAllVoteDecisions(projectId: ProjectId): List<VoteDecisionModel> {
+    requirePermissions { readProjectVotes(projectId) }
+    return with(PROJECT_VOTE_DECISIONS) {
+      dslContext
+          .select(PROJECT_VOTE_DECISIONS.asterisk())
+          .from(PROJECT_VOTE_DECISIONS)
+          .where(PROJECT_ID.eq(projectId))
+          .orderBy(PROJECT_ID, PHASE_ID)
```
sgrimm: No need to order by project ID here since there will never be more than one value.

src/main/kotlin/com/terraformation/backend/accelerator/db/VoteStore.kt
```diff
@@ -34,13 +35,27 @@ class VoteStore(
...
+          .fetch { VoteDecisionModel.of(it) }
+    }
+  }
+
+  fun delete(projectId: ProjectId, phase: CohortPhase, userId: UserId? = null): VoteDecisionModel {
```
sgrimm, line 54: Might be worth a documentation comment clarifying that this deletes _voters_ from the project+phase, as opposed to deleting (setting to null) their _votes_.

src/test/kotlin/com/terraformation/backend/accelerator/db/VoteStoreTest.kt
```diff
@@ -566,5 +634,84 @@ class VoteStoreTest : DatabaseTest(), RunsAsUser {
         store.upsert(projectId, phase, newUser, null, null)
       }
     }
+
+    @Test
+    fun `computes vote decision on creating one vote`() {
```
sgrimm, line 639: I think the common scenario will be that voters are added with no vote initially (maybe populated from the default voters list). Might be worth testing that when a voter with no vote is added, the decision ends up null.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -410,6 +412,7 @@ abstract class DatabaseTest {
   protected val projectReportSettingsDao: ProjectReportSettingsDao by lazyDao()
   protected val projectsDao: ProjectsDao by lazyDao()
   protected val projectVotesDao: ProjectVotesDao by lazyDao()
+  protected val projectVoteDecisionDao: ProjectVoteDecisionsDao by lazyDao()
```
sgrimm, line 414: Minor nit: This should be one line higher so it's in alphabetical order.

sgrimm (changes requested):
This looks pretty good! I have some comments but they're all minor.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -408,8 +410,9 @@ abstract class DatabaseTest {
   protected val plantingZonesDao: PlantingZonesDao by lazyDao()
   protected val projectDocumentSettingsDao: ProjectDocumentSettingsDao by lazyDao()
   protected val projectReportSettingsDao: ProjectReportSettingsDao by lazyDao()
-  protected val projectsDao: ProjectsDao by lazyDao()
   protected val projectVotesDao: ProjectVotesDao by lazyDao()
+  protected val projectsDao: ProjectsDao by lazyDao()
+  protected val projectVoteDecisionDao: ProjectVoteDecisionsDao by lazyDao()
```
sgrimm: This still seems out of order to me ("s" is before "v" and "d" is before "s").

sgrimm (approved):
One minor nit about alphabetical order you can fix before merging, but this looks good other than that!

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
tommylau523: [facepalm](https://app.graphite.dev/settings/meme-library?org=terraware)

PR #1761, opened 2024-03-14
Rename a few project votes API payload properties

karthikbtf:
- observed candidates for rename when I synced FE with project votes API types
- renamed user to userId for consistency
- set project votes payload in a root property so it can be extracted without cherry picking properties (consistent with existing pattern in other API)


karthikbtf:
@tommylau523 I didn't catch these until I merged types in the FE.

tommylau523:
Looks good to me. Likely there will be one more change incoming this week, to include vote decision computation in responses. Will keep you synced!

tommylau523 (approved):

karthikbtf:
> Looks good to me. Likely there will be one more change incoming this week, to include vote decision computation in responses. Will keep you synced!

Sounds good, I have an FE ticket open to sync with the vote decision changes.

PR #1763, opened 2024-03-14
SW-5056 Created default voters table

tommylau523:


tommylau523:
* **#1763** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1763?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/db/migration/0250/V253__DefaultVoters.sql
```diff
@@ -0,0 +1,6 @@
+-- Add scoring records
+CREATE TABLE accelerator.default_voters (
+    user_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+    PRIMARY KEY (user_id)
+);
+
```
karthikbtf, line 6: Do we need a global roles constraint? 
If so, not sure what should happen if a default voter has all global roles removed.

tommylau523: Good call! Let me put that in! Like we can delete the default_voters in case global role is removed. This will not alter existing votes but remove the voter from being added in the future.

sgrimm: I don't think it's practical to do that as a database constraint (at least without writing custom trigger code). `REFERENCES` constraints can only point to unique/primary keys.

tommylau523: I had just run into this issue basically. I am considering folding this into an additional global role. 

https://terraformationteam.slack.com/archives/C01C69KA23Y/p1710458929796029

sgrimm: If we add a way to add default voters via the admin UI, we could make it show a warning if you add someone who doesn't have a suitable global role. I think that'll be totally adequate as a solution.

The list of voters is going to be quite small, won't change that often, and will consist entirely of people who need to be able to access the accelerator console for other reasons anyway. If they can't see anything about participants or cohorts at all, we'll discover that and fix it long before they'd need to vote on anything.

sgrimm:
Test failure here is unrelated to this change; there are a couple tests for the random plot selection code that have a very tiny but nonzero chance of failing if the random number generator happens to pick particular values a few dozen times in a row. I've kicked off another test run.

sgrimm (approved):

PR #1764, opened 2024-03-14
SW-5035 Implement project score logic

sgrimm:
Add real implementations of the `/api/v1/accelerator/projects/*/scores`
endpoints.

This also simplifies the payload definitions, removing response fields that
wouldn't be used by clients.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresController.kt
@@ -1,14 +1,21 @@
 package com.terraformation.backend.accelerator.api
 
+import com.terraformation.backend.accelerator.db.ProjectScoreStore
+import com.terraformation.backend.accelerator.model.NewProjectScoreModel
+import com.terraformation.backend.accelerator.model.ProjectScoreModel
 import com.terraformation.backend.api.AcceleratorEndpoint
 import com.terraformation.backend.api.ApiResponse200
 import com.terraformation.backend.api.ApiResponse404
+import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.accelerator.CohortPhase
 import com.terraformation.backend.db.accelerator.ScoreCategory
 import com.terraformation.backend.db.default_schema.ProjectId
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.Valid
+import jakarta.validation.constraints.Max
+import jakarta.validation.constraints.Min
 import java.math.BigDecimal
 import java.time.Instant
 import org.springframework.web.bind.annotation.GetMapping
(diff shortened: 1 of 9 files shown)


sgrimm:
* **#1764** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresController.kt
```diff
@@ -21,68 +28,77 @@ import org.springframework.web.bind.annotation.RestController
...
-    val phases: List<PhaseScores>
-) : SuccessResponsePayload
-
-data class UpsertProjectScoresResponsePayload(
-    val projectId: ProjectId,
-    val phaseId: CohortPhase,
-    val scores: List<Score>,
-    val systemScore: BigDecimal? = null,
-) : SuccessResponsePayload
+data class GetProjectScoresResponsePayload(val phases: List<PhaseScores>) : SuccessResponsePayload
```
sgrimm, line 104: I removed the project ID from this because it's always going to be the one from the request URL, and project name because it's not really related to scores and because I believe the client will have already needed to fetch the project information before it can get far enough along to call this endpoint. If I'm wrong about that, we can add it back, of course.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresController.kt
```diff
@@ -21,68 +28,77 @@ import org.springframework.web.bind.annotation.RestController
...
+    val value: Int? = null,
+) {
+  fun toModel() = NewProjectScoreModel(category, null, qualitative, value)
+}
 
 data class PhaseScores(
     val phase: CohortPhase,
     val scores: List<Score>,
-    val systemScore: BigDecimal? = null,
+    val totalScore: BigDecimal? = null,
```
sgrimm, line 96: Changed this name to reflect the fact that we're now calling this the "total score" per the requirements.

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectScoreStore.kt
```diff
@@ -0,0 +1,100 @@
...
+              .set(CREATED_BY, userId)
+              .set(CREATED_TIME, now)
+              .set(MODIFIED_BY, userId)
+              .set(MODIFIED_TIME, now)
+              .set(PHASE_ID, phase)
+              .set(PROJECT_ID, projectId)
+              .set(QUALITATIVE, score.qualitative)
+              .set(SCORE, score.score)
+              .set(SCORE_CATEGORY_ID, score.category)
+              .onConflict()
```
tommylau523, line 75: Does this work the same as `onDuplicateKeyUpdate`?  I actually think `onConflict().doUpdate()` reads better!

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectScoreStore.kt
```diff
@@ -0,0 +1,100 @@
...
+      projectId: ProjectId,
+      phase: CohortPhase,
+      scores: Collection<NewProjectScoreModel>
+  ) {
+    requirePermissions { updateProjectScores(projectId) }
+
+    val now = clock.instant()
+    val userId = currentUser().userId
+
+    val currentPhase =
```
tommylau523: pretty minor but maybe we can refactor this part since it is shared by Votes too. some AcceleratorProjectUtil

tommylau523 (approved):
Looks great! Just a few questions for my understanding.

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectScoreStore.kt
sgrimm: The "on conflict" syntax is the native PostgreSQL way to do this. "On duplicate key" is the MySQL syntax (maybe MySQL picked it up from somewhere else originally, not sure) but jOOQ translates it to an equivalent PostgreSQL statement at runtime.

PR #1767, opened 2024-03-15
SW-4961 Get User and Search Users API

nickgraz:
- Add "get user" API, only supports "ordinary" users for now. 
- Add "search users" API, only supports search by email for now.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
@@ -1,6 +1,8 @@
 package com.terraformation.backend.customer.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.terraformation.backend.api.ApiResponse200
+import com.terraformation.backend.api.ApiResponse404
 import com.terraformation.backend.api.ArbitraryJsonObject
 import com.terraformation.backend.api.CustomerEndpoint
 import com.terraformation.backend.api.SimpleSuccessResponsePayload
@@ -8,6 +10,8 @@ import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.UserStore
 import com.terraformation.backend.customer.model.IndividualUser
+import com.terraformation.backend.db.UserNotFoundException
+import com.terraformation.backend.db.UserNotFoundForEmailException
 import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.UserId
@@ -19,6 +23,7 @@ import java.time.ZoneId
 import java.util.Locale
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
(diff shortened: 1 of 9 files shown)


nickgraz:
* **#1767** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
```diff
@@ -104,6 +99,38 @@ class UsersController(private val userStore: UserStore) {
...
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping
+  @Operation(summary = "Gets a user by some criteria, for now only email is available")
+  fun searchUsers(
+      @RequestParam
+      @Schema(description = "The email to use when searching for a user")
+      email: String
+  ): GetUserResponsePayload {
+    val user = userStore.fetchByEmail(email)
```
sgrimm: I think we need a wrapper function for this in `UserStore` that checks permissions. The `fetchByEmail` function is used internally in the auth code so it can't check permissions all the time (it's called in contexts where there's no "current user" yet, I believe) but we need to not allow random users to look up other random users' information.

For now, just checking for a global role is probably okay here. A more sophisticated permission check could also allow looking up people who are in one of the current user's organizations.

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
```diff
@@ -104,6 +99,38 @@ class UsersController(private val userStore: UserStore) {
...
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{userId}")
+  @Operation(summary = "Get a user by ID, if they exist, only ordinary users are supported.")
+  fun getUser(
+      @PathVariable("userId") userId: UserId,
+  ): GetUserResponsePayload {
+    val user = userStore.fetchOneById(userId)
```
sgrimm: Similar to previous comment: this needs to prevent exposing all users' data to all other users.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
nickgraz: Adding the wrapper function now, do you think I should also verify that the found user has an `@terraformation.com` email, otherwise throw "not found"?

sgrimm: I think we'll probably find it useful to have "get other user who I have permission to view" as a general-purpose endpoint down the road, so I'd say probably not.

nickgraz: ~Also, the permission I am thinking of adding is "canReadTerraformationUser", that way I can apply it to both `fetchOneById` and `fetchByEmail` wrapper functions. Does that sound alright or should I have a separate permission for each one?~

sgrimm: I think `canReadUser(userId)` is probably a more general design. And for now, the implementation of that permission just checks if you have a global role. For the email search, it'll be a little different: do the fetch first, then do the permission check on the looked-up user's user ID. That is, in both cases the permission is "can I see this specific user" rather than "can I search for users like this."

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -2602,6 +2618,17 @@ internal class PermissionTest : DatabaseTest() {
       }
     }
 
+    fun expect(
```
sgrimm, line 2621: Need to call this in `andNothingElse` too so it tests that the permission is false for regular users.

sgrimm (changes requested):
This looks good! Just missing one bit in `PermissionTest` (and we want to verify that the test passes with that change, hence me marking this Request Changes instead of approving).

PR #1769, opened 2024-03-15
SW-5047 Add createdTime to IndividualUser and UserWithGlobalRolesPayload

nickgraz:
- In order to show the "Date Added" column in the "list people" accelerator view, add `createdTime` to the `IndividualUser` and `UserWithGlobalRolesPayload`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
@@ -46,7 +46,7 @@ class GlobalRolesController(
 }
 
 data class UserWithGlobalRolesPayload(
-    val createdTime: Instant?,
+    val createdTime: Instant,
     val id: UserId,
     val email: String,
     val firstName: String?,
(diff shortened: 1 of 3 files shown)


nickgraz:
* **#1769** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
```diff
@@ -45,6 +46,7 @@ class GlobalRolesController(
 }
 
 data class UserWithGlobalRolesPayload(
+    val createdTime: Instant?,
```
sgrimm: This should never be null, right?

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -74,6 +75,7 @@ import org.springframework.security.core.userdetails.UserDetails
  * will be cached afterwards.
  */
 data class IndividualUser(
+    val createdTime: Instant?,
```
sgrimm: I believe this can be non-nullable since the underlying column in the users table is.

nickgraz: That is what I originally tried, but the `UsersRow` is nullable. `build/generated-jooq/com/terraformation/backend/db/default_schema/tables/pojos/UsersRow.kt:29`



Perhaps I need to use the `usersRow.createdTime!!` when creating `IndividualUser`?

nickgraz: Never mind, I see what you did there for the other non-nullable fields and applied the same pattern.

sgrimm (approved):
Looks good!

PR #1774, opened 2024-03-16
Updated Votes API to not include project IDs in payloads

tommylau523:
projectId is already in the path of the API.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectVotesController.kt
@@ -29,7 +29,9 @@ import org.springframework.web.bind.annotation.RestController
 @AcceleratorEndpoint
 @RequestMapping("/api/v1/accelerator/projects/{projectId}/votes")
 @RestController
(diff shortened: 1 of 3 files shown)


tommylau523:
* **#1774** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (commented):
What's the motivation here? The PR description says what the change does, but not why.

Generally we return IDs and let the clients look up further information using those IDs as needed; in a lot of cases the clients will already have the information on hand anyway. We can return names alongside the IDs as an optimization if there are going to be lots of IDs that clients aren't likely to have already looked up, but for things like turning the names into clickable links, the IDs are usually still needed.

Is there a particular reason we definitely _don't_ want to know the project IDs, just the names?

tommylau523:
One nit thing is that projectId is already in the request path. We are returning duplicated information here. Another thing is that an accelerator admin with global roles may not have access to project names via APIs. Let's move this second point to Slack to discuss.

sgrimm (approved):
Makes sense.

PR #1779, opened 2024-03-18
SW-5047 API for bulk removing global roles from users

nickgraz:
- Add API `DELETE /api/v1/globalRoles/users` for bulk removing global roles from a given set of user IDs
  - Receives a payload of `{"userIds: [$userId1,$userId2]}`
- Update the `UserStore.updateGlobalRoles` function to accept a list of userIds to perform the operation on.
  - The operation is considered "all or nothing", IE no users supplied in the list will be updated if there is an error with any of them (for example if some of the users are not `@terraformation.com` users)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
@@ -11,6 +11,7 @@ import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.db.default_schema.UserId
 import io.swagger.v3.oas.annotations.Operation
 import java.time.Instant
+import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
@@ -43,6 +44,18 @@ class GlobalRolesController(
 
     return SimpleSuccessResponsePayload()
   }
+
+  @ApiResponse200
+  @ApiResponse404
+  @DeleteMapping("/users/globalRoles")
+  @Operation(summary = "Remove global roles from the supplied users.")
+  fun removeGlobalRoles(
+    @RequestBody payload: RemoveGlobalRolesRequestPayload,
+  ): SuccessResponsePayload {
+//    userStore.updateGlobalRoles(userId, payload.globalRoles)
+
+    return SimpleSuccessResponsePayload()
+  }
 }
 
 data class UserWithGlobalRolesPayload(
@@ -66,6 +79,10 @@ data class UserWithGlobalRolesPayload(
 
 data class GlobalRoleUsersListResponsePayload(val users: List<UserWithGlobalRolesPayload>) :
     SuccessResponsePayload
+
+data class RemoveGlobalRolesRequestPayload(
+    val userIds: Set<UserId>,
+)
 
 data class UpdateGlobalRolesRequestPayload(
     val globalRoles: Set<GlobalRole>,
--- a/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
@@ -40,7 +40,7 @@ class GlobalRolesController(
       @PathVariable("userId") userId: UserId,
       @RequestBody payload: UpdateGlobalRolesRequestPayload,
   ): SuccessResponsePayload {
-    userStore.updateGlobalRoles(userId, payload.globalRoles)
+    userStore.updateGlobalRoles(setOf(userId), payload.globalRoles)
 
     return SimpleSuccessResponsePayload()
   }
@@ -50,9 +50,9 @@ class GlobalRolesController(
   @DeleteMapping("/users/globalRoles")
   @Operation(summary = "Remove global roles from the supplied users.")
   fun removeGlobalRoles(
-    @RequestBody payload: RemoveGlobalRolesRequestPayload,
+      @RequestBody payload: RemoveGlobalRolesRequestPayload,
   ): SuccessResponsePayload {
-//    userStore.updateGlobalRoles(userId, payload.globalRoles)
+    userStore.updateGlobalRoles(payload.userIds, emptySet())
 
     return SimpleSuccessResponsePayload()
   }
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#1779** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1779?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz:
@sgrimm would you mind taking a look when you have a moment?

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
```diff
@@ -39,7 +40,19 @@ class GlobalRolesController(
...
+    userStore.updateGlobalRoles(setOf(userId), payload.globalRoles)
+
+    return SimpleSuccessResponsePayload()
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @DeleteMapping("/users/globalRoles")
+  @Operation(summary = "Remove global roles from the supplied users.")
+  fun removeGlobalRoles(
```
sgrimm: Looks like all the other DELETE endpoint function names use the word "delete" rather than "remove," so maybe it'd be good to use that word here for consistency.

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
```diff
@@ -39,7 +40,19 @@ class GlobalRolesController(
...
   ): SuccessResponsePayload {
-    userStore.updateGlobalRoles(userId, payload.globalRoles)
+    userStore.updateGlobalRoles(setOf(userId), payload.globalRoles)
+
+    return SimpleSuccessResponsePayload()
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @DeleteMapping("/users/globalRoles")
```
sgrimm: I think this might be better as `/globalRoles/users` so that (a) it's not defining another special literal user ID under `/api/v1/users` in addition to "me" and (b) because it is kind of symmetrical to `GET /api/v1/globalRoles/users` in that it operates on lists of users that have global roles.

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -363,25 +363,33 @@ class UserStore(
...
-      dslContext.deleteFrom(USER_GLOBAL_ROLES).where(USER_GLOBAL_ROLES.USER_ID.eq(userId)).execute()
+      dslContext
+          .deleteFrom(USER_GLOBAL_ROLES)
+          .where(USER_GLOBAL_ROLES.USER_ID.`in`(userIds))
+          .execute()
 
       if (roles.isNotEmpty()) {
-        val records = roles.map { UserGlobalRolesRecord(userId = userId, globalRoleId = it) }
+        userIds.forEach { userId ->
+          val records = roles.map { UserGlobalRolesRecord(userId = userId, globalRoleId = it) }
```
sgrimm: Could combine the `userIds` and `roles` iterations here and do one insert, something like
```kotlin
val records = users.flatMap { userId ->
    roles.map { UserGlobalRolesRecord(userId = userId, globalRoleId = it) }
}

dslContext.insertInto(USER_GLOBAL_ROLES).set(records).execute()
```

src/test/kotlin/com/terraformation/backend/customer/db/UserStoreTest.kt
```diff
@@ -753,18 +753,44 @@ internal class UserStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      assertEquals(emptyList<Any>(), userGlobalRolesDao.findAll())
+    }
+
+    @Test
+    fun `can remove all global roles from a set of users`() {
+      val userId1 = insertUser(10)
+      val userId2 = insertUser(11)
+
+      userStore.updateGlobalRoles(setOf(userId1, userId2), emptySet())
```
sgrimm: The previous test case has this problem too, but I'm just now noticing it: this doesn't actually test removing global roles because the test never inserts any global roles for `updateGlobalRoles` to remove.

src/test/kotlin/com/terraformation/backend/customer/db/UserStoreTest.kt
```diff
@@ -753,18 +753,44 @@ internal class UserStoreTest : DatabaseTest(), RunsAsUser {
...
+              UserGlobalRolesRow(userId = userId2, globalRoleId = GlobalRole.SuperAdmin)),
+          userGlobalRolesDao.findAll())
+    }
+
     @Test
     fun `can remove all global roles from a user`() {
       val userId = insertUser(10)
 
-      userStore.updateGlobalRoles(userId, emptySet())
+      userStore.updateGlobalRoles(setOf(userId), emptySet())
```
sgrimm: Can probably lose this test case since its functionality is covered by the new multi-user-ID one.

src/test/kotlin/com/terraformation/backend/customer/db/UserStoreTest.kt
```diff
@@ -773,7 +799,32 @@ internal class UserStoreTest : DatabaseTest(), RunsAsUser {
     fun `throws exception if user does not have a Terraformation email address`() {
       val userId = insertUser(10, email = "test@elsewhere.com")
 
-      assertThrows<AccessDeniedException> { userStore.updateGlobalRoles(userId, emptySet()) }
+      assertThrows<AccessDeniedException> { userStore.updateGlobalRoles(setOf(userId), emptySet()) }
```
sgrimm: Can remove this test case since the newly-added one covers the behavior implicitly.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
nickgraz: I wrestled with that for a moment, ended up going with what I did because I generally try to put "the resource" (user.globalRoles) in question at the end. I will change it up.

nickgraz: I generally try to put the resource (I know we aren't RESTful, but I can't help myself) at the end. But you make a good point about the new special literal ID, I'll move it.

src/main/kotlin/com/terraformation/backend/customer/api/GlobalRolesController.kt
```diff
@@ -47,9 +47,9 @@ class GlobalRolesController(
 
   @ApiResponse200
   @ApiResponse404
-  @DeleteMapping("/users/globalRoles")
+  @DeleteMapping("/globalRoles/users")
   @Operation(summary = "Remove global roles from the supplied users.")
-  fun removeGlobalRoles(
+  fun deleteGlobalRoles(
       @RequestBody payload: RemoveGlobalRolesRequestPayload,
```
sgrimm: Maybe rename the payload class too for consistency?

sgrimm (approved):
Looks good! Just one small nit but no need for re-review for that.

PR #1772, opened 2024-03-15
SW-5058 Auto-assign default voters

sgrimm:
We want the list of voters to automatically be populated for the participant
projects that are eligible for voting in a particular phase.

Add logic to automatically populate the projects' voter rosters with the list
of default voters when:

- A project is added to a participant that is already in a cohort.
- A participant that already has projects is added to a cohort.
- A cohort moves to a new phase.

PR #1780, opened 2024-03-18
Implemented default voters controller

tommylau523:
Added default voters API endpoints in `/api/v1/accelerator/voters`

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/DefaultVotersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/DefaultVotersController.kt
@@ -9,6 +9,8 @@ import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.default_schema.UserId
 import io.swagger.v3.oas.annotations.Operation
 import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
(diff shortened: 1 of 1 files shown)


tommylau523:
* **#1781** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1781?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1780** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1772** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1772?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

sgrimm (commented):
I'm not sure we need this. If we only expose the default voters list via the admin UI, nothing will consume these endpoints, right?

tommylau523:
> I'm not sure we need this. If we only expose the default voters list via the admin UI, nothing will consume these endpoints, right?
> 
> 

I was working with the assumptions that the admin UI will consume these APIs. Maybe that means that I have a gap of understanding of what the admin UI actually is.

tommylau523:
> > I'm not sure we need this. If we only expose the default voters list via the admin UI, nothing will consume these endpoints, right?
> > 
> > 
> 
> I was working with the assumptions that the admin UI will consume these APIs. Maybe that means that I have a gap of understanding of what the admin UI actually is. 

Another consideration is that when we update the list of default voters, we may want to propagate those changes to existing projects and their cohort phases. That is reflected in the downstream PR.

PR #1785, opened 2024-03-19
Created admin default voters page

tommylau523:

PR #1786, opened 2024-03-19
SW-5058 Added updateExisting checkbox, and send event if box is checked

tommylau523:

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/VoteService.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/VoteService.kt
@@ -5,6 +5,8 @@ import com.terraformation.backend.accelerator.db.ParticipantStore
 import com.terraformation.backend.accelerator.db.VoteStore
 import com.terraformation.backend.accelerator.event.CohortParticipantAddedEvent
 import com.terraformation.backend.accelerator.event.CohortPhaseUpdatedEvent
+import com.terraformation.backend.accelerator.event.DefaultVoterAddedEvent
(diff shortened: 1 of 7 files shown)


tommylau523:
* **#1786** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1786?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1785** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1785?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

sgrimm (changes requested):
"Update existing" may be too coarse-grained to use in production once we have multiple cohorts in play. A couple possible approaches:

* Add a multi-select list of cohorts and let the admin pick which ones to update. They can select just the ones that actually need the change in voters.
* Exclude cohorts where all the voters have already voted on all the projects in the current phase. That way changing the default list won't mess with completed votes.

Or combine the two where there's a multi-select list and by default only the cohorts with incomplete votes in their current phases are selected.

Eventually maybe we'll need to be able tweak this on a per-project level rather than a per-cohort level but something like the above will at least make it harder to retroactively reopen votes on older cohorts.

Also, in terms of the architecture here, I think it's fine to have the admin controller directly call a `VoteService` method rather than using an event. Usually we use events for this kind of thing when we want to avoid circular dependencies or dependencies between store classes, but it's fine for a controller to depend on a service. That'll maybe make it a little less awkward to model the "choose which cohorts to update" part of my suggestion.

PR #1788, opened 2024-03-20
SW-5089 Implemented admin unfiltered search

tommylau523:
TODO:
- determine if existing search functions will be impacted. Add strict organization id filter when applicable

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
@@ -133,6 +133,8 @@ data class DeviceManagerUser(
 
   override fun canReadTimeseries(deviceId: DeviceId): Boolean = canAccessDevice(deviceId)
 
+  override fun canSearchUnfiltered(): Boolean = false
+
   override fun canSendAlert(facilityId: FacilityId): Boolean = canAccessFacility(facilityId)
 
   override fun canTriggerAutomation(automationId: AutomationId): Boolean =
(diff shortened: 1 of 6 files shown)


tommylau523:
* **#1788** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1788?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (changes requested):
A couple concerns about this approach, but I think it's a good start.

First, we'd need to make sure the frontend code isn't relying on the implicit filter for the user's organizations. I think in the majority of places, it's limiting its queries to the currently-selected organization, but if there are places where it isn't doing that, they're suddenly going to return results for all the orgs on the system if you're logged in as a super admin.

One approach to mitigate that problem, if it actually is a problem, would be to add a flag to the search payload that you have to set to true to do a systemwide search. We'd set that flag on the accelerator console searches but not in the user-facing web app. Obviously that flag would only work if the user had permission.

Second, this may need at least three access levels rather than a simple yes/no. Super-admin users need to be able to see all organizations so that they can tag organizations as being in the accelerator. Accelerator admins and other global roles need to see organizations they're not in, but probably shouldn't be able to see non-accelerator orgs.

tommylau523:
> A couple concerns about this approach, but I think it's a good start.
> 
> First, we'd need to make sure the frontend code isn't relying on the implicit filter for the user's organizations. I think in the majority of places, it's limiting its queries to the currently-selected organization, but if there are places where it isn't doing that, they're suddenly going to return results for all the orgs on the system if you're logged in as a super admin.
> 
> One approach to mitigate that problem, if it actually is a problem, would be to add a flag to the search payload that you have to set to true to do a systemwide search. We'd set that flag on the accelerator console searches but not in the user-facing web app. Obviously that flag would only work if the user had permission.
> 
> Second, this may need at least three access levels rather than a simple yes/no. Super-admin users need to be able to see all organizations so that they can tag organizations as being in the accelerator. Accelerator admins and other global roles need to see organizations they're not in, but probably shouldn't be able to see non-accelerator orgs.
> 
> 

That sounds good Steve. I am taking a look now in each table and how the visibility conditions get inherited. I will push a new patch to modify these conditions by roles.

PR #1800, opened 2024-03-22
SW-5058 Added in default voter admin screen options to propagate to projects.

tommylau523:

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminDefaultVotersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminDefaultVotersController.kt
@@ -1,12 +1,22 @@
 package com.terraformation.backend.admin
 
 import com.terraformation.backend.accelerator.db.DefaultVoterStore
+import com.terraformation.backend.accelerator.db.VoteStore
 import com.terraformation.backend.api.RequireGlobalRole
 import com.terraformation.backend.customer.db.UserStore
 import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.accelerator.CohortPhase
+import com.terraformation.backend.db.accelerator.VoteOption
+import com.terraformation.backend.db.accelerator.tables.references.COHORTS
+import com.terraformation.backend.db.accelerator.tables.references.PARTICIPANTS
+import com.terraformation.backend.db.accelerator.tables.references.PROJECT_VOTE_DECISIONS
 import com.terraformation.backend.db.default_schema.GlobalRole
+import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.default_schema.tables.references.PROJECTS
 import com.terraformation.backend.log.perClassLogger
+import org.jooq.DSLContext
+import org.jooq.Record
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
 import org.springframework.validation.annotation.Validated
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#1800** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/templates/admin/defaultVoters.html
```diff
@@ -12,14 +12,21 @@
...
+    projects with incomplete votes will inherit the voter change.
 </p>
 
 <table>
     <thead>
     <tr>
         <th>Email</th>
+        <th>Action</th>
+        <th>Projects:</th>
+        <th th:each="project : ${projects}"
```
sgrimm, line 26: Would it make sense to use a `<select multiple>` for this instead of a one-column-per-project approach? I'm concerned about what this UI will look like when there are a couple dozen projects.

Alternately, maybe rotate this 90 degrees so the voters are the columns and the projects are the rows.

src/main/kotlin/com/terraformation/backend/admin/AdminDefaultVotersController.kt
```diff
@@ -64,11 +83,18 @@ class AdminDefaultVotersController(
...
   ): String {
+    requirePermissions { updateDefaultVoters() }
+
+    val projects = getAcceleratorProjects()
+
     try {
       defaultVoterStore.delete(userId)
       redirectAttributes.successMessage = "Default voter removed."
+
+      projectIds?.forEach { voteStore.upsert(it, getProjectPhase(it)!!, userId) }
```
sgrimm, line 97: If I'm following the code correctly, when you remove a default voter and leave the checkboxes checked for their existing projects, it will keep the user on the voter list of those projects but will set their votes and conditional text to null. Do I have that right? Should this be calling `voteStore.delete()` instead?

src/main/resources/templates/admin/defaultVoters.html
```diff
@@ -34,6 +41,16 @@ <h2>Default Voters</h2>
...
             </form>
         </td>
+        <td> </td>
+        <td th:each="project : ${projects}">
+            <input type="checkbox"
+                   name="projectIds"
+                   th:form="|remove-${user.userId}|"
+                   th:value="|${project.projectId}|"
+                   th:title="|${project.projectId}|"
+                   th:checked="${project.voteDecision == null}"
```
sgrimm, line 51: Am I correct that the checkbox for a given user on a given project's row doesn't reflect whether or not the user is currently a voter on that project? I feel like that's maybe a bit counterintuitive since the grid layout suggests to me that I'm seeing per-user, per-project values. Maybe others will disagree with me on this, though.

sgrimm (commented):
Will look at this again later (I'm working on updating some map data right now) but had a few comments after an initial skim.

src/main/kotlin/com/terraformation/backend/admin/AdminDefaultVotersController.kt
tommylau523: That's right..

src/main/resources/templates/admin/defaultVoters.html
tommylau523: I am thinking of folding this into a manage voters screen. And when a user gets assigned to be default voters, instead of propagating to some subsets of projects under the hood, we will show that in the table UI directly. Better for clarity I believe. I am putting out a PR right now for that.

tommylau523:
Closing since I am combining default voter admin UI with voters admin UI

PR #1805, opened 2024-03-25
SW-5058 Implemented voter management screen and migrated default voter management

tommylau523:
View voters by projects:

Bulk Removal:

Add/remove one voter to projects:


tommylau523:
* **#1805** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1805?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523:
Moved over default voter management, and instead of guessing what logic we need to assign default voters to projects, simply display it all. 

Currently deciding whether to add default voters as columns. Could get messy if there are a lot of default voters.

src/main/kotlin/com/terraformation/backend/admin/AdminVotersController.kt
```diff
@@ -0,0 +1,251 @@
...
+    val userIdForEmail = email?.let { userStore.fetchByEmail(it)?.userId }
+
+    val projects = getAcceleratorProjects()
+    val projectVoters =
+        projects.associateBy(
+            { it.projectId },
+            { project ->
+              voteStore
+                  .fetchAllVotes(project.projectId)
+                  .filter { it.phase === project.phase }
```
sgrimm: Should be `==` instead of `===`

sgrimm: But would it make sense to have an optional phase argument for `fetchAllVotes` to do the filtering at the database level?

src/main/kotlin/com/terraformation/backend/admin/AdminVotersController.kt
```diff
@@ -0,0 +1,251 @@
...
+      @RequestParam email: String?,
+      model: Model,
+      redirectAttributes: RedirectAttributes,
+  ): String {
+    requirePermissions { updateDefaultVoters() }
+
+    val userIdForEmail = email?.let { userStore.fetchByEmail(it)?.userId }
+
+    val projects = getAcceleratorProjects()
+    val projectVoters =
```
sgrimm: Maybe throw an explicit type on here just for clarity. I think it's `Map<ProjectId, Set<UserId>>` but figuring that out takes a little effort.

src/main/kotlin/com/terraformation/backend/admin/AdminVotersController.kt
```diff
@@ -0,0 +1,251 @@
...
+    } catch (e: Exception) {
+      log.error("Failed to add voter $userId to project $projectId", e)
+      redirectAttributes.failureMessage = "Failed to add voter: ${e.message}"
+    }
+
+    return redirectToVoters(email)
+  }
+
+  @PostMapping("/voters/remove")
+  fun removeVoter(
```
sgrimm, line 91: Can this endpoint be combined with the bulk one? Seems like it could always take a list of user IDs, and for the single-user case the list will have one entry rather than multiple.

Same with the default voter remove endpoints.

src/main/kotlin/com/terraformation/backend/admin/AdminVotersController.kt
```diff
@@ -0,0 +1,251 @@
...
+        .from(PROJECTS)
+        .join(PARTICIPANTS)
+        .on(PARTICIPANTS.ID.eq(PROJECTS.PARTICIPANT_ID))
+        .join(COHORTS)
+        .on(PARTICIPANTS.COHORT_ID.eq(COHORTS.ID))
+        .leftJoin(PROJECT_VOTE_DECISIONS)
+        .on(PROJECT_VOTE_DECISIONS.PROJECT_ID.eq(PROJECTS.ID))
+        .orderBy(COHORTS.ID, PROJECTS.ID)
+        .fetch { AcceleratorProject.of(it) }
+  }
```
sgrimm: Not the worst thing ever, but I think it's generally good for database queries to live in store classes rather than in controllers just to keep the layering clean, make unit testing easier, and promote reuse.

src/main/resources/templates/admin/voters.html
```diff
@@ -0,0 +1,185 @@
...
+        </td>
+    </tr>
+    <tr th:each="project : ${projects}" class="striped">
+        <td th:text="${project.projectName}">Project Name</td>
+        <td th:text="${project.phaseName}">Project Phase</td>
+        <td th:text="${project.voteDecision ?: 'Incomplete'}">Yes</td>
+        <td>
+            <select name="userIds"
+                    th:id="|project-${project.projectId}|"
+                    th:form="|remove-${project.projectId}|"
```
sgrimm: Can this `<select>` live inside the `<form>` rather than referencing it by ID?

sgrimm (changes requested):
Ah, this is going to be much nicer than the older version. Some comments inline but I think this is definitely the right direction.

src/main/kotlin/com/terraformation/backend/admin/AdminVotersController.kt
tommylau523: I agree. I am going to refactor this into a AcceleratorService and add in any permissions needed.

src/main/kotlin/com/terraformation/backend/accelerator/AcceleratorProjectService.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.accelerator.tables.references.COHORTS
+import com.terraformation.backend.db.accelerator.tables.references.PARTICIPANTS
+import com.terraformation.backend.db.accelerator.tables.references.PROJECT_VOTE_DECISIONS
+import com.terraformation.backend.db.default_schema.tables.references.PROJECTS
+import jakarta.inject.Named
+import org.jooq.DSLContext
+
+@Named
+class AcceleratorProjectService(
```
tommylau523, line 13: We can likely replace this class with just SearchService functionalities. We will need to add Cohort and Participant search tables.

sgrimm: I think this is a bit better since (a) it gives us a strongly-typed return value which means IntelliJ can do autocomplete and such when we're editing templates, and (b) the query is more compact and less costly both to generate and to run.

sgrimm (approved):
Sweet!

PR #1818, opened 2024-03-28
Add fields missing in the project details response, fix a bug with the data store fetch one by ID

nickgraz:
- Added `createdBy`, `createdTime`, `modifiedBy`, `modifiedTime`, `projectName` to the `ProjectAcceleratorDetailsModel`
- Fixed a bug with `ProjectAcceleratorDetailsStore.fetchOneById`, it was returning multiple rows

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
@@ -41,13 +41,19 @@ class ProjectAcceleratorDetailsStore(
             landUseModelTypesMultiset,
             PROJECT_ACCELERATOR_DETAILS.asterisk(),
             PROJECTS.COUNTRY_CODE,
+            PROJECTS.CREATED_BY,
+            PROJECTS.CREATED_TIME,
             PROJECTS.ID,
+            PROJECTS.MODIFIED_BY,
+            PROJECTS.MODIFIED_TIME,
+            PROJECTS.NAME,
         )
         .from(PROJECTS)
         .leftJoin(PROJECT_ACCELERATOR_DETAILS)
         .on(PROJECTS.ID.eq(PROJECT_ACCELERATOR_DETAILS.PROJECT_ID))
         .leftJoin(COUNTRIES)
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#1818** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):
Ah, so we're not fetching the generic project info at all, just this? Do we need to include the other fields that are in the project payload (description, organization ID) too?

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
```diff
@@ -41,13 +41,19 @@ class ProjectAcceleratorDetailsStore(
...
+            PROJECTS.MODIFIED_BY,
+            PROJECTS.MODIFIED_TIME,
+            PROJECTS.NAME,
         )
         .from(PROJECTS)
         .leftJoin(PROJECT_ACCELERATOR_DETAILS)
         .on(PROJECTS.ID.eq(PROJECT_ACCELERATOR_DETAILS.PROJECT_ID))
         .leftJoin(COUNTRIES)
         .on(PROJECTS.COUNTRY_CODE.eq(COUNTRIES.CODE))
+        .where(PROJECTS.ID.eq(projectId))
```
sgrimm, line 56: The thing that bugs me most about this getting through is that it wasn't caught in a unit test.

nickgraz: I was able to reproduce by adding another project in the test, perhaps in the future we should always ensure there are other records in "fetch one" tests.

nickgraz:
> Ah, so we're not fetching the generic project info at all, just this? Do we need to include the other fields that are in the project payload (description, organization ID) too?
> 
> 

I might end up going that route... I just noticed that I also need the organization name. I'm half and half on adding another join to the store fetch, what do you think?

sgrimm:
I think there's a slightly broader discussion to be had about this; will post in the team Slack channel.

PR #1827, opened 2024-04-02
Add cohort details to GET project API

nickgraz:
- Add method to fetch cohort details for a project ID
- Insert cohort details into GET project response if the project is associated to a cohort

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
@@ -1,11 +1,11 @@
 package com.terraformation.backend.customer.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.terraformation.backend.accelerator.model.ProjectCohortData
 import com.terraformation.backend.api.CustomerEndpoint
 import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.customer.ProjectService
-import com.terraformation.backend.customer.db.ProjectCohortData
 import com.terraformation.backend.customer.db.ProjectStore
 import com.terraformation.backend.customer.model.ExistingProjectModel
 import com.terraformation.backend.customer.model.NewProjectModel
--- a/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
@@ -5,9 +5,13 @@ import com.terraformation.backend.api.CustomerEndpoint
 import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.customer.ProjectService
+import com.terraformation.backend.customer.db.ProjectCohortData
 import com.terraformation.backend.customer.db.ProjectStore
 import com.terraformation.backend.customer.model.ExistingProjectModel
 import com.terraformation.backend.customer.model.NewProjectModel
+import com.terraformation.backend.db.accelerator.CohortId
+import com.terraformation.backend.db.accelerator.CohortPhase
+import com.terraformation.backend.db.accelerator.ParticipantId
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#1827** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
```diff
@@ -138,7 +149,7 @@ data class ProjectPayload(
       modifiedTime = model.modifiedTime,
       name = model.name,
       organizationId = model.organizationId,
-  )
+      participantId = model.participantId)
```
sgrimm: Trailing comma would be good here to keep future diffs one line smaller.

src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
```diff
@@ -420,4 +421,40 @@ class ProjectStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      every { user.canReadCohort(any()) } returns false
+
+      val actual = store.fetchCohortData(projectId1)
+      assertEquals(null, actual)
+    }
+
+    @Test
+    fun `returns null if the project is not associated to a cohort`() {
+      every { user.canReadCohort(any()) } returns false
```
sgrimm: This test name seems slightly inaccurate:

- It's testing a project with no participant, rather than one with a participant that has no cohort. (Could test both of those cases.)
- The permission should probably be true, not false, so that lack of permission isn't part of what's being tested here. Permissions are covered in the previous case.

src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
```diff
@@ -420,4 +421,40 @@ class ProjectStoreTest : DatabaseTest(), RunsAsUser {
...
+
+      val actual = store.fetchCohortData(projectId)
+      assertEquals(null, actual)
```
sgrimm: Could use `assertNull()` here.

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -182,3 +201,17 @@ class ProjectStore(
...
+) {
+  companion object {
+    fun of(record: Record): ProjectCohortData {
+      return ProjectCohortData(
+          cohortId = record[COHORTS.ID]!!,
+          cohortPhase = record[COHORTS.PHASE_ID]!!,
+      )
+    }
+  }
+}
```
sgrimm: Maybe move this to a separate file in the `model` package.

sgrimm (changes requested):
Some small comments but nothing major.

src/main/kotlin/com/terraformation/backend/customer/api/ProjectsController.kt
sgrimm: Oops, you beat me to this one!

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
nickgraz: When is it okay and not? I see `data class`es in other non-model packages.

src/test/kotlin/com/terraformation/backend/customer/db/ProjectStoreTest.kt
nickgraz: I removed that permission check, updated the name of the test, and added another test

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
sgrimm: > When is it okay and not? I see data classes in other non-model packages.

Fair question, and I'm probably inconsistent about it myself. I guess the pattern the code _mostly_ follows -- and I'd be totally down with fixing this where it's not true and writing it down as a style rule if people agree -- is:

* If a class is only used within a single source file, never passed to code in another file, it's fine to colocate. Example: the payload classes for controllers.
* If there are a number of similar classes and they're all reasonably small, they can live together. Example: the exception classes in the various `Exceptions.kt` files.

sgrimm: I guess a third rule, or an exception, would be that the first two rules are relaxed for tests.

sgrimm (approved):
Looks good!

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
nickgraz: Excellent, I think those rules make a lot of sense, thanks!

PR #1834, opened 2024-04-05
SW-3683 Remove /api/v1/seedbank/values/all

sgrimm:
The seedbank API has two endpoints to list all the values of a field. One of
them lists all the values that are currently present on any accessions, and
the other lists all possible values, even if they're not used.

The web and mobile apps no longer use the "list all possible values" endpoint,
so there's no need to maintain it. Removing it means we can remove a method
from `SearchTable` and its subclasses that only existed to support that endpoint.


sgrimm:
* **#1834** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

PR #1838, opened 2024-04-10
SW-5179 Created Atlassian Client for support tickets

tommylau523:
JIRA is no longer supporting a client library so we implemented a thin HTTP layer using ktor to submit REST APIs.

An AtlassianResource interface is used to build the request including path, header and request body and each endpoint will implement this interface. This is to maintain strong typing and ensure that the payload/request are correct.

The tests should demonstrate the use cases of this client.


tommylau523:
* **#1838** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1838?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

ATLASSIAN.md
```diff
@@ -0,0 +1,78 @@
+# Configuring support tickets with Atlassian
+
+Feature requests and bug reports submitted are triaged as JIRA service issues on Atlassian. This requires configuring terraware-server with the right Atlassian API credentials. 
+
+## Prerequisites
+
+- Atlassian Cloud site with JIRA
```
sgrimm: FYI, the official spelling is "Jira" rather than "JIRA" -- it used to be all caps but they changed it a while back.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/CreateServiceDeskRequest.kt
```diff
@@ -0,0 +1,116 @@
...
+      setBody(body)
+    }
+  }
+
+  override suspend fun parseResponse(response: HttpResponse): PostServiceDeskRequestResponse {
+    return response.body()
+  }
+
+  /** A builder class to inject configs or build information into the resource. */
+  @Named
```
sgrimm: This will cause the class to be instantiated at application start and made available for dependency injection, which we don't want for a builder class; every caller is going to want a fresh new instance of the builder.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/** Submits API requests to interact with Atlassian services. */
+@Named
+class AtlassianHttpClient(private val config: TerrawareServerConfig) {
+  private val httpClient: HttpClient by lazy { createHttpClient() }
+
+  suspend fun <T> makeRequest(resource: AtlassianResource<T>): T {
+    val response = httpClient.request { resource.buildRequest(this) }
+    return resource.parseResponse(response)
+  }
```
sgrimm: The Ktor client uses suspend functions, but in our application code we generally stick to synchronous functions. The `runBlocking` could live here instead of the callers having to do it.

My assumption is that we're not going to be making tons of concurrent requests such that we'd need to worry about the thread count exploding.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/CreateServiceDeskRequest.kt
```diff
@@ -0,0 +1,116 @@
...
+  class Builder(private val config: TerrawareServerConfig) {
```
sgrimm: Can we get away with not having a builder class at all?

In Java we needed this kind of thing thanks to the lack of named parameters, but this could be a secondary constructor on the outer class to do the request type mapping, plus an `init` block to do the validation checks.

The big advantage is you get compile-time checking of required arguments. WIth a builder, forgetting to set one of the mandatory values is a runtime error (the `!!` will throw an exception) rather than a compile failure.

sgrimm (changes requested):
I like the detailed README here! Some comments inline.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/CreateServiceDeskRequest.kt
tommylau523: I agree. I was struggling a bit to inject TerrawareServer config (and maybe sometime in the future things like BuildVersion) into the request constructor. I do think Builder is not the correct pattern. I am leaning towards just a RequestFactory class. Will follow up in the next revision and get your feedback on that.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/AtlassianResourceFactory.kt
```diff
@@ -0,0 +1,31 @@
+package com.terraformation.backend.support.atlassian.resource
+
+import com.terraformation.backend.config.TerrawareServerConfig
+import com.terraformation.backend.support.atlassian.SupportRequestType
+import javax.inject.Named
+
+/** A factory class to inject configs or build information into the resource. */
+@Named
+class AtlassianResourceFactory(private val config: TerrawareServerConfig) {
```
sgrimm: This is definitely nicer than the builder class.

What will the call sites look like? If it's always going to be the equivalent of `client.makeRequest(factory.createServiceDeskRequest(...))` like in the test cases, and the callers will never do anything with these objects other than immediately pass them to `AtlassianHttpClient.makeRequest`, then would it make sense to turn this class into a service that actually sends the request and returns the response? Then callers don't need to have both this class and `AtlassianHttpClient` as dependencies, and the request classes become internal implementation details that callers never need to see.

tommylau523: This makes a lot of sense! I originally wanted to split up the service layer and the http client layer. The caller in this case would be a `SupportService` class. But I agree with you it just ends up complicating things. I will keep working on this PR to include permissions and use this as the service layer too.

tommylau523:
@sgrimm Another thing I will need to figure out, is to configure the necessary envs for our CI pipeline, staging and prod env. 

The latter 2 can happen when the entire feature is completed and API controllers implemented. 

The CI one would be a really nice to have to track regressions and such.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -106,6 +109,43 @@ class TerrawareServerConfig(
...
+      @DefaultValue("false") val enabled: Boolean = false,
+
+      /** Jira Bug Report type ID */
+      val bugReportTypeId: Int? = null,
+
+      /** Jira Feature Request type ID */
+      val featureRequestTypeId: Int? = null,
+
+      /** Service Desk ID */
+      val serviceDeskId: Int? = null,
```
sgrimm: Could we look these IDs up at runtime (say, with a `by lazy` function so we only need to do it once per server process) rather than requiring them to be configured? Maybe have config options for the type _names_ with defaults of `"Bug Report"` and `"Feature Request"`.

Obviously configuring the IDs is also workable; I'm just thinking about how to reduce the number of setup steps and the amount of config we need for local dev and for CI.

src/main/kotlin/com/terraformation/backend/config/TerrawareServerConfig.kt
```diff
@@ -106,6 +109,32 @@ class TerrawareServerConfig(
...
+      val apiHostname: String? = null,
+
+      /** Atlassian api key */
+      val apiToken: String? = null,
+
+      /** Enabled flag */
+      @DefaultValue("false") val enabled: Boolean = false,
+
+      /** Service Desk Tag */
+      val serviceDeskKey: String? = null,
```
sgrimm: Minor nit: The comment says this is a tag, but the variable name says it's a key.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -273,6 +273,8 @@ data class IndividualUser(
   override fun canDeleteSubLocation(subLocationId: SubLocationId) =
       isAdminOrHigher(parentStore.getFacilityId(subLocationId))
 
+  override fun canDeleteSupportIssue(): Boolean = false
```
sgrimm, line 277: So no human user can delete issues, not even super-admins?

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+            serviceDeskId = serviceDeskId(),
+        ))
+  }
+
+  private fun serviceDeskId(): Int =
+      if (serviceDesk == null) {
+        throw IllegalStateException("Atlassian configuration is invalid")
+      } else {
+        serviceDesk!!.id
+      }
```
sgrimm: Would it make sense to make `serviceDesk` non-nullable and throw this exception in `findServiceDesk()` instead? Then this method can go away and callers can access `serviceDesk.id` directly.

src/main/kotlin/com/terraformation/backend/support/atlassian/model/ServiceDeskModel.kt
```diff
@@ -0,0 +1,10 @@
+package com.terraformation.backend.support.atlassian.model
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ServiceDeskModel(
+    val id: Int,
+    val projectName: String,
```
sgrimm: Is this used?

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/AtlassianResource.kt
```diff
@@ -0,0 +1,17 @@
...
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.client.statement.HttpResponse
+
+/**
+ * Every Atlassian API call will implement this interface. The intended workflow is that every time
+ * an API call is made, an instance of this interface will be created. The [AtlassianHttpClient]
+ * will invoke the [buildRequest] method so the caller will not be able to erroneously or
+ * maliciously set the request body.
+ */
+interface AtlassianResource<T> {
```
sgrimm: From the names, it wasn't clear to me what the difference between a "resource" and a "model" was. Would it make sense to rename this to `AtlassianRequest` and rename its package to `request` given that the word "request" is used in a couple of the implementation classes already? And then maybe make sure all the implementation names end with "Request" for consistency.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+import jakarta.inject.Named
+import kotlinx.coroutines.runBlocking
+
+/** Submits API requests to interact with Atlassian services. */
+@Named
+class AtlassianHttpClient(private val config: TerrawareServerConfig) {
+  private val httpClient: HttpClient by lazy { createHttpClient() }
+  private val serviceDesk: ServiceDeskModel? by lazy { findServiceDesk() }
+
+  val requestTypes: Set<ServiceRequestTypeModel> by lazy { getServiceRequestTypes() }
```
sgrimm, line 34: Could this be `Map<Int, ServiceRequestTypeModel>` or even `Set<Int>` instead? Or will it eventually be used for more than just validating that the request type ID exists?

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    val response = runBlocking {
+      val httpResponse = httpClient.request { resource.buildRequest(this) }
+      resource.parseResponse(httpResponse)
+    }
+    return response
+  }
+
+  private fun createHttpClient(): HttpClient {
+    return HttpClient(Java) {
```
sgrimm, line 81: Could check the `enabled` flag here instead of in `makeRequest`. It can't change on the fly so checking it once should be sufficient.

src/main/kotlin/com/terraformation/backend/support/atlassian/model/ServiceRequestFieldsModel.kt
```diff
@@ -0,0 +1,9 @@
+package com.terraformation.backend.support.atlassian.model
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
sgrimm, line 19: These model classes could use one-line KDoc comments to say what they're for. At first glance, it's not obvious what the difference between `ServiceRequestFieldsModel` and `ServiceRequestTypeModel` is other than that one has a name instead of a summary and one has an ID.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/CreateServiceRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * The POST request to create a service ticket.
+ *
+ * @param requestFieldValues the title of the Jira issue
+ * @param requestTypeId the ID of the Jira issue type
+ * @param reporter the email address of the reporter
+ * @param serviceDeskId the ID for the service desk
+ */
+class CreateServiceDeskRequest(
```
sgrimm: Minor nit: Filename doesn't match this class name. (Technically not required, of course, but the names are close enough that it seems like they were intended to be the same.)

sgrimm (changes requested):
The new config setup seems much easier to deal with; nice. Some feedback inline but this is definitely shaping up.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
tommylau523: Yes. I think the intention is that if a human needs to delete a support issue, they can do so on the Jira page itself, instead of going through an API. The only use case of deleting an issue is simply for cleaning up tests for now. 

Down the line we can maybe expose ways to transition a ticket to close or complete. Even then I am guessing there is not many use cases, perhaps for a user to cancel a support request.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
tommylau523: I am hoping that this will cause an error during a service call time, instead of during server initialization time. I don't want our whole server to be not running because the Atlassian API is unreachable for example. 

Perhaps I am overthinking it and not understanding the Spring @named components lifecycle well.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
tommylau523: Instead of hard coding the request types, I am planning on exposing the allowed ticket types and descriptions via an API. Our support forms can populate the support type and description.

tommylau523: So I am imagining there will be a GET /support that will return the list of support request types configured on the Jira, and PUT /support that will create an issue. Routes are subject to changed.

src/main/kotlin/com/terraformation/backend/support/atlassian/resource/AtlassianResource.kt
tommylau523: Was hoping to follow Jira vocabs, but even they are quite inconsistent. I do think requests or httpRequest is a better name. Will rename for consistency.

src/main/kotlin/com/terraformation/backend/support/atlassian/model/ServiceDeskModel.kt
tommylau523: Will remove for now. I don't think we will need it.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
sgrimm: The `by lazy` is what stops that from happening. The value won't be initialized until it's referenced, which won't happen until something calls into the class. Net effect should be the same as what's in the PR right now: you get an exception at runtime when you call one of the methods that makes a request.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  fun createServiceDeskRequest(
+      description: String,
+      summary: String,
+      requestTypeId: Int,
+      reporter: String,
+  ): PostServiceDeskRequestResponse {
+    // No required permission
+    if (requestTypes.find { it.id == requestTypeId } == null) {
```
sgrimm: Could be `requestTypes.none { it.id == requestTypeId }`.

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
tommylau523: Thanks!!

src/main/kotlin/com/terraformation/backend/support/atlassian/AtlassianHttpClient.kt
```diff
@@ -0,0 +1,103 @@
...
+
+  fun createServiceDeskRequest(
+      description: String,
+      summary: String,
+      requestTypeId: Int,
+      reporter: String,
+  ): PostServiceDeskRequestResponse {
+    // No required permission
+    if (requestTypes.none { it.id == requestTypeId }) {
+      throw RuntimeException("Request ID type not recognized")
```
sgrimm: `IllegalArgumentException` would be a more descriptive exception to throw here.

src/main/kotlin/com/terraformation/backend/support/atlassian/request/ListServiceRequestTypesHttpRequest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.terraformation.backend.support.atlassian.model.ServiceRequestTypeModel
+import io.ktor.client.call.body
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.client.statement.HttpResponse
+import io.ktor.http.HttpMethod
+import io.ktor.http.path
+
+class ListServiceRequestTypes(serviceDeskId: Int) :
```
sgrimm: `ListServiceRequestTypesHttpRequest`?

sgrimm (approved):
Could use a couple minor tweaks before merging, but this seems good otherwise.

.github/workflows/main.yml
```diff
@@ -112,6 +112,10 @@ jobs:
       if: env.IS_CD == 'false'
       run: ./gradlew test --tests='*ExternalTest'
```
tommylau523, line 113: @sgrimm I can't seem to get the ExternalTest to work. Even though I can run the tests successfully if I set the env variables locally.

PR #1840, opened 2024-04-11
SW-5099 Added data tables for modules and events

tommylau523:


tommylau523:
* **#1840** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/db/migration/0250/V259__Modules.sql
```diff
@@ -0,0 +1,39 @@
...
+    end_time TIMESTAMP WITH TIME ZONE,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+
+    UNIQUE(id, module_id, event_type_id)
+);
+
+CREATE TABLE accelerator.event_projects (
```
tommylau523: One thing I did not add here is a constraint to make sure a project is not scheduled for multiple events of the same type in the same module.

i.e. it is possible for a project to be in two 1:1 sessions in one module for example. 

We can check for this on the service/controller level I believe. Also give us flexibility if any project does need to be in multiple events (maybe they missed a workshop and joined one later).

src/main/resources/db/migration/0250/V259__Modules.sql
```diff
@@ -0,0 +1,39 @@
+ALTER TABLE accelerator.modules ADD COLUMN overview TEXT;
+ALTER TABLE accelerator.modules ADD COLUMN preparation_materials TEXT;
+ALTER TABLE accelerator.modules ADD COLUMN live_session_description TEXT;
+ALTER TABLE accelerator.modules ADD COLUMN workshop_description TEXT;
+ALTER TABLE accelerator.modules ADD COLUMN private_session_description TEXT;
+ALTER TABLE accelerator.modules ADD COLUMN additional_resources TEXT;
+
+ALTER TABLE accelerator.cohort_modules ADD COLUMN phase_id INTEGER REFERENCES accelerator.cohort_phases;
```
sgrimm: Should this live in `modules` instead of `cohort_modules`? Or do we need to support the same module being in phase 1 for one cohort and phase 2 for a different cohort?

src/main/resources/db/migration/0250/V259__Modules.sql
```diff
@@ -0,0 +1,39 @@
...
+    UNIQUE (name)
+);
+
+CREATE TABLE accelerator.events (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    module_id  BIGINT NOT NULL REFERENCES accelerator.modules ON DELETE CASCADE,
+    event_type_id INTEGER NOT NULL REFERENCES accelerator.event_types,
+    meeting_link TEXT,
+    slides_link TEXT,
+    recording_link TEXT,
```
sgrimm: If you use `_url` instead of `_link` as the suffix, these will turn into `URI` objects in the generated classes (rather than strings).

src/main/resources/db/migration/0250/V259__Modules.sql
sgrimm: I suspect you're right that people will occasionally want to be able to put a project in multiple events of the same type, so this seems fine and we might not even want to check for it in the application code.

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -107,6 +107,12 @@ VALUES (1, 'Boreal forests/Taiga'),
        (14, 'Tundra')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
+INSERT INTO accelerator.event_types (id, name)
+VALUES (1, 'Private Session'),
+       (2, 'Workshop'),
+       (3, 'Group Session')
```
sgrimm: I don't recognize the terms "private session" and "group session" -- did we change the terminology? If not, I think we should stick with the terminology the accelerator team is using. "1-on-1" can be "One-on-One" so the enum value doesn't start with a digit.

src/main/resources/db/migration/0250/V259__Modules.sql
tommylau523: Let me get product to answer that. I made this assumption because of one comment I saw a while back.

src/main/resources/db/migration/R__TypeCodes.sql
tommylau523: Yes should have been live session. I was trying to get around 1-on-1 but your method makes more sense!

src/main/resources/db/migration/0250/V259__Modules.sql
```diff
@@ -0,0 +1,39 @@
...
+    slides_url TEXT,
+    recording_url TEXT,
+    start_time TIMESTAMP WITH TIME ZONE,
+    end_time TIMESTAMP WITH TIME ZONE,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+
+    UNIQUE(id, module_id, event_type_id)
```
sgrimm: Is this needed? ID is already unique and there isn't a foreign key reference to these three columns.

src/main/resources/db/migration/0250/V259__Modules.sql
```diff
@@ -0,0 +1,39 @@
...
+);
+
+CREATE TABLE accelerator.event_projects (
+    event_id BIGINT NOT NULL REFERENCES accelerator.events ON DELETE CASCADE,
+    project_id BIGINT REFERENCES projects ON DELETE CASCADE,
+    PRIMARY KEY (event_id, project_id)
+);
```
sgrimm: A trailing newline would be good just so editors/formatters don't add one later on and cause the file to show up as edited.

src/main/resources/db/migration/0250/V259__Modules.sql
tommylau523: Not needed at all. I was experimenting earlier with setting the constraint mentioned in another comment. My mistake on this oversight.

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -468,7 +468,13 @@ COMMENT ON COLUMN accelerator.deliverables.is_sensitive IS 'If true, the data us
 
 COMMENT ON TABLE accelerator.document_stores IS '(Enum) Locations where uploaded documents are stored.';
 
-COMMENT ON TABLE accelerator.modules IS 'Possible steps in the workflow of a cohort.';
+COMMENT ON TABLE accelerator.events IS 'Events with meeting links and time within an acclerator module.';
```
sgrimm: Minor nit: `events` should sort below `event_`.

sgrimm (changes requested):
I merged #1842 which included a new migration, so the migration here will need to be renamed to V260.

sgrimm (approved):

tommylau523:
### Merge activity

* **Apr 15, 1:44 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1840).

PR #1853, opened 2024-04-15
Add Spring Boot Actuator dependency

sgrimm:
JobRunr 7 removed the dependency on Spring Boot's starter bundle for
Actuator, which is used for load balancer health checks; we were relying on that
transitive dependency. Add it as an explicit dependency.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -81,6 +81,7 @@ dependencies {
   // kapt("org.springframework.boot:spring-boot-configuration-processor")
 
   implementation("org.springframework.boot:spring-boot-starter-web")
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#1853** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm:
This is to fix the "no healthy hosts" alert we're getting from the load balancer.

tommylau523 (approved):

PR #1855, opened 2024-04-16
SW-5123 Modules Store

tommylau523:
Skeleton structure for stores and permissions for modules

Implemented datetime constraints on tables, and module store


tommylau523:
* **#1864** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1855** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1855?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ModulesController.kt
```diff
@@ -0,0 +1,3 @@
+package com.terraformation.backend.accelerator.api
+
+class ModulesController {}
```
sgrimm: Is it useful to merge this yet? Could wait to create it until it has some contents.

src/main/kotlin/com/terraformation/backend/customer/model/TerrawareUser.kt
```diff
@@ -230,6 +230,8 @@ interface TerrawareUser : Principal {
 
   fun canReadInternalTags(): Boolean
 
+  fun canReadModulesForProject(projectId: ProjectId): Boolean
```
sgrimm: This name feels a little inconsistent with the other permission names to me. What would you think of `canReadProjectModules`, to follow the same pattern as `canReadProjectVotes` and friends?

src/test/kotlin/com/terraformation/backend/accelerator/db/DeliverableStoreTest.kt
```diff
@@ -97,9 +98,9 @@ class DeliverableStoreTest : DatabaseTest(), RunsAsUser {
       val deliverableId3 = insertDeliverable(id = 3, moduleId = 2)
       insertDeliverableDocument(templateUrl = "https://example.com/")
 
-      insertCohortModule(cohortId1, moduleId1)
-      insertCohortModule(cohortId1, moduleId2)
-      insertCohortModule(cohortId2, moduleId2)
+      insertCohortModule(cohortId1, moduleId1, LocalDate.of(2024, 1, 1), LocalDate.of(2024, 1, 31))
```
sgrimm: Would it make sense to add a default value for the end date in `insertCohortModule`, say 6 days after the start date so the module is a week long, just to simplify call sites like this where we don't care about the specific dates?

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -2066,6 +2086,46 @@ abstract class DatabaseTest {
     cohortModulesDao.insert(row)
   }
 
+  fun insertEvent(
+      id: Any? = null,
+      moduleId: Any = inserted.moduleId,
+      eventType: EventType = EventType.Workshop,
+      meetingUrl: URI? = null,
+      slidesUrl: URI? = null,
+      recordingUrl: URI? = null,
```
sgrimm: Might be convenient to make these `Any?` and then do `meetingUrl?.let { URI("$it") }` so callers can pass these as strings.

src/main/kotlin/com/terraformation/backend/accelerator/model/ModuleModel.kt
```diff
@@ -0,0 +1,80 @@
...
+          name = record[MODULES.NAME]!!,
+          phase = record[MODULES.PHASE_ID]!!,
+          additionalResources = record[MODULES.ADDITIONAL_RESOURCES],
+          cohorts =
+              listOf(
+                  CohortModuleModel(
+                      record[COHORT_MODULES.COHORT_ID]!!,
+                      record[COHORT_MODULES.START_DATE]!!,
+                      record[COHORT_MODULES.END_DATE]!!,
+                  )),
```
sgrimm: At the risk of overengineering, I wonder if, to reduce duplication, this could be structured as something like
```kotlin
private fun of(
    record: Record,
    eventsField: Field<Map<EventType, List<EventModel>>>,
    getCohorts: (Record) -> List<CohortModuleModel>
): ModuleModel {
  return ModuleModel(
      id = record[MODULES.ID]!!,
      // ...
      cohorts = getCohorts(record),
      // ...
    )
}
```
and then the two externally-visible `of` methods only differ in how to get the cohort list, like:
```kotlin
fun of(
    record: Record,
    eventsField: Field<Map<EventType, List<EventModel>>>,
): ModuleModel {
  return of(record, eventsField) {
    CohortModuleModel(
        it[COHORT_MODULES.COHORT_ID]!!,
        it[COHORT_MODULES.START_DATE]!!,
        it[COHORT_MODULES.END_DATE]!!,
    )
}
```

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleStore.kt
```diff
@@ -0,0 +1,108 @@
...
+      projectId: ProjectId? = null
+  ): Field<Map<EventType, List<EventModel>>> {
+    val projectCondition = projectId?.let { EVENT_PROJECTS.PROJECT_ID.eq(it) }
+    val projectsField =
+        if (projectId == null) {
+          with(EVENT_PROJECTS) {
+            DSL.multiset(DSL.select(PROJECT_ID).from(this).where(EVENT_ID.eq(EVENTS.ID)))
+                .convertFrom { result -> result.map { it.value1() }.toSet() }
+          }
+        } else null
```
sgrimm: Braces around the `null` here would make it clearer this is a two-branch `if` expression.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleStore.kt
```diff
@@ -0,0 +1,108 @@
...
+    return with(EVENTS) {
+      DSL.multiset(
+              DSL.selectDistinct(asterisk(), projectsField)
```
sgrimm: Instead of the "distinct" here (which might be a little expensive on a multiset, though admittedly the number of rows we're dealing with will be small), how about a "where in" structure, like
```kotlin
DSL.select(asterisk(), projectsField)
    .from(this)
    .where(ID.`in`(
        DSL.select(EVENT_PROJECTS.EVENT_ID)
            .from(EVENT_PROJECTS)
            .where(projectCondition)))
    .and(MODULE_ID.eq(MODULES.ID))
    .orderBy(START_TIME)
```
That's just off the top of my head, so it might not be quite right, but the idea is to move the multiset construction to after the row filtering.

sgrimm (changes requested):
Some comments inline, but the main thing is that I think it'd be good to have some test coverage of the fetch functions since they're doing nontrivial stuff with subqueries and it'd be easy to get things slightly wrong and return duplicate rows and such.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleStore.kt
tommylau523: This was really helpful! Thanks! I was out of ideas on how to get around the join step which will return multiple rows.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
```diff
@@ -539,6 +539,17 @@ internal class PermissionRequirementsTest : RunsAsUser {
 
   @Test fun readInternalTags() = allow { readInternalTags() } ifUser { canReadInternalTags() }
 
+  @Test
+  fun readModulesForProject() {
```
sgrimm: This should be `readProjectModules` now.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
```diff
@@ -0,0 +1,1210 @@
...
+    }
+
+    @Test
+    fun `returns empty list of modules if module not associated with cohort`() {
+      insertModule()
+      assertEquals(emptyList<ModuleModel>(), store.fetchModulesForProject(projectId))
+    }
+
+    @Test
+    fun `returns list of modules with null cohorts for module associated with cohort`() {
```
sgrimm: This doesn't seem to be looking for null cohorts. Would a better name be something like `does not return projects list for cohort`?

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
```diff
@@ -0,0 +1,1210 @@
...
+                      ),
+                  eventSessions = emptyMap(),
+                  overview = "<h> Overview </h>",
+                  preparationMaterials = "<i> Preps </i>",
+              )),
+          store.fetchModulesForProject(projectId))
+    }
+
+    @Test
+    fun `returns list of modules according to project scopes`() {
```
sgrimm: Is the previous test case needed given this one exists? It seems like this asserts the same thing the previous test does, but also verifies that it works if there are multiple cohorts.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
```diff
@@ -0,0 +1,1210 @@
...
+                      mapOf(
+                          EventType.LiveSession to "Live session lectures",
+                          EventType.OneOnOneSession to "1:1 meetings",
+                          EventType.Workshop to "Workshop ideas",
+                      ),
+                  eventSessions = emptyMap(),
+                  overview = "<h> Overview </h>",
+                  preparationMaterials = "<i> Preps </i>",
+              )),
+          store.fetchAllModules())
```
sgrimm, line 225: It seems like this exact `insertModule()` call is happening a lot in this test class, and then there are a lot of `assertEquals` with the exact same model aside from the ID. 

Could we extract this insert call into a helper method, and also have a helper method that takes a module ID and  returns a model with the default values from the insert call? Then tests can either use that model verbatim, or call its `copy()` method to tweak the values as needed.

Alternately, maybe a small number of test cases could populate all the optional fields, just enough to make sure all the fields are being populated from the right columns, and they could be left out (defaulting to null) in most of the tests.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
```diff
@@ -0,0 +1,1210 @@
...
+                  id = otherModuleId,
+                  name = "OtherTestModule",
+                  phase = CohortPhase.Phase0DueDiligence,
+              ),
+          ),
+          store.fetchAllModules())
+    }
+
+    @Test
+    fun `returns list of modules with multiple modules and multiple cohorts`() {
```
sgrimm: This could test the "module is used by more than one cohort" case as well if one of the modules is in both cohorts and the other is only in one cohort.

At which point the test method above this one could probably go away since this covers that functionality too.

sgrimm (changes requested):
The comprehensive test coverage here is great, but I think the tests are going to be kind of hard to maintain just because there's a fair amount of overlapping coverage and they're a bit on the verbose side. A couple suggestions inline to maybe make the test code more succinct.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
tommylau523: I went with the latter approach. The test cases now explicitly look and check for specific fields. I think it ends up being more readable, and describes the tests better.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleStore.kt
```diff
@@ -0,0 +1,114 @@
+package com.terraformation.backend.accelerator.db
```
tommylau523, line 1: Looking at the rest of the package, I believe it is better to move all of modules/events stuff out of the accelerator package, since it is focused on consoles.

Perhaps we can create a `com.terraformation.backend.participant` package, for the project-side of accelerator services.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
```diff
@@ -0,0 +1,433 @@
...
+      val otherProjectId = insertProject(participantId = otherParticipantId)
+
+      insertCohortModule(cohortId, moduleId)
+      insertCohortModule(otherCohortId, otherModuleId)
+
+      val fetchModules = store.fetchModulesForProject(projectId)
+      val otherFetchModules = store.fetchModulesForProject(otherProjectId)
+
+      assertEquals(fetchModules.map { it.id }, listOf(moduleId))
+      assertEquals(otherFetchModules.map { it.id }, listOf(otherModuleId))
```
sgrimm: I think the arguments are swapped here.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleStoreTest.kt
sgrimm: Much easier to follow now, thanks.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -2063,9 +2085,50 @@ abstract class DatabaseTest {
             endDate = endDate,
         )
 
+    nextCohortModuleStartDate = endDate.plusDays(1)
```
sgrimm, line 2088: Would it make sense to add 7 days instead of 1 day here, so you can insert two cohort modules in a row using default start/end dates without overlaps?

sgrimm (approved):
One minor thing to fix before merging, and one suggestion you can take or leave, but this looks good!

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
tommylau523: This uses end date, so it actually will never overlap even if we inputted a custom end date, then go back to using defaults

sgrimm: Oops, right you are! Ignore my comment, then.

tommylau523:
### Merge activity

* **Apr 17, 7:33 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1855).

PR #1856, opened 2024-04-16
SW-5187 Add undo dates to search API

sgrimm:
In the web app, we need to show the date of the original withdrawal as part of the
withdrawal log row for an undo operation. Currently, the client has to request the
dates from the server after getting the initial set of search results. Add fields
for the dates on both sides of an undo operation to avoid the extra API calls.


sgrimm:
* **#1856** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/db/migration/0250/V261__WithdrawalUndoDates.sql
```diff
@@ -0,0 +1,48 @@
+DROP VIEW nursery.withdrawal_summaries;
+
+-- Add undoes_withdrawal_date and undone_by_withdrawal_date
```
tommylau523, line 3: I am wondering if we should include the old migration file name in a comment so it is easier to find the original version for diffs (this one was easy to find!)

sgrimm: Can't hurt!

PR #1859, opened 2024-04-16
SW-5215 Fetch organization user's created_time when querying for organization users

karthikbtf:
- previously was looking up the user's created time
- updated code to fetch the org user's created time (date/time when user was added to the org)
- updated existing tests to capture this scenario

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
@@ -957,7 +957,7 @@ abstract class DatabaseTest {
           .set(CREATED_BY, createdBy)
           .set(CREATED_TIME, createdTime)
           .set(MODIFIED_BY, createdBy)
-          .set(MODIFIED_TIME, Instant.EPOCH)
+          .set(MODIFIED_TIME, createdTime)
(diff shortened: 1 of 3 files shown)


src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
```diff
@@ -949,12 +949,13 @@ abstract class DatabaseTest {
...
+      createdTime: Instant = Instant.EPOCH,
   ) {
     with(ORGANIZATION_USERS) {
       dslContext
           .insertInto(ORGANIZATION_USERS)
           .set(CREATED_BY, createdBy)
-          .set(CREATED_TIME, Instant.EPOCH)
+          .set(CREATED_TIME, createdTime)
           .set(MODIFIED_BY, createdBy)
           .set(MODIFIED_TIME, Instant.EPOCH)
```
sgrimm: Should set modified time to the created time too.

sgrimm (approved):
One little thing to tweak, but looks good otherwise.

src/test/kotlin/com/terraformation/backend/db/DatabaseTest.kt
karthikbtf: Oops, yes good catch.

PR #1864, opened 2024-04-17
SW-5210 Added project module controller

tommylau523:


tommylau523:
* **#1864** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1855** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1855?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -0,0 +1,106 @@
...
+      @PathVariable("projectId") projectId: ProjectId,
+  ): GetProjectModulesResponsePayload {
+    val models = moduleStore.fetchModulesForProject(projectId)
+    return GetProjectModulesResponsePayload(models)
+  }
+}
+
+data class ProjectModuleEventSession(
+    val id: EventId,
+    // Times
```
sgrimm: Are the comments like this one just to split the fields up into sections visually? I'm not sure they're adding any information; hopefully it's self-evident that `startTime` and `endTime` are times.

If we want to document what the fields are, we could add `@Schema` annotations so the docs appear in the OpenAPI schema rather than just the source code. Though again, I think most of the field names in these payloads are pretty self-explanatory.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -0,0 +1,106 @@
...
+            ProjectModuleEvent(
+                it.value,
+                model.eventSessions[it.key]?.map { event -> ProjectModuleEventSession(event) }
+                    ?: emptyList(),
+            )
+          })
+}
+
+data class GetProjectModulesResponsePayload(
+    val modules: Map<CohortPhase, List<ProjectModule>> = emptyMap(),
```
sgrimm: Grouping the modules by phase here, as opposed to returning the results as a flat list, seems like it's pretty specific to a particular client-side usage pattern. If a client wants to iterate through all the modules across all phases or group them by something other than phase, it has to essentially undo the grouping in the API response, which is more of a pain than doing the grouping on the client side when it's needed.

If the client only cares about a single phase, maybe we could make the endpoint take an optional phase argument to limit the response to just that phase's modules.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -0,0 +1,97 @@
...
+}
+
+data class ProjectModuleEvent(
+    val eventDescription: String,
+    val sessions: List<ProjectModuleEventSession>,
+)
+
+data class ProjectModule(
+    val id: ModuleId,
+    val name: String,
```
nickgraz, line 67: Do we also need a "module title" or something like that? We need "Module 1" and "Kick-off" in the frontend.

tommylau523: I was thinking that the Module X part comes from the ordering of modules. "Kick-off" is the name in this case. 

The query for project modules currently sort modules by startDate. And the table is constraint so that there is no overlapping dates. 

We may want to add an isActive boolean that comes from the backend.

tommylau523: And any given module can be in multiple cohorts, in any order. So we can't hard-wire the Module X title either.

nickgraz: Would they be in multiple cohorts? How would work with the dates?

tommylau523: We have a table modules, and cohort_modules that associate modules with cohorts

modules:
- name
- description
- html text 
etc. most of the contents

cohort_modules:
- module_id
- cohort_id
- start_date
- end_date (constraint date collisions and end_date > start_date)

tommylau523: And adding to before, the isActive boolean will be there to make sure we are respecting server clock and dates as opposed to browser clock which can be set to anything.

nickgraz: Oh I thought the content of the modules could change from cohort to cohort

tommylau523: We are going with every time a module changes it becomes a brand new module.

tommylau523: We discussed with product and they didn't want accelerator admins to deal with the complexity of content management. So if module has any variations, it is a brand new module.

nickgraz: Sounds good, thanks for the info

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+data class ProjectModuleEventSession(
+    val id: EventId,
+    val startTime: Instant?,
+    val endTime: Instant?,
+    val meetingURL: URI?,
+    val recordingURL: URI?,
+    val slidesURL: URI?,
```
sgrimm: We use `fooUrl` instead of `fooURL` elsewhere in the API, so probably good to rename these for consistency.

sgrimm (approved):
Looks good -- just one minor thing to change before merging.

tommylau523:
### Merge activity

* **Apr 19, 3:24 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1864).

PR #1866, opened 2024-04-18
SW-5205 Implemented module event notifications

tommylau523:


tommylau523:
* **#1877** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1866** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/EventNotifier.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val log = perClassLogger()
+  }
+
+  /** Schedules an "event is starting" notification according to preset time. */
+  @EventListener
+  fun on(event: ModuleEventScheduledEvent) {
+    val moduleEvent = moduleStore.fetchEventById(event.eventId)
+    if (moduleEvent?.startTime != null) {
+      val notifyTime = moduleEvent.startTime.minus(notificationLeadTime)
+      if (notifyTime.isAfter(clock.instant())) {
```
tommylau523: Per requirement in design, we are choosing to not show notifications if event changes is happening within the next 15 minutes. 

Likely accelerator admins will be warned before making changes to events too soon or in the past (in the admin UI), and will be responsible for contacting attendants outside of Terraware should that happen.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -337,6 +341,27 @@ class AppNotificationService(
     }
   }
 
+  @EventListener
+  fun on(event: ModuleEventStartingEvent) {
+    val moduleEvent = moduleStore.fetchEventById(event.eventId)
+    if (moduleEvent != null && event.startTime == moduleEvent.startTime) {
```
tommylau523: This check may cause multiple notifications to fire still, if an event changed from an original time, to a different time, then back to an original time. 

One potential fix is to keep a version number in the database, and increment it every time event time is updated.

tommylau523: But at that point, it may be better to delete the scheduled job and add a new one, and save the job ID in the table.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventProjectsHelper.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.accelerator.tables.references.EVENTS
+import com.terraformation.backend.db.accelerator.tables.references.EVENT_PROJECTS
+import com.terraformation.backend.db.default_schema.ProjectId
+import org.jooq.Field
+import org.jooq.impl.DSL
+
+class ModuleEventProjectsHelper {
+  fun eventProjectsMultiset(): Field<Set<ProjectId>> {
```
sgrimm: Could this be a top-level function or even a top-level constant?

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventProjectsHelper.kt
```diff
@@ -0,0 +1,20 @@
...
+    requirePermissions { readModuleEventParticipants() }
```
sgrimm: This is subjective, but this function feels like it's too low-level for an exception-throwing permission check. I'm having a hard time articulating why, though. Maybe because it doesn't actually do any work, just returns a fragment of SQL. So I guess consider this a weak bit of feedback to ask if moving this permission check to the caller would be reasonable.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.springframework.context.ApplicationEventPublisher
+
+@Named
+class ModuleEventStore(
+    private val clock: InstantSource,
+    private val dslContext: DSLContext,
+    private val eventPublisher: ApplicationEventPublisher,
+    private val eventsDao: EventsDao,
+) {
+  private val eventProjectsHelper: ModuleEventProjectsHelper by lazy { ModuleEventProjectsHelper() }
```
sgrimm: This shouldn't need the `by lazy` -- the helper's constructor doesn't do any work that would need to be deferred until time of first use. (That's assuming the function stays in a class rather than moving to the top level where it can be invoked without needing to instantiate a helper class.)

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
```diff
@@ -0,0 +1,163 @@
...
+
+  fun fetchOneById(eventId: EventId, projectId: ProjectId? = null): EventModel {
+    requirePermissions { readModuleEvent(eventId) }
+
+    val projectsField =
+        if (projectId == null) {
+          eventProjectsHelper.eventProjectsMultiset()
+        } else {
+          requirePermissions { readModuleEvent(eventId) }
```
sgrimm: This is already checked on line 36.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
```diff
@@ -0,0 +1,163 @@
...
+@Named
+class ModuleEventStore(
+    private val clock: InstantSource,
+    private val dslContext: DSLContext,
+    private val eventPublisher: ApplicationEventPublisher,
+    private val eventsDao: EventsDao,
+) {
+  private val eventProjectsHelper: ModuleEventProjectsHelper by lazy { ModuleEventProjectsHelper() }
+
+  fun fetchOneById(eventId: EventId, projectId: ProjectId? = null): EventModel {
```
sgrimm: The code paths for null and non-null project IDs are different enough that I wonder if this would be cleaner as two methods, one that takes just an event ID and one that takes an event and project ID.

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -147,6 +149,19 @@ class WebAppUrls(
...
   }
 
+  fun moduleEvent(
+      moduleId: ModuleId,
+      eventId: EventId,
+      organizationId: OrganizationId,
+      projectId: ProjectId
+  ): URI {
+    return UriBuilder.fromUri(config.webAppUrl)
+        .path("/modules/$moduleId/events/$eventId")
```
sgrimm: For in-app notifications, we don't use fully qualified URLs, just local paths, so the first two lines of this would be replaced with
```kotlin
return UriBuilder.fromPath("/modules/$moduleId/events/$eventId")
```

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -147,6 +149,19 @@ class WebAppUrls(
...
+        .queryParam("organizationId", organizationId)
+        .queryParam("organizationId", projectId)
```
sgrimm: Parameter name should be `projectId`, I believe.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -141,6 +143,19 @@ class Messages {
...
+      leadTime: Duration,
+      moduleName: String,
+  ): NotificationMessage =
+      NotificationMessage(
+          title =
+              getMessage(
+                  "notification.module.eventStarting.title",
+                  eventType.getDisplayName(currentLocale()),
+                  leadTime.toMinutes().toString()),
+          body = getMessage("notification.accession.dryingEnd.app.title", moduleName))
```
sgrimm: Wrong message identifier here.

src/main/resources/db/migration/0250/V264__EventRevision.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE accelerator.events ADD COLUMN revision INTEGER NOT NULL DEFAULT 1
```
sgrimm: A trailing newline would be good, as would a semicolon at the end of the statement.

src/main/resources/i18n/Enums_en.properties
```diff
@@ -19,6 +19,9 @@ accelerator.DeliverableCategory.SupplementalFiles=Supplemental Files
 accelerator.DeliverableCategory.VerraNonPermanenceRiskToolNPRT=Verra Non-Permanence Risk Tool (NPRT)
 # Noun, as in a legal document
 accelerator.DeliverableType.Document=Document
+accelerator.EventType.LiveSession=LiveSession
```
sgrimm: The value here should probably be `Live Session` or else notifications will say `Your LiveSession will start in 15 minutes`.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -97,6 +97,8 @@ notification.email.html.footer.2=PO Box 3470, PMB 15777
 notification.email.html.footer.3=Honolulu, HI 96801-3470
 notification.email.html.manageSettings=Manage your [notification settings].
 notification.email.text.footer=Manage your notification settings\: {0}\n\nTerraformation Inc.\nPO Box 3470, PMB 15777, Honolulu, HI 96801-3470\n\nhttps\://twitter.com/TF_Global\nhttps\://www.linkedin.com/company/terraformation/\nhttps\://www.instagram.com/globalterraform/\nhttps\://www.facebook.com/GlobalTerraform\nhttps\://terraformation.com/
+notification.module.eventStarting.title=Your {0} will start in {1} minutes
+notification.module.eventStarting.body=Click the join live session call for {0}
```
sgrimm: Is this notification body from the requirements? "Click the join live session call for 1:1 Session" feels like pretty awkward wording to me.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -97,6 +97,8 @@ notification.email.html.footer.2=PO Box 3470, PMB 15777
 notification.email.html.footer.3=Honolulu, HI 96801-3470
 notification.email.html.manageSettings=Manage your [notification settings].
 notification.email.text.footer=Manage your notification settings\: {0}\n\nTerraformation Inc.\nPO Box 3470, PMB 15777, Honolulu, HI 96801-3470\n\nhttps\://twitter.com/TF_Global\nhttps\://www.linkedin.com/company/terraformation/\nhttps\://www.instagram.com/globalterraform/\nhttps\://www.facebook.com/GlobalTerraform\nhttps\://terraformation.com/
+notification.module.eventStarting.title=Your {0} will start in {1} minutes
```
sgrimm, line 101: It'd probably be good to add a comment here to tell translators what kind of thing `{0}` is, in case a language has different words for "start" depending on what's starting.

src/test/kotlin/com/terraformation/backend/accelerator/ModuleEventNotifierTest.kt
```diff
@@ -0,0 +1,132 @@
...
+  @Test
+  fun `schedules notification at given lead time of module event`() {
+    clock.instant = Instant.EPOCH.plusSeconds(3600)
+    val startTime = clock.instant.plusSeconds(3600)
+    val notifyTime = startTime.minus(ModuleEventNotifier.notificationLeadTime)
+
+    val moduleEventId = insertEvent(revision = 1, startTime = startTime)
+    val notificationEvent = ModuleEventScheduledEvent(moduleEventId, 1)
+
+    assertTrue(notifyTime.isAfter((clock.instant)))
```
sgrimm: `clock.instant` has an extra set of parentheses around it.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStoreTest.kt
```diff
@@ -0,0 +1,314 @@
...
+  }
+  private lateinit var cohortId: CohortId
+  private lateinit var moduleId: ModuleId
+
+  @BeforeEach
+  fun setUp() {
+    insertUser()
+    insertOrganization()
+    cohortId = insertCohort()
+    insertParticipant(cohortId = inserted.cohortId)
```
sgrimm: Could use `cohortId` instead of `inserted.cohortId`

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
```diff
@@ -0,0 +1,163 @@
...
+        val overlap = updated.projects!!.intersect(existing.projects!!)
+        val toAdd = updated.projects.minus(overlap)
+        val toRemove = existing.projects.minus(overlap)
+
+        with(EVENT_PROJECTS) {
+          dslContext
+              .deleteFrom(this)
+              .where(EVENT_ID.eq(eventId))
+              .and(PROJECT_ID.`in`(toRemove))
+              .execute()
```
sgrimm: Maybe only do this if `toRemove` is nonempty.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
```diff
@@ -0,0 +1,163 @@
...
+    val updated = updateFunc(existing)
+    val userId = currentUser().userId
+    val now = clock.instant()
+
+    val revision =
+        if (updated.startTime == existing.startTime) {
+          existing.revision
+        } else {
+          existing.revision + 1
+        }
```
sgrimm, line 115: I don't think it's likely enough to be worth worrying about, but just noting that this will produce unpredictable results if two different updates of the same event happen at the same time. (They'll both see the same existing revision number.)

src/main/kotlin/com/terraformation/backend/accelerator/event/ModuleEventStartingEvent.kt
```diff
@@ -0,0 +1,6 @@
+package com.terraformation.backend.accelerator.event
+
+import com.terraformation.backend.db.accelerator.EventId
+
+/** Published when an event is starting. Ignored if event has been rescheduled. */
```
sgrimm: The second sentence here doesn't seem accurate to me -- doesn't the rescheduling check happen before this event gets published?

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -337,6 +342,26 @@ class AppNotificationService(
     }
   }
 
+  @EventListener
+  fun on(event: ModuleEventStartingEvent) {
+    systemUser.run {
+      val moduleEvent = moduleEventStore.fetchOneById(event.eventId)
+      val renderMessage = {
+        messages.moduleEventStartingNotification(
+            parentStore.getEventType(event.eventId)!!,
```
sgrimm: Could the event type live in `EventModel`? Then we don't need to fetch it separately.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -337,6 +342,26 @@ class AppNotificationService(
...
+            ModuleEventNotifier.notificationLeadTime,
+            parentStore.getModuleName(event.eventId)!!)
```
sgrimm: How about having the module ID in `EventModel` (which seems like it'd be generally useful anyway) and then fetching the module from the module store to get its name?

My concern here is mostly that `ParentStore` could easily turn into a grab-bag of "fetch this random field on this random table" functions. Which isn't the end of the world, but the reason it exists is so the other `Store` classes can navigate relationships without having to have each other as dependencies, which can lead to dependency cycles. Service classes like this one are free to depend on store classes as needed, no concerns about cycles.

We probably ought to document more of this kind of architectural thinking, admittedly; there's no great way to discover any of it right now.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
tommylau523: Should we look at a locking mechanism, or maybe is there a PostgreSQL atomic increment query that handles that on the DB side?

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventProjectsHelper.kt
tommylau523: I am opting to keep this for now just to make sure the caller is not missing this permission check. I am going to add a redundant permission check as well on the caller end to make it clear and readable.

src/main/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStore.kt
sgrimm: There are a few ways you could do this if you wanted.

1. Locking, like you said. Start the transaction right away and do `SELECT FOR UPDATE` to read the events row.
2. Add a condition like `.and(REVISION.eq(existing.revision))` to the update. Then if the revision doesn't match, the update won't do anything. In that case, you will no longer be able to assume that the update touching 0 rows means the event doesn't exist.
3. Let the database do the incrementing for you: `.set(REVISION, REVISION.plus(1))` and then instead of `.execute()`, you'll want `.returning(REVISION).fetchOne()` to get back the updated revision.

That said, I wouldn't spend a lot of effort on this unless you want to do it as an exercise or something. Only admins can edit events and I think it's sufficiently unlikely there'd be two different updates to the same event at the exact same time that handling it is pretty low priority.

src/main/kotlin/com/terraformation/backend/db/Exceptions.kt
```diff
@@ -122,6 +123,9 @@ open class KeycloakRequestFailedException(
 /** Keycloak couldn't find a user that we expected to be able to find. */
 class KeycloakUserNotFoundException(message: String) : EntityNotFoundException(message)
 
+class ModuleNotFoundException(moduleId: ModuleId) :
+    EntityNotFoundException("Module $moduleId not found")
```
sgrimm: This could live in `accelerator/db/Exceptions.kt` instead.

src/test/kotlin/com/terraformation/backend/accelerator/ModuleEventNotifierTest.kt
```diff
@@ -76,7 +76,7 @@ class ModuleEventNotifierTest : DatabaseTest(), RunsAsUser {
     val moduleEventId = insertEvent(revision = 1, startTime = startTime)
     val notificationEvent = ModuleEventScheduledEvent(moduleEventId, 1)
 
-    assertTrue(notifyTime.isAfter((clock.instant)))
+    assertTrue(notifyTime.isAfter((clock.instant())))
```
sgrimm: Oops, I was unclear in my previous comment: this had _too many_ parentheses, not too few. That is, it should be
```kotlin
assertTrue(notifyTime.isAfter(clock.instant))
```
rather than
```kotlin
assertTrue(notifyTime.isAfter((clock.instant)))
```

sgrimm: Also, a description of the assertion would be good to differentiate it from the other `assertTrue` in this test.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -2068,6 +2129,7 @@ internal class PermissionTest : DatabaseTest() {
     private val uncheckedDraftPlantingSites = draftPlantingSiteIds.toMutableSet()
     private val uncheckedFacilities = facilityIds.toMutableSet()
     private val uncheckedModuleEvents = moduleEventIds.toMutableSet()
+    private val uncheckedModules = moduleIds.toMutableSet()
```
sgrimm, line 2136: Need to add a statement to `andNothingElse()` to verify that the permission is false for any IDs that weren't listed in an `expect()` call in a test case.

sgrimm (changes requested):
Getting close!

sgrimm (approved):

tommylau523:
### Merge activity

* **Apr 24, 4:51 PM EDT**: @tommylau523 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1866).
* **Apr 24, 4:51 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1866).

PR #1873, opened 2024-04-22
SW-5213 Generate some documentation on in-app and email notifications

alexTerraformation:
Parse the notifications from the messages file to generate some documentation that can get added to the auto-generated docs.

### DIFF
--- a/.github/scripts/notifications.py
+++ b/.github/scripts/notifications.py
@@ -1,0 +1,106 @@
+def extract_notifications(file_path):
+    notifications = {'app': {}, 'email': {}}
+    ignore_keys = ['buttonIntro', 'linkIntro', 'buttonLabel', 'footer', 'manageSettings', 'html']
+    with open(file_path, 'r', encoding='utf-8') as file:
+        for line in file:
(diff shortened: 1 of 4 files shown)


.github/scripts/notifications.py
```diff
@@ -0,0 +1,106 @@
+def extract_notifications(file_path):
+    notifications = {'app': {}, 'email': {}}
+    ignore_keys = ['buttonIntro', 'linkIntro', 'buttonLabel', 'footer', 'manageSettings', 'html']
+    with open(file_path, 'r', encoding='utf-8') as file:
+        for line in file:
+            line = line.strip()
+            if '=' in line:
```
sgrimm: Should also check for the line not starting with `#` since `=` can occur in a comment.

sgrimm (approved):
One minor thing that it'd be good to change, but this seems fine otherwise.

If the notifications have HTML tags in them, the tags won't be escaped in the docs. But that's probably a feature, not a bug, if the goal is "see what the notifications look like to end users."

sgrimm (approved):

PR #1874, opened 2024-04-23
Put the event type in the ProjectModuleEventSession response object

nickgraz:
- This allows the enum to be carried to the frontend
- The OpenAPI type generation was ignoring the `EventType` as the key and converting it to `string`
- The property is needed in the context of the session in the frontend, we don't ever display an "event", we display the session instead.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
@@ -39,28 +39,30 @@ class ProjectModulesController(
 }
 
 data class ProjectModuleEventSession(
-    val id: EventId,
-    val startTime: Instant?,
     val endTime: Instant?,
+    val id: EventId,
     val meetingUrl: URI?,
     val recordingUrl: URI?,
     val slidesUrl: URI?,
+    val startTime: Instant?,
+    val type: EventType,
 ) {
   constructor(
-      model: EventModel
+      model: EventModel,
+      type: EventType,
   ) : this(
-      id = model.id,
-      startTime = model.startTime,
       endTime = model.endTime,
+      id = model.id,
       meetingUrl = model.meetingUrl,
       recordingUrl = model.recordingUrl,
       slidesUrl = model.slidesUrl,
+      startTime = model.startTime,
+      type = type,
   )
 }
 
 data class ProjectModuleEvent(
     val description: String,
-    val type: EventType,
     val sessions: List<ProjectModuleEventSession>,
 )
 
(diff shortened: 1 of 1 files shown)


nickgraz:
* **#1874** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):
Seems reasonable. Client can easily reconstruct the map if it wants.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -39,27 +39,30 @@ class ProjectModulesController(
...
-      id = model.id,
-      startTime = model.startTime,
       endTime = model.endTime,
+      id = model.id,
       meetingUrl = model.meetingUrl,
       recordingUrl = model.recordingUrl,
       slidesUrl = model.slidesUrl,
+      startTime = model.startTime,
+      type = type,
   )
```
nickgraz, line 61: Besides the addition of `type`, the rest of the changes are alphabetizing the arguments.

nickgraz:
Sorry for the re-request @sgrimm. I decided to move the property into the session since we don't really display an "event" to people on the frontend.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -39,28 +39,30 @@ class ProjectModulesController(
...
 data class ProjectModuleEventSession(
-    val id: EventId,
-    val startTime: Instant?,
     val endTime: Instant?,
+    val id: EventId,
     val meetingUrl: URI?,
     val recordingUrl: URI?,
     val slidesUrl: URI?,
+    val startTime: Instant?,
+    val type: EventType,
```
sgrimm, line 48: Should this go in `EventModel` instead? It won't vary from session to session for the same event, right?

sgrimm: Oh, hang on, I misread this. Ignore my comment.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -84,11 +87,12 @@ data class ProjectModule(
       overview = model.overview,
       preparationMaterials = model.preparationMaterials,
       events =
-          model.eventDescriptions.mapValues {
+          model.eventDescriptions.map {
```
sgrimm: This might be clearer with explicit variable names, e.g., `.map { (eventType, description) ->` rather than using `it.key` and `it.value`.

sgrimm (approved):
Mentioned one thing you can change if you feel like it, or not.

PR #1876, opened 2024-04-23
SW-5264 Use PlantingSiteStore to create site in importer

sgrimm:
In preparation for keeping a history of planting site map updates, change
`PlantingSiteImporter` to create sites via `PlantingSiteStore` rather than by
directly inserting them into the database. The logic to maintain the map
history will then only need to exist in one place.

One difference between the two classes was that only `PlantingSiteImporter`
populated the area and grid origin fields. Move the calculations of those
values into `PlantingSiteStore` as well, which has the added benefit that
they'll be populated for site maps that aren't created via shapefile import.


sgrimm:
* **#1876** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1876?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Nice refactor

PR #1877, opened 2024-04-24
SW-5201 Module Events Admin UI

tommylau523:
Table for all modules

Module Preview

Event List 

Module Events multi-select for updating participants

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
@@ -1,9 +1,13 @@
 package com.terraformation.backend.admin
 
+import com.terraformation.backend.accelerator.db.CohortStore
 import com.terraformation.backend.accelerator.db.DeliverablesImporter
+import com.terraformation.backend.accelerator.db.ModuleStore
 import com.terraformation.backend.accelerator.db.ModulesImporter
 import com.terraformation.backend.api.RequireGlobalRole
 import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.db.accelerator.EventType
+import com.terraformation.backend.db.accelerator.ModuleId
 import com.terraformation.backend.db.accelerator.tables.references.MODULES
 import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.importer.CsvImportFailedException
@@ -13,6 +17,7 @@ import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
 import org.springframework.validation.annotation.Validated
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestPart
@@ -24,21 +29,42 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes
 @RequireGlobalRole([GlobalRole.SuperAdmin, GlobalRole.AcceleratorAdmin])
 @Validated
 class AdminModulesController(
+    private val cohortStore: CohortStore,
     private val deliverablesImporter: DeliverablesImporter,
     private val dslContext: DSLContext,
     private val modulesImporter: ModulesImporter,
+    private val moduleStore: ModuleStore,
 ) {
   private val log = perClassLogger()
 
   @GetMapping("/modules")
   fun modulesHome(model: Model): String {
     val hasModules = dslContext.fetchExists(MODULES)
+    val cohortNames = cohortStore.findAll().associateBy { it.id }.mapValues { it.value.name }
+    val modules = moduleStore.fetchAllModules()
 
+    model.addAttribute("cohortNames", cohortNames)
     model.addAttribute("canManageDeliverables", currentUser().canManageDeliverables())
     model.addAttribute("canManageModules", currentUser().canManageModules())
     model.addAttribute("hasModules", hasModules)
+    model.addAttribute("modules", modules)
 
     return "/admin/modules"
+  }
+
+  @GetMapping("/modules/{moduleId}")
+  fun moduleView(model: Model, @PathVariable moduleId: String): String {
+    val module = moduleStore.fetchAllModules().firstOrNull { it.id == ModuleId(moduleId) }
+    val hasModule = module != null
+
+    model.addAttribute("hasModule", hasModule)
+    model.addAttribute("canManageModules", currentUser().canManageModules())
+    model.addAttribute("module", module)
+    model.addAttribute("workshopInfo", module?.eventDescriptions?.get(EventType.Workshop))
+    model.addAttribute("oneOnOneInfo", module?.eventDescriptions?.get(EventType.OneOnOneSession))
+    model.addAttribute("liveSessionInfo", module?.eventDescriptions?.get(EventType.LiveSession))
+
+    return "/admin/moduleView"
   }
 
   @PostMapping("/uploadDeliverables")
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
@@ -6,7 +6,6 @@ import com.terraformation.backend.accelerator.db.ModuleEventStore
 import com.terraformation.backend.accelerator.db.ModuleStore
 import com.terraformation.backend.accelerator.db.ModulesImporter
 import com.terraformation.backend.accelerator.db.ParticipantStore
-import com.terraformation.backend.accelerator.model.EventModel
 import com.terraformation.backend.api.RequireGlobalRole
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.ProjectStore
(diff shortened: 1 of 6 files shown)


tommylau523:
* **#1877** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1866** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1866?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
```diff
@@ -24,23 +42,154 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes
...
+      model: Model,
+      @PathVariable moduleId: String,
+      redirectAttributes: RedirectAttributes
+  ): String {
+    val module =
+        try {
+          moduleStore.fetchOneById(ModuleId(moduleId))
+        } catch (e: ModuleNotFoundException) {
+          log.warn("Module not found")
+          redirectAttributes.failureMessage = "Import failed: ${e.message}"
```
sgrimm: Should the error message here be "module not found" too?

src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
```diff
@@ -24,23 +42,154 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes
...
+
+    val cohortNames =
+        module.cohorts.associate { it.cohortId to cohortStore.fetchOneById(it.cohortId).name }
+
+    val projectNames = moduleProjects.associateWith { projectStore.fetchOneById(it).name }
+
+    // cohort name - project name
+    val cohortProjectNames =
+        moduleProjects.associateWith {
+          val project = projectStore.fetchOneById(it)
```
sgrimm: Would it make sense to fetch the project models once (maybe store them in a map by ID) and use them for both the project names and the cohort project names? Obviously the list of projects is going to be short but may as well not double-fetch if it's easy to avoid.

src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
```diff
@@ -24,23 +42,154 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes
...
+    eventStore.create(
+        ModuleId(moduleId),
+        EventType.forId(eventTypeId) ?: throw IllegalArgumentException("Event Type not recognized"),
+        dateStringToInstant(startTime),
+        dateStringToInstant(endTime),
+        meetingUrl,
+        recordingUrl,
+        slidesUrl,
+        toAdd?.toSet() ?: emptySet())
+    return redirectToModule(ModuleId(moduleId))
```
sgrimm, line 143: Setting the success message here to acknowledge the creation (and maybe to include the event ID) would be good so the user knows it succeeded.

It'd also be good to wrap the create call in a try block and set a failure message if it throws an exception for whatever reason.

Same with the update and delete endpoints.

src/main/kotlin/com/terraformation/backend/admin/AdminModulesController.kt
```diff
@@ -24,23 +42,154 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes
...
+      @RequestParam toAdd: List<ProjectId>?,
+      @RequestParam toRemove: List<ProjectId>?,
+      redirectAttributes: RedirectAttributes
+  ): String {
+    val event = eventStore.fetchOneById(id)
+
+    val projects = event.projects?.toMutableSet() ?: mutableSetOf()
+
+    toAdd?.let { projects += it.toSet() }
+    toRemove?.let { projects -= it.toSet() }
```
sgrimm: Alternately, something like:
```kotlin
val projects = (event.projects ?: emptySet<ProjectId>())
    .plus(toAdd ?: emptySet())
    .minus(toRemove ?: emptySet())
```
The way it is now is totally fine, though!

src/main/resources/templates/admin/moduleView.html
```diff
@@ -0,0 +1,535 @@
...
+
+            <tbody>
+            <tr th:each="event : ${oneOnOneEvents}" class="striped">
+                <td th:text="${event.id}">Event ID</td>
+                <td>
+                    <input type="datetime-local"
+                           name="startTime"
+                           required
+                           th:form="|update-${event.id}|"
+                           th:value="${#temporals.format(event.startTime.atZone('UTC'), dateFormat)}" />
```
sgrimm: Is it going to be asking for trouble to make people do the conversions between UTC and local timezones themselves? I could see that being a common source of user error, but maybe people are already used to doing it and it's not actually a big deal.

The alternative, I guess, would be to put a time zone selector on the page.

src/main/resources/templates/admin/moduleView.html
```diff
@@ -0,0 +1,535 @@
...
+<body>
+
+<span th:replace="~{/admin/header :: top}"/>
+
+<a href="/admin/">Home</a>
+-
+<a href="/admin/modules">Modules</a>
+
+<th:block th:if="${canManageModules}">
+    <h2 th:text="|Module ${module.name}|">Module (name)</h2>
```
sgrimm: Might be good to also include the module ID here since names aren't required to be unique (and probably won't be since we create new modules for module content edits).

src/main/resources/templates/admin/moduleView.html
```diff
@@ -0,0 +1,535 @@
...
+        <ul th:each="projectId : ${cohort.projects}">
+            <li th:text="${projectNames.get(projectId)}"> Project Name </li>
+        </ul>
+    </div>
+
+
+    <th:block th:if="${module.overview}">
+        <hr>
+        <h3> Overview </h3>
+        <div data-th-utext="${module.overview}"> </div>
```
sgrimm: Use `th:utext` instead of `data-th-utext` for consistency with the other Thymeleaf attributes; this file isn't valid HTML5 anyway.

src/main/resources/templates/admin/moduleView.html
```diff
@@ -0,0 +1,535 @@
...
+                        <input type="submit"
+                               value="Create" />
+                    </form>
+                </td>
+            </tr>
+            </tbody>
+        </table>
+    </th:block>
+
+    <th:block th:if="${workshopInfo}">
```
sgrimm: Feels like a fair bit of duplication here. Could this maybe iterate over a list of event types and look each type's description and events up from the maps in the module model, rather than the server splitting them out into separate template variables?

src/main/resources/templates/admin/voters.html
```diff
@@ -71,7 +71,8 @@ <h2>Voters</h2>
                         id="defaultVoters"
                         multiple>
                     <option th:each="userId : ${defaultVoters}"
-                            th:text="${users.get(userId).email}">
+                            th:text="${users.get(userId).email}"
+                            th:value="${userId}">
```
sgrimm: Seems unrelated to the rest of the PR.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStoreTest.kt
```diff
@@ -210,6 +211,35 @@ class ModuleEventStoreTest : DatabaseTest(), RunsAsUser {
...
+              moduleId, EventType.Workshop, startTime, projects = setOf(project1, project2))
+
+      assertNotNull(eventsDao.fetchOneById(model.id))
+
+      assertEquals(
+          listOf(
+              EventProjectsRow(model.id, project1),
+              EventProjectsRow(model.id, project2),
+          ),
+          eventProjectsDao.findAll())
```
sgrimm: The order of results from `findAll()` is undefined, so this should do `.toSet()` and compare against a set of expected results.

src/test/kotlin/com/terraformation/backend/accelerator/db/ModuleEventStoreTest.kt
```diff
@@ -333,4 +363,32 @@ class ModuleEventStoreTest : DatabaseTest(), RunsAsUser {
...
+      val eventId = insertEvent()
+      assertThrows<AccessDeniedException> { store.delete(eventId) }
+    }
+
+    @Test
+    fun `deletes the event`() {
+      val event1 = insertEvent()
+      val event2 = insertEvent()
+
+      assertTrue(eventsDao.findAll().isNotEmpty())
```
sgrimm: The assertion failure message would be more useful if this was something like
```kotlin
assertNotEquals(emptyList<ModuleEventsRow>(), eventsDao.findAll(), "Events before deletion")
```
And similarly for the other two assertions below. Testing for equality with a value will mean that if the assertion fails, the test output includes the expected and actual values, and including a message will make it easier to see which assertion failed if you're looking at the output from a CI build.

sgrimm (changes requested):
This will be useful! Some comments inline.

src/main/resources/templates/admin/moduleView.html
tommylau523: Opting to keep it in UTC for now for simplicity. 

If product team likes to have this feature we can add it then. Likely we can expand this to show times in project timezones as a separate column as well. But there will be a lot more data interdependencies and more javascript.

sgrimm (approved):

tommylau523:
### Merge activity

* **Apr 24, 4:51 PM EDT**: @tommylau523 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1877).
* **Apr 24, 4:52 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1877) rebased this pull request as part of a merge.
* **Apr 24, 5:00 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1877).

PR #1878, opened 2024-04-24
SW-5293 Implemented GET one module API

tommylau523:


tommylau523:
* **#1901** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1878** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1884** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -21,20 +21,36 @@ import org.springframework.web.bind.annotation.RequestMapping
...
+
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{moduleId}")
+  @Operation(summary = "Gets one module for a project.")
+  fun getModule(
+      @PathVariable("projectId") projectId: ProjectId,
+      @PathVariable moduleId: ModuleId,
+  ): GetProjectModuleResponsePayload {
+    val models = moduleStore.fetchModulesForProject(projectId)
```
tommylau523: Cannot use fetch one here because we do need the list of modules for ordering. 

There is probably a SQL query that can do something like `join select row_num() from .... ` but it is pretty complex and will require changes to our data models.

tommylau523: if the SQL method is preferred we can track in as a separate task.

sgrimm: I guess you mean you need the index? What's that used for? Could the start date (which is unique and sortable) serve the same purpose?

tommylau523: The start date is unique and sortable. I think this has something to do with how our client selector is set up. @nickgraz can you give some context?

nickgraz: We need to know what the modules "title" is. For example "name: Kick-off, title: Module 1". When we load a module, we need to know what the title is, but the data isn't stored in a field it seems, it is derived from the position in the list instead.

sgrimm: I'd kind of be inclined to say that the user-facing title of a module should be an explicit value we specify in the spreadsheet, rather than something we calculate based on date ordering.

I could easily see us, say, wanting "Module 3A" and "Module 3B" at some point.

tommylau523: Let me add the title in cohorts_modules table then.

sgrimm (changes requested):
It seems like the admin UI changes got folded into this PR, along with the schema change to support user-supplied titles; can you split the changes up so this one is just the read path? It'll be easier to review if it's not changing multiple things at once.

tommylau523:
> It seems like the admin UI changes got folded into this PR, along with the schema change to support user-supplied titles; can you split the changes up so this one is just the read path? It'll be easier to review if it's not changing multiple things at once.
> 
> 

Done! Most of the changes are to do with the title column, which I had split into a separate PR.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectModulesController.kt
```diff
@@ -21,20 +21,32 @@ import org.springframework.web.bind.annotation.RequestMapping
...
-    return GetProjectModulesResponsePayload(models.map { ProjectModule(it) })
+    return GetProjectModulesResponsePayload(models.map { model -> ProjectModule(model) })
+  }
+
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{moduleId}")
+  @Operation(summary = "Gets one module for a project.")
+  fun getModule(
+      @PathVariable("projectId") projectId: ProjectId,
```
sgrimm: No need for the name argument on the annotation here since the default is the parameter name. Doesn't hurt anything, just looks inconsistent since it's not present on the other parameter's annotation.

I know the existing code is inconsistent about specifying names; part of it is historical reasons (the server started off not using Spring Boot, and the other framework we used needed the explicit names) and part of it is just me being sloppy. At some point we should go clean it up across the code base.

src/main/kotlin/com/terraformation/backend/accelerator/db/CohortStore.kt
```diff
@@ -118,6 +116,31 @@ class CohortStore(
...
+                .set(START_DATE, it.startDate)
+                .set(END_DATE, it.endDate)
+                .onDuplicateKeyUpdate()
+                .set(TITLE, it.title)
+                .set(START_DATE, it.startDate)
+                .set(END_DATE, it.endDate)
+                .execute()
+          }
+        }
+      }
```
sgrimm: This feels like it belongs in another PR -- it has nothing to do with the GET endpoint, right?

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,171 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.servlet.mvc.support.RedirectAttributes
+
+@Controller
+@RequestMapping("/admin")
+@RequireGlobalRole(
+    [GlobalRole.SuperAdmin, GlobalRole.AcceleratorAdmin, GlobalRole.TFExpert, GlobalRole.ReadOnly])
+@Validated
+class AdminCohortsController(
```
sgrimm: This is both here and in #1884 but it seems like it should only be in #1884.

src/main/resources/db/migration/0250/V266__CohortModuleTitle.sql
```diff
@@ -0,0 +1,11 @@
...
+ALTER TABLE accelerator.cohort_modules ADD COLUMN title TEXT;
+UPDATE accelerator.cohort_modules t
+SET title = CONCAT('Module ', s.row_num::TEXT)
+    FROM (
+        SELECT cohort_id, start_date, ROW_NUMBER() OVER (PARTITION BY cohort_id ORDER BY start_date) AS row_num
+        FROM accelerator.cohort_modules
+    ) AS s
+WHERE t.cohort_id = s.cohort_id AND t.start_date = s.start_date;
+
+ALTER TABLE accelerator.cohort_modules ALTER COLUMN title SET NOT NULL;
```
sgrimm: These two migrations seem to have leaked in from other PRs.

sgrimm (changes requested):
Seems like the PRs have gotten a bit jumbled.

tommylau523:
Sorry looks like I accidentally did a force push without stacking. Should be resolved now!

sgrimm (approved):
Ah, much smaller now! Looks good.

tommylau523:
### Merge activity

* **Apr 29, 3:07 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1878) rebased this pull request as part of a merge.
* **Apr 29, 3:15 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1878).

PR #1880, opened 2024-04-24
SW-5269 - Add species_growth_forms table

nickgraz:
- Adds a table `species_growth_forms`, used to create a many-to-many relationship between the `species` table and `growth_forms` enum table.
- Migrate all `species.growth_form` into this new table
- Drop the `species.growth_form` column
- Update all tests to check for the new rows when checking species growth form data
- Update all controllers to return `species.growth_forms` instead of `species.growth_form`.
- Add `growthForms` sublist to `species` search table

TODO:
- [x] Update frontend - [PR](https://github.com/terraware/terraware-web/pull/2551)
- [x] Make sure consumers of reports with species are OK with this field changing from `growthForm` to `growthForms`
- [x] Determine if this type of dependent deploy is OK, or if we want to make it backwards compatible while the frontend is done.
  - FE is done, there seem to be no mobile considerations (this is being verified), so we can deploy them simultaneously with the next release.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
@@ -1,0 +1,37 @@
+package com.terraformation.backend.search.table
+
+import com.terraformation.backend.db.default_schema.SpeciesId
+import com.terraformation.backend.db.default_schema.tables.references.SPECIES
+import com.terraformation.backend.db.default_schema.tables.references.SPECIES_GROWTH_FORMS
+import com.terraformation.backend.search.SearchTable
+import com.terraformation.backend.search.SublistField
+import com.terraformation.backend.search.field.SearchField
+import org.jooq.Record
+import org.jooq.SelectJoinStep
+import org.jooq.TableField
+
+class SpeciesGrowthFormsTable(private val tables: SearchTables) : SearchTable() {
+  override val primaryKey: TableField<out Record, out Any?>
+    get() = SPECIES_GROWTH_FORMS.SPECIES_ID
+
+  override val sublists: List<SublistField> by lazy {
+    with(tables) {
+      listOf(
+          species.asSingleValueSublist("species", SPECIES_GROWTH_FORMS.SPECIES_ID.eq(SPECIES.ID)),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          idWrapperField("species_id", SPECIES_GROWTH_FORMS.SPECIES_ID) { SpeciesId(it) },
+          enumField("growth_form", SPECIES_GROWTH_FORMS.GROWTH_FORM_ID),
+      )
+
+  override val inheritsVisibilityFrom: SearchTable
+    get() = tables.species
+
+  override fun <T : Record> joinForVisibility(query: SelectJoinStep<T>): SelectJoinStep<T> {
+    return query.join(SPECIES).on(SPECIES_GROWTH_FORMS.SPECIES_ID.eq(SPECIES.ID))
+  }
+}
--- a/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
@@ -1,6 +1,5 @@
 package com.terraformation.backend.search.table
 
-import com.terraformation.backend.db.default_schema.SpeciesId
 import com.terraformation.backend.db.default_schema.tables.references.SPECIES
 import com.terraformation.backend.db.default_schema.tables.references.SPECIES_GROWTH_FORMS
 import com.terraformation.backend.search.SearchTable
@@ -24,8 +23,7 @@ class SpeciesGrowthFormsTable(private val tables: SearchTables) : SearchTable() 
 
   override val fields: List<SearchField> =
       listOf(
-          idWrapperField("species_id", SPECIES_GROWTH_FORMS.SPECIES_ID) { SpeciesId(it) },
-          enumField("growth_form", SPECIES_GROWTH_FORMS.GROWTH_FORM_ID),
+          enumField("growthForm", SPECIES_GROWTH_FORMS.GROWTH_FORM_ID),
       )
 
   override val inheritsVisibilityFrom: SearchTable
--- a/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
@@ -11,7 +11,7 @@ import org.jooq.TableField
 
 class SpeciesGrowthFormsTable(private val tables: SearchTables) : SearchTable() {
   override val primaryKey: TableField<out Record, out Any?>
-    get() = SPECIES_GROWTH_FORMS.SPECIES_ID
+    get() = SPECIES_GROWTH_FORMS.SPECIES_GROWTH_FORM_ID
 
   override val sublists: List<SublistField> by lazy {
     with(tables) {
--- a/src/main/resources/db/migration/0250/V265__SpeciesGrowthForm.sql
+++ b/src/main/resources/db/migration/0250/V265__SpeciesGrowthForm.sql
@@ -6,3 +6,11 @@ CREATE TABLE public.species_growth_forms (
 );
 
 CREATE INDEX ON public.species_growth_forms (species_id);
+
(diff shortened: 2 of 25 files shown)


nickgraz:
* **#1893** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1893?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1892** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1892?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1890** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1890?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1880** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
```diff
@@ -0,0 +1,37 @@
...
+    with(tables) {
+      listOf(
+          species.asSingleValueSublist("species", SPECIES_GROWTH_FORMS.SPECIES_ID.eq(SPECIES.ID)),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          idWrapperField("species_id", SPECIES_GROWTH_FORMS.SPECIES_ID) { SpeciesId(it) },
```
nickgraz: For some reason this line `SPECIES_GROWTH_FORMS.SPECIES_ID` is saying it needs to be `Int` instead of `SpeciesId`. I was hoping [this line](https://github.com/terraware/terraware-server/blob/main/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt#L232) would figure out the correct type for the field since it is `species_growth_forms.species_id`. @sgrimm  Am I missing something?

src/main/resources/db/migration/0250/V265__SpeciesGrowthForm.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE public.species_growth_forms (
+    species_id INTEGER REFERENCES public.species ON DELETE CASCADE,
```
nickgraz: This field does not seem to be getting the ID wrapped correctly within jOOQ generated classes.

sgrimm: Try `BIGINT` instead of `INTEGER`.

src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
sgrimm: Changing the column type should fix the problem with this being `Int`, but another problem here is the field name. Periods and underscores are reserved characters in the search API. `species_id` means, "the `id` field from the `species` sublist."

You have two options here.

* Get rid of this field entirely since you already have `species` as a sublist, and then use `species_id` which will join with the species table and give you the ID. That's the approach we use for foreign keys in most of the search tables.
* Keep an explicit ID field but call it `speciesId` so it's not parsed as a sublist field reference. This is slightly more performant if you're doing a lot of queries where you only care about species IDs but you don't care about any other species fields. If you also want, say, species names, then having an explicit `speciesId` field won't make things any faster since the query will have to join with the species table regardless.

My hunch is that you'll mostly be accessing this table as a sublist of the species table, rather than starting from here and looking up species information. In that case, you definitely won't need an explicit ID field since nothing would access it anyway.

src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
```diff
@@ -0,0 +1,37 @@
...
+          enumField("growth_form", SPECIES_GROWTH_FORMS.GROWTH_FORM_ID),
```
sgrimm: You'd want this name to be `growthForm`, similar to the above comment: `growth_form` gets parsed as, "the `form` field in the `growth` sublist" and of course there is no sublist named `growth`.

src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
sgrimm: You'll also want to add this as a sublist on the species search table so you can query the growth forms for a species.

sgrimm: I realized there was no validation of field names to make sure they don't run into ambiguity problems, so I added some test coverage in #1882 .

nickgraz: Thanks for the info, I went with the first option.

nickgraz: > You'll also want to add this as a sublist on the species search table so you can query the growth forms for a species.

Is [this change](https://github.com/terraware/terraware-server/pull/1880/files#diff-6a2f0959bf379022d4c70bec7f494557b26cf0f1978f6c5266f942cfaa2fd417R34-R35) sufficient for that, or did I miss something else?

sgrimm: Oh, oops, somehow I managed to not see that. Yes, that's totally sufficient.

src/main/kotlin/com/terraformation/backend/search/table/SpeciesGrowthFormsTable.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.terraformation.backend.search.SearchTable
+import com.terraformation.backend.search.SublistField
+import com.terraformation.backend.search.field.SearchField
+import org.jooq.Record
+import org.jooq.SelectJoinStep
+import org.jooq.TableField
+
+class SpeciesGrowthFormsTable(private val tables: SearchTables) : SearchTable() {
+  override val primaryKey: TableField<out Record, out Any?>
+    get() = SPECIES_GROWTH_FORMS.SPECIES_ID
```
sgrimm: Because this table has a compound primary key, you'll need to tell jOOQ to create a synthetic column that combines the species and growth form IDs, and use that column as the primary key here. At the bottom of `Config.kt`, you'll find a variable `EMBEDDABLES` that has a bunch of existing examples you can follow.

The reason we need this is because of what the primary key is used for in the search API. Basically it creates a SQL query that's structured like
```sql
SELECT field1, field2, field3
FROM table_from_search_prefix
WHERE primary_key IN (
    SELECT primary_key
    FROM table_from_search_prefix
    WHERE conditions_from_search_request
)
```
which only works if the primary key is guaranteed to refer to a single row. With a compound primary key, if you only use one of the columns, that query will return more results than expected.

sgrimm: CC @tommylau523 on the above since you're adding a compound primary key to a table too (though your table isn't exposed via the search API yet).

nickgraz:
@sgrimm the way this is currently written, the frontend must be updated and they need to be deployed more or less at the same time. Is that alright? It seems like a change that is small enough to where that might be OK. 

If you would rather this is made backward compatible, release, do the frontend, then remove backward compatibility in a future release, just let me know.

sgrimm:
> the frontend must be updated and they need to be deployed more or less at the same time. Is that alright? It seems like a change that is small enough to where that might be OK.

What about the mobile app? I know it uses the species API endpoints, but maybe it never references the growth form field at all.

nickgraz:
Cool, the frontend is already completed in another PR. I will double check with Josh and make the updates in the app where necessary.

nickgraz:
Looks like it is [only referenced in the types](https://github.com/search?q=repo%3Aterraware%2Fterraware-mobile-app%20growthForm&type=code) coming back from the server, and no where else in the codebase.

src/main/resources/db/migration/0250/V266__MigrateSpeciesGrowthForm.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO public.species_growth_forms (species_id, growth_form_id)
+SELECT id, growth_form_id
+FROM public.species
+WHERE growth_form_id IS NOT null
+ON CONFLICT DO NOTHING;
+
+ALTER TABLE public.species DROP COLUMN growth_form_id;
```
sgrimm: May as well combine these into one migration since we're not doing a phased backward-compatible rollout for this change.

sgrimm (approved):

PR #1884, opened 2024-04-25
SW-5201 Cohort Module Admin UI

tommylau523:


tommylau523:
* **#1901** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1878** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1884** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/model/CohortModel.kt
```diff
@@ -12,19 +12,21 @@ data class CohortModel<ID : CohortId?>(
...
+        participantIdsField: Field<Set<ParticipantId>>? = null,
+        cohortModulesField: Field<List<CohortModuleModel>>? = null,
     ): ExistingCohortModel {
       return ExistingCohortModel(
           id = record[COHORTS.ID]!!,
           name = record[COHORTS.NAME]!!,
           participantIds = participantIdsField?.let { record[it] } ?: emptySet(),
           phase = record[COHORTS.PHASE_ID]!!,
-      )
+          modules = cohortModulesField?.let { record[it] } ?: emptyList())
```
sgrimm: Maybe move this up so it's in alphabetical order? Easier to find at a glance that way.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.servlet.mvc.support.RedirectAttributes
+
+@Controller
+@RequestMapping("/admin")
+@RequireGlobalRole([GlobalRole.SuperAdmin, GlobalRole.AcceleratorAdmin])
```
sgrimm: Should TF Expert or Read Only users be able to view cohort info, even if they can't update it?

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,163 @@
...
+    model.addAttribute("cohorts", cohorts)
+    model.addAttribute("canUpdateCohorts", canUpdateCohorts)
+
+    return "/admin/cohorts"
+  }
+
+  @GetMapping("/cohorts/{cohortId}")
+  fun cohortView(
+      model: Model,
+      @PathVariable cohortId: String,
```
sgrimm: For path variables or request params, Spring should be able to do the type conversion for you. That is, this can be `@PathVariable cohortId: CohortId`.

The only time we need to do the conversion ourselves is in endpoints that accept multipart form data for file uploads, which for some reason Spring doesn't handle automatically.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,163 @@
...
+          cohortStore.fetchOneById(id, CohortDepth.Cohort, CohortModuleDepth.Module)
+        } catch (e: CohortNotFoundException) {
+          log.warn("Cohort not found", e)
+          redirectAttributes.failureMessage = "Cohort not found: ${e.message}"
+          return redirectToCohortHome()
+        }
+    val cohortModules = cohort.modules.map { it.moduleId }
+
+    val modules = moduleStore.fetchAllModules()
+    val moduleNames = modules.associate { it.id to "(" + it.id.toString() + ") " + it.name }
```
sgrimm: I take it you put the ID first because of the expectation that there will be lots of duplicate module names?

The value here might be easier to read as `"(${it.id}) ${it.name}"` rather than as a concatenation expression.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,163 @@
...
+    return "/admin/cohortView"
+  }
+
+  @PostMapping("/cohorts/{cohortId}/addModule")
+  fun addModule(
+      model: Model,
+      @PathVariable cohortId: String,
+      @RequestParam moduleId: ModuleId,
+      @RequestParam startDate: String,
+      @RequestParam endDate: String,
```
sgrimm: You should be able to declare these as `LocalDate` here.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,163 @@
...
+      redirectAttributes: RedirectAttributes
+  ): String {
+    val id = CohortId(cohortId)
+    try {
+      cohortStore.update(id) {
+        it.copy(
```
sgrimm: Declaring a variable for the cohort model would make this a little easier to follow, I think.

src/main/resources/db/migration/0250/V265__CohortModulePrimaryKey.sql
```diff
@@ -0,0 +1,10 @@
+-- Delete duplicated cohort_module assignments, keep only last inserted
+DELETE FROM accelerator.cohort_modules a USING accelerator.cohort_modules b
```
sgrimm: Nice!

src/main/resources/db/migration/0250/V265__CohortModulePrimaryKey.sql
```diff
@@ -0,0 +1,10 @@
...
+    WHERE b.id < a.id
+    AND a.cohort_id = b.cohort_id
+    AND a.module_id = b.module_id;
+
+-- Make (cohort_id, module_id) primary key and unique
+ALTER TABLE accelerator.cohort_modules DROP CONSTRAINT cohort_modules_pkey;
+ALTER TABLE accelerator.cohort_modules ADD PRIMARY KEY (cohort_id, module_id);
```
sgrimm: This made me realize that there were no indexes on cohort or module ID before, which would have made the foreign key constraints more expensive to enforce if a cohort or module is deleted (it'll have to do a full scan of `cohort_modules`). This primary key constraint will create an index the database can use for the `cohort_id` foreign key; may as well create an index on module ID in this migration too.

src/main/resources/templates/admin/cohortView.html
```diff
@@ -0,0 +1,110 @@
...
+<head th:replace="~{/admin/header :: head}"/>
+<body>
+
+<span th:replace="~{/admin/header :: top}"/>
+
+<a href="/admin/">Home</a>
+ -
+<a href="/admin/cohorts">Cohorts</a>
+
+<th:block th:if="${canUpdateCohort}">
```
sgrimm: If you decide to make this available in read-only form to TF Expert and Read Only users, this conditional will probably need to turn into something like a series of conditionals on the submit buttons to grey them out.

src/test/kotlin/com/terraformation/backend/accelerator/db/CohortStoreTest.kt
```diff
@@ -234,6 +269,83 @@ class CohortStoreTest : DatabaseTest(), RunsAsUser {
...
+      store.update(cohortId) {
+        it.copy(
+            modules =
+                listOf(
+                    CohortModuleModel(
+                        cohortId,
+                        module1,
+                        LocalDate.of(2024, 2, 1),
+                        LocalDate.of(2024, 2, 3),
+                    ),
```
sgrimm: Maybe insert this one beforehand too, so the test covers the three cases of delete, keep, and add rather than just delete and add.

src/test/kotlin/com/terraformation/backend/accelerator/db/CohortStoreTest.kt
```diff
@@ -234,6 +269,83 @@ class CohortStoreTest : DatabaseTest(), RunsAsUser {
...
+              CohortModulesRow(
+                  cohortId, module1, LocalDate.of(2024, 2, 1), LocalDate.of(2024, 2, 3)),
+              CohortModulesRow(
+                  cohortId, module2, LocalDate.of(2024, 3, 1), LocalDate.of(2024, 3, 3))),
+          cohortModulesDao.findAll(),
+          "Cohort modules after update.")
+    }
+
+    @Test
+    fun `deletes cohort modules if modules are removed`() {
```
sgrimm: This is covered by the previous test, I think.

sgrimm (changes requested):
Some feedback, but this seems pretty good at a high level.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,169 @@
...
+    private val cohortStore: CohortStore,
+    private val moduleStore: ModuleStore,
+) {
+  private val log = perClassLogger()
+
+  @GetMapping("/cohorts")
+  fun cohortsHome(model: Model): String {
+    val cohorts = cohortStore.findAll()
+
+    val canReadCohorts = cohorts.all { currentUser().canReadCohort(it.id) }
```
sgrimm: It'd be more consistent with the rest of the admin UI to make this throw an access denied exception if the user doesn't have permission to view the page's contents, at which point the template doesn't need this value.

At some point it might be worth talking about whether we want to make "you visited a page you shouldn't have been able to even see a link to" do something else in the admin UI in general, but in the meantime, may as well have it behave consistently.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
```diff
@@ -0,0 +1,169 @@
...
+        val updatedModules =
+            it.modules.plus(CohortModuleModel(cohortId, moduleId, startDate, endDate))
+        it.copy(modules = updatedModules)
+      }
+      redirectAttributes.successMessage = "Cohort module added."
+    } catch (e: Exception) {
+      log.warn("Add cohort module failed")
+      redirectAttributes.failureMessage =
+          when (e) {
+            is DataIntegrityViolationException -> "Add module failed. Dates are invalid."
```
sgrimm, line 98: Worth explaining the constraints in this error message so people will have a better idea what they did wrong? This will usually be user error rather than a system failure.

When we add this functionality to the real web app, we'll probably want to make the failure checking a little more fine-grained, e.g., by catching this in `CohortStore` and querying to see which modules' dates overlap so we can give the user more specific feedback.

src/test/kotlin/com/terraformation/backend/accelerator/db/CohortStoreTest.kt
```diff
@@ -174,10 +142,77 @@ class CohortStoreTest : DatabaseTest(), RunsAsUser {
...
+          startDate = LocalDate.of(2024, 2, 1),
+          endDate = LocalDate.of(2024, 2, 3))
+
+      // No depth, defaults to "cohort"
+      assertEquals(
+          ExistingCohortModel(
+              id = cohortId,
+              name = "Cohort Test",
+              phase = CohortPhase.Phase0DueDiligence,
+              participantIds = setOf(),
```
sgrimm, line 180: Slightly more idiomatic to use `emptySet()` instead of `setOf()`.

sgrimm (approved):
Noticed a couple minor things, but this seems solid in general.

src/main/kotlin/com/terraformation/backend/admin/AdminCohortsController.kt
tommylau523: I will follow up in a separate PR to add this in `CohortStore`. 

I think since modules may be allowed to overlap, we may remove the dates overlapping constraints. Will updates documentations once all of those requirements are finalized.

tommylau523:
### Merge activity

* **Apr 29, 2:52 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1884).

PR #1895, opened 2024-04-26
SW-5303 Added Event status columns and modified store and controllers 

tommylau523:

PR #1894, opened 2024-04-26
SW-5303 Added Event Status tables and Enums

tommylau523:

PR #1891, opened 2024-04-26
Added module title column to store, controller, and updated admin UI to update those

tommylau523:

PR #1901, opened 2024-04-29
SW-5223 Module Events Search Tables

tommylau523:
Use cases:
1) Participant project user can search for events associated with the project, and modules associated with the events, (mainly for event descriptions)

2) Accelerator admin user can search for every event under a module. 

Out of scope:
1) Associations of cohorts to modules
2) Associations of deliverables, submissions to modules


tommylau523:
* **#1901** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1901?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1878** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1878?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1884** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
```diff
@@ -0,0 +1,51 @@
...
+  override val localize: Boolean
+    get() = false
+
+  override val supportedFilterTypes: Set<SearchFilterType>
+    get() = EnumSet.of(SearchFilterType.Exact, SearchFilterType.PhraseMatch)
+
+  override fun getCondition(fieldNode: FieldNode): Condition {
+    val valuesWithoutScheme =
+        fieldNode.values.map { it?.removePrefix("https://")?.removePrefix("http://") }
+    val uris = valuesWithoutScheme.mapNotNull { URI("https://$it") }
```
sgrimm: If someone includes a scheme, why would we want to ignore what they asked us to search for and force it to HTTPS? And if someone doesn't include a scheme, it should be fine because the search is already doing a substring match rather than a prefix match or an exact match.

But it's possible I'm just not seeing the use case here.

src/main/kotlin/com/terraformation/backend/search/table/ModulesTable.kt
```diff
@@ -0,0 +1,63 @@
...
+    with(tables) {
+      listOf(
+          events.asMultiValueSublist("events", EVENTS.MODULE_ID.eq(MODULES.ID)),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          textField("additionalResources", MODULES.ADDITIONAL_RESOURCES, nullable = false),
```
sgrimm: Aren't the text fields on this table all nullable except for `name`? The `nullable` flags here seem like they're inverted.

src/main/kotlin/com/terraformation/backend/search/table/EventsTable.kt
```diff
@@ -0,0 +1,61 @@
...
+              .and(PROJECTS.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)))
+    }
+  }
+
+  private val eventProjectsCondition =
+      DSL.exists(
+          DSL.selectOne()
+              .from(EVENT_PROJECTS)
+              .where(EVENT_PROJECTS.PROJECT_ID.eq(PROJECTS.ID))
+              .and(EVENT_PROJECTS.EVENT_ID.eq(EVENTS.ID)))
```
sgrimm, line 64: Interesting approach! Took me a moment to realize that this was to avoid having to expose the linking table as an entity in the API.

src/main/kotlin/com/terraformation/backend/search/table/ProjectsTable.kt
```diff
@@ -80,4 +84,17 @@ class ProjectsTable(tables: SearchTables) : SearchTable() {
...
+          DSL.selectOne()
+              .from(EVENT_PROJECTS)
+              .where(EVENT_PROJECTS.PROJECT_ID.eq(PROJECTS.ID))
+              .and(EVENT_PROJECTS.EVENT_ID.eq(EVENTS.ID)))
+
+  private val cohortModulesCondition: Condition =
+      COHORT_MODULES.COHORT_ID.`in`(
+          DSL.select(PARTICIPANTS.COHORT_ID)
+              .from(PARTICIPANTS)
+              .where(PARTICIPANTS.ID.eq(PROJECTS.PARTICIPANT_ID)))
```
sgrimm: Is this used?

sgrimm (commented):
Commenting rather than approving/requesting changes because there's one part of this I'm not clear on.

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
tommylau523: I was thinking about the case of creating a URI object without a scheme, since the contains operation expects one. 

What I think would suit the use cases is to treat the table field as string. But jooq is giving me a hard time with that. I was looking into some casting or coercing functions but difficult. 

I updated this part to use regex to perform substring matching and this works well so far.

Let me know if you have any advice on treating the table field as a string, which will enable Fuzzy searches as well!

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
```diff
@@ -0,0 +1,48 @@
...
+    override val nullable: Boolean = true,
+) : SingleColumnSearchField<URI>() {
+  override val localize: Boolean
+    get() = false
+
+  override val supportedFilterTypes: Set<SearchFilterType>
+    get() = EnumSet.of(SearchFilterType.Exact, SearchFilterType.PhraseMatch)
+
+  override fun getCondition(fieldNode: FieldNode): Condition {
+    val valuesRegex = fieldNode.values.map { "$it" }
```
sgrimm: What is this doing? The list of values is already `List<String?>` so this seems like it would be a no-op for actual values, but would turn null values into `"null"` which doesn't seem like what we'd want.

The fact that the variable name says this is a regex, but the value is a list of non-regex strings, makes me think the intention was something else here.

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
```diff
@@ -0,0 +1,48 @@
...
+          throw IllegalArgumentException("Fuzzy search is not supported for URI fields")
+      SearchFilterType.Range ->
+          throw IllegalArgumentException("Range search is not supported for URI fields")
+    }
+  }
+
+  override val orderByField: Field<*>
+    get() = databaseField.collate(currentLocale().collation)
+
+  // Text fields are always raw.
```
sgrimm: s/Text/URI/

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceNestedFieldsTest.kt
```diff
@@ -806,7 +806,6 @@ internal class SearchServiceNestedFieldsTest : SearchServiceTest() {
     // the organization users tables is a special case: a single-value sublist that
     // refers to a child, not a parent. Include it explicitly.
     val usersFieldNames = tables.users.getAllFieldNames("members.user.")
```
sgrimm: The blank line here was because the comment doesn't apply to the line after this.

sgrimm (changes requested):
Looks pretty good, but I think the handling of null search values isn't quite right.

Also, it would be worthwhile to add some test coverage of the sublist relationships that are being added here. This is introducing a brand-new pattern for many-to-many relationships (joining on EXISTS) and we'll want to verify that it returns the results we expect.

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
tommylau523: Ahhh I left this with some escaped operations earlier. I needed to modify this to be filter non nulls basically.

src/main/kotlin/com/terraformation/backend/search/field/UriField.kt
```diff
@@ -25,13 +23,13 @@ class UriField(
...
   override fun getCondition(fieldNode: FieldNode): Condition {
-    val valuesRegex = fieldNode.values.map { "$it" }
+    val nonNullValues = fieldNode.values.filterNotNull()
     return when (fieldNode.type) {
       SearchFilterType.PhraseMatch,
       SearchFilterType.Exact ->
           DSL.or(
               listOfNotNull(if (fieldNode.values.any { it == null }) databaseField.isNull else null)
-                  .plus(valuesRegex.map { databaseField.likeRegex(it) }))
+                  .plus(nonNullValues.map { databaseField.likeRegex(it) }))
```
sgrimm: Not sure `likeRegex` is the operation we want here. The search API doesn't expect regexes as search strings in other places; it treats the search terms as literal strings even if they include `.` or `*` or other regex special characters.

Is the client expecting to be able to pass in actual regexes for searches on URI fields?

src/test/kotlin/com/terraformation/backend/accelerator/ModuleEventSearchTest.kt
```diff
@@ -0,0 +1,292 @@
...
+    val event2 = insertEvent(startTime = Instant.ofEpochSecond(800))
+    val event3 = insertEvent(startTime = Instant.ofEpochSecond(100))
+
+    val prefix = SearchFieldPrefix(searchTables.events)
+    val fields = listOf("id").map { prefix.resolve(it) }
+
+    val expected =
+        SearchResults(
+            listOf(
+                mapOf("id" to event3.toString()),
```
sgrimm: `.toString()` is totally correct, so not saying this needs to change. Another option is `"$event3"`.

src/test/kotlin/com/terraformation/backend/accelerator/ModuleEventSearchTest.kt
```diff
@@ -0,0 +1,292 @@
...
+        SearchResults(
+            listOf(
+                mapOf("id" to event1.toString()),
+                mapOf("id" to event2.toString()),
+                mapOf("id" to event3.toString())),
+            null)
+    val eventActual = searchService.search(eventPrefix, eventFields, NoConditionNode())
+
+    assertJsonEquals(projectExpected, projectActual)
+    assertJsonEquals(eventExpected, eventActual)
```
sgrimm: These could have assertion messages so it's easier to see which one is failing in the test output.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/accelerator/ModuleEventSearchTest.kt
```diff
@@ -0,0 +1,286 @@
...
+            ),
+            null)
+
+    val actual = searchService.search(prefix, fields, NoConditionNode())
+
+    assertJsonEquals(expected, actual)
+  }
+
+  @Test
+  fun `can search for events by projects`() {
```
sgrimm: What this test is checking for is good. When I read this test name, I though the test was going to check whether you could have a _filter_ with a project ID, that is, flipping the structure of this and searching on the events table with a `FieldNode` that searches `projects.id`.

I think it'd be good to include that as well, again because this is introducing a new pattern for the many-to-many relationships and we should verify that the logic for filtering across this kind of relationship works as expected, in addition to verifying the logic for fetching across the relationship (which this test and the others in the PR already cover nicely).

src/main/kotlin/com/terraformation/backend/search/SearchExceptions.kt
```diff
@@ -7,3 +7,7 @@ import jakarta.ws.rs.core.Response
 class SearchFieldNotExportableException(fieldName: String) :
     ClientErrorException(
         "The search field $fieldName is not exportable", Response.Status.BAD_REQUEST)
+
+/** An exception thrown when a search path contains the same table multiple times. */
+class SearchTableRevisitedException(table: String) :
+    IllegalArgumentException("The search path visited $table multiple times")
```
sgrimm: This isn't needed any more, correct?

sgrimm (changes requested):
The changes for `getAllFields` could have been a separate PR since they're kind of logically distinct from adding the new search tables, but no need to split this PR, just a consideration for future changes.

sgrimm (approved):
Looks good!

PR #1914, opened 2024-05-01
Fix site size validation on shapefile upload

sgrimm:
We were setting the SRID on the geometries we got back from GeoTools' shapefile
parser, but we weren't setting it in such a way that `Geometry.getEnvelope()`
would return a geometry with the same SRID.

This was causing the maximum envelope size validation check to bomb out with
actual shapefiles, though it worked with the programmatically-constructed
"shapefiles" in our unit tests since they don't go through the parser.


sgrimm:
* **#1914** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1914?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Nice fix

PR #1915, opened 2024-05-01
Support tiny simple planting site maps

sgrimm:
The planting sites table has a constraint that requires the site's area in
hectares to be greater than 0, because we divide by the site area when
calculating certain statistics.

Previously, we didn't calculate the areas of simple planting sites at all, so
site maps could be arbitrarily tiny without violating the constraint. Commit
af1e716b started calculating areas for all sites, not just detailed ones, and
this caused creation of tiny simple sites to bomb out because their areas were
rounded down to 0.

Restore the previous behavior if a simple site's boundary is tiny: the area is
not stored at all. For larger simple sites, we continue to store it.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -267,9 +267,14 @@ class PlantingSiteStore(
       throw PlantingSiteMapInvalidException(problems)
     }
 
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#1915** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
A nice, simple solution

PR #1919, opened 2024-05-01
Add a controller and function to fetch a participant project species by ID

nickgraz:
Add a controller and function to fetch a participant project species by ID

Fix URL

PR #1917, opened 2024-05-01
SW-5318 Participant Project Species permissions

nickgraz:
- Adds 4 new permissions: 
  - `canCreateParticipantProjectSpecies`
  - `canDeleteParticipantProjectSpecies`
  - `canReadParticipantProjectSpecies`
  - `canUpdateParticipantProjectSpecies`

PR #1916, opened 2024-05-01
SW-5318 Create the participant_project_species table

nickgraz:
- Create the `participant_project_species table`
- It links together `projects` and `species`
- Also used to store a `SubmissionStatus` and `feedback` for deliverable management

PR #1924, opened 2024-05-02
SW-5318 Add create store functionality for ParticipantProjectSeries

nickgraz:
- Adds validation to make sure the given project has a participant

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
@@ -1,19 +1,55 @@
 package com.terraformation.backend.accelerator.db
 
+import com.terraformation.backend.accelerator.event.CohortParticipantAddedEvent
 import com.terraformation.backend.accelerator.model.ExistingParticipantProjectSpeciesModel
+import com.terraformation.backend.accelerator.model.NewParticipantProjectSpeciesModel
 import com.terraformation.backend.accelerator.model.ParticipantProjectSpeciesModel
+import com.terraformation.backend.accelerator.model.toModel
 import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.accelerator.ParticipantProjectSpeciesId
+import com.terraformation.backend.db.accelerator.tables.daos.ParticipantProjectSpeciesDao
+import com.terraformation.backend.db.accelerator.tables.pojos.ParticipantProjectSpeciesRow
+import com.terraformation.backend.db.accelerator.tables.pojos.ParticipantsRow
 import com.terraformation.backend.db.accelerator.tables.references.PARTICIPANT_PROJECT_SPECIES
 import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.default_schema.tables.daos.ProjectsDao
 import jakarta.inject.Named
 import org.jooq.Condition
 import org.jooq.DSLContext
 
 @Named
 class ParticipantProjectSpeciesStore(
-    private val dslContext: DSLContext,
+  private val dslContext: DSLContext,
+  private val participantProjectSpeciesDao: ParticipantProjectSpeciesDao,
+  private val projectsDao: ProjectsDao,
 ) {
+  fun create(model: NewParticipantProjectSpeciesModel): ExistingParticipantProjectSpeciesModel {
+    requirePermissions { createParticipantProjectSpecies(model.projectId) }
+
+    // Participant project species can only be associated
+    // to projects that are associated to a participant
+    val project = projectsDao.fetchOneById(model.projectId);
+    if (project?.participantId == null) {
+      throw ProjectNotInParticipantException(model.projectId);
+    }
+
+    return dslContext.transactionResult { _ ->
+      val row =
+          ParticipantProjectSpeciesRow(
+              feedback = model.feedback,
+              projectId = model.projectId,
+              rationale = model.rationale,
+              speciesId = model.speciesId,
+              submissionStatusId = model.submissionStatus,
+          )
+
+      participantProjectSpeciesDao.insert(row)
+
+      row.toModel()
+    }
+  }
+
   fun fetchOneById(
       participantProjectSpeciesId: ParticipantProjectSpeciesId
   ): ExistingParticipantProjectSpeciesModel {
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
@@ -1,6 +1,5 @@
 package com.terraformation.backend.accelerator.db
 
-import com.terraformation.backend.accelerator.event.CohortParticipantAddedEvent
 import com.terraformation.backend.accelerator.model.ExistingParticipantProjectSpeciesModel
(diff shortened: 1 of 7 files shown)


nickgraz:
* **#1937** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1935** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1935?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1934** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1933** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1932** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1932?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1924** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1919** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1918** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -1,19 +1,53 @@
...
     private val dslContext: DSLContext,
+    private val participantProjectSpeciesDao: ParticipantProjectSpeciesDao,
+    private val projectsDao: ProjectsDao,
 ) {
+  fun create(model: NewParticipantProjectSpeciesModel): ExistingParticipantProjectSpeciesModel {
+    requirePermissions { createParticipantProjectSpecies(model.projectId) }
+
+    // Participant project species can only be associated
+    // to projects that are associated to a participant
+    val project = projectsDao.fetchOneById(model.projectId)
```
nickgraz, line 30: Does this seem like an alright place to put this? Or would this make more sense within a "service"? It seems alright to me since the project ID is a direct field on the entity.

sgrimm: Yeah, I think this is fine. We don't want `Store` classes to depend on each other because of the risk of dependency cycles, but it's fine for a `Store` to depend on multiple `Dao` classes.

Another option (but to be clear: what you're doing here is fine) is to add a `getParticipantId(projectId)` method to `ParentStore`.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -1,19 +1,53 @@
...
+    }
+
+    return dslContext.transactionResult { _ ->
+      val row =
+          ParticipantProjectSpeciesRow(
+              feedback = model.feedback,
+              projectId = model.projectId,
+              rationale = model.rationale,
+              speciesId = model.speciesId,
+              submissionStatusId = model.submissionStatus,
```
sgrimm: Will we ever want to create one of these with an initial status other than "not submitted"? If not, maybe hardwiring that here would be reasonable.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
```diff
@@ -102,4 +104,55 @@ class ParticipantProjectSpeciesStoreTest : DatabaseTest(), RunsAsUser {
...
+            NewParticipantProjectSpeciesModel(
+                feedback = "feedback",
+                id = null,
+                projectId = projectId,
+                rationale = "rationale",
+                speciesId = speciesId,
+                submissionStatus = SubmissionStatus.NotSubmitted))
+      }
+    }
+  }
```
sgrimm, line 175: For completeness' sake, could use a test case for throwing an exception if the user doesn't have the right permission.

sgrimm (changes requested):
Looks good; just one thing missing and one potential change you could consider.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: I don't believe so, I will hardwire it.

sgrimm:
Noticed the PR title says "Series" instead of "Species."

sgrimm (changes requested):
Seems like this needs to be rebased.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -1,19 +1,54 @@
...
+    requirePermissions { createParticipantProjectSpecies(model.projectId) }
+
+    // Participant project species can only be associated
+    // to projects that are associated to a participant
+    val project = projectsDao.fetchOneById(model.projectId)
+    if (project?.participantId == null) {
+      throw ProjectNotInParticipantException(model.projectId)
+    }
+
+    return dslContext.transactionResult { _ ->
```
sgrimm: This is just inserting one row, so no need to wrap it in a transaction.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -1,19 +1,54 @@
...
+      val row =
+          ParticipantProjectSpeciesRow(
+              feedback = model.feedback,
+              projectId = model.projectId,
+              rationale = model.rationale,
+              speciesId = model.speciesId,
+              submissionStatusId = model.submissionStatus ?: SubmissionStatus.NotSubmitted,
```
sgrimm: It might be better to put the default value on the property in the model class so the field can stay non-nullable. That is, in the model class, declare it as `val submissionStatusId: SubmissionStatus = SubmissionStatus.NotSubmitted`.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionRequirementsTest.kt
```diff
@@ -366,6 +372,13 @@ internal class PermissionRequirementsTest : RunsAsUser {
 
   @Test fun createParticipant() = allow { createParticipant() } ifUser { canCreateParticipant() }
 
+  @Test
+  fun createParticipantProjectSpecies() {
+    grant { user.canReadProject(projectId) }
+    grant { user.canCreateParticipantProjectSpecies(projectId) }
+    requirements.createParticipantProjectSpecies(projectId)
+  }
```
sgrimm: This is missing assertions for the permission-denied cases.

sgrimm (changes requested):
A couple minor things, but this is close.

sgrimm (approved):
Looks good, but the PR title says "Series" instead of "Species".

PR #1934, opened 2024-05-06
SW-5318 Add update functionality for ParticipantProjectSpecies

nickgraz:
- Add update function to `ParticipantProjectSpeciesStore`
- Add controller endpoint for updating a participant project species

PR #1933, opened 2024-05-06
SW-5318 Add endpoints for create and assign (create many) for ParticipantProjectSpecies

nickgraz:
- POST `/api/v1/accelerator/projects/species` endpoint for creating a single entry
- POST `/api/v1/accelerator/projects/species/assign` endpoint for bulk assigning species to projects

PR #1932, opened 2024-05-06
SW-5318 Add createMany function to ParticipantProjectSpeciesStore

nickgraz:
- This will be used so assign a bulk list of species to any number of projects

PR #1935, opened 2024-05-06
SW-5318 Add deleteMany functionality to ParticipantProjectSpecies

nickgraz:
- Add `deleteMany` functionality to `ParticipantProjectSpeciesStore`
- Add `DELETE  /api/v1/accelerator/projects/species` endpoint

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
@@ -33,6 +33,9 @@ class ParticipantNotFoundException(id: ParticipantId) :
 class ParticipantProjectSpeciesNotFoundException(id: ParticipantProjectSpeciesId) :
     EntityNotFoundException("Participant Project Species $id not found")
 
+class ParticipantProjectSpeciesSetNotFoundException(ids: Set<ParticipantProjectSpeciesId>) :
+    EntityNotFoundException("Participant Project Species ${ids.joinToString { ", " }} not found")
+
 class ProjectDocumentSettingsNotConfiguredException(id: ProjectId) :
     MismatchedStateException("Project $id document upload settings have not been configured")
 
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
@@ -33,9 +33,6 @@ class ParticipantNotFoundException(id: ParticipantId) :
 class ParticipantProjectSpeciesNotFoundException(id: ParticipantProjectSpeciesId) :
     EntityNotFoundException("Participant Project Species $id not found")
 
-class ParticipantProjectSpeciesSetNotFoundException(ids: Set<ParticipantProjectSpeciesId>) :
-    EntityNotFoundException("Participant Project Species ${ids.joinToString { ", " }} not found")
-
 class ProjectDocumentSettingsNotConfiguredException(id: ProjectId) :
     MismatchedStateException("Project $id document upload settings have not been configured")
 
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
@@ -79,29 +79,15 @@ class ParticipantProjectSpeciesStore(
     }
   }
 
-  fun deleteMany(participantProjectSpeciesIds: Set<ParticipantProjectSpeciesId>) {
+  fun delete(participantProjectSpeciesIds: Set<ParticipantProjectSpeciesId>) {
     participantProjectSpeciesIds.forEach {
       requirePermissions { deleteParticipantProjectSpecies(it) }
     }
 
-    dslContext.transactionResult { _ ->
-      val rowsDeleted =
-          dslContext
-              .deleteFrom(PARTICIPANT_PROJECT_SPECIES)
-              .where(PARTICIPANT_PROJECT_SPECIES.ID.`in`(participantProjectSpeciesIds))
-              .execute()
-
-      if (rowsDeleted != participantProjectSpeciesIds.size) {
-        val failed =
-            dslContext
-                .select(PARTICIPANT_PROJECT_SPECIES.ID)
-                .from(PARTICIPANT_PROJECT_SPECIES)
-                .where(PARTICIPANT_PROJECT_SPECIES.ID.`in`(participantProjectSpeciesIds))
-                .fetchSet(PARTICIPANT_PROJECT_SPECIES.ID.asNonNullable())
-
-        throw ParticipantProjectSpeciesSetNotFoundException(failed)
-      }
-    }
+    dslContext
+        .deleteFrom(PARTICIPANT_PROJECT_SPECIES)
+        .where(PARTICIPANT_PROJECT_SPECIES.ID.`in`(participantProjectSpeciesIds))
+        .execute()
   }
 
   fun fetchOneById(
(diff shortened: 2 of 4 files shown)


nickgraz:
* **#1937** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1935** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1934** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1933** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1932** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1932?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1924** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1924?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1919** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1918** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1918?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1917?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -78,6 +79,31 @@ class ParticipantProjectSpeciesStore(
...
+              .where(PARTICIPANT_PROJECT_SPECIES.ID.`in`(participantProjectSpeciesIds))
+              .execute()
+
+      if (rowsDeleted != participantProjectSpeciesIds.size) {
+        val failed =
+            dslContext
+                .select(PARTICIPANT_PROJECT_SPECIES.ID)
+                .from(PARTICIPANT_PROJECT_SPECIES)
+                .where(PARTICIPANT_PROJECT_SPECIES.ID.`in`(participantProjectSpeciesIds))
+                .fetchSet(PARTICIPANT_PROJECT_SPECIES.ID.asNonNullable())
```
sgrimm: Won't this always return an empty set?

src/main/kotlin/com/terraformation/backend/accelerator/db/Exceptions.kt
```diff
@@ -33,6 +33,9 @@ class ParticipantNotFoundException(id: ParticipantId) :
 class ParticipantProjectSpeciesNotFoundException(id: ParticipantProjectSpeciesId) :
     EntityNotFoundException("Participant Project Species $id not found")
 
+class ParticipantProjectSpeciesSetNotFoundException(ids: Set<ParticipantProjectSpeciesId>) :
+    EntityNotFoundException("Participant Project Species ${ids.joinToString { ", " }} not found")
+
```
sgrimm: This could be combined with the previous exception. Have it always take `Collection<ParticipantProjectSpeciesId>`, and add a secondary constructor that takes one ID and wraps it in a list or set.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -78,6 +79,31 @@ class ParticipantProjectSpeciesStore(
     }
   }
 
+  fun deleteMany(participantProjectSpeciesIds: Set<ParticipantProjectSpeciesId>) {
```
sgrimm: Bikeshedding a bit, and feel free to ignore this if you like the current name better: this could be named `delete`. There is no single-ID delete method and the fact that this takes a collection argument already documents that it can take more than one ID.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
```diff
@@ -265,4 +265,102 @@ class ParticipantProjectSpeciesStoreTest : DatabaseTest(), RunsAsUser {
...
+    fun `throws exception if an entry does not exist`() {
+      val participantId = insertParticipant()
+      val projectId = insertProject(participantId = participantId)
+      val speciesId = insertSpecies()
+      val participantProjectSpeciesId =
+          insertParticipantProjectSpecies(projectId = projectId, speciesId = speciesId)
+
+      every { user.canDeleteParticipantProjectSpecies(any()) } returns true
+
+      assertThrows<ParticipantProjectSpeciesSetNotFoundException> {
```
sgrimm: This test could do a more precise assertion on the exception to verify that its list of IDs is as expected.

If you do that, it might make sense to make the set of IDs a `val` property of the exception, so it gets retained after the exception object is constructed the test can look at the IDs directly rather than having to rely on the exception message being formatted a certain way.

sgrimm: In case you haven't run across this pattern before, it'd look something like
```kotlin
try {
  store.deleteMany(...)
  fail("Expected exception")
} catch (e: ParticipantProjectSpeciesSetNotFoundException) {
  assertEquals(setOf(ParticipantProjectSpeciesId(9999)), e.ids, "IDs that couldn't be deleted")
}
```

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: Good catch

nickgraz: The idea here is to verify that the IDs were deleted. Elsewhere I see to just check the row count and throw an error if it isn't what is expected. In this case I thought it would be nice to indicate which IDs failed to delete, but it seems like that would require a select before the delete to see which IDs wouldn't be deleted. Is that worth it, or am I adding unnecessary complexity?

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
nickgraz: Making the assertion more precise shows that the delete really wasn't pulling in the non-deleted IDs!

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
sgrimm: Probably unnecessary complexity, I agree. You _can_ do it in one query (the `RETURNING` clause lets you specify a list of columns to return from affected rows on a write operation) but probably not worth the bother.

nickgraz: I got it to work as expected, and satisfy the test case below, but will remove for simplicity.

sgrimm (approved):

nickgraz:
### Merge activity

* **May 8, 12:09 PM EDT**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1935).
* **May 8, 12:33 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1935) rebased this pull request as part of a merge.
* **May 8, 12:42 PM EDT**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1935).

PR #1942, opened 2024-05-07
Add 'Species List' Deliverable Type to deliverable importer

nickgraz:
- Throws a descriptive error if the value is not "document" or "species".

Here's the CSV I used to test with

[DD Phase - Deliverable and Module Data - Deliverables - 240507.csv](https://github.com/terraware/terraware-server/files/15241217/DD.Phase.-.Deliverable.and.Module.Data.-.Deliverables.-.240507.csv)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverablesImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverablesImporter.kt
@@ -33,16 +33,19 @@ class DeliverablesImporter(
     private const val COLUMN_CATEGORY = COLUMN_MODULE_ID + 1
     private const val COLUMN_SENSITIVE = COLUMN_CATEGORY + 1
     private const val COLUMN_REQUIRED = COLUMN_SENSITIVE + 1
-    private const val DELIVERABLE_TYPE = COLUMN_REQUIRED + 1
-    private const val NUM_COLUMNS = DELIVERABLE_TYPE + 1
+    private const val COLUMN_DELIVERABLE_TYPE = COLUMN_REQUIRED + 1
+    private const val NUM_COLUMNS = COLUMN_DELIVERABLE_TYPE + 1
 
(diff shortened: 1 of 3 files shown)


nickgraz:
* **#1942** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverablesImporter.kt
```diff
@@ -73,6 +74,16 @@ class DeliverablesImporter(
...
+        val deliverableType =
+            values[DELIVERABLE_TYPE]?.lowercase()?.let {
+              when (it) {
+                "document" -> DeliverableType.Document
+                "species" -> DeliverableType.Species
+                else ->
+                    throw IllegalArgumentException(
+                        "Invalid Document Type on row $rowNumber - $it, must be one of: 'Document', 'Species'")
+              }
+            }
```
sgrimm: Is it valid to leave the field blank so this ends up being null? I would think we'd want to make the deliverable type mandatory, in which case following a pattern like we do for category might be better, where there's a map of lowercase names to enums and this becomes something like
```kotlin
val deliverableType = values[DELIVERABLE_TYPE]?.lowerCase()?.let { deliverableTypesByName[it] }
    ?: throw IllegalArgumentException("same message you have in the PR")
```

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverablesImporter.kt
nickgraz: I ended up following the pre-existing pattern for validating as seen with the category column

sgrimm (approved):

PR #1945, opened 2024-05-10
Ensure a species deliverable submission when a species are added to a project

nickgraz:
- Adds a function to get the active species deliverable for a project if it exists
- Added service class `ParticipantProjectSpeciesService` which ensures a deliverable submission exists for the project when a species is added to a project with an active species deliverable

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
@@ -1,0 +1,45 @@
+package com.terraformation.backend.accelerator
+
+import com.terraformation.backend.accelerator.db.ParticipantProjectSpeciesStore
+import com.terraformation.backend.accelerator.db.SubmissionStore
+import com.terraformation.backend.accelerator.model.ExistingParticipantProjectSpeciesModel
+import com.terraformation.backend.accelerator.model.NewParticipantProjectSpeciesModel
+import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.default_schema.SpeciesId
+import jakarta.inject.Named
+
+@Named
+class ParticipantProjectSpeciesService(
+    private val participantProjectSpeciesStore: ParticipantProjectSpeciesStore,
+    private val submissionStore: SubmissionStore,
+) {
+  /** Creates a new participant project species, possibly creating a deliverable submission. */
+  fun create(model: NewParticipantProjectSpeciesModel): ExistingParticipantProjectSpeciesModel {
+    val existingModel = participantProjectSpeciesStore.create(model)
+
+    // If a submission doesn't exist for the deliverable, create one
+    val deliverableSubmission =
+        submissionStore.fetchActiveSpeciesDeliverableSubmission(model.projectId)
+    if (deliverableSubmission.submissionId == null) {
+      submissionStore.createSubmission(deliverableSubmission.deliverableId, model.projectId)
+    }
+
+    return existingModel
+  }
+
+  /**
+   * Creates a participant project species for each projectId - speciesId combination and create a
+   * species deliverable submission for each project that doesn't have one
+   */
+  fun create(projectIds: Set<ProjectId>, speciesIds: Set<SpeciesId>): Unit {
+    participantProjectSpeciesStore.create(projectIds, speciesIds)
+
+    projectIds.forEach {
+      // A submission must exist for every project that is getting a new species assigned
+      val deliverableSubmission = submissionStore.fetchActiveSpeciesDeliverableSubmission(it)
+      if (deliverableSubmission.submissionId == null) {
+        submissionStore.createSubmission(deliverableSubmission.deliverableId, it)
+      }
+    }
+  }
+}
--- a/src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
@@ -7,24 +7,28 @@ import com.terraformation.backend.accelerator.model.NewParticipantProjectSpecies
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.SpeciesId
 import jakarta.inject.Named
+import org.jooq.DSLContext
 
 @Named
 class ParticipantProjectSpeciesService(
+    private val dslContext: DSLContext,
     private val participantProjectSpeciesStore: ParticipantProjectSpeciesStore,
     private val submissionStore: SubmissionStore,
 ) {
   /** Creates a new participant project species, possibly creating a deliverable submission. */
   fun create(model: NewParticipantProjectSpeciesModel): ExistingParticipantProjectSpeciesModel {
-    val existingModel = participantProjectSpeciesStore.create(model)
+    return dslContext.transactionResult { _ ->
+      val existingModel = participantProjectSpeciesStore.create(model)
 
-    // If a submission doesn't exist for the deliverable, create one
-    val deliverableSubmission =
-        submissionStore.fetchActiveSpeciesDeliverableSubmission(model.projectId)
-    if (deliverableSubmission.submissionId == null) {
-      submissionStore.createSubmission(deliverableSubmission.deliverableId, model.projectId)
+      // If a submission doesn't exist for the deliverable, create one
+      val deliverableSubmission =
+          submissionStore.fetchActiveSpeciesDeliverableSubmission(model.projectId)
+      if (deliverableSubmission.submissionId == null) {
+        submissionStore.createSubmission(deliverableSubmission.deliverableId, model.projectId)
+      }
+
+      existingModel
     }
-
-    return existingModel
   }
 
   /**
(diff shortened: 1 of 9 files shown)


nickgraz:
* **#1945** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
```diff
@@ -27,11 +31,61 @@ class SubmissionStore(
     private val dslContext: DSLContext,
     private val eventPublisher: ApplicationEventPublisher,
 ) {
-  fun fetchOneById(
```
nickgraz, line 30: Removed this because it wasn't being used anywhere

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -23,6 +23,7 @@ class ParticipantProjectSpeciesStore(
     private val dslContext: DSLContext,
     private val participantProjectSpeciesDao: ParticipantProjectSpeciesDao,
     private val projectsDao: ProjectsDao,
+    private val submissionStore: SubmissionStore,
```
sgrimm: I normally try to avoid having store classes depend on each other both to keep the layering of components more consistent and to reduce the risk of dependency cycles.

There are two approaches the server uses to avoid this kind of layering violation:

1. Introduce a service class that depends on both store classes and calls into both.
2. Publish an application event on one side and handle it on the other.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: Makes sense, I will make a service class, thanks.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -0,0 +1,45 @@
...
+  /** Creates a new participant project species, possibly creating a deliverable submission. */
+  fun create(model: NewParticipantProjectSpeciesModel): ExistingParticipantProjectSpeciesModel {
+    val existingModel = participantProjectSpeciesStore.create(model)
+
+    // If a submission doesn't exist for the deliverable, create one
+    val deliverableSubmission =
+        submissionStore.fetchActiveSpeciesDeliverableSubmission(model.projectId)
+    if (deliverableSubmission.submissionId == null) {
+      submissionStore.createSubmission(deliverableSubmission.deliverableId, model.projectId)
+    }
```
sgrimm: Wrapping this (and the other one) in a transaction would probably be good so you don't have a dangling participant project species if creating the submission fails for some reason.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -36,7 +36,7 @@ class ParticipantProjectSpeciesController(
   fun assignParticipantProjectSpecies(
       @RequestBody payload: AssignParticipantProjectSpeciesPayload
   ): SimpleSuccessResponsePayload {
-    participantProjectSpeciesStore.createMany(payload.projectIds, payload.speciesIds)
+    participantProjectSpeciesStore.create(payload.projectIds, payload.speciesIds)
```
sgrimm: Should this call the service instead?

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
```diff
@@ -27,11 +31,70 @@ class SubmissionStore(
...
+            .from(DELIVERABLES)
+            .join(MODULES)
+            .on(DELIVERABLES.MODULE_ID.eq(MODULES.ID))
+            .join(COHORT_MODULES)
+            .on(MODULES.ID.eq(COHORT_MODULES.MODULE_ID))
+            .join(PARTICIPANTS)
+            .on(COHORT_MODULES.COHORT_ID.eq(PARTICIPANTS.COHORT_ID))
+            .join(PROJECTS)
+            .on(PARTICIPANTS.ID.eq(PROJECTS.PARTICIPANT_ID))
+            .fullOuterJoin(SUBMISSIONS)
```
sgrimm: Why does this need to be a full outer join instead of a regular (left) outer join? I believe the condition on deliverable type would filter out any of the extra rows you'd get from a full join.

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
```diff
@@ -27,11 +31,70 @@ class SubmissionStore(
...
+            .on(
+                DELIVERABLES.ID.eq(SUBMISSIONS.DELIVERABLE_ID),
+                PROJECTS.ID.eq(SUBMISSIONS.PROJECT_ID))
+            .join(ORGANIZATIONS)
+            .on(PROJECTS.ORGANIZATION_ID.eq(ORGANIZATIONS.ID))
+            .where(COHORT_MODULES.START_DATE.lessOrEqual(today))
+            .and(COHORT_MODULES.END_DATE.greaterOrEqual(today))
+            .and(PROJECTS.ID.eq(projectId))
+            .and(DELIVERABLES.DELIVERABLE_TYPE_ID.eq(DeliverableType.Species))
+            .orderBy(DELIVERABLES.ID, PROJECTS.ID)
```
sgrimm: Project ID is a constant so won't affect ordering. But if there's more than one species list deliverable in the module, `fetchOne` will throw an exception. So no need for an `orderBy` here at all.

src/main/kotlin/com/terraformation/backend/accelerator/model/SpeciesDeliverableSubmissionModel.kt
```diff
@@ -0,0 +1,26 @@
...
+data class ExistingSpeciesDeliverableSubmissionModel(
+    val deliverableId: DeliverableId,
+    val submissionId: SubmissionId? = null
+) {
+  companion object {
+    fun of(record: Record): ExistingSpeciesDeliverableSubmissionModel {
+      if (record[SUBMISSIONS.ID] == null) {
+        return ExistingSpeciesDeliverableSubmissionModel(
+            deliverableId = record[DELIVERABLES.ID]!!, submissionId = null)
+      }
```
sgrimm: Removing the `!!` from the submission ID value in the other `return` statement would do this too.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -0,0 +1,120 @@
...
+    every { user.canCreateParticipantProjectSpecies(any()) } returns true
+    every { user.canCreateSubmission(any()) } returns true
+    every { user.canReadProject(any()) } returns true
+    every { user.canReadProjectDeliverables(any()) } returns true
+  }
+
+  @Nested
+  inner class Create {
+    @Test
+    fun `creates a submission for the project and deliverable if one does not exist for the active module when a species is added to a project`() {
```
sgrimm, line 48: Maybe worth a test to make sure it handles cases where there already is a deliverable.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
nickgraz: Yes, yes it should!

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
nickgraz: Left outer join works as expected.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -0,0 +1,51 @@
...
+      existingModel
+    }
+  }
+
+  /**
+   * Creates a participant project species for each projectId - speciesId combination and create a
+   * species deliverable submission for each project that doesn't have one
+   */
+  fun create(projectIds: Set<ProjectId>, speciesIds: Set<SpeciesId>): Unit {
+    return dslContext.transaction { _ ->
```
sgrimm: No need for the `: Unit` or the `return` here.

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
```diff
@@ -27,11 +31,69 @@ class SubmissionStore(
...
+            .join(PARTICIPANTS)
+            .on(COHORT_MODULES.COHORT_ID.eq(PARTICIPANTS.COHORT_ID))
+            .join(PROJECTS)
+            .on(PARTICIPANTS.ID.eq(PROJECTS.PARTICIPANT_ID))
+            .leftOuterJoin(SUBMISSIONS)
+            .on(
+                DELIVERABLES.ID.eq(SUBMISSIONS.DELIVERABLE_ID),
+                PROJECTS.ID.eq(SUBMISSIONS.PROJECT_ID))
+            .join(ORGANIZATIONS)
+            .on(PROJECTS.ORGANIZATION_ID.eq(ORGANIZATIONS.ID))
```
sgrimm: Is joining to `ORGANIZATIONS` needed? It doesn't look like the rest of the query references anything from it.

src/test/kotlin/com/terraformation/backend/accelerator/db/SubmissionStoreTest.kt
```diff
@@ -29,46 +30,114 @@ class SubmissionStoreTest : DatabaseTest(), RunsAsUser {
...
+      val deliverableIdActive =
+          insertDeliverable(moduleId = moduleIdActive, deliverableTypeId = DeliverableType.Species)
+
+      // Module goes from epoch + 12 days -> epoch + 18 days
+      val moduleIdFuture = insertModule()
+      insertCohortModule(cohortId = cohortId, moduleId = moduleIdFuture)
+      insertDeliverable(moduleId = moduleIdFuture, deliverableTypeId = DeliverableType.Species)
+
+      // epoch + 7 days
+      clock.instant = Instant.EPOCH.plusSeconds(60 * 60 * 24 * 7)
```
sgrimm: Alternately, if you like: `Instant.EPOCH.plus(7, ChronoUnit.DAYS)`

src/main/kotlin/com/terraformation/backend/accelerator/db/SubmissionStore.kt
nickgraz: It is not needed, removed.

sgrimm (approved):

PR #1946, opened 2024-05-10
Add modifiedTime to project species entity

nickgraz:
Add modifiedTime to project species entity

Add modified time

Put modifiedTime back into tests


nickgraz:
* **#1946** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1946?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -89,6 +94,31 @@ class ParticipantProjectSpeciesStore(
...
+    val lastUpdatedTime =
+        with(PARTICIPANT_PROJECT_SPECIES) {
+          dslContext
+              .select(ID, MODIFIED_TIME)
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .orderBy(MODIFIED_TIME.desc())
+              .limit(1)
+              .fetch()
+              .filter { user.canReadParticipantProjectSpecies(it[ID]!!) }
```
sgrimm: Obviously in the current implementation this will always return true if the user has permission to read the project at all, but in theory, this filter will cause the method to have some unexpected behavior:

1. You call this method, get back a timestamp
2. Someone else modifies a participant project species you don't have permission to read
3. You call this method, it throws a "not found" exception even though there are still other species you can read

Three possible solutions here, I think:

1. Get rid of the per-species permission check entirely, and say that if you have permission to read the project, you have permission to ask when its most recent participant project species update was
2. Get rid of the `limit(1)` so it'll return the most recent timestamp that's visible to you (this will of course cause the query to fetch the full list of results but hopefully the list won't be massive)
3. Bake the permission filtering into the query

Those are in descending order of my personal preference but any of them would work.

sgrimm (changes requested):
May as well throw the other three columns in too, yeah. They're very low cost, so better to have them and not end up using them than to need them later and have no way to backfill accurate data.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: I didn't think about that, thanks. I will go with option 1.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -89,6 +106,28 @@ class ParticipantProjectSpeciesStore(
...
+    val lastUpdatedTime =
+        with(PARTICIPANT_PROJECT_SPECIES) {
+          dslContext
+              .select(ID, MODIFIED_TIME)
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .orderBy(MODIFIED_TIME.desc())
+              .limit(1)
+              .fetch()
+              .map { it[MODIFIED_TIME]!! }
```
sgrimm: Alternately:
```kotlin
dslContext
    .select(DSL.max(MODIFIED_TIME))
    .from(this)
    .where(PROJECT_ID.eq(projectId))
    .fetchOne(MODIFIED_TIME)
```

src/main/resources/db/migration/0250/V276__NewParticipantProjectSpeciesFields.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE accelerator.participant_project_species ADD COLUMN created_by BIGINT REFERENCES users (id);
+ALTER TABLE accelerator.participant_project_species ADD COLUMN created_time TIMESTAMP WITH TIME ZONE NOT NULL;
```
sgrimm: This migration will fail if there are any existing rows in the table (which there are on staging; I just checked.)

Adding a non-nullable column needs three steps:
```sql
ALTER TABLE foo ADD COLUMN bar TIMESTAMP WITH TIME ZONE;
UPDATE foo SET bar = NOW();
ALTER TABLE foo ALTER COLUMN bar SET NOT NULL;
```

sgrimm (changes requested):

src/main/resources/db/migration/0250/V276__NewParticipantProjectSpeciesFields.sql
nickgraz: Thanks for that, I see plenty of examples for the same treatment for user ID too.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: For some reason, when running tests I now get:
```
java.lang.IllegalArgumentException: Field ("accelerator"."participant_project_species"."modified_time") is not contained in Row (max("accelerator"."participant_project_species"."modified_time"))
```

I can see the `modified_time` being inserted correctly when I debug the test though...

nickgraz: Ah, had to update the fetch to `.fetchOne(DSL.max(MODIFIED_TIME))`, works like a charm, thanks.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -89,6 +107,22 @@ class ParticipantProjectSpeciesStore(
         .execute()
   }
 
+  fun fetchLastUpdatedSpeciesTime(projectId: ProjectId): Instant {
+    requirePermissions { readProject(projectId) }
+
+    val lastUpdatedTime =
```
sgrimm, line 113: Could do `return with(...)` instead of declaring the variable.

src/main/resources/db/migration/0250/V276__NewParticipantProjectSpeciesFields.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE accelerator.participant_project_species ADD COLUMN created_by BIGINT REFERENCES users (id);
+ALTER TABLE accelerator.participant_project_species ADD COLUMN modified_by BIGINT REFERENCES users (id);
+UPDATE accelerator.participant_project_species SET (created_by, modified_by) = (
+    SELECT organizations.created_by, organizations.modified_by
```
sgrimm: What would you think of setting these to the system user's user ID instead?

sgrimm (approved):
Had a couple take-it-or-leave-it suggestions, but this looks good.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: I ended up doing this in the notifications PR [here](https://github.com/terraware/terraware-server/pull/1951/files#diff-8b96e3384ac4817c8be17f0fe30f860368647f9f6410550b609275bf7c912b06R192).

src/main/resources/db/migration/0250/V276__NewParticipantProjectSpeciesFields.sql
nickgraz: I suppose it is less of a lie that way. I will update.

PR #1951, opened 2024-05-13
Send notifications when project species are added or edited

nickgraz:
- Create `SpeciesNotifier` which handles debouncing (10 minutes) of species related notifications. 
- Send out in app notifications and emails when a species is added to a participant project.
- Send out in app notifications and emails when a species is modified if the status of the species is already approved.


nickgraz:
* **#1951** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/SpeciesNotifier.kt
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Publishes [DeliverableReadyForReviewEvent] if no documents have been uploaded for a submission
+   * since the one referenced by the event.
+   */
+  private fun notifyIfNoNewerUpdates(event: ParticipantProjectSpeciesAddedEvent) {
+    systemUser.run {
+      val lastUpdateTime =
+          participantProjectSpeciesStore.fetchLastUpdatedSpeciesTime(event.projectId)
+
+      if (lastUpdateTime == event.modifiedTime) {
```
sgrimm: I'm a little nervous about a strict equality check here just because `Instant` might have higher resolution than the `TIMESTAMP` type in PostgreSQL, and the event modified time here comes from a `clock.instant()` call.

One option would be to make this `<=` instead of `==` -- if the database value is lower resolution, it's probably always going to round down so as to avoid storing timestamps in the future.

Another option would be to make `ParticipantProjectSpeciesStore.create()` re-fetch the row from the database after inserting it, so the event is populated with the timestamp as it exists in the database.

src/main/kotlin/com/terraformation/backend/accelerator/SpeciesNotifier.kt
```diff
@@ -0,0 +1,97 @@
...
+    if (old.submissionStatus == SubmissionStatus.Approved && old != new) {
+      scheduler.schedule<SpeciesNotifier>(clock.instant().plus(notificationDelay)) {
+        notifyIfNoNewerUpdates(event)
+      }
+    }
+  }
+
+  /**
+   * Publishes [DeliverableReadyForReviewEvent] if no documents have been uploaded for a submission
+   * since the one referenced by the event.
```
sgrimm: Copypasta?

src/main/kotlin/com/terraformation/backend/accelerator/SpeciesNotifier.kt
```diff
@@ -0,0 +1,97 @@
...
+   */
+  private fun notifyIfNoNewerUpdates(event: ParticipantProjectSpeciesAddedEvent) {
+    systemUser.run {
+      val lastUpdateTime =
+          participantProjectSpeciesStore.fetchLastUpdatedSpeciesTime(event.projectId)
```
sgrimm: Should this be looking at creation time instead? Otherwise if you add a species and then edit an existing one, you'll never get the add notification since the latest modification time in the project will be newer. (Or maybe that's the desired notification behavior?)

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -148,24 +151,33 @@ class ParticipantProjectSpeciesStore(
...
+            ?: throw ParticipantProjectSpeciesNotFoundException(participantProjectSpeciesId)
+
+    val oldParticipantProjectSpecies =
+        ExistingParticipantProjectSpeciesModel.of(participantProjectSpecies)
+
+    participantProjectSpecies.feedback = updated.feedback
+    participantProjectSpecies.modifiedBy = currentUser().userId
+    participantProjectSpecies.modifiedTime = modifiedTime
+    participantProjectSpecies.rationale = updated.rationale
+    participantProjectSpecies.submissionStatusId = updated.submissionStatus
```
sgrimm: Happy to see someone else using `Record` instead of `Row` classes for this! Might be a little prettier to wrap this in `with(participantProjectSpecies)` so the assignments become `feedback = updated.feedback` and so on, though you'd need to rename the local `modifiedTime` variable to avoid name collision.

src/main/kotlin/com/terraformation/backend/accelerator/event/ParticipantProjectSpeciesAddedToProjectEvent.kt
```diff
@@ -0,0 +1,12 @@
+package com.terraformation.backend.accelerator.event
+
+import com.terraformation.backend.db.accelerator.DeliverableId
+import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.default_schema.SpeciesId
+
+/** Published when a participant project species is added to a project */
+data class ParticipantProjectSpeciesAddedToProjectEvent(
```
sgrimm: Reading this whole PR, I see the distinction between this event and the other one, but in isolation, they're kind of confusing. `ParticipantProjectSpeciesAddedEvent` also includes a project ID so the comment here would be true of it as well.

Maybe it'd be clearer to rename this to something that specifically mentions "notification due" so people don't think they can listen for this when they really want the other one.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -261,6 +265,71 @@ class AppNotificationService(
         setOf(Role.Owner, Role.Admin))
   }
 
+  @EventListener
+  fun on(event: ParticipantProjectSpeciesApprovedSpeciesEditedEvent) {
+    systemUser.run {
```
sgrimm: This is harmless but I don't think it's needed; the event will be published inside a `systemUser.run` block.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -167,6 +167,14 @@ notification.observation.upcoming.email.body.text.3=To complete observations, do
...
 notification.organization.email.buttonIntro=Please click the button below to go to your Terraware account where you can view the organization.
 notification.organization.email.linkIntro=Please click the link below to go to your Terraware account where you can view the organization.
+notification.participantProjectSpecies.added.ctaLabel=Review in Terraware
+notification.participantProjectSpecies.added.longBody={0} has submitted {1} for use for {2} and is ready for approval.
+notification.participantProjectSpecies.added.shortBody={0} has been submitted for use for {1}.
+notification.participantProjectSpecies.added.subject=A species has been added to a project for {0}.
+notification.participantProjectSpecies.edited.ctaLabel=Review in Terraware
+notification.participantProjectSpecies.edited.longBody={0} has edited {1} and is ready for approval.
+notification.participantProjectSpecies.edited.shortBody={0} has been edited and is ready for approval.
+notification.participantProjectSpecies.edited.subject=An approved species has been edited for {0}.
```
sgrimm: This string naming convention is new, yes? Would it be worth having a separate discussion on changing the naming of notification strings? I think it's preferable to have a consistent naming convention.

I guess my preference (but happy to discuss it) would be to use the existing naming convention for these strings and if we decide to switch to the scheme in this PR, we can update all the strings at once.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -160,6 +170,31 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
                   projectId = projectId2,
                   submissionStatusId = SubmissionStatus.NotSubmitted)),
           submissionsDao.fetchByDeliverableId(deliverableId).map { it.copy(id = null) })
+
+      eventPublisher.assertEventPublished(
```
sgrimm: There is also `TestEventPublisher.assertEventsPublished()` that takes a set of events to verify.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/SpeciesNotifier.kt
nickgraz: Good call, yes, I think it should be looking at the creation time for this one. I added the debounce for added was based on an assumption, which is yet to be confirmed or denied. Either way, the behavior I was going for is, in fact, based on the created time for "species added" events.

src/main/kotlin/com/terraformation/backend/accelerator/SpeciesNotifier.kt
nickgraz: I'll go with the second option, thanks

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: Pairs nicely with the other change since I no longer need to assign `modifiedTime` to a variable

src/main/resources/i18n/Messages_en.properties
nickgraz: Yea I wasn't really sure what the correct way to do it was with notifications that are both emailed and sent in app, without duplicating entries, so I just made it up. I will ask in Slack and figure out the correct way.

nickgraz: Going to put this back to the pre-existing conventions

src/main/kotlin/com/terraformation/backend/accelerator/event/ParticipantProjectSpeciesAddedToProjectEvent.kt
nickgraz: Good call, I was hoping making the name a bit more specific would indicate the notification intended, but you're right, it isn't obvious.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
nickgraz: I ended up having to add a permission allowance for reading global roles in the service test.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -35,17 +45,36 @@ class ParticipantProjectSpeciesService(
...
+        }
+
+        val deliverableSubmission =
+            submissionStore.fetchActiveSpeciesDeliverableSubmission(it.projectId)
         if (deliverableSubmission.submissionId == null) {
-          submissionStore.createSubmission(deliverableSubmission.deliverableId, it)
+          submissionStore.createSubmission(deliverableSubmission.deliverableId, it.projectId)
         }
+
+        checkedProjectIds.plus(it.projectId)
```
sgrimm: I suspect `checkedProjectIds` should be a mutable set, and then this would be `add` instead of `plus`. `plus` returns a new set with the added element, and its return value is immediately discarded here.

But the tests are passing despite this, so either there's a missing test case or this isn't needed at all.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -35,17 +45,36 @@ class ParticipantProjectSpeciesService(
...
+
+        eventPublisher.publishEvent(
+            ParticipantProjectSpeciesAddedEvent(
+                deliverableId = deliverableSubmission.deliverableId,
+                participantProjectSpecies = it))
```
sgrimm: Since the lambda function here is several lines long, it'd be slightly clearer to declare an explicit variable name on the `forEach` rather than using `it`.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -176,4 +182,29 @@ class ParticipantProjectSpeciesStore(
...
+    }
+  }
+
+  private fun fetchOneRecordById(
+      participantProjectSpeciesId: ParticipantProjectSpeciesId
+  ): ParticipantProjectSpeciesRecord =
+      dslContext
+          .selectFrom(PARTICIPANT_PROJECT_SPECIES)
+          .where(PARTICIPANT_PROJECT_SPECIES.ID.eq(participantProjectSpeciesId))
+          .fetchOne()
```
sgrimm: Alternately:
```kotlin
dslContext.fetchOne(
    PARTICIPANT_PROJECT_SPECIES,
    PARTICIPANT_PROJECT_SPECIES.ID.eq(participantProjectSpeciesId))
```
But this way is totally fine too.

src/main/kotlin/com/terraformation/backend/accelerator/event/ParticipantProjectSpeciesAddedToProjectNotificationDueEvent.kt
```diff
@@ -0,0 +1,12 @@
+package com.terraformation.backend.accelerator.event
+
+import com.terraformation.backend.db.accelerator.DeliverableId
+import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.default_schema.SpeciesId
+
+/** Published when a participant project species is added to a project */
```
sgrimm: Comment should probably mention something about this being a deferred, throttled event rather than one that's published immediately when a participant project species is added to a project.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -167,6 +167,28 @@ notification.observation.upcoming.email.body.text.3=To complete observations, do
...
 notification.organization.email.buttonIntro=Please click the button below to go to your Terraware account where you can view the organization.
 notification.organization.email.linkIntro=Please click the link below to go to your Terraware account where you can view the organization.
+# {0} is the scientific name of a species. {1} is the project name it was added to.
+notification.participantProjectSpecies.added.app.body={0} has been submitted for use for {1}.
+# {0} is the participant name.
+notification.participantProjectSpecies.added.app.title=A species has been added to a project for {0}.
+# {0} is the scientific name of a species.
+notification.participantProjectSpecies.edited.app.body={0} has been edited and is ready for approval.
+# {0} is the participant name.
+notification.participantProjectSpecies.edited.app.title=An approved species has been edited for {0}.
```
sgrimm: Just noting that Phrase will move these down to be in alphabetical order when it publishes its PR to add translations.

src/main/resources/templates/email/participantProjectSpecies/added/body.txt.ftl
```diff
@@ -0,0 +1,10 @@
+<#-- @ftlvariable name="" type="com.terraformation.backend.email.model.ParticipantProjectSpeciesAdded" -->
+${strings("notification.participantProjectSpecies.added.email.subject", participantName)}
+
+${strings("notification.participantProjectSpecies.added.email.body.1", speciesName, projectName)}
```
sgrimm: Should this be `body.1` and `body.2` instead of `subject` and `body`? Same with the plaintext file for the edit event.

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -57,14 +62,15 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
...
 
-      service.create(
-          NewParticipantProjectSpeciesModel(
-              feedback = "feedback",
-              id = null,
-              modifiedTime = Instant.EPOCH,
-              projectId = projectId,
-              rationale = "rationale",
-              speciesId = speciesId))
+      val existingModel =
```
sgrimm, line 72: Would it make sense to assert that this model looks like it's supposed to, either in addition to or instead of the assertion on `SubmissionsRow`?

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
```diff
@@ -42,7 +46,37 @@ class ParticipantProjectSpeciesStoreTest : DatabaseTest(), RunsAsUser {
...
+      val participantId = insertParticipant()
+      val projectId = insertProject(participantId = participantId)
+      val speciesId1 = insertSpecies()
+      val speciesId2 = insertSpecies()
+      val speciesId3 = insertSpecies()
+
+      insertParticipantProjectSpecies(
+          createdTime = Instant.EPOCH, projectId = projectId, speciesId = speciesId1)
+      insertParticipantProjectSpecies(
+          createdTime = Instant.EPOCH.plusSeconds(2), projectId = projectId, speciesId = speciesId2)
```
sgrimm: Maybe set `modifiedTime` to a different value to make sure the function is actually returning the created time?

sgrimm (changes requested):
Just some minor stuff; this is close.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
nickgraz: Yes this was supposed to save a request to `fetchActiveSpeciesDeliverableSubmission` for each additional species within a single project. It is idempotent so it doesn't matter if it runs many times, but I will fix it up so _actually_ does what I intend and make sure it is proven in a test.

src/main/resources/templates/email/participantProjectSpecies/added/body.txt.ftl
nickgraz: Good catch!

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
nickgraz: I figured that was covered adequately by the store tests, but certainly wouldn't hurt to add it here

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -52,29 +53,42 @@ class ParticipantProjectSpeciesService(
...
+      val projectDeliverableIds = mutableMapOf<ProjectId, DeliverableId>()
 
-      existingModels.forEach {
-        // A submission must exist for every project that is getting a new species assigned
-        if (checkedProjectIds.contains(it.projectId)) {
+      existingModels.forEach { participantProjectSpecies ->
+        if (projectDeliverableIds.contains(participantProjectSpecies.projectId)) {
+          publishEditEvent(
+              projectDeliverableIds[participantProjectSpecies.projectId]!!,
+              participantProjectSpecies)
```
sgrimm: This is totally fine as-is. Two alternate ideas, in case you prefer one of them:
```kotlin
if (participantProjectSpecies.projectId in projectDeliverableIds) { ... }
```
```kotlin
projectDeliverableIds[participantProjectSpecies.projectId]?.let {
  publishEditEvent(it, participantProjectSpecies)
  return@forEach
}
```

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -52,29 +53,42 @@ class ParticipantProjectSpeciesService(
...
+            deliverableSubmission.deliverableId
+        publishEditEvent(
+            projectDeliverableIds[participantProjectSpecies.projectId]!!, participantProjectSpecies)
       }
 
       existingModels
     }
   }
+
+  private fun publishEditEvent(
```
sgrimm: Maybe call this `publishAddedEvent` since there is also an "edited" event which people might assume this publishes.

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -177,6 +197,11 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
             ParticipantProjectSpeciesAddedEvent(
                 deliverableId = deliverableId, participantProjectSpecies = it)
           })
+
+      io.mockk.verify(exactly = 1) {
```
sgrimm: We normally use static imports for MockK functions, so this would be `verify(exactly=1)` here.

sgrimm: Also, there could be a comment here that this is to test a performance optimization, rather than to test correctness; if we introduce some form of application-wide caching at some point, this might start failing even though the end result is correct, and a comment will make it easy for someone to adjust/remove this check as needed.

sgrimm (approved):
Had a couple minor comments, but nothing that requires a re-review; this looks good.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
nickgraz: Thanks, that second one looks like the way to do it in Kotlin

PR #1952, opened 2024-05-14
Add DSL for constructing planting sites in tests

sgrimm:
To support tests that need to construct planting site maps in specific
configurations, add a builder class that acts as a planting site DSL.
This allows site maps to be specified succinctly, which makes the test cases
easier to follow.

This is mainly motivated by the upcoming support for planting site map
editing, the tests for which require constructing lots of site maps. But it
also applies to some existing tests. `PlantingSiteModelTest` is updated to
use it.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
@@ -121,8 +121,8 @@ private constructor(
   fun zone(
       x: Int = nextZoneX,
       y: Int = 0,
-      width: Int = this.width + this.x - x,
-      height: Int = this.height + this.y - y,
+      width: Int = this.width - (x - this.x),
+      height: Int = this.height - (y - this.y),
       name: String? = null,
       func: ZoneBuilder.() -> Unit = {}
   ): ExistingPlantingZoneModel {
@@ -168,8 +168,8 @@ private constructor(
     fun subzone(
         x: Int = nextSubzoneX,
         y: Int = 0,
-        width: Int = this.width + this.x - x,
-        height: Int = this.height + this.y - y,
+        width: Int = this.width - (x - this.x),
+        height: Int = this.height - (y - this.y),
         name: String? = null,
         func: SubzoneBuilder.() -> Unit = {},
     ): ExistingPlantingSubzoneModel {
--- a/src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
@@ -47,14 +47,18 @@ import org.locationtech.jts.geom.PrecisionModel
  *
  * Returns a planting site with:
  * - A boundary of 1000 by 500 meters
- * - Zone Z1 with a boundary of 400 by 500 meters on the west edge of the site
- *     - Subzone S1 with a boundary of 150 by 500 meters on the west edge of the zone
+ * - Zone Z1 with a boundary of 400 by 500 meters whose southwest corner is the southwest corner of
+ *   the site's boundary
+ *     - Subzone S1 with a boundary of 150 by 500 meters whose southwest corner is the southwest
+ *       corner of the zone (and thus of the site)
  *         - Cluster 1 with plots 1, 2, 3, 4 arranged counterclockwise starting at the southwest
- *           corner of the subzone
- *         - Plot 5 immediately to the east of plot 2 (50 meters east of the subzone's SW corner)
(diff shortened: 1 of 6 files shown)


sgrimm:
* **#1956** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1955** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1954** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1953** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1953?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1952** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
```diff
@@ -0,0 +1,270 @@
...
+
+/**
+ * DSL for building planting site models with simple maps to test operations on site/zone/subzone
+ * geometry. The different layers of the map are all rectangular and are laid out horizontally
+ * starting at the southwest corner of the parent area.
+ *
+ * The DSL aims for succinctness and will try to build a simple but complete site if the caller
+ * doesn't specify otherwise.
+ *
+ * By default, each layer fills the remaining area of its parent. If a site has no zones, one is
```
alexTerraformation, line 24: I think client might also be doing this.  Would be good to double-check to see if we have redundancy @tommylau523

sgrimm: Not sure I understand the comment: are you saying we should make sure the client does this too, or that because the client does it, we don't need it in the code in this PR?

alexTerraformation: Sorry I should've been more thorough in my response.  I think the client is already doing this (or will be once our user-drawn planting site feature is merged) because the UI prompts users for the "simple" approach or the "detailed" approach to drawing, which determines whether the user is asked to draw zones/subzones or not.  Obviously this doesn't cover shapefile uploads though.

Mainly I was pointing it out in case we wanted to remove the code from either client or server.

sgrimm: Got it. I think there isn't any actual redundancy because the code in this PR is part of the test suite. There's no client code involved in the tests, and the code in this PR isn't included in the server.

alexTerraformation: Ah, my misunderstanding then.  Should've read more carefully.

src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
```diff
@@ -0,0 +1,270 @@
...
+          numPermanentClusters = numPermanentClusters,
+          numTemporaryPlots = numTemporaryPlots,
+          plantingSubzones = plantingSubzones.ifEmpty { listOf(subzone()) },
+      )
+    }
+
+    fun subzone(
+        x: Int = nextSubzoneX,
+        y: Int = 0,
+        width: Int = this.width + this.x - x,
```
tommylau523: I know this is mathematically the same, but I think it reads just a bit easier with `this.width - (x - this.x)`. So it becomes a bit more obvious that we are subtracting the `x` offset from the zone width.

tommylau523 (approved):
Looks good! Just one minor comment.

src/test/kotlin/com/terraformation/backend/tracking/model/PlantingSiteBuilder.kt
```diff
@@ -0,0 +1,270 @@
...
+ *           plot()
+ *         }
+ *         subzone()
+ *       }
+ *       zone()
+ *     }
+ *
+ * Returns a planting site with:
+ * - A boundary of 1000 by 500 meters
+ * - Zone Z1 with a boundary of 400 by 500 meters on the west edge of the site
```
nickgraz: Is it worth mentioning the Y edge here?

nickgraz (approved):
This is great, looks easy to use

sgrimm:
### Merge activity

* **May 16, 1:27 PM EDT**: @sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1952).

PR #1953, opened 2024-05-14
SW-5228 Calculate changes to site maps

sgrimm:
In preparation for allowing editing of planting site maps, add logic to
calculate the differences between an existing planting site and a site map
requested by the user, and to detect edits that are disallowed according to the
product requirements.

The internal representation of a site edit will be used as the basis of both the
confirmation step in the UI ("this edit would do X, Y, and Z") and as input to
an update function that will apply the requested edits.

The code in this change only concerns itself with site geometry, not with any
of the other edit behaviors related to active observations or the presence of
planting data. Upcoming changes will cover those requirements.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
@@ -262,10 +262,10 @@ class PlantingSiteEditCalculatorTest {
           }
 
       assertHasProblem(
-          existing,
-          desired,
           PlantingSiteEditProblem.ZoneBoundaryChanged(
-              conflictsWith = setOf("Z1", "Z2"), zoneName = "Z1"))
+              conflictsWith = setOf("Z1", "Z2"), zoneName = "Z1"),
+          existing,
+          desired)
     }
 
     @Test
@@ -286,10 +286,10 @@ class PlantingSiteEditCalculatorTest {
           }
 
       assertHasProblem(
-          existing,
-          desired,
           PlantingSiteEditProblem.SubzoneBoundaryChanged(
-              conflictsWith = setOf("S1", "S2"), subzoneName = "S1", zoneName = "Z1"))
+              conflictsWith = setOf("S1", "S2"), subzoneName = "S1", zoneName = "Z1"),
+          existing,
+          desired)
     }
 
     @Test
@@ -302,10 +302,10 @@ class PlantingSiteEditCalculatorTest {
           }
 
       assertHasProblem(
-          existing,
-          desired,
           PlantingSiteEditProblem.CannotSplitZone(
-              conflictsWith = setOf("N1", "N2"), zoneName = "Z1"))
+              conflictsWith = setOf("N1", "N2"), zoneName = "Z1"),
+          existing,
+          desired)
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


sgrimm:
* **#1956** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1956?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1955** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1954** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1954?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1953** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1953?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1952** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1952?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
```diff
@@ -0,0 +1,371 @@
...
+        site,
+        site.toNew(),
+        PlantingSiteEdit(
+            BigDecimal("0.0"), site.boundary!!, null, PlantingSiteId(1), emptyList(), emptyList()))
+  }
+
+  @Nested
+  inner class Validation {
+    @Test
+    fun `detects change to boundary between existing zones`() {
```
alexTerraformation, line 370: Do these cases assume plants have been allocated to the site?

src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
```diff
@@ -0,0 +1,371 @@
...
+                                ))))))
+  }
+
+  @Test
+  fun `adds permanent clusters in proportion to size of added area`() {
+    val existing = existingSite(width = 10000)
+    val doubleSize = newSite(width = 20000)
+    val halfAgainSize = newSite(width = 15000)
+    val slightlyLargerSize = newSite(width = 10051)
+
```
alexTerraformation, line 270: Might be good to have a test where the boundary stays the same but an exclusion area is removed

src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
alexTerraformation, line 370: Worth it to have a test that tries to completely remove a zone/subzone?

sgrimm: Yes, this logic would only run if the site has plants. If it doesn't, my plan is to just delete the old map data entirely (saving it in history, of course) and create the new map since there's no need to preserve IDs and the fact that we allow arbitrary edits in that case means that the logic for matching old and new regions might fail even on valid edits.

sgrimm: Ah, good catch. Yes, should have added a case for this.

tommylau523 (approved):

sgrimm:
### Merge activity

* **May 16, 1:27 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1953) rebased this pull request after merging its parent, because this pull request is set to merge when ready.
* **May 16, 1:42 PM EDT**: @sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1953).

PR #1961, opened 2024-05-17
Add a couple more Participant Project Species fields

nickgraz:
- Add `internalComments` field to the entity
- Add enum table for `species_native_categories` selection, and add `species_native_category_id` to the entity


nickgraz:
* **#1961** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/db/migration/0250/V277__NewParticipantProjectSpeciesFields.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE accelerator.participant_project_species ADD COLUMN internal_comment TEXT;
+
+CREATE TABLE native_non_native (
```
sgrimm: The name here seems a bit too specific for this to be an enum. If we add a third choice at some point, then `native_non_native` might become a misleading name for it. If we never add a third choice, the native status could be a boolean column on `participant_project_species` instead of an enum.

If we keep it as an enum, something like `species_native_types` or `species_native_categories` would still be accurate if there are more choices later, and (more subjectively, I admit) names like those also make it clearer at a glance that the table is an enum table that relates to species.

sgrimm (changes requested):
The meat of this change seems fine; I just have feedback on a naming choice.

src/main/resources/db/migration/0250/V277__NewParticipantProjectSpeciesFields.sql
nickgraz: Good call. I asked about this in the PRD but haven't gotten an answer about a more general name for it. What you suggest seems like a good alternative.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -129,6 +132,8 @@ data class DeleteParticipantProjectSpeciesPayload(
 data class ParticipantProjectSpeciesPayload(
     val feedback: String?,
     val id: ParticipantProjectSpeciesId,
+    val internalComment: String?,
+    val speciesNativeCategory: SpeciesNativeCategory?,
```
sgrimm: Do you want these alphabetized?

sgrimm (approved):
Change looks good, but the rename messed up the alphabetical ordering of the field/table names.

PR #1974, opened 2024-05-21
Add two new endpoints for Participant Project Species 

nickgraz:
Add two new endpoints (in lieu of using the Search API) for:
- Retrieving projects associated to species through the Participant Project Species (PPS)
  - An active deliverable ID (according to the cohort module dates) will be included if it exists
- Retrieving species associated to a project through PPS

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
@@ -4,6 +4,8 @@ import com.terraformation.backend.accelerator.ParticipantProjectSpeciesService
 import com.terraformation.backend.accelerator.db.ParticipantProjectSpeciesStore
 import com.terraformation.backend.accelerator.model.ExistingParticipantProjectSpeciesModel
 import com.terraformation.backend.accelerator.model.NewParticipantProjectSpeciesModel
+import com.terraformation.backend.accelerator.model.ParticipantProjectsForSpecies
+import com.terraformation.backend.accelerator.model.SpeciesForParticipantProject
 import com.terraformation.backend.api.AcceleratorEndpoint
 import com.terraformation.backend.api.ApiResponse200
 import com.terraformation.backend.api.ApiResponse404
@@ -12,7 +14,6 @@ import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.db.accelerator.DeliverableId
 import com.terraformation.backend.db.accelerator.ParticipantProjectSpeciesId
 import com.terraformation.backend.db.accelerator.SubmissionStatus
-import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.SpeciesId
 import com.terraformation.backend.db.default_schema.SpeciesNativeCategory
@@ -24,11 +25,10 @@ import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
 @AcceleratorEndpoint
-@RequestMapping("/api/v1/accelerator/")
+@RequestMapping("/api/v1/accelerator")
 @RestController
 class ParticipantProjectSpeciesController(
     private val participantProjectSpeciesService: ParticipantProjectSpeciesService,
@@ -95,24 +95,12 @@ class ParticipantProjectSpeciesController(
       summary =
           "Gets all participant projects associated to a species with active deliverable information if applicable.")
   fun getProjectsForSpecies(
-      @PathVariable speciesId: SpeciesId,
-      @RequestParam organizationId: OrganizationId
+      @PathVariable speciesId: SpeciesId
   ): GetParticipantProjectsForSpeciesResponsePayload {
-    val results =
-        participantProjectSpeciesStore.fetchParticipantProjectsForSpecies(organizationId, speciesId)
+    val results = participantProjectSpeciesStore.fetchParticipantProjectsForSpecies(speciesId)
 
     return GetParticipantProjectsForSpeciesResponsePayload(
-        results.map {
-          ParticipantProjectForSpeciesPayload(
-              activeDeliverableId = it.activeDeliverableId,
-              participantProjectSpeciesId = it.participantProjectSpeciesId,
-              participantProjectSpeciesSubmissionStatus =
-                  it.participantProjectSpeciesSubmissionStatus,
-              projectId = it.projectId,
-              projectName = it.projectName,
-              speciesId = it.speciesId,
-          )
-        })
+        results.map { ParticipantProjectForSpeciesPayload(it) })
   }
 
   @ApiResponse200
@@ -125,18 +113,7 @@ class ParticipantProjectSpeciesController(
     val results = participantProjectSpeciesStore.fetchSpeciesForParticipantProjects(projectId)
 
     return GetSpeciesForParticipantProjectsResponsePayload(
-        results.map {
-          SpeciesForParticipantProjectPayload(
-              participantProjectSpeciesId = it.participantProjectSpeciesId,
-              participantProjectSpeciesRationale = it.participantProjectSpeciesRationale,
-              participantProjectSpeciesSubmissionStatus =
-                  it.participantProjectSpeciesSubmissionStatus,
-              projectId = it.projectId,
-              speciesId = it.speciesId,
-              speciesCommonName = it.speciesCommonName,
-              speciesScientificName = it.speciesScientificName,
-          )
-        })
+        results.map { SpeciesForParticipantProjectPayload(it) })
   }
 
   @ApiResponse200
@@ -205,7 +182,17 @@ data class ParticipantProjectForSpeciesPayload(
     val projectId: ProjectId,
     val projectName: String,
     val speciesId: SpeciesId,
-)
+) {
+  constructor(
+      model: ParticipantProjectsForSpecies
+  ) : this(
+      activeDeliverableId = model.activeDeliverableId,
+      participantProjectSpeciesId = model.participantProjectSpeciesId,
+      participantProjectSpeciesSubmissionStatus = model.participantProjectSpeciesSubmissionStatus,
+      projectId = model.projectId,
+      projectName = model.projectName,
+      speciesId = model.speciesId)
+}
 
 data class GetParticipantProjectsForSpeciesResponsePayload(
     val participantProjectsForSpecies: List<ParticipantProjectForSpeciesPayload>,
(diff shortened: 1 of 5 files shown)


nickgraz:
* **#1979** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1979?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1974** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -22,17 +24,18 @@ import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
 @AcceleratorEndpoint
-@RequestMapping("/api/v1/accelerator/projects/species")
+@RequestMapping("/api/v1/accelerator/")
```
sgrimm: No need for the trailing slash here.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -87,7 +90,58 @@ class ParticipantProjectSpeciesController(
...
   @ApiResponse200
   @ApiResponse404
-  @PutMapping("/{participantProjectSpeciesId}")
+  @GetMapping("/species/{speciesId}/projects")
+  @Operation(
+      summary =
+          "Gets all participant projects associated to a species with active deliverable information if applicable.")
+  fun getProjectsForSpecies(
+      @PathVariable speciesId: SpeciesId,
+      @RequestParam organizationId: OrganizationId
```
sgrimm: Species are per-organization so you should always be able to derive the organization ID from the species ID; no need to make the client pass it in.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -87,7 +90,58 @@ class ParticipantProjectSpeciesController(
...
+    return GetParticipantProjectsForSpeciesResponsePayload(
+        results.map {
+          ParticipantProjectForSpeciesPayload(
+              activeDeliverableId = it.activeDeliverableId,
+              participantProjectSpeciesId = it.participantProjectSpeciesId,
+              participantProjectSpeciesSubmissionStatus =
+                  it.participantProjectSpeciesSubmissionStatus,
+              projectId = it.projectId,
+              projectName = it.projectName,
+              speciesId = it.speciesId,
```
sgrimm: This could also be a secondary constructor on the payload class.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -144,6 +198,33 @@ data class GetParticipantProjectSpeciesResponsePayload(
...
+)
+
+data class GetParticipantProjectsForSpeciesResponsePayload(
+    val participantProjectsForSpecies: List<ParticipantProjectForSpeciesPayload>,
+) : SuccessResponsePayload
+
+data class SpeciesForParticipantProjectPayload(
+    val participantProjectSpeciesId: ParticipantProjectSpeciesId,
+    val participantProjectSpeciesRationale: String?,
+    val participantProjectSpeciesSubmissionStatus: SubmissionStatus,
```
sgrimm, line 204: These names seem pretty long, but I assume this is to clearly distinguish the per-species submission status from the deliverable-level one.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -119,6 +132,79 @@ class ParticipantProjectSpeciesStore(
...
+            PARTICIPANT_PROJECT_SPECIES.ID,
+            PARTICIPANT_PROJECT_SPECIES.SUBMISSION_STATUS_ID,
+            SPECIES.ID)
+        .from(SPECIES)
+        .join(PARTICIPANT_PROJECT_SPECIES)
+        .on(SPECIES.ID.eq(PARTICIPANT_PROJECT_SPECIES.SPECIES_ID))
+        .join(PROJECTS)
+        .on(PARTICIPANT_PROJECT_SPECIES.PROJECT_ID.eq(PROJECTS.ID))
+        .join(ORGANIZATIONS)
+        .on(PROJECTS.ORGANIZATION_ID.eq(ORGANIZATIONS.ID))
```
sgrimm: Is joining with `organizations` necessary here?

As I mentioned in an earlier comment, the organization is implied by the species, so this function doesn't need to take an organization ID argument at all. But if it were to take one, `.where(PROJECTS.ORGANIZATION_ID.eq(organizationId))` would work without this join.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -119,6 +132,79 @@ class ParticipantProjectSpeciesStore(
...
+            DELIVERABLES.MODULE_ID.eq(MODULES.ID),
+            DELIVERABLES.DELIVERABLE_TYPE_ID.eq(DeliverableType.Species))
+        .where(ORGANIZATIONS.ID.eq(organizationId))
+        .and(SPECIES.ID.eq(speciesId))
+        .fetch {
+          if (user.canReadProjectDeliverables(it[PROJECTS.ID]!!)) {
+            ParticipantProjectsForSpecies.of(it)
+          } else {
+            it[DELIVERABLES.ID] = null
+            ParticipantProjectsForSpecies.of(it)
```
sgrimm: Rather than modifying the result jOOQ returns, how about `ParticipantProjectsForSpecies.of(record).copy(activeDeliverableId = null)`?

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -119,6 +132,79 @@ class ParticipantProjectSpeciesStore(
...
+            COHORT_MODULES.START_DATE.lessOrEqual(today))
+        .leftOuterJoin(MODULES)
+        .on(COHORT_MODULES.MODULE_ID.eq(MODULES.ID))
+        .leftOuterJoin(DELIVERABLES)
+        .on(
+            DELIVERABLES.MODULE_ID.eq(MODULES.ID),
+            DELIVERABLES.DELIVERABLE_TYPE_ID.eq(DeliverableType.Species))
+        .where(ORGANIZATIONS.ID.eq(organizationId))
+        .and(SPECIES.ID.eq(speciesId))
+        .fetch {
```
sgrimm: This could use a declared variable name since the body of the lambda is a nontrivial multi-line expression.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
```diff
@@ -119,6 +132,79 @@ class ParticipantProjectSpeciesStore(
...
+            SPECIES.ID,
+            SPECIES.COMMON_NAME,
+            SPECIES.SCIENTIFIC_NAME)
+        .from(SPECIES)
+        .join(PARTICIPANT_PROJECT_SPECIES)
+        .on(SPECIES.ID.eq(PARTICIPANT_PROJECT_SPECIES.SPECIES_ID))
+        .join(PROJECTS)
+        .on(PARTICIPANT_PROJECT_SPECIES.PROJECT_ID.eq(PROJECTS.ID))
+        .join(PARTICIPANTS)
+        .on(PROJECTS.PARTICIPANT_ID.eq(PARTICIPANTS.ID))
```
sgrimm: Is this just a guard against cases where there's a participant project species for a project that isn't in a participant? If so, it'd also work (and would remove an unnecessary join) to do `.and(PROJECTS.PARTICIPANT_ID.isNotNull)` in the "where" clause.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
```diff
@@ -177,6 +181,199 @@ class ParticipantProjectSpeciesStoreTest : DatabaseTest(), RunsAsUser {
...
+  @Nested
+  inner class FetchSpeciesForParticipantProject {
+    @Test
+    fun `fetches the species and participant project species data associated to a participant project`() {
+      val cohortId = insertCohort()
+      val participantId = insertParticipant(cohortId = cohortId)
+      val projectId = insertProject(participantId = participantId)
+
+      val speciesId1 = insertSpecies(scientificName = "Acacia Kochi")
+      val speciesId2 = insertSpecies(scientificName = "Acacia Koa")
```
sgrimm: Any reason this uses two very similar-looking species names?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
nickgraz: Exactly, this model is really a partial combination of two models, specific to a UI view, so I wanted it to be extra clear. Another option is to just pass back both of the entire models in the payload like: 

```
{
  participantProjectSpecies: { ... },
  species: { ... },
}
```

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: No need for the join, thanks

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: Excellent, thanks, I was wondering how you would recommend cleaning this up because it did feel a bit crunchy at the time.

src/main/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStore.kt
nickgraz: Yes, that is the purpose. It _should_ be impossible for it to exist on a project that isn't associated to a participant, so perhaps neither join or condition are necessary.

src/test/kotlin/com/terraformation/backend/accelerator/db/ParticipantProjectSpeciesStoreTest.kt
nickgraz: Not any good reason

nickgraz:
### Merge activity

* **May 22, 1:02 PM EDT**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1974).

PR #1975, opened 2024-05-21
SW-5183 Added controller for support requests, integrated with emails

tommylau523:


tommylau523:
* **#1975** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1975?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/support/SupportService.kt
```diff
@@ -0,0 +1,39 @@
...
+    private val userStore: UserStore,
+) {
+  fun submitServiceRequest(description: String, summary: String, requestTypeId: Int) {
+    val user = userStore.fetchOneById(currentUser().userId) as IndividualUser
+    val response =
+        atlassianHttpClient.createServiceDeskRequest(
+            description, summary, requestTypeId, user.email)
+    val requestType =
+        atlassianHttpClient.requestTypes[requestTypeId]
+            ?: throw IllegalArgumentException("Request ID type not recognized")
```
sgrimm, line 24: Should this come before the `createServiceDeskRequest` call so we don't make a creation request with a bad request type?

src/main/kotlin/com/terraformation/backend/support/api/SupportController.kt
```diff
@@ -0,0 +1,40 @@
...
+  @Operation(summary = "Submit support request types.")
+  fun submitRequest(
+      @RequestBody payload: SubmitSupportRequestPayload
+  ): SimpleSuccessResponsePayload {
+    service.submitServiceRequest(payload.description, payload.summary, payload.id)
+    return SimpleSuccessResponsePayload()
+  }
+}
+
+data class ListSupportRequestTypesResponsePayload(val types: List<ServiceRequestTypeModel>)
```
sgrimm: It's a bit more boilerplate, but I think this should define a payload class for service request types and use a list of those rather than a list of the internal models. That way if we change the model for internal use, it doesn't cause the API to change.

src/main/kotlin/com/terraformation/backend/support/api/SupportController.kt
```diff
@@ -0,0 +1,40 @@
...
+
+data class SubmitSupportRequestPayload(
+    val id: Int,
```
sgrimm: Maybe call this `requestTypeId` or something like that? Otherwise at first glance it looks like maybe the client is supposed to generate an ID for the new support request.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -242,6 +242,19 @@ notification.seedBank.unknownAutomationTriggered.app.title={0} triggered at {1}
...
+notification.support.requestSubmitted.email.body.4=In the meantime, you may find helpful information in the [Terraware - Knowledge Base].
+notification.support.requestSubmitted.email.body.5=For your reference, here are the details of your request\:
+# {0} is the issue key
+notification.support.requestSubmitted.email.body.6=Request Number\: {0}
+# {0} is the summary of the request
+notification.support.requestSubmitted.email.body.7=Summary\: {0}
+# {0} is the description of the request
+notification.support.requestSubmitted.email.body.8=Description\: {0}
+# {0} is the name of the request type, {1} is the key of the request
+notification.support.requestSubmitted.email.subject={0} Received: {1}
```
sgrimm, line 257: This may be the best we can do for now, but I think long term we will probably want a separate string in this file for each request type since in some languages (Arabic, for example) the verb in this phrase would need to change depending on the noun.

Also, am I correct in assuming that at the moment, the request type will always be in English here even if the user's language is set to Spanish, since the name is an opaque string we get back from Atlassian?

src/main/resources/i18n/Messages_en.properties
```diff
@@ -242,6 +242,19 @@ notification.seedBank.unknownAutomationTriggered.app.title={0} triggered at {1}
 notification.seedBank.unknownAutomationTriggered.email.body.1=Automation {0} has been triggered at {1}.
 notification.seedBank.unknownAutomationTriggered.email.body.2=Additional details\: {0}
 notification.seedBank.unknownAutomationTriggered.email.subject={0} triggered at {1}
+notification.support.requestSubmitted.email.body.1=Thank you for contacting Terraware support.
+notification.support.requestSubmitted.email.body.2=We rely on our users to help make Terraware better and appreciate hearing from you.
+notification.support.requestSubmitted.email.body.3=Within the next few days, we may reach out to obtain additional information that would help us investigate the issue or to provide you with the status of a bug fix.
+notification.support.requestSubmitted.email.body.4=In the meantime, you may find helpful information in the [Terraware - Knowledge Base].
```
sgrimm: The hyphen in the link here looks a little weird to me; isn't it just the "Terraware Knowledge Base"?

sgrimm (changes requested):

src/main/resources/i18n/Messages_en.properties
tommylau523: Yes that name is always in English for now. I think it gets tricky since we will need to translate on the fly.

src/main/resources/i18n/Messages_en.properties
tommylau523: This is from Zapier triggers for now but let me check with PMs.

src/main/resources/i18n/Messages_en.properties
tommylau523: Maybe down the line, once the Request types are more finalized, we can take the reverse approach, and have Terraware server actually create the request types in Jira via API if they don't already exist. 

At that point, we can define all the service request types we want to support in a table, and translate any enum names, etc. 

Right now it is hard to translate everything from Jira because we are treating that as the ground truth.

src/main/kotlin/com/terraformation/backend/support/api/SupportController.kt
```diff
@@ -1,40 +1,52 @@
...
 class SupportController(private val service: SupportService) {
+
+  @ApiResponse(responseCode = "200")
   @GetMapping
   @Operation(summary = "Lists support request types.")
   fun listRequestTypes(): ListSupportRequestTypesResponsePayload {
-    return ListSupportRequestTypesResponsePayload(service.listServiceRequestTypes())
+    return ListSupportRequestTypesResponsePayload(
+        service.listServiceRequestTypes().map {
+          ServiceRequestType(it.id, it.name, it.description)
```
sgrimm: This is totally fine, but alternately, could make a secondary constructor for this class that takes a model argument.

tommylau523:
### Merge activity

* **May 22, 12:18 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/1975).

PR #1989, opened 2024-05-23
SW-5414 Added API endpoint for uploading attachments, added attachment IDs to existing endpoint

tommylau523:


tommylau523:
* **#1989** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/support/api/SupportController.kt
```diff
@@ -32,9 +43,41 @@ class SupportController(private val service: SupportService) {
       @RequestBody payload: SubmitSupportRequestPayload
   ): SubmitSupportRequestResponsePayload {
     val issueKey =
-        service.submitServiceRequest(payload.description, payload.summary, payload.requestTypeId)
+        service.submitServiceRequest(
+            requestTypeId = payload.requestTypeId,
+            summary = payload.summary,
+            description = payload.description,
+            attachmentIds = payload.attachmentIds ?: emptyList(),
```
sgrimm, line 50: It makes me nervous to pass an arbitrary client-supplied identifier through to an external service with no validation or permission checking (nothing to stop people from attaching other people's files to their issues!) but I suppose it's probably what the existing support form already does anyway, so this isn't making anything worse. Plus I guess the worst possible outcome of the client forging an ID here is that a Terraformation employee sees the wrong attachment.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/support/api/SupportController.kt
tommylau523: one future improvement we can implement is maybe store the file ID and assign that to a user in our table. Seems heavy for our purpose for now.

PR #1993, opened 2024-05-24
SW-5421 Update the SpeciesForParticipantProject model to include all data

nickgraz:
In order to more easily get all species data associated to a project, include the entirety of the `project`, `species`, and `participant_project_species` data within the model that is returned when fetching all species for a project.

PR #1994, opened 2024-05-24
SW-5421 Save submission snapshot on approval of species list deliverable

nickgraz:
- Add `submissionId` added to the `DeliverableStatusUpdatedEvent`
- Create `submission_snapshots` table which links a submission to a file stored through the `FileService`
- Listen to this event and save a species list submission snapshot when the submission moves to the "approved" state.
  - This only happens for "species list" deliverables
- Entry is added to the `submission_snapshots` table that links submission to file


nickgraz:
* **#2003** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1994** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1994?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1993** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/1993?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/db/SchemaDocsGenerator.kt
```diff
@@ -154,6 +154,7 @@ class SchemaDocsGenerator : DatabaseTest() {
                   "score_categories" to setOf(ALL, ACCELERATOR),
                   "submission_documents" to setOf(ALL, ACCELERATOR),
                   "submission_statuses" to setOf(ALL, ACCELERATOR),
+                  "submission_snapshots" to setOf(ALL, ACCELERATOR),
```
sgrimm, line 156: Nitpick: Move this up one line

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -283,4 +302,133 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
...
+      service.on(
+          DeliverableStatusUpdatedEvent(
+              deliverableId = deliverableId,
+              projectId = projectId,
+              oldStatus = SubmissionStatus.InReview,
+              newStatus = SubmissionStatus.Approved,
+              submissionId = submissionId))
+
+      val submissionSnapshot =
+          submissionSnapshotsDao.fetchBySubmissionId(submissionId).firstOrNull()
```
sgrimm: Using `first()` instead of `firstOrNull()` would make the variable non-nullable and will throw an exception if the list is empty (as opposed to the `!!` operator throwing an exception later on).

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -283,4 +302,133 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
...
+
+      assertEquals(
+          SubmissionSnapshotsRow(fileId = null, id = null, submissionId = submissionId),
+          submissionSnapshot.let { it!!.copy(fileId = null, id = null) })
```
sgrimm: Maybe `assertEquals(submissionId, submissionSnapshot.submissionId, "Submission ID")` since only the submission ID is being checked here.

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -283,4 +302,133 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
...
+      val expectedRow1 =
+          "\"$projectId\",\"$speciesId1\",\"Approved\",\"\",\"\",\"\",\"Non-native\",\"Species 1\",\"\""
+      val expectedRow2 =
+          "\"$projectId\",\"$speciesId2\",\"In Review\",\"It is a great tree\",\"\",\"\",\"Native\",\"Species 2\",\"\""
+      val expectedRow3 =
+          "\"$projectId\",\"$speciesId3\",\"Rejected\",\"\",\"Need to know native status\",\"\",\"\",\"Species 3\",\"Common name 3\""
+
+      assertEquals(
+          "$expectedHeaders\r\n$expectedRow1\r\n$expectedRow2\r\n$expectedRow3\r\n",
+          String(stream.readAllBytes()))
```
sgrimm: A couple things would make this cleaner.

Since the lines aren't being treated as individual values, they don't need to be separate variables; put the `\r\n` at the end of each string here and concatenate them.

Use the `"""` string literal format to avoid having to escape all the double quotes.

So you'd end up with something like
```kotlin
val expected = """"Project ID","Species ID","Status",...\r\n""" +
    """"$projectId1","$speciesId1","Approved",...\r\n"""
```

src/main/resources/db/migration/0250/V279__SubmissionSnapshots.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE accelerator.submission_snapshots (
+id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
```
sgrimm: Indent this to line up with the other columns

src/main/kotlin/com/terraformation/backend/accelerator/event/DeliverableStatusUpdatedEvent.kt
```diff
@@ -10,6 +11,7 @@ data class DeliverableStatusUpdatedEvent(
     val projectId: ProjectId,
     val oldStatus: SubmissionStatus,
     val newStatus: SubmissionStatus,
+    val submissionId: SubmissionId
```
sgrimm: May as well put a trailing comma here.

src/main/resources/db/migration/0250/V279__SubmissionSnapshots.sql
```diff
@@ -0,0 +1,5 @@
...
+    file_id BIGINT REFERENCES files ON DELETE SET NULL,
```
sgrimm: Why `ON DELETE SET NULL` instead of `ON DELETE CASCADE` here? What does it mean for a row in this table to have a null file ID?

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -109,12 +131,74 @@ class ParticipantProjectSpeciesService(
...
+            CSVWriter.DEFAULT_ESCAPE_CHARACTER,
+            CSVWriter.RFC4180_LINE_END,
+        )
+        .use { csvWriter ->
+          csvWriter.writeNext(rows.first().map { it.first }.toTypedArray())
+          rows.forEach { row -> csvWriter.writeNext(row.map { it.second }.toTypedArray()) }
+        }
+
+    val inputStream = ByteArrayInputStream(stream.toByteArray())
+    val filename = "${event.deliverableId}-${event.projectId}"
```
sgrimm: Can the same submission have more than one snapshot? If so, do we want a timestamp or something in the filename in case people download multiple snapshots?

Should the filename have a `.csv` extension?

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -109,12 +131,74 @@ class ParticipantProjectSpeciesService(
...
+
+    val speciesData =
+        participantProjectSpeciesStore.fetchSpeciesForParticipantProject(event.projectId)
+    val rows: List<List<Pair<String, String>>> =
+        speciesData.map {
+          listOf(
+              "Project ID" to it.project.id.toString(),
+              "Species ID" to it.species.id.toString(),
+              "Status" to it.participantProjectSpecies.submissionStatus.jsonValue,
+              "Rationale" to (it.participantProjectSpecies.rationale ?: ""),
```
sgrimm: I believe CSVWriter will render nulls as zero-length strings, so maybe the `?: ""` isn't needed.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -109,12 +131,74 @@ class ParticipantProjectSpeciesService(
...
+    CSVWriter(
+            streamWriter,
+            CSVWriter.DEFAULT_SEPARATOR,
+            CSVWriter.DEFAULT_QUOTE_CHARACTER,
+            CSVWriter.DEFAULT_ESCAPE_CHARACTER,
+            CSVWriter.RFC4180_LINE_END,
+        )
+        .use { csvWriter ->
+          csvWriter.writeNext(rows.first().map { it.first }.toTypedArray())
+          rows.forEach { row -> csvWriter.writeNext(row.map { it.second }.toTypedArray()) }
```
sgrimm: There is an extension method `writeNext()` in `Csv.kt` that lets you pass a `List<Any?>` instead of an `Array<String?>`, which would remove the need for the `.toTypedArray()` here.

sgrimm (changes requested):

src/main/resources/db/migration/0250/V279__SubmissionSnapshots.sql
nickgraz: Good catch, there was no logical reason

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
nickgraz: Good catch, I will throw a timestamp in there for convenience but the thing I was missing to make it unique was the submission ID.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -67,16 +86,21 @@ class ParticipantProjectSpeciesService(
         // A submission must exist for every project that is getting a new species assigned
         val deliverableSubmission =
             submissionStore.fetchActiveSpeciesDeliverableSubmission(
-                participantProjectSpecies.projectId)
+                participantProjectSpecies.projectId,
```
sgrimm: No need for a trailing comma here, I'd say, since this only takes one argument and is unlikely to ever take more.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -109,12 +133,74 @@ class ParticipantProjectSpeciesService(
     }
   }
 
+  /**
+   * When a species deliverable status is updated to "approved", we save a snapshot of the species
+   * list and reference it in the deliverable submission
+   */
+  @EventListener
+  fun on(event: DeliverableStatusUpdatedEvent) {
+    if (event.newStatus !== SubmissionStatus.Approved) {
```
sgrimm: `!=` instead of `!==`

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -109,12 +133,74 @@ class ParticipantProjectSpeciesService(
...
+            CSVWriter.DEFAULT_ESCAPE_CHARACTER,
+            CSVWriter.RFC4180_LINE_END,
+        )
+        .use { csvWriter ->
+          csvWriter.writeNext(rows.first().map { it.first })
+          rows.forEach { row -> csvWriter.writeNext(row.map { it.second }) }
+        }
+
+    val inputStream = ByteArrayInputStream(stream.toByteArray())
+    val filename = "species-list-snapshot-${event.submissionId}-${Instant.now()}.csv"
```
sgrimm: The default string representation of `Instant` includes colons which I believe would make this filename invalid on Windows. Probably want to use a custom time format here.

src/main/resources/db/migration/0250/V279__SubmissionSnapshots.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE accelerator.submission_snapshots (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    file_id BIGINT REFERENCES files ON DELETE CASCADE,
+    submission_id BIGINT NOT NULL REFERENCES accelerator.submissions
```
sgrimm, line 4: Putting an index on `file_id` would be good so that when files are deleted, the database doesn't have to do a full scan of this table to see if they're referenced.

Same with `submission_id` though I guess we don't support deleting submissions so it's less important.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
```diff
@@ -182,7 +183,10 @@ class ParticipantProjectSpeciesService(
         }
 
     val inputStream = ByteArrayInputStream(stream.toByteArray())
-    val filename = "species-list-snapshot-${event.submissionId}-${Instant.now()}.csv"
+    val timestamp =
+        DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss")
+            .format(Instant.now().atZone(ZoneId.of("UTC")))
```
sgrimm: It'd be better to have a clock constructor parameter for this class, and then use it to get the current time. That way the dates stay consistent with the application's clock if people are messing with time during testing. (Sorry I didn't catch this on the previous round of review!)

FYI, there's a system constant for UTC, so you don't need to look it up at runtime; you can do
```kotlin
clock.instant().atZone(ZoneOffset.UTC)
```
But even easier in this case is to declare the clock dependency as `Clock` rather than `InstantSource` and then you can do
```kotlin
ZonedDateTime.now(clock)
```
since the clock is already set to the UTC time zone.

sgrimm (approved):
One small suggestion about generating the timestamp, but good to go other than that.

src/main/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesService.kt
nickgraz: Thanks for the info. I was wondering about that, I figured it would be best to not add another dependency for something small, but you make some great points.

PR #1995, opened 2024-05-24
SW-5433 Add native category to payloads

constanzauanini:


src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
```diff
@@ -179,6 +180,7 @@ data class ParticipantProjectForSpeciesPayload(
     val activeDeliverableId: DeliverableId?,
     val participantProjectSpeciesId: ParticipantProjectSpeciesId,
     val participantProjectSpeciesSubmissionStatus: SubmissionStatus,
+    val participantProjectSpeciesNativeCategory: SpeciesNativeCategory?,
```
nickgraz: You can ditch this and line 194, in [this PR](https://github.com/terraware/terraware-server/pull/1993) I return back the entire `participantProjectSpecies` model in this response payload.

sgrimm:
The build failure here is just due to formatting. If you run `./gradlew spotlessApply` it will reformat the code, equivalent to the way `yarn format` works in the frontend repo. You'll need Java on your system; you can install it from Homebrew if you don't have it already.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
constanzauanini: Just saw you did that! Nice!

constanzauanini:
> The build failure here is just due to formatting. If you run `./gradlew spotlessApply` it will reformat the code, equivalent to the way `yarn format` works in the frontend repo. You'll need Java on your system; you can install it from Homebrew if you don't have it already.

Thanks Steve! I will update it

sgrimm (approved):
This conflicts with #1993 but also includes functionality that isn't in that PR. Either one could reasonably be merged first, and then the other will need to be rebased to resolve the conflicts.

sgrimm:
Thinking about it a bit more, since #1993 is part of a stack whose second PR has changes requested, it probably makes more sense to merge this one first since the stack will be changing regardless.

src/main/kotlin/com/terraformation/backend/accelerator/api/ParticipantProjectSpeciesController.kt
constanzauanini: @nickgraz I think I still need this change, because I see now that the `participantProjectSpecies` is added in `SpeciesForParticipantProjectPayload` but I need it in `ParticipantProjectForSpeciesPayload` too

PR #2001, opened 2024-05-28
Stop using deprecated Jackson API

sgrimm:
Our custom deserializer for geometry objects throws an exception when a GeoJSON
value has an unexpected or missing geometry type. The exception includes the
location in the JSON object where the failure happened.

We were using an old Jackson API method to retrieve the location. In the next
Jackson version, that method is marked as deprecated, which will cause a build
failure when we upgrade. Switch to the newer version of the method.


sgrimm:
* **#2001** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2001?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Looks good

PR #2002, opened 2024-05-28
Update coding conventions docs

sgrimm:
We've been updating the Kotlin coding conventions page in Confluence but haven't
been pulling those changes into the conventions docs in the actual code base.
Bring the docs in the repo up to date.

### DIFF
--- a/docs/CONVENTIONS.md
+++ b/docs/CONVENTIONS.md
@@ -1,10 +1,10 @@
 # Coding conventions
 
-The code should be written in idiomatic Kotlin and should generally follow the guidelines in the official Kotlin [Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) document.
+Code should be written in idiomatic Kotlin and should generally follow the guidelines in the official Kotlin [Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) document.
 
 It runs on the JVM, and thus makes extensive use of Java libraries. There are no plans to support multiplatform builds. If you find a Java library that does just what you need, use it rather than reinventing the wheel!
 
-Currently, the build targets the Java 20 JVM.
+Currently, the build targets the Java 21 JVM. We use the Amazon Corretto JVM and will generally target the most recent Corretto release.
 
 ## Formatting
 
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2002** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

docs/CONVENTIONS.md
```diff
@@ -19,11 +19,149 @@ There isn't currently a way to make IntelliJ's real-time formatting adhere stric
...
+
+Models. The store classes can accept and return models rather than rows.
+
+Sorting. The DAO classes have methods to fetch multiple rows, but they don’t guarantee a sort order.
+
+More sophisticated operations. Sometimes we want to do things that aren’t trivial single-table CRUD operations.
+
+Stores should not depend on other stores! See the next section for more. But it’s fine for a store to depend on DAOs and low-level helper objects.
+
+### Rows vs. Records
```
sgrimm, line 132: This section is likely to change in favor of using Record classes more, but keeping it as is for now since it describes how the code currently looks.

nickgraz: Thank you for adding this section!

docs/CONVENTIONS.md
```diff
@@ -19,11 +19,149 @@ There isn't currently a way to make IntelliJ's real-time formatting adhere stric
...
+
+TL;DR: Generally prefer the `Row` classes over the `Record` classes. The rest of this section talks about the differences between the two.
```
nickgraz, line 134: To be clear, this statement "generally prefer the `Row` classes" is how it currently is, but we will likely change in favor of using `Record` in the future. So if I am doing something now, I should use `Record` more often, is that right? 

When I originally read this, I went back to rework something in a current PR, because I am using the `Record`. Since I was just accessing some data as opposed to potentially interacting with the table again, it seemed like I should use the `Row`. But if we are likely to change in favor of using `Record`, should I keep it as is?

nickgraz (approved):

docs/CONVENTIONS.md
sgrimm: Correct. We've been using the `Row` classes but they are kind of second-class citizens in jOOQ and the `Record` ones are more full-featured. For now, you should use whichever is more convenient, but at some point I want to come up with a plan to migrate over to `Record`s. The reason I'm not pushing for it too much in new code right now is just that I think it'll make sense to come up with a usage/style guide so we use them consistently everywhere.

Once we've moved over, we can turn off code generation for the `Row` classes and we can get rid of a bunch of `Dao` dependencies all over the place since all you need is a `DSLContext` to read and write `Record`s.

PR #2028, opened 2024-06-03
Added ControllerIntegrationTest from pdd-backend 

nickgraz:
- Added `ControllerIntegrationTest` from `pdd-backend` and split up the bootstrap of the `DatabaseTest` from the internals, which were moved to a superclass called `DatabaseBackedTest`


nickgraz:
* **#2030** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2030?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2029** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2028** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/api/ControllerIntegrationTest.kt
```diff
@@ -0,0 +1,57 @@
+package com.terraformation.backend.api
```
nickgraz, line 1: This is taken directly from [here](https://github.com/terraware/pdd-backend/blob/main/server/src/test/kotlin/com/terraformation/pdd/ControllerIntegrationTest.kt)

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -0,0 +1,2482 @@
...
+ *   compare against. So subclasses can override [tablesToResetSequences] with a list of tables
+ *   whose primary key sequences should be reset before each test method.
+ */
+@ActiveProfiles("test")
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ContextConfiguration(
+    initializers = [DatabaseBackedTest.DockerPostgresDataSourceInitializer::class])
+@EnableConfigurationProperties(TerrawareServerConfig::class)
+@Testcontainers
+@Transactional
```
nickgraz: This annotation was added

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -0,0 +1,2482 @@
...
+@ComponentScan(basePackageClasses = [UsersDao::class])
+abstract class DatabaseBackedTest {
+  @Autowired lateinit var dslContext: DSLContext
+
+  /**
+   * List of tables from which sequences are to be reset before each test method. Sequences used
+   * here belong to the primary key in the table.
+   */
+  protected open val tablesToResetSequences: List<Table<out Record>>
```
nickgraz, line 322: I had to open this value to override it in child classes. I am not really sure why it is implicitly open in `pdd-backend`

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -0,0 +1,2482 @@
+package com.terraformation.backend.db
```
nickgraz, line 1: Other than comments below, this is an exact copy of `DatabaseTest` before this PR.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
tommylau523: I thought protected implied open!

tommylau523 (approved):

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
nickgraz: Yea I actually ended up being able to remove this later when I added the `Transactional` annotation so perhaps it isn't needed here after all. Going to try to remove it in this PR.

tommylau523: Can it be abstract val perhaps? I also think open is good since it gives clear indication that it is meant to be overriden. abstract further confirms that.

nickgraz: Problem with abstract is that it is initialized with a default value

tommylau523: Makes sense. I think I just have gaps in my Kotlin understanding! `open` seems totally fine and thanks for explaining it!

nickgraz: To be honest I don't really understand why it "just works" with the `@Transactional` annotation so that is the thing that is puzzling me

sgrimm: It "just works" because we use a compiler plugin that automatically opens variables/functions in classes with certain Spring annotations. That's needed (or at least helpful) because Kotlin makes things `final` by default, but Spring needs to be able to construct subclasses on the fly in order to implement stuff like `@Transactional`. The way `@Transactional` works is that Spring basically builds a little subclass that wraps each of the methods with code to manage the transactions.

Making this `open` is the right fix here.

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
nickgraz: Ah, I see, thanks for the explainer

sgrimm:
Oh, and as for `open` vs. `protected`, they're related but somewhat independent.

`open` means you can override it in a subclass, and `protected` means it is visible in subclasses but not in unrelated classes. You can have open functions/properties that are public rather than protected, and you can have protected functions/properties that subclasses aren't allowed to override but are still able to access.

PR #2029, opened 2024-06-03
Copy net-new PDD Backend into codebase

nickgraz:
**No changes have been made to any file under `src/main/kotlin` or `src/test`**. These files are identical to their current state in `pdd-backend:main`. This code does not compile. The intention with this pull request is that the contents of the files can be largely ignored, since they are identical to already reviewed application code from `pdd-backend`.

### Package restructuring
Application code was copied over into `terraware-server` into a new package `com.terraformation.backend.documentproducer`. The package structure in `pdd-backend` was flattened into this new package. 
```
pdd-backend folder                                     || terraware-server folder
src/main/kotlin/com/terraformation/pdd/document        -> src/main/kotlin/com/terraformation/backend/documentproducer
src/main/kotlin/com/terraformation/pdd/document/api    -> src/main/kotlin/com/terraformation/backend/documentproducer/api
src/main/kotlin/com/terraformation/pdd/document/db     -> src/main/kotlin/com/terraformation/backend/documentproducer/db
src/main/kotlin/com/terraformation/pdd/document/model  -> src/main/kotlin/com/terraformation/backend/documentproducer/model
src/main/kotlin/com/terraformation/pdd/methodology/api -> src/main/kotlin/com/terraformation/backend/documentproducer/api
src/main/kotlin/com/terraformation/pdd/variable        -> src/main/kotlin/com/terraformation/backend/documentproducer
src/main/kotlin/com/terraformation/pdd/variable/api    -> src/main/kotlin/com/terraformation/backend/documentproducer/api
src/main/kotlin/com/terraformation/pdd/variable/db     -> src/main/kotlin/com/terraformation/backend/documentproducer/db
src/main/kotlin/com/terraformation/pdd/variable/model  -> src/main/kotlin/com/terraformation/backend/documentproducer/model

The one exception to this rule is the admin controller, since all 
admin controllers exist in the same folder in `terraware-server`

src/main/kotlin/com/terraformation/pdd/admin/AdminController.kt -> src/main/kotlin/com/terraformation/backend/admin/AdminDocumentsController.kt
```

All test code was flattened in the same way.

### Migrations
- All necessary migrations from `pdd-backend` were brought over and concatenated into a single migration file at [V280](src/main/resources/db/migration/0250/V280__DocumentProducerTables.sql). With the exception of the modifications to the `files` table at the top of the migration, these migrations are an exact copy from `pdd-backend`.

PR #2030, opened 2024-06-03
Get the document producer compiling and passing tests

nickgraz:
The purpose of this pull request is to expand on the "copy pdd into terraware" work, and accomplish code compilation and tests passing. 

A very large portion of this code is changing of package names and imports. Other than that, there are no functional changes to any application code. 

Notable changes to pre-existing code: 
1. Add test configuration ([1](https://github.com/terraware/terraware-server/pull/2030/files#diff-dd438244ca2c3de883647882b9a9a26dab5296782a4cc22d8846653d5be32e75R8-R21), [2](https://github.com/terraware/terraware-server/pull/2030/files#diff-dd438244ca2c3de883647882b9a9a26dab5296782a4cc22d8846653d5be32e75R2-R3)) needed by the mock MVC used in controller tests
2. Add the [insertions](https://github.com/terraware/terraware-server/pull/2030/files#diff-128d9d95ce52f80b9b54c1653bb313fab85fcdcff5bd582791889fdb932636f8R2394-R2878) for the new document producer entities used in database tests
3. Copy over [BeanTestDouble](https://github.com/terraware/terraware-server/pull/2030/files#diff-71ee57534d78d03087d5284f78fa607beb7ff81408232764d800fe5cd4c16c43R1) and implement into the [TestClock](https://github.com/terraware/terraware-server/pull/2030/files#diff-2a5027746674a8991c58152e8305ce2f1dbf59c495b88f945510a324288e4025R16-R25) and [InMemoryFileStore](https://github.com/terraware/terraware-server/pull/2030/files#diff-a625d4f7c93e2019a91bda7e5bf4b0a874e36e33d56dc2d53de7522766bfd1aaR18-R28)
4. Use DI to force usage of `InMemoryFileStore` during tests ([1](https://github.com/terraware/terraware-server/pull/2030/files#diff-a625d4f7c93e2019a91bda7e5bf4b0a874e36e33d56dc2d53de7522766bfd1aaR16-R17), [2](https://github.com/terraware/terraware-server/pull/2030/files#diff-aab0b2a180e6625814bfe5058c69aa3c6db81542e7644d3ac3dee30dac21b46fR34), [3](https://github.com/terraware/terraware-server/pull/2030/files#diff-10b0905048ec592e9e8c77fe6b5b167213cd04cf4979dc315b2835a52f41b76bR33))
5. **TODO: verify this change works with FE** - Configure Jackson to exclude `null` fields from API responses ([1](https://github.com/terraware/terraware-server/pull/2030/files#diff-b4c0421bcb3280a8e738550c830dba197057c515bcdd9579a52cbfe9adca2061R46-R47))
6. Tell jOOQ to jenerate the schema for `docprod` ([1](https://github.com/terraware/terraware-server/pull/2030/files#diff-1e5ed62cd5ce3937d30cb4cdac8f461442ba03a978751cb38bc1860e87f013cdR39-R424))
7. Override Keycloak info in tests, needed for mock MVC ([1](https://github.com/terraware/terraware-server/pull/2030/files#diff-d33324ff9e92179d90f6fdc67b1d3d1dc622c0d10e50848b72f7360cecfb01bfR23), [2](https://github.com/terraware/terraware-server/pull/2030/files#diff-d33324ff9e92179d90f6fdc67b1d3d1dc622c0d10e50848b72f7360cecfb01bfR50))
8. Add permissions (canCreateDocument, canCreateVariableManifest, etc...) to users within `src/main/kotlin/com/terraformation/backend/customer/model`


nickgraz:
* **#2030** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2029** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2033** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2033?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 14 other dependent PRs ([#2064](https://github.com/terraware/terraware-server/pull/2064) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2064?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2065](https://github.com/terraware/terraware-server/pull/2065) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2065?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2068](https://github.com/terraware/terraware-server/pull/2068) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2068?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 11 others)
* `main`

<h2></h2>

nickgraz:
@sgrimm I have been trying for way too long to figure this out to no avail: whenever a test runs that uses `ControllerIntegrationTest`, subsequent tests that insert time zones fail. There are is no time zone insertion in tests in `pdd-backend`. 

The way I am seeing this locally is by moving `src/test/kotlin/com/terraformation/backend/documentproducer/api` out of the codebase, after doing this all the tests pass. Putting even a single one of the API tests back, however, produces over 100 failed tests all for the same reason, some variation of:
```
    org.springframework.dao.DuplicateKeyException: jOOQ; SQL [insert into "time_zones" ("time_zone") values (?) returning "time_zones"."time_zone"]; ERROR: duplicate key value violates unique constraint "time_zones_pkey"
      Detail: Key (time_zone)=(Pacific/Honolulu) already exists.
```

I'll keep digging to see if I can resolve this, but if you have any ideas I'd love to hear them.

sgrimm:
Time zones get inserted by `TimeZonePopulator` as part of application startup, so what's probably happening is that those tests are expecting to have to populate the time zones table explicitly (since in the current Terraware test environment, nothing populates it) which is failing when `ControllerIntegrationTest` is in the picture because the tests are running after the application initialization has happened.

A couple possible fixes:

* Turn time zones table initialization into a numbered SQL migration rather than doing it programmatically, so tests can rely on it being populated and we no longer need `TimeZonePopulator` at all. We'll then potentially have to add new time zones in the future (assuming we want to support them in our app) but new time zone names getting defined isn't super frequent. This will also shave a few milliseconds off application startup since it won't have to check whether any new time zones need to be inserted.
* Make the tests do `ON CONFLICT DO NOTHING` when they try to insert their zones, so they work whether or not the time zones table is populated already.

nickgraz:
Interesting, I was stepping through that class when I was debugging because it seemed like it was the culprit but when checking the table directly in subsequent tests, sometimes it would be empty and sometimes it would just have the one timezone that is conflicting. Removing both the class and test though solves the issue, so I will try to do the first suggestion.

sgrimm:
Moving the time zones to a SQL migration would be reasonable to do as a regular PR rather than as part of the feature branch, since it (slightly) benefits the existing Terraware code too. But not a huge deal either way.

src/main/kotlin/com/terraformation/backend/admin/AdminDocumentController.kt
```diff
@@ -22,7 +22,7 @@ import org.springframework.web.multipart.MultipartFile
 import org.springframework.web.servlet.mvc.support.RedirectAttributes
 
 @Controller
-@RequestMapping("/admin")
+@RequestMapping("/admin/document-producer")
```
sgrimm, line 25: Should this have a `@RequireGlobalRole` annotation too like the other terraware-server admin UI endpoints?

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -196,6 +197,8 @@ data class IndividualUser(
 
   override fun canCreateDeviceManager() = isSuperAdmin()
 
+  override fun canCreateDocument() = true
```
sgrimm, line 200: Document creation isn't going to be end-user-accessible at first, right? Should these document permissions require TF Expert instead of being open to everyone?

Or is this just to get the tests passing initially without having to muck around with permissions?

src/main/kotlin/com/terraformation/backend/documentproducer/db/CsvValidator.kt
```diff
@@ -1,8 +1,8 @@
-package com.terraformation.backend.document.db
+package com.terraformation.backend.documentproducer.db
```
sgrimm, line 1: Not a task for this PR, but we can probably combine this with the existing Terraware `CsvValidator` (the pdd-backend one started off as a copy of the Terraware one anyway).

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -65,68 +65,62 @@ COMMENT ON TABLE device_templates IS 'Canned device configurations for use in ca
 
 COMMENT ON TABLE devices IS 'Hardware devices managed by the device manager at a facility.';
 
-COMMENT ON TABLE document_saved_versions IS 'Saved versions of document variable values. A saved version is conceptually just a reference to a particular point in the edit history of the document; to restore that version, we ignore any later edits.';
+COMMENT ON TABLE docprod.document_saved_versions IS 'Saved versions of document variable values. A saved version is conceptually just a reference to a particular point in the edit history of the document; to restore that version, we ignore any later edits.';
```
sgrimm, line 68: Can we move these down to their own section of this file like we do for the nursery and accelerator and tracking tables? (At some point we should do the same for the seedbank tables so they aren't jumbled together with the public ones.)

src/main/resources/i18n/Enums_en.properties
```diff
@@ -36,6 +36,27 @@ accelerator.SubmissionStatus.Rejected=Rejected
...
+docprod.VariableType.Date=Date
+docprod.VariableType.Image=Image
+docprod.VariableType.Link=Link
+docprod.VariableType.Number=Number
+docprod.VariableType.Section=Section
+docprod.VariableType.Select=Select
+docprod.VariableType.Table=Table
+docprod.VariableType.Text=Text
+docprod.VariableUsageType.Injection=Injection
+docprod.VariableUsageType.Reference=Reference
```
sgrimm, line 59: I wonder if we should make these non-localizable since I don't see us exposing variable manifests directly to end users any time soon. No harm making them localizable, though.

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
```diff
@@ -188,15 +194,15 @@ class DocumentsControllerTest : ControllerIntegrationTest() {
     fun `updates a document by id`() {
       val otherMethodologyId = insertMethodology()
       val otherVariableManifestId = insertVariableManifest(methodologyId = otherMethodologyId)
-      val otherUserId = insertUser()
+      val otherUserId = insertUser(UserId(101))
```
sgrimm, line 197: FYI, you can do `insertUser(101)` here for brevity (that's why the insert methods use `Any` as the types for their ID arguments).

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
```diff
@@ -589,8 +604,10 @@ class DocumentsControllerTest : ControllerIntegrationTest() {
 
     @Test
     fun `returns not found error for nonexistent manifest`() {
+      val documentId = insertDocument()
```
sgrimm, line 607: Out of curiosity, why does this need to insert a document explicitly rather than using the default `cannedDocumentId`?

src/test/kotlin/com/terraformation/backend/documentproducer/db/ManifestImporterTest.kt
```diff
@@ -79,14 +78,9 @@ class ManifestImporterTest : DatabaseTest() {
...
-      val context = SecurityContextHolder.getContext()
-
-      oldAuthentication = context.authentication
-      context.authentication = PreAuthenticatedAuthenticationToken(user, "N/A", user.authorities)
-
-      every { permissionChecks.canCreateVariableManifest() } returns true
-      every { permissionChecks.invoke(any()) } answers { nothing }
+      every { user.canCreateVariableManifest() } returns true
+      // TODO is this needed?
+      // every { requirePermissions.invoke(any()) } answers { nothing }
```
sgrimm, line 83: Apparently not needed!

src/test/kotlin/com/terraformation/backend/file/InMemoryFileStore.kt
```diff
@@ -10,11 +13,20 @@ import java.time.Instant
...
-class InMemoryFileStore(private val pathGenerator: PathGenerator? = null) : FileStore {
+@Named
+@Priority(1) // Higher priority than the non-test implementations of FileStore
+class InMemoryFileStore(private val pathGenerator: PathGenerator? = null) :
+    FileStore, BeanTestDouble {
   private var counter = 0
-  private val files = mutableMapOf<URI, ByteArray>()
   private val deletedFiles = mutableSetOf<URI>()
 
+  val files = mutableMapOf<URI, ByteArray>()
```
sgrimm, line 23: Why can't this be private any more?

sgrimm (approved):
A few minor questions/comments but this seems good to merge into the feature branch.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
nickgraz: This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/main/resources/i18n/Enums_en.properties
nickgraz: Good point. I think I'd rather just leave them here if that's alright since the work is already done.

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
nickgraz: This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/main/resources/db/migration/R__Comments.sql
nickgraz: This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/test/kotlin/com/terraformation/backend/file/InMemoryFileStore.kt
nickgraz: It is used in a [test](https://github.com/terraware/terraware-server/pull/2030/files#diff-b218b08c6f9e7c40e78a31250114a35a42f6200602aebb2771044ff4fca4b9f1R126). I figured it would be OK since this is only used in test code.

[PDD source](https://github.com/terraware/pdd-backend/blob/main/server/src/test/kotlin/com/terraformation/pdd/file/InMemoryFileStore.kt#L23)

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
nickgraz: This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
nickgraz: Wasn't needed after all. I put it in there because I was thinking it was returning a 404 for missing document since it isn't a "real" document, but it is in fact returning the nonexistent manifest error.

This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/test/kotlin/com/terraformation/backend/documentproducer/db/ManifestImporterTest.kt
nickgraz: This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/main/kotlin/com/terraformation/backend/admin/AdminDocumentController.kt
nickgraz: Good catch.

This will be addressed in another PR into the feature branch, for now it is captured as a TODO in the main feature branch PR description.

src/test/kotlin/com/terraformation/backend/documentproducer/api/DocumentsControllerTest.kt
nickgraz: This ended up staying in place as all the "canned" variables were ditched in favor of the predominant pattern in `DatabaseBackedTest`

PR #2033, opened 2024-06-04
Document Producer 

nickgraz:
This pull request brings in the "document producer" functionality from our `pdd-backend` repository. 

Notable changes:
- "pdd" has been replaced with "document". Since this tool is general purpose now, we can produce PDDs, but we can also produce other documents like "Feasibility Study". 
- "methodology" has been replaced with "document template". Methodologies are specific to PDDs, and are a container for a set of variables that belong in a PDD. Now we have "document templates" which are a collection of variables that belong to a particular "document".
- All of the database tables and functions that were brought over now exist under the `docprod` schema
- An "all variable" import has been created (and admin UI for it), which is where we can import a CSV that contains all variables that can exist in the system. We have effectively split up the previous import process into 2 imports. One import is for all of the variables that exist and their particular configuration (validation data, options data, etc...) and eventually we will modify the "manifest importer" so that it is only concerned with the sequencing and order of variables within a particular variable manifest for a document template. One of the biggest changes in this regard is that the "section" type variables can only exist within the manifest CSV. Since sections are specific to a particular manifest (read: document template), these are excluded from the "all variable" manifest.
- Variable values (user input) are now associated to a project instead of a document. Since we will have multiple document types (feasibility study, project summary, etc...), this change allows us to reuse values for a given project across all documents.
- APIs for interacting with variable values for a given project were created.
- Variables now can have "owners", which is how we will inform internal TF users who is responsible for answering/validating/etc... values for a given variable.
- Additional fields related to deliverables are added to the "all variable" import. This allows us to create variables in the system that are associated to deliverables within the accelerator program. 
- Additional fields have been added to the "all variables" import that allows us to introduce some basic logic to conditionally show/hide variables based on their dependency to other variables.
- An additional field was added that allows us to indicate "internal only" variables
- The Jackson API configuration has been updated to only render object fields with non-null values by default


nickgraz:
* **#2033** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2033?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 15  dependent PRs ([#2029](https://github.com/terraware/terraware-server/pull/2029) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2064](https://github.com/terraware/terraware-server/pull/2064) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2064?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2065](https://github.com/terraware/terraware-server/pull/2065) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2065?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 12 others) 👈
* `main`

<h2></h2>

docs/index.md
```diff
@@ -28,6 +28,7 @@ The database has some subject-area-specific schemas as well as a default public
 * [Nursery schema](schema/all/nursery/relationships.html) (`nursery`)
 * [Seed bank schema](schema/all/seedbank/relationships.html) (`seedbank`)
 * [Tracking schema](schema/all/tracking/relationships.html) (`tracking`)
+* Document Producer schema (`documentproducer`)
```
sgrimm: What do you think of using a shorter name for this schema? Maybe `docprod`, though the specific name isn't important.

I'm thinking about when we're writing ad-hoc SQL queries to debug stuff; a long name is potentially going to be kind of tiresome to have to type out each time you reference a table in this schema.

nickgraz: Sounds good, `docprod` it is!

src/main/resources/db/migration/0250/V282__DocumentProducerTables.sql
```diff
@@ -0,0 +1,552 @@
...
+
+CREATE TABLE variable_value_citations (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    variable_value_id BIGINT NOT NULL REFERENCES variable_values ON DELETE CASCADE,
+    name TEXT NOT NULL,
+    url TEXT
+);
+
+CREATE INDEX ON variable_value_citations (variable_value_id);
+CREATE FUNCTION reject_delete() RETURNS TRIGGER AS $$
```
nickgraz: @sgrimm Should these functions be moved to the `docprod` schema?

sgrimm (approved):
I think this is good to merge. Let's do a regular git merge on this one rather than a squash merge so we preserve the history of PRs on the branch.

PR #2037, opened 2024-06-05
Move the population of the timezones to a numbered migration file

nickgraz:
- Remove `TimeZonePopulator` and populated the timezones as part of a normal migration. This allows the table to be pre-populated instead of needing to be populated during runtime, which creates failures for tests that run subsequent to any tests that use `ControllerIntegrationTest`
- The timezones were taken from the `TimeZonePopulator` [here](https://github.com/terraware/terraware-server/pull/2037/files#diff-355bfb929d05c9a8f05602d27f4b3c2aaa152c77746d440f1059008d1fb77cf5L24)


nickgraz:
* **#2037** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2037?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/resources/db/migration/0250/V280__TimeZones.sql
```diff
@@ -0,0 +1,605 @@
...
+       ('US/Hawaii'),
+       ('US/Indiana-Starke'),
+       ('US/Michigan'),
+       ('US/Mountain'),
+       ('US/Pacific'),
+       ('US/Samoa'),
+       ('UTC'),
+       ('W-SU'),
+       ('WET'),
+       ('Z'),
```
nickgraz, line 604: This is the only timezone that I had to add manually, because it is expected [here](https://github.com/terraware/terraware-server/blob/main/src/test/kotlin/com/terraformation/backend/tracking/db/plantingSiteStore/PlantingSiteStoreUpdateSeasonsTest.kt#L264) but doesn't seem to exist within the zone IDs.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -2091,11 +2090,7 @@ abstract class DatabaseBackedTest {
     return rowWithDefaults.id!!
   }
 
-  fun insertTimeZone(timeZone: Any = ZoneId.of("Pacific/Honolulu")): ZoneId {
-    val zoneId = if (timeZone is ZoneId) timeZone else ZoneId.of("$timeZone")
-    timeZonesDao.insert(TimeZonesRow(zoneId))
-    return zoneId
-  }
+  fun insertTimeZone(timeZone: Any = ZoneId.of("Pacific/Honolulu")): ZoneId = ZoneId.of("$timeZone")
```
nickgraz: Since time zones are already inserted, just return back the supplied or default value. I went this route so I didn't have to go through all test code and replace `insertTimeZone` calls with a hard coded value, if you'd rather I do that just let me know.

src/main/resources/db/migration/0250/V280__TimeZones.sql
```diff
@@ -0,0 +1,605 @@
...
+       ('Zulu');
```
sgrimm: Add a trailing newline for consistency with the other migration scripts.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
sgrimm: I think it'd be better to update the callers and get rid of this so we aren't left with a misleading function name. The "inline function" refactoring in IntelliJ will do the bulk of the work, so it shouldn't be too painful.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/customer/db/FacilityStoreTest.kt
```diff
@@ -84,7 +84,7 @@ internal class FacilityStoreTest : DatabaseTest(), RunsAsUser {
     every { user.canUpdateSubLocation(any()) } returns true
     every { user.canUpdateTimeseries(any()) } returns true
 
-    timeZone = insertTimeZone()
+    timeZone = ZoneId.of("${ZoneId.of("Pacific/Honolulu")}")
```
sgrimm: This seems a little ugly but is harmless for now; we can clean it up later.

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/customer/db/FacilityStoreTest.kt
nickgraz: Easy fix

PR #2043, opened 2024-06-05
SW-5496 Add user fields for cookie consent

sgrimm:
In preparation for switching to our own implementation of cookie consent
management, add a cookie consent flag and timestamp to the users table.

These values are returned as fields in the user profile payload, and can
be set using a new dedicated endpoint `PUT /api/v1/users/me/cookies`.


sgrimm:
* **#2043** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -87,6 +87,8 @@ data class IndividualUser(
     val firstName: String?,
     val lastName: String?,
     val countryCode: String?,
+    val cookiesConsented: Boolean?,
```
nickgraz, line 90: Should this be a map? In most places I see these consent banners, the consent is either "all", "none", or some subset of a set of cookies. Perhaps that is not necessary yet though?

nickgraz (approved):

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
sgrimm: The requirements doc specifically says it should be a boolean. Our setup is very simple: we only have essential cookies (no consent needed) and analytics ones, no other categories.

PR #2048, opened 2024-06-06
SW-5439 Welcome Email: Enhancements - new email template & content updates

constanzauanini:


src/main/resources/templates/email/logo.ftlh.mjml
```diff
@@ -1,5 +1,5 @@
 <!-- <#-- @ftlvariable name="" type="com.terraformation.backend.email.model.EmailTemplateModel" --> -->
 <mj-hero mj-class="logo-wrapper">
-    <mj-image mj-class="logo-img" src="${webAppUrl}/assets/logo-terraformation.png">
+    <mj-image mj-class="logo-img" src="${webAppUrl}/assets/tw-tf-logo.svg">
```
sgrimm: Are we confident that SVG images work in non-web-based email clients (Outlook, Apple Mail, etc.)?

src/main/resources/i18n/Messages_en.properties
```diff
@@ -257,11 +257,25 @@ notification.support.requestSubmitted.email.body.8=Description\: {0}
...
 notification.user.addedToOrganization.email.subject=You''ve been added to a new organization\!
 notification.user.addedToTerraware.email.body.1=You are now a member of {0}. Welcome to Terraware\!
 notification.user.addedToTerraware.email.body.2=Added by {0} ({1})
+notification.user.addedToTerraware.email.body.3=Welcome to Terraware\!
+notification.user.addedToTerraware.email.body.4=Click the button below to create your Terraware account where you can view this organization.
+notification.user.addedToTerraware.email.body.5=Once you set up your account, you will be able to use Terraware to understand and optimize any stage of a forestry project from seed collection to carbon calculation.
+notification.user.addedToTerraware.email.body.item.1=Collect seeds more efficiently and know their viability
+notification.user.addedToTerraware.email.body.item.2=Ensure your nursery inventory is on track for planting season
+notification.user.addedToTerraware.email.body.item.3=Know what has been planted where, and monitor survival
+notification.user.addedToTerraware.email.body.item.4=Streamline the MRV process using in-field data collection tools
```
sgrimm: Would it make sense to use a different prefix for these since they're shared between two different notification messages? Reading the MJML, I found it a little unexpected that the string references were jumping back and forth between `notification.user.addedToOrganization` and `notification.user.addedToTerraware`.

Maybe these could be something like `notification.email.terrawareBenefits.1`.

tommylau523 (approved):

src/main/resources/templates/email/user/addedToTerraware/body.ftlh.mjml
```diff
@@ -3,18 +3,49 @@
...
     <mj-body>
-        <mj-include path="../../logo.ftlh.mjml"/>
+        <mj-section padding="24px 0px" background-color="#F9F8F7">
+            <mj-column padding="0">
+                <mj-image mj-class="logo-img" src="${webAppUrl}/assets/tw-tf-logo.png">
+                </mj-image>
+            </mj-column>
+            <mj-column padding="0">
+                <mj-text mj-class="text-header01">
+                    ${strings("notification.user.addedToOrganization.email.header.1")}</mj-text>
```
sgrimm: Aside from things like the page footer and button labels, we've generally erred on the side of having a separate set of strings for each notification type even if some of them are duplicates of other notifications. That way we're able to tweak the wording on just one of the notifications without accidentally affecting others, and it's easier to look at the messages file and see the contents of a notification. Mixing references to different notifications means you have to go back and forth between the MJML file and the strings file to see what's in a message.

Which is to say: I think we should have all the strings for this notification under `notification.user.addedToTerraware` (aside from shared components like the "Terraware benefits" list and footers/button labels) even though some of them are the same as the "added to organization" strings.

sgrimm (changes requested):
Seems pretty good, but a few bits of feedback.

src/main/resources/templates/email/logo.ftlh.mjml
constanzauanini: Oh no, we need a png. I forgot to change it here, thanks!

src/main/resources/i18n/Messages_en.properties
constanzauanini: Sounds goods! Will update it

sgrimm (approved):
Looks good!

sgrimm:
Is there a Jira task for this we can put in the PR title?

constanzauanini:
> Is there a Jira task for this we can put in the PR title?

Yes, I will update title correctly, I created this PR as a draft and didn't update it when it was ready

PR #2049, opened 2024-06-06
SW-5484 Check detected file Mime type in addition to client header

tommylau523:
This additional check will prevent a user from changing the file extension before uploading the file, or use an incorrect header, to bypass the request header content type check. 

The intention is that this will prevent malicious files from being uploaded to our support requests.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/support/SupportService.kt
+++ b/src/main/kotlin/com/terraformation/backend/support/SupportService.kt
@@ -63,13 +63,15 @@ class SupportService(
   ): List<TemporaryAttachmentModel> {
     val tikaContentType = MediaType.parseMediaType(Tika().detect(sizedInputStream))
     sizedInputStream.contentType?.let {
-      !isContentTypeSupported(it) &&
-          throw NotSupportedException(
-              "$it is not a supported content type. Must be one of $SUPPORTED_CONTENT_TYPES_STRING")
-    }
-    !isContentTypeSupported(tikaContentType) &&
+      if (!isContentTypeSupported(it)) {
         throw NotSupportedException(
-            "File detected to be $tikaContentType, which is not supported. Must be one of $SUPPORTED_CONTENT_TYPES_STRING")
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#2049** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2049?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/support/SupportService.kt
```diff
@@ -59,11 +61,15 @@ class SupportService(
...
   ): List<TemporaryAttachmentModel> {
+    val tikaContentType = MediaType.parseMediaType(Tika().detect(sizedInputStream))
     sizedInputStream.contentType?.let {
       !isContentTypeSupported(it) &&
           throw NotSupportedException(
               "$it is not a supported content type. Must be one of $SUPPORTED_CONTENT_TYPES_STRING")
     }
+    !isContentTypeSupported(tikaContentType) &&
+        throw NotSupportedException(
+            "File detected to be $tikaContentType, which is not supported. Must be one of $SUPPORTED_CONTENT_TYPES_STRING")
```
sgrimm: Didn't notice this in the existing code when I reviewed it, but I think this, and the one above it, would be clearer as `if` statements than as short-circuit boolean expressions.

sgrimm (approved):
Seems okay other than one syntax suggestion, but I'm unclear on what problem we're solving here, so I can't really review this from an "is this the right way to solve the problem" perspective. There's no PR description and the Jira task doesn't tell me much.

tommylau523:
I added a brief description in the PR.

tommylau523:
### Merge activity

* **Jun 7, 3:27 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2049).

PR #2050, opened 2024-06-07
Rename "methodologies" -> "document templates"

nickgraz:
Since the document producer is meant to be a general purpose document creator, and methodologies are specific to Verra "project design documents", rename this entity to "document template" which is more generic.

- Get rid of the "canned" IDs used in document producer testing and use the "inserted" way of handling test IDs/entities. ([1](https://github.com/terraware/terraware-server/compare/ng/SW-5425-get-tests-passing...ng/SW-5425-rename-methodologies?expand=1#diff-128d9d95ce52f80b9b54c1653bb313fab85fcdcff5bd582791889fdb932636f8L375-L377))


tommylau523 (approved):
Looks good! I supposed we can edit the migrations since this is merging into a feature branch instead of main!

src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
```diff
@@ -12,7 +12,7 @@ data class ExistingDocumentModel(
     val createdBy: UserId,
     val createdTime: Instant,
     val id: DocumentId,
-    val methodologyId: MethodologyId,
+    val documentTemplateId: DocumentTemplateId,
```
sgrimm, line 14: Nit: Keep this and the other classes in the PR alphabetized? Or maybe that's better left for a followup sweep through the code (if we care about it at all) after all the other renaming and reorganization is done.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -2391,22 +2387,22 @@ abstract class DatabaseBackedTest {
   protected fun insertDocument(
       createdBy: UserId = inserted.userId,
       createdTime: Instant = Instant.EPOCH,
-      methodologyId: MethodologyId = cannedMethodologyId,
+      documentTemplateId: DocumentTemplateId = inserted.documentTemplateId,
```
sgrimm, line 2390: I have zero objection to this, but out of curiosity, why the move away from the `canned` approach? To make it more consistent with the existing terraware-server tests?

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
nickgraz: Mostly to make it consistent, but also because after removing [these lines](https://github.com/terraware/terraware-server/pull/2050/files#diff-e911a186ebc1e23d7bef146f87e2215e9937047ad719f93b5184883af2e57833L102), some controller integration tests starting failing.

src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
nickgraz: Ah yes, I will go back through this and make sure everything is alphabetized. I forgot to do that after the find and replace.

PR #2055, opened 2024-06-10
SW-5516 Add new permanent clusters to observation

sgrimm:
When area is added to a planting site with an observation in progress, we need to
create new permanent clusters in the newly-added area and include them in the
in-progress observation.

We were creating them, but not adding them to the observation because they were
created as part of applying the map edit; the code that updates the observation
plots was only adding plots that needed to be created to replace previous ones
because of a zone shrinking.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
@@ -412,6 +412,8 @@ class ObservationService(
       val observationPlots = observationStore.fetchObservationPlotDetails(observationId)
       val plantedSubzoneIds = plantingSiteStore.countReportedPlantsInSubzones(plantingSiteId).keys
 
+      val existingPermanentPlotIds =
+          observationPlots.filter { it.model.isPermanent }.map { it.model.monitoringPlotId }
       val removedObservationPlots =
           observationPlots.filter {
             it.model.monitoringPlotId in event.monitoringPlotReplacements.removedMonitoringPlotIds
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2055** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2055?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -447,7 +448,10 @@ class ObservationService(
             zone.plantingSubzones
                 .flatMap { subzone ->
                   subzone.monitoringPlots.filter {
-                    it.id in newPermanentPlotIds &&
+                    it.isAvailable &&
+                        it.id !in existingPermanentPlotIds &&
+                        it.permanentCluster != null &&
+                        it.permanentCluster <= zone.numPermanentClusters &&
```
nickgraz: This is a bit hard to grok, but the val name is very specific, which helps a lot here. I wonder if the val name should be updated since the underlying filter condition has changed.

nickgraz (approved):
Looks good to me, one small comment about readability.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
sgrimm: Updated the name and also broke the expression up into pieces with hopefully meaningful names.

nickgraz (approved):
Thanks!

PR #2060, opened 2024-06-12
Variable table changes

nickgraz:
In order to support separate variable and document template imports, we need to move some columns from the `variable_manifest_entries` table into the `variables` table. 

Here's a [slack thread](https://terraformationteam.slack.com/archives/C01C69KA23Y/p1718040574228279) with more context.

As a result of this change, a new variable will be created if the description or name fields are changed in the "all variable" import CSV (coming later). For now the import process has been untouched except for the modification that looks at name and description to determine if a variable can be reused.


sgrimm (approved):
Seems consistent with what we discussed. The changes to the import code are going to be the interesting part of this.

PR #2062, opened 2024-06-13
Updated tika file type detection to only occur if mark/reset are supported

tommylau523:
Currently, attachment uploads are failing because `Tika().detect()` reads bytes off the input stream, and results in "IO Error: Too few bytes" when streaming the file to Jira. 

Two changes to fix this:
1) Only perform Tika detection if the input stream can be marked and reset
2) Cast file stream to be a `BufferedInputStream`, so that it supports marking and resetting.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/support/SupportService.kt
+++ b/src/main/kotlin/com/terraformation/backend/support/SupportService.kt
@@ -61,16 +61,23 @@ class SupportService(
       filename: String,
       sizedInputStream: SizedInputStream,
   ): List<TemporaryAttachmentModel> {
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#2062** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):
This is a reasonable approach. We'll probably want to keep an eye on file sizes since this will cause the entire file to be loaded into memory (possibly twice, if the HTTP server or Spring decide to buffer the whole thing in memory before calling the controller). If people are commonly uploading really large files, we may want to increase the server's memory capacity.

PR #2070, opened 2024-06-14
SW-5523 Implemented a timeout limit for Atlassian endpoint

tommylau523:
This is mostly applicable for the file upload endpoint, which can sometimes hang.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/support/atlassian/Constants.kt
+++ b/src/main/kotlin/com/terraformation/backend/support/atlassian/Constants.kt
@@ -1,0 +1,4 @@
+package com.terraformation.backend.support.atlassian
+
+/** Timeout limit in milliseconds for an API request to Atlassian */
+const val REQUEST_TIMEOUT_MS: Long = 3 * 60 * 1000 // Three minutes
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#2070** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2070?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/support/atlassian/Constants.kt
```diff
@@ -0,0 +1,4 @@
+package com.terraformation.backend.support.atlassian
+
+/** Timeout limit in milliseconds for an API request to Atlassian */
+const val REQUEST_TIMEOUT_MS: Long = 3 * 60 * 1000 // Three minutes
```
sgrimm, line 4: This is happening synchronously, right? We should make sure 3 minutes isn't so long that the user's request times out on its own (browser, load balancer, etc.) even if the server ends up succeeding at the last second.

Was 3 minutes just an arbitrary starting point, or is it based on seeing how long Atlassian sometimes takes to return a response?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/support/atlassian/Constants.kt
tommylau523: I simply set this based off some back of napkin math, like in a case of uploading a 50MB file (very very large for our purpose), assuming 300 kbps. 

It is just an arbitrary starting point but I am happy to adjust to whatever you think.

tommylau523: As for the first part, I am not really sure. Ideally this catches all the timeout cases on the server side, but on the client front, as long as the client sees an Error, instead of getting stuck in Pending for a long time, it is an okay behavior. 

Likely something to adjust over time.

sgrimm:
3 minutes seems fine to start with. We can test in staging and adjust as needed.

PR #2065, opened 2024-06-14
Keep track of table column order

sgrimm:
In the original PDD schema, we used the manifest entry's position to determine
the order of table columns. In the absence of a manifest entry, we need to track
column order explicitly. Add a position field for that.

PR #2064, opened 2024-06-14
Fold document producer migrations into final result

sgrimm:
To make it easier to quickly look at the table definitions, update the document
producer migration SQL script to fold all the concatenated migrations into the
initial set of CREATE statements.

This produces a slightly different end result than before:

- The `files` table no longer has a new `caption` column. It was in the initial
  version of the PDD files table, but was dropped in a later migration.
- Trigger functions are created in the `docprod` schema.

PR #2071, opened 2024-06-14
All variable import

nickgraz:
- Create a new "all variable" import
- Mostly a copy of the manifest importer, except everything related to sections is removed
- Adds an admin endpoint to receive the CSV
- Adds a form in the document producer admin for the all variable import upload
- Adds the following new fields used in the "all variables" csv import
```
deliverable_id
deliverable_question
dependency_variable_stable_id
dependency_condition_id
dependency_value
internal_only
```
- Adds a new enum table for "dependency condition" and some basic conditions
- Variable store is modified to have a fetch context for variable manifest imports as well as all variable imports
- Within a single variable, there is a "dependency configuration" which is dictated by the 3 `dependency_*` fields. If any of them are filled out, they must all be filled out.

- [x] Add tests
- [x] Determine where to put the section variable manifest validation - **I split up the two imports so they each have their own validation now, section validation will stay in the variable manifest importer**
- [ ] Investigate failed tests after removing table column check in `fetchManifestVariables` - [thread](https://github.com/terraware/terraware-server/pull/2071/files/bf383267c23d4cd83175a6f4320169fb46e9211c#r1647848943)


nickgraz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2071?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2071** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2071?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2033** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2033?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 14 other dependent PRs ([#2029](https://github.com/terraware/terraware-server/pull/2029) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2064](https://github.com/terraware/terraware-server/pull/2064) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2064?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2065](https://github.com/terraware/terraware-server/pull/2065) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2065?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 11 others)
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/admin/AdminDocumentProducerController.kt
```diff
@@ -61,6 +63,39 @@ class AdminDocumentProducerController(
...
   }
 
+  @Operation(
+      summary = "Upload a variable manifest.",
+      description = "The uploaded file must be in CSV format.",
+      externalDocs =
+          ExternalDocumentation(
+              description = "Current Google sheet",
+              url =
+                  "https://docs.google.com/spreadsheets/d/16ybKFseSwFubxZ_i6ENxsF2XT0RqajhdumefJOoHQo8/edit#gid=1050189377"))
```
sgrimm: We should make sure the permissions on this document are locked down tight since terraware-server is a public repo.

nickgraz: It is currently only accessible to people with a Terraformation email

sgrimm: Looks that way to me too; we'll just need to make sure we're careful with sharing that doc from now on since we don't have security-by-obscurity as an added hurdle.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -85,7 +93,23 @@ class VariableStore(
...
+   * Returns information about a variable. Eagerly fetches other variables that are related to the
+   * requested variable, e.g., table columns.
+   *
+   * @throws VariableNotFoundException The variable didn't exist.
+   * @throws VariableIncompleteException The variable didn't have all the information required for
+   *   its type.
+   * @throws CircularReferenceException There is a cycle in the graph of the variable and its
+   *   children.
+   */
+  fun fetchVariableForStableId(stableId: String, variableId: VariableId): Variable {
```
sgrimm: Why are both the stable and variable IDs needed here? Can't you look up the stable ID given just the variable ID?

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -255,7 +298,128 @@ class VariableStore(
...
+                .from(VARIABLE_SECTION_RECOMMENDATIONS)
+                .where(SECTION_VARIABLE_ID.eq(base.id))
+                .and(VARIABLE_MANIFEST_ID.eq(manifestId))
+                .fetch(RECOMMENDED_VARIABLE_ID.asNonNullable())
+          }
+
+      return SectionVariable(base, sectionRow.renderHeading!!, children, recommends)
+    }
+
+    protected fun fetchTable(base: BaseVariableProperties): TableVariable {
```
sgrimm: Should this be in the other fetch class? Tables should be entirely in the all-variables sheet, I believe.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/AllVariableCsvVariable.kt
```diff
@@ -0,0 +1,110 @@
...
+    /** Column 12/L - Header? (for table columns) */
+    val isHeader: Boolean,
+    /** Column 13/M - Notes */
+    val notes: String?,
+    /** Column 14/N - Deliverable ID */
+    val deliverableId: DeliverableId?,
+    /** Column 15/O - Deliverable Question */
+    val deliverableQuestion: String?,
+    /** Column 16/P - Dependency - Variable ID */
+    val dependencyVariableId: VariableId?,
```
sgrimm: Wouldn't this need to be a stable ID, not a variable ID? If you're adding a new variable and another new variable that depends on it, there won't be a variable ID for the first one yet.

sgrimm (commented):
Not done looking yet, but a few initial comments.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/AllVariableCsvVariable.kt
nickgraz: Good point, will fix it up

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: That is correct, thanks for the catch

nickgraz: Looks like I had it in both places, will remove from this spot

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: That is true, I was trying to repurpose the code used for the manifest import, but it seems like it is unnecessary as you point out

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -67,6 +67,14 @@ class VariableStore(
    */
   private val variables = ConcurrentHashMap<Pair<VariableManifestId, VariableId>, Variable>()
 
+  /**
+   * Cache of information about variables by their stable ID. Each stable ID will point to the
+   * latest version of the variable for that ID. This just holds variable definitions, not values!
+   * There is currently no mechanism to limit the size of this cache; the assumption for now is that
+   * all the variables in the system will easily fit in memory.
+   */
+  private val stableIdVariables = ConcurrentHashMap<Pair<String, VariableId>, Variable>()
```
sgrimm: I suspect this will be changing or going away based on the discussion on `fetchVariableForStableId`, but if we're keeping it in some form, we'll need to be careful  about cache invalidation. The variables cache (by variable ID) doesn't need invalidation because variables are immutable, but the variable a stable ID points to can change over time.

It's not enough to just remove the entry from this map as part of the all-variables import, because in staging/prod, we're running two terraware-server instances each of which will have its own in-memory cache. Just removing the entry from the map on whichever instance is handling the import request would leave a stale entry on the other instance.

My recommendation for now would be to avoid trying to cache which variable a stable ID points to. If it turns out to be a performance issue, we can look at adding a cache but I'd rather avoid having to do distributed cache management unless we really need to.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -114,6 +138,28 @@ class VariableStore(
...
 
+  /**
+   * Fetches the list of non-section variables, using the latest version of each based on their
+   * stable ID. Table columns are not included in the top-level list, but rather in the containing
+   * variables' lists of children.
+   */
+  fun fetchAllNonSectionVariables(): List<Variable> {
+    return with(VARIABLES) {
+      dslContext
+          .select(ID, STABLE_ID)
```
sgrimm: There needs to be an aggregation function on `ID` for the `groupBy` to be valid.

Try `select(DSL.max(ID), STABLE_ID)` and remove the `orderBy`.

sgrimm (commented):
A couple more comments; will wait for the next revision before looking at it in more detail.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: I removed the stable ID from this, so now it is just variable ID -> variable. Let me know if that is alright.

sgrimm: Yeah, that solves the invalidation problem.

I ended up having to modify this in #2075 too, but I took a somewhat different approach. We should discuss which way we want to go.

sgrimm: My thought was that for a given variable, it should either always have a manifest ID (if it's a section) or never have one (if it's any other kind of variable). So I kept it in one map and made the manifest ID nullable in the key.

nickgraz: I see, that seems like a good way to go. I will update my PR so it is closer to yours.

nickgraz:
@sgrimm I think this is ready for a re-review

src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverableStore.kt
```diff
@@ -28,6 +28,14 @@ import org.jooq.impl.DSL
 class DeliverableStore(
     private val dslContext: DSLContext,
 ) {
+  fun deliverableIdExists(id: DeliverableId): Boolean {
+    requirePermissions { readAllDeliverables() }
+
+    return with(DELIVERABLES) {
+      dslContext.select(ID).from(this).where(ID.eq(id)).fetchOne()?.value1()?.let { true } ?: false
```
sgrimm: jOOQ has a convenience method for this:
```kotlin
dslContext.fetchExists(DELIVERABLES, DELIVERABLES.ID.eq(id))
```

src/main/kotlin/com/terraformation/backend/admin/AdminDocumentProducerController.kt
sgrimm: It's harmless to have, but we don't include `/admin` endpoints in the OpenAPI schema since they're not intended for use outside the admin UI, so this documentation will be invisible.

It might be convenient to put a link to the spreadsheet in the admin UI itself so people can navigate to it from the admin UI easily. (I should have mentioned the OpenAPI thing on the earlier PR that added the endpoint below this one.)

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -87,10 +87,24 @@ class VariableStore(
    * @throws CircularReferenceException There is a cycle in the graph of the variable and its
    *   children.
    */
-  fun fetchVariable(variableId: VariableId, manifestId: VariableManifestId? = null): Variable {
+  fun fetchVariableForManifest(manifestId: VariableManifestId, variableId: VariableId): Variable {
```
sgrimm: Curious why this is being split into two methods. In #2075 I had changed it to make the manifest ID an optional parameter (and put it after the variable ID in the argument list) such that you could do `fetchVariable(variableId, manifestId)` to get the behavior of this method, or `fetchVariable(variableId)` to get the behavior of the other method.

It's not a huge deal either way, just curious why you didn't like that design.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -187,11 +222,10 @@ class VariableStore(
    * Logic for recursively fetching a variable and the variables it's related to. Variable fetching
    * is stateful because we want to detect cycles.
    */
-  private inner class FetchContext(private val manifestId: VariableManifestId?) {
+  private inner class FetchContext(val manifestId: VariableManifestId? = null) {
```
sgrimm: Why the change from private to public here?

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -113,7 +127,28 @@ class VariableStore(
                   .and(VARIABLE_SECTIONS.PARENT_VARIABLE_ID.isNotNull))
           .orderBy(POSITION)
           .fetch(VARIABLE_ID.asNonNullable())
-          .map { fetchVariable(it, manifestId) }
+          .map { fetchVariableForManifest(manifestId, it) }
```
sgrimm: The not-exists check on `VARIABLE_TABLE_COLUMNS` can probably go away since sections can't ever be table columns.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -203,8 +237,19 @@ class VariableStore(
...
+            } else {
+              null
+            }
+        val recommendedBy =
+            if (manifestId != null) {
+              fetchRecommendedBy(variableId)
+            } else {
+              emptyList()
+            }
+
```
sgrimm: Will this now always be an empty list? A section will never be recommended by another section.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -276,15 +317,49 @@ class VariableStore(
...
+          base.manifestId
+              ?: with(VARIABLE_MANIFEST_ENTRIES) {
+                dslContext
+                    .select(VARIABLE_MANIFEST_ID)
+                    .from(this)
+                    .where(VARIABLE_ID.eq(base.id))
+                    .fetchOne(VARIABLE_MANIFEST_ID.asNonNullable())
+                    ?: throw VariableManifestNotFoundForVariableException(base.id)
+              }
+
```
sgrimm, line 315: Under what circumstances would the query happen here? Either you're fetching non-manifest variables in which case you won't be fetching any sections, or you're fetching manifest variables in which case the manifest ID is already in the fetch context.

That aside, a couple issues here:

This variable name shadows the one in `FetchContext` which is probably going to cause confusion down the road. Better to pick a distinct name.

If you upload a revised version of a manifest, any sections that didn't change will have two `variable_manifest_entries` rows and the `fetchOne()` here will fail because the query returns more than one result.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -263,7 +304,7 @@ class VariableStore(
       }
     }
 
-    private fun fetchNumber(base: BaseVariableProperties): NumberVariable {
+    protected fun fetchNumber(base: BaseVariableProperties): NumberVariable {
```
sgrimm: Why the change from `private` to `protected` on the fetch methods? There don't appear to be any subclasses of `FetchContext` that would need to inherit these methods.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableCsvValidator.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    existingDeliverableIds.add(deliverableId)
+  }
+
+  private fun validateDependencyVariableStableId(value: String?, field: String) {
+    if (value.isNullOrBlank()) {
+      return
+    }
+
+    if (value !in existingStableIds) {
```
sgrimm, line 110: Not a problem, but noting that this will mean you have to have the dependencies listed in the CSV before the dependent variables. It would be weird to list them the other way around anyway, though, and this has the nice side effect of preventing cycles. Just something to make sure the people editing the variable list are aware of.

sgrimm: Though maybe this could use a sanity check to detect a variable depending on itself (since at this point the variable's stable ID will have been added to `existingStableIds` already).

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableCsvValidator.kt
```diff
@@ -0,0 +1,272 @@
...
+    if (!configSupplied) {
+      return
+    }
+
+    dependencyConfig.forEach { (value, message) ->
+      if (value == null) {
+        addError(null, null, message)
+      }
+    }
+  }
```
sgrimm: Obviously this isn't a hot code path, but constructing all three error messages for every row in the spreadsheet feels inefficient to me. How about something like this?
```kotlin
val columns =
    listOf(
        VARIABLE_CSV_COLUMN_INDEX_DEPENDENCY_VARIABLE_STABLE_ID,
        VARIABLE_CSV_COLUMN_INDEX_DEPENDENCY_CONDITION,
        VARIABLE_CSV_COLUMN_INDEX_DEPENDENCY_VALUE)
val missingColumns = columns.filter { values[it] == null }

if (missingColumns.isNotEmpty() && missingColumns.size != columns.size) {
    missingColumns.forEach { index ->
        val columnName = getColumnName(index)
        addError(columnName, null, messages.variablesCsvDependencyConfigIncomplete(columnName))
    }
}
```
One other thing I did there was pass the column name to `addError()`. So it should no longer be necessary to also include the column name in the actual error message, that is, `Messages.variablesCsvDependencyConfigIncomplete()` can stop taking an argument.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableImporter.kt
```diff
@@ -0,0 +1,464 @@
...
+          // Import all non tables, non sections
+          importAllVariableCsvVariables(
+              csvVariables = csvVariables,
+              ignoreVariableTypes =
+                  listOf(HierarchicalVariableType.Table, HierarchicalVariableType.Section))
+
+          // Import all sections
+          importAllVariableCsvVariables(
+              csvVariables = csvVariables, importVariableType = HierarchicalVariableType.Section)
+
```
sgrimm: Will there ever be sections to import?

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/AllVariableCsvVariable.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    fun create(input: String): AllVariableCsvVariableType =
+        byValue[input] ?: throw IllegalArgumentException(LOCALIZED_ERROR_KEY_UNKNOWN_DATA_TYPE)
+  }
+}
+
+enum class HierarchicalVariableType(val value: String) {
+  Table("Table"),
+  Section("Section")
+}
```
sgrimm: Maybe this should move to its own top-level file since it's used by both the all-variables and the manifest importer.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableImporter.kt
```diff
@@ -0,0 +1,464 @@
...
+        section: SectionVariable
+    ): Boolean {
+      val csvSubsections = csvVariablesByParentPath[csvVariable.variablePath] ?: emptyList()
+
+      return csvSubsections.size == section.children.size &&
+          section.children.zip(csvSubsections).all { (subsection, csvSubsection) ->
+            canReuseExistingVariable(csvSubsection, subsection)
+          }
+    }
+
```
sgrimm: Is this needed?

src/main/kotlin/com/terraformation/backend/documentproducer/model/Variable.kt
```diff
@@ -441,15 +416,22 @@ data class SectionVariable(
         throw VariableValueInvalidException(id, "Display style is required for variable injections")
       }
 
-      val usedVariableId = value.value.usedVariableId
+      // TODO validate this somewhere else
```
sgrimm: What was wrong with the existing validation setup?

src/test/kotlin/com/terraformation/backend/documentproducer/db/VariableImporterTest.kt
```diff
@@ -0,0 +1,821 @@
...
+    @Test
+    fun `detects duplicate stable IDs`() {
+      val testCsv =
+          header +
+              "\nName X,Duplicate ID,,Number,Yes,,,,,,,,,,,,,," +
+              "\nName Y,Duplicate ID,,Number,Yes,,,,,,,,,,,,,,"
+
+      val importResult = importer.import(sizedInputStream(testCsv))
+
+      Assertions.assertEquals(
```
sgrimm: We generally use static imports for the assertion methods in `Assertions`.

src/main/resources/db/migration/0250/V282__DocumentProducerTables.sql
```diff
@@ -24,6 +29,12 @@ CREATE TABLE docprod.variables (
     stable_id TEXT NOT NULL,
     name TEXT NOT NULL,
     description TEXT,
+    deliverable_id BIGINT REFERENCES accelerator.deliverables ON DELETE SET NULL,
+    deliverable_question TEXT,
+    dependency_variable_stable_id TEXT,
```
sgrimm, line 34: We'll need to keep track of the order of the questions.

src/test/kotlin/com/terraformation/backend/documentproducer/model/VariableTest.kt
```diff
@@ -205,19 +205,19 @@ class VariableTest {
...
+    //        testValidate(
+    //            SectionVariable(baseVariable(), true),
+    //            NewSectionValue(
+    //                baseValue(),
+    //                SectionValueVariable(
+    //                    nonexistentVariableId,
+    //                    VariableUsageType.Injection,
+    //                    VariableInjectionDisplayStyle.Block)))
+    //      }
+    //    }
```
sgrimm: Might be worth a TODO here if the intent is to update this test. But it should work as-is, shouldn't it? The variable won't exist since nothing has inserted it.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/admin/AdminDocumentProducerController.kt
nickgraz: OK that makes sense, I will remove it.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: I will put it back the way you had it, much cleaner. It was lost in the shuffle when I was resolving merge conflicts.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: Another change lost in the shuffle, thanks for the catch.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: Good point... it seems I should remove this condition and also remove the `.and(VARIABLE_MANIFEST_ID.eq(manifestId))` condition in `fetchRecommendedBy`. Does this mean there is a missing test or a false positive?

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: Removing those lines makes these two tests ([1](https://github.com/terraware/terraware-server/pull/2033/files#diff-b86832f3f41fd25f07ca39b2e5cf533b98b5b04550d2f2e8f32c75f2a3ae1361R152), [2](https://github.com/terraware/terraware-server/pull/2033/files#diff-abe3f21b9a94e358017d9e6a623c4354d4edae5bdf85d3bc3cae6ec832dfbd3dR103)) fail for some reason.

I will take a look at this after resolving the rest of your comments.

Diff looks like this for reference:
```
diff --git a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
index ecc654c3257..91050eb20ca 100644
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -102,10 +102,6 @@ class VariableStore(
           .select(VARIABLE_ID)
           .from(VARIABLE_MANIFEST_ENTRIES)
           .where(VARIABLE_MANIFEST_ID.eq(manifestId))
-          .andNotExists(
-              DSL.selectOne()
-                  .from(VARIABLE_TABLE_COLUMNS)
-                  .where(VARIABLE_ID.eq(VARIABLE_TABLE_COLUMNS.VARIABLE_ID)))
           .andNotExists(
               DSL.selectOne()
                   .from(VARIABLE_SECTIONS)

```

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: Never mind, I see that it will just return an empty list if there is no manifest ID.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
nickgraz: These were needed when I had the subclasses, will put it back.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableCsvValidator.kt
nickgraz: Good call out, thanks

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableImporter.kt
nickgraz: Not in this import!

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableImporter.kt
nickgraz: Not used, not needed!

src/test/kotlin/com/terraformation/backend/documentproducer/model/VariableTest.kt
nickgraz: I put this back, it was related to removing `fetchVariable` as an argument to all of the variable validation functions, but I was able to put that back.

src/main/kotlin/com/terraformation/backend/documentproducer/model/Variable.kt
nickgraz: I had removed this when I thought I needed to split up the fetch contexts, I put it all back and everything is still working.

src/main/resources/db/migration/0250/V282__DocumentProducerTables.sql
nickgraz: I'm planning on adding that in another PR since this one has already ballooned quite a bit.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
sgrimm: I think my comment above is still valid.

src/main/kotlin/com/terraformation/backend/documentproducer/db/variable/VariableCsvValidator.kt
sgrimm: Sanity check would still be good here.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -381,10 +381,47 @@ class Messages {
...
+
+  fun variableCsvDeliverableDoesNotExist() = getMessage("variablesCsvDeliverableDoesNotExist")
+
+  fun variablesCsvDependencyConfigIncomplete() =
+      getMessage("variablesCsvDependencyConfigIncomplete")
+
+  fun variablesCsvDependencyVariableStableIdDoesNotExist() =
+      getMessage("variablesCsvDependencyVariableStableIdDoesNotExist")
+
+  fun variablesCsvNameLineBreak() = getMessage("variablesCsvNameLineBreak")
```
sgrimm, line 423: This seems redundant with line 386.

src/test/kotlin/com/terraformation/backend/documentproducer/db/VariableImporterTest.kt
```diff
@@ -0,0 +1,820 @@
...
+      val actualVariableOAL = getVariableByName("Organization Name As List")
+      val actualVariableTextOAL = variableTextsDao.fetchOneByVariableId(actualVariableOAL.id!!)
+      val actualVariablePS = getVariableByName("Prior Scenario")
+      val actualVariableTextPS = variableTextsDao.fetchOneByVariableId(actualVariablePS.id!!)
+      val actualVariablePSAL = getVariableByName("Prior Scenario As List")
+      val actualVariableTextPSAL = variableTextsDao.fetchOneByVariableId(actualVariablePSAL.id!!)
+
+      assertEquals(emptyList<String>(), importResult.errors, "no errors")
+
+      Assertions.assertFalse(actualVariableO.isList!!, "single input without list is not a list")
```
sgrimm, line 255: Should use a static import for this and the other assert methods. When I end up with this (usually after I use IntelliJ to duplicate a class and it decides to replace the existing static imports with qualified names) I usually search for `Assertions.` and walk through the file seeing what assertions I need to import.

sgrimm (approved):
I think this is good enough to merge into the feature branch, though there are a few things that could stand to be changed. Up to you if you want to address them here or in a followup.

nickgraz:
> I think this is good enough to merge into the feature branch, though there are a few things that could stand to be changed. Up to you if you want to address them here or in a followup.

Thanks @sgrimm , I'll merge it in and address these comments in another PR

PR #2079, opened 2024-06-18
Add user to "added to organization" email model

sgrimm:
Add a `user` object to the model that's accessible by the "user added to
organization" email templates. This allows the notificaiton email to greet the
user by name.

PR #2080, opened 2024-06-19
Use user firstName in the "added to organization" email

constanzauanini:
This needs to be merged after https://github.com/terraware/terraware-server/pull/2079


src/main/resources/templates/email/footer.ftlh.mjml
```diff
@@ -1,6 +1,6 @@
 <!-- <#-- @ftlvariable name="" type="com.terraformation.backend.email.model.EmailTemplateModel" --> -->
 <mj-section padding="0px" mj-class="footer-wrapper">
     <mj-column>
-        <mj-text mj-class="text-footer">${strings("notification.email.html.footer.1")}</mj-text>
+        <mj-text mj-class="text-footer"><span>&#169;</span>${strings("notification.email.html.footer.1")}</mj-text>
```
sgrimm, line 4: This seems fine for now. At some point we should figure out why it wasn't working to have the copyright symbol (rather than the `&#169;`) in the strings file; that ought to have worked.

sgrimm (approved):
Looks good, assuming the tests pass now that it has been rebased.

src/main/resources/templates/email/footer.ftlh.mjml
constanzauanini: It worked locally, but not in staging

PR #2082, opened 2024-06-20
SW-5566 Add variable owners

sgrimm:
Allow variables to be assigned to owners on a per-project basis.

Initially, this will be used to assign numbered sections to owners during the
feasibility study authoring process, but this implementation is more general
and allows any variable to have an owner.

Variable ownership is not considered part of the variable's value; it's a
separate piece of metadata that can exist even for variables that don't yet
have values in a project.

In anticipation of changing the variables API to stop requiring documents
(which will be necessary to support questionnaire deliverables), the new
endpoints added here use project IDs rather than document IDs.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -461,6 +461,8 @@ data class IndividualUser(
 
   override fun canReadProjectScores(projectId: ProjectId) = isReadOnlyOrHigher()
 
+  override fun canReadProjectVariableOwners(projectId: ProjectId) = isReadOnlyOrHigher()
+
   override fun canReadProjectVotes(projectId: ProjectId) = isReadOnlyOrHigher()
 
   override fun canReadReport(reportId: ReportId) =
@@ -613,6 +615,8 @@ data class IndividualUser(
   override fun canUpdateProjectDocumentSettings(projectId: ProjectId) = isAcceleratorAdmin()
 
   override fun canUpdateProjectScores(projectId: ProjectId): Boolean = isTFExpertOrHigher()
+
+  override fun canUpdateProjectVariableOwners(projectId: ProjectId) = isTFExpertOrHigher()
 
   override fun canUpdateProjectVotes(projectId: ProjectId): Boolean = isTFExpertOrHigher()
 
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/api/VariableOwnersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/api/VariableOwnersController.kt
@@ -1,0 +1,61 @@
+package com.terraformation.backend.documentproducer.api
+
+import com.terraformation.backend.api.InternalEndpoint
+import com.terraformation.backend.api.SimpleSuccessResponsePayload
+import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.docprod.VariableId
+import com.terraformation.backend.documentproducer.db.VariableOwnerStore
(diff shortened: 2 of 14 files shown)


sgrimm:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2082?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2082** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2082?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2033** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2033?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 14 other dependent PRs ([#2029](https://github.com/terraware/terraware-server/pull/2029) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2064](https://github.com/terraware/terraware-server/pull/2064) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2064?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2065](https://github.com/terraware/terraware-server/pull/2065) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2065?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 11 others)
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -461,6 +461,8 @@ data class IndividualUser(
 
   override fun canReadProjectScores(projectId: ProjectId) = isReadOnlyOrHigher()
 
+  override fun canReadProjectVariableOwners(projectId: ProjectId) = isReadOnlyOrHigher()
```
nickgraz, line 464: Is `projectId` needed here?

src/main/kotlin/com/terraformation/backend/documentproducer/api/VariableOwnersController.kt
```diff
@@ -0,0 +1,61 @@
...
+@RequestMapping("/api/v1/document-producer/projects/{projectId}/owners")
+@RestController
+class VariableOwnersController(
+    private val variableOwnerStore: VariableOwnerStore,
+) {
+  @GetMapping
+  @Operation(
+      summary = "List the owners of a project's variables.",
+      description = "Only variables that actually have owners are returned.")
+  fun listVariableOwners(@PathVariable projectId: ProjectId): ListVariableOwnersResponsePayload {
```
nickgraz, line 29: I wonder if we can optionally add the ownership information into the "get variables" response for admins. It will be much more straightforward on the FE if we don't need to coalesce the two API responses.

nickgraz (approved):
This all looks great, I do have a question about the usage of the `listVariableOwners` endpoint, but isn't strictly necessary to resolve in this PR

src/main/kotlin/com/terraformation/backend/documentproducer/api/VariableOwnersController.kt
sgrimm: I didn't put it there because that endpoint doesn't take a project ID as input, so there'd be no way to know which owners to return.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
sgrimm: Not needed per se, more following the pattern of making permissions fine-grained at the level of this API so we can easily make the logic more complex without having to change all the call sites (e.g., maybe we'll eventually have a team that only deals with African projects and doesn't have permission to see Asian ones).

PR #2109, opened 2024-06-27
SW-5498 Fix the POST URL used in the admin form for uploading all variables CSV

nickgraz:

PR #2110, opened 2024-06-27
SW-5498 Add function to VariableStore for fetching variable IDs for a given deliverable

nickgraz:
We already have APIs available for fetching values. Instead of rewriting that to accept deliverable ID, and couple those two areas more than needed, we will simply fetch the variable IDs for a deliverable and (in a later PR) use those variable IDs to explicitly fetch the values for a given project.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.documentproducer.db
 
+import com.terraformation.backend.db.accelerator.DeliverableId
 import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.docprod.VariableId
 import com.terraformation.backend.db.docprod.VariableManifestId
@@ -76,6 +77,13 @@ class VariableStore(
    * list.
    */
   private val replacements = ConcurrentHashMap<VariableId, List<VariableId>>()
+
+  fun fetchDeliverableVariableIds(deliverableId: DeliverableId): List<VariableId> =
+      dslContext
+          .select(VARIABLES.ID)
+          .from(VARIABLES)
+          .where(VARIABLES.DELIVERABLE_ID.eq(deliverableId))
+          .fetch(VARIABLES.ID.asNonNullable())
 
   /**
    * Returns information about a variable. Eagerly fetches other variables that are related to the
--- a/src/main/resources/templates/admin/documentProducer.html
+++ b/src/main/resources/templates/admin/documentProducer.html
@@ -41,7 +41,7 @@
 
 <h2>Import All Variable CSV</h2>
 
-<form method="POST" action="/admin/document-producer/uploadVariables" enctype="multipart/form-data">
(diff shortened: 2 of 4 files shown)


nickgraz:
* **#2112** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2112?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2111** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2111?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2110** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2109** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2109?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -77,6 +78,13 @@ class VariableStore(
    */
   private val replacements = ConcurrentHashMap<VariableId, List<VariableId>>()
 
+  fun fetchDeliverableVariableIds(deliverableId: DeliverableId): List<VariableId> =
+      dslContext
+          .select(VARIABLES.ID)
+          .from(VARIABLES)
+          .where(VARIABLES.DELIVERABLE_ID.eq(deliverableId))
+          .fetch(VARIABLES.ID.asNonNullable())
```
sgrimm, line 86: Would `orderBy(VARIABLES.DELIVERABLE_POSITION)` make sense here so the IDs are guaranteed to be in the correct order?

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -77,6 +78,13 @@ class VariableStore(
    */
   private val replacements = ConcurrentHashMap<VariableId, List<VariableId>>()
 
+  fun fetchDeliverableVariableIds(deliverableId: DeliverableId): List<VariableId> =
```
nickgraz, line 81: I wasn't able to figure out what permissions this should require. We don't really have any "able to view X deliverable permissions", which would seem appropriate here. 

In the PRDs there are permissions related to the _values_ of the variables, which we will fetch in a subsequent PR. But seeing that those permissions are basically wide open, I thought it would be OK to leave this permission-less. Let me know if that assumption is incorrect.

sgrimm: Yeah, I think right now we kind of consider deliverables to be public information. I could see maybe restricting them to people in accelerator-tagged orgs (and internal users, of course) but that seems out of scope for this change, if it's even needed at all.

sgrimm (commented):
The implementation is fine, but I'm not sure the high-level tradeoff here is worth it. This approach (referring here to the stack as a whole) seems more complex and less efficient than referencing the variables table in the "list values" code. Variable values are already tightly coupled with variables and we already reference the variables table in `VariableValueStore`, and in this context it seems fair to think of deliverable ID as just a property of a variable, same as its type or its name.

PR #2123, opened 2024-07-01
SW-5594 Reduce boilerplate when adding permissions

sgrimm:
Currently, when you add a new permission, you have to add implementations to the
system and device manager user classes, but these nearly always return true and
false respectively.

Add the concept of a "default permission" and make the permission checking methods
return it by default.

This should make it a bit less of a chore to add new permissions, at the cost of
the compiler no longer being able to detect if you add a permission method to
`TerrawareUser` and forget to implement it in `IndividualUser`. In that case,
an exception will be thrown which will cause any code that checks the new
permission to fail, so the risk should be acceptably low.


sgrimm:
* **#2123** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2123?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
This is awesome!

PR #2127, opened 2024-07-02
SW-5605 Stop using hardwired user IDs in tests

sgrimm:
As a step toward being able to run our JUnit tests in parallel, stop inserting
hardwired user IDs in tests and instead let the database allocate IDs. This will
avoid deadlocks when multiple tests try to insert the same user ID in different
transactions.

The default auth ID and email addresses were based on the hardwired user IDs;
replace those with random UUIDs which should help avoid unique-key collisions
on those as well.

Many of the references to hardwired user IDs were in the form of fetching the
user ID from the object returned by `mockUser()`. Add setup methods to
automatically insert a user into the database and make the mock return its ID;
tests can continue to reference `user.userId` as before but now it will return
a dynamically-assigned value instead of a fixed one. As a side benefit, tests
no longer need to call `insertUser()` explicitly to insert the mock user into
the database.

Some tests still use hardwired non-unique email addresses; those can be fixed
in a later change.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/MockUser.kt
+++ b/src/test/kotlin/com/terraformation/backend/MockUser.kt
@@ -2,20 +2,18 @@ package com.terraformation.backend
 
 import com.terraformation.backend.auth.CurrentUserHolder
 import com.terraformation.backend.customer.model.IndividualUser
-import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.db.default_schema.UserType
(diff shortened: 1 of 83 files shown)


sgrimm:
* **#2127** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Looks great!

PR #2134, opened 2024-07-08
SW-5506 Make the 'variable manifest' CSV work with the latest CSV format

nickgraz:
- Modified for the latest version of the CSV here:

[feasibility-study-variable-manifest-rev1.csv](https://github.com/user-attachments/files/16149305/feasibility-study-variable-manifest-rev1.csv)

Notable changes:
- Removed all mention of "variable type" from the manifest importer, since we are only importing section variables in there now
- Since names are no longer unique for variables (they used to be unique in their "manifest entry"), the "default text" for sections needs to include the variable's "stable ID", in order to link it to the default text. I updated the regex so it accepts an easier to read default text variable template. The new format is `{{Variable name - StableId}}`, `{{Project Size - 1001}}` for example. The variable name is not validated and does not hold any bearing on the variable referenced, **it is only there in the CSV to make it easier to read** which variables you are using in the sheet. **This is subject to change** - I will ask product to make sure this is alright. Other options are: to additionally validate the variable name, remove variable name entirely, come up with a different format other than ` - ` as the separator.
  - These are all valid for loading the variable with the stable ID 1111:
```
{{1111}}
{{Project Size -1111}}
{{Project Size- 1111}}
{{Project Size-1111}}
{{Anything-1111}}
  ```
- Remove all of the other variable imports from the manifest importer

### CODE REVIEW

**Comment by nickgraz** [comment]
* **#2134** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2135** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

**Comment by nickgraz** [comment]
Putting back to draft while I make the default text parser and validator

File: src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt [hunk 1]
```diff
@@ -95,6 +95,16 @@ class VariableStore(
     return variables[manifestId to variableId] ?: FetchContext(manifestId).fetchVariable(variableId)
   }
 
+  fun fetchByStableId(stableId: String): Variable? =
+      with(VARIABLES) {
+        dslContext
+            .select(ID)
+            .from(VARIABLES)
+            .where(STABLE_ID.eq(stableId))
+            .fetchOne(VARIABLES.ID.asNonNullable())
```
sgrimm: Don't need the `asNonNullable` here since the result can be null anyway if the ID isn't found.

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestCsvValidator.kt
```diff
@@ -70,6 +44,13 @@ class ManifestCsvValidator(
     return messages.manifestCsvColumnName(position)
   }
 
+  private fun validateDefaultSectionText(value: String?, field: String) {
+    // TODO
+    if (value.isNullOrBlank()) {
+      //      addError(field, value, messages.manifestCsvNameRequired())
+    }
```
sgrimm: Sections aren't required to have default text, so null/blank is a valid value, right?

But this could probably validate that the references are all syntactically valid, e.g., detecting if someone forgets the closing `}` on a reference.

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
```diff
@@ -175,9 +150,7 @@ class ManifestImporter(
             }
           }
 
-          if (variable is TableVariable) {
-            processVariables(variable.columns.map { it.variable })
-          } else if (variable is SectionVariable) {
+          if (variable is SectionVariable) {
```
sgrimm: Would it make sense for this function to take `List<SectionVariable>` instead of `List<Variable>` since there are no other variable types to process?

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
```diff
@@ -289,14 +203,15 @@ class ManifestImporter(
               renderHeading = !csvVariable.isNonNumberedSection))
 
       if (csvVariable.defaultSectionText != null) {
-        val regex = Regex("(.*?)(?:\\{\\{([^}]+)}}|\$)", RegexOption.DOT_MATCHES_ALL)
+        val regex =
+            Regex("(.*?)(?:\\{\\{(?:[^{]+-\\s*)?([0-9]+)}}|\$)", RegexOption.DOT_MATCHES_ALL)
```
sgrimm: Should the bit before the hyphen be `[^}]+` instead of `[^{]+`? The way it is now would stop at the beginning of the next variable reference rather than the end of the current one.

Also, maybe `*` instead of `+` so a reference like `{- 1234}` works (right now it would fail because there needs to be at least one character before the hyphen).

sgrimm (changes requested):
Seems generally reasonable, just a few relatively small comments.

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestCsvValidator.kt
nickgraz: Ah yes, I need to remove. Once I pulled in the missing PDD code I noticed this validation is done during the import.

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
```diff
@@ -311,26 +227,31 @@ class ManifestImporter(
...
-                        if (referencedCsvVariable != null) {
+                      if (variableStableId.isNotEmpty()) {
+                        val referencedVariable =
+                            if (defaultTextVariableByStableId.containsKey(variableStableId)) {
+                              defaultTextVariableByStableId[variableStableId]
+                            } else {
+                              defaultTextVariableByStableId[variableStableId] =
+                                  variableStore.fetchByStableId(variableStableId)
+                              defaultTextVariableByStableId[variableStableId]
+                            }
```
sgrimm: Alternative:
```kotlin
defaultTextVariableByStableId.computeIfAbsent(variableStableId) {
  variableStore.fetchByStableId(variableStableId)
}
```

sgrimm (approved):

PR #2136, opened 2024-07-09
Add missing `pdd-backend` commit for #135

nickgraz:
Adds in the missing code from `pdd-backend` when the code was initially copied into `terraware-server`. This change seems like it will affect the way the frontend reads server responses from the backend.

If this breaks the frontend, we may have to solve the [original bug](https://terraformation.atlassian.net/browse/SW-4564) in another way if possible.


nickgraz:
* **#2136** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (changes requested):
I don't think this is a safe change. The reason we added this in the first place was that the web app has a ton of places where it sends empty strings to the server for optional fields (e.g., copying the value from an empty text input) and expects the server to treat them as null values. It was messing up stuff like search API calls that search for null field values.

PR #2137, opened 2024-07-09
SW-5500 Send notifications for variables reviews

tommylau523:


tommylau523:
* **#2137** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
```diff
@@ -94,4 +108,28 @@ class SubmissionNotifier(
       }
     }
   }
+
+  /**
+   * Publishes [QuestionsDeliverableStatusUpdatedEvent] if not user-visible feedback or status has
```
sgrimm: Nit: "not" should be "no"

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
```diff
@@ -94,4 +108,28 @@ class SubmissionNotifier(
...
+   * been updated since the one referenced by the event.
+   */
+  fun notifyIfNoNewerReviews(event: QuestionsDeliverableReviewedEvent) {
+    systemUser.run {
+      val deliverableVariableWorkflowsForProject =
+          variableWorkflowStore.fetchCurrentForProject(event.projectId).filterKeys {
+            variableStore.fetchVariable(it).deliverableId == event.deliverableId
+          }
```
sgrimm: This seems like it'll be fetching a bunch of data only to throw most of it away. Alternative would be to add a fetch method that takes both a project and a deliverable ID and does this filtering in the database query.

(But maybe this isn't needed at all; see my comment below about an alternative way of detecting recent changes.)

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
```diff
@@ -94,4 +108,28 @@ class SubmissionNotifier(
...
+
+      if (event.currentWorkflows.keys == deliverableVariableWorkflowsForProject.keys &&
+          deliverableVariableWorkflowsForProject.all {
+            val deliverableWorkflow = event.currentWorkflows[it.key]
```
sgrimm: Instead of `it.key` and `it.value`, this could use destructuring to declare named variables, like `.all { (variableId, historyModel) -> ... }`

src/main/kotlin/com/terraformation/backend/documentproducer/event/QuestionsDeliverableReviewedEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.terraformation.backend.documentproducer.model.ExistingVariableWorkflowHistoryModel
+
+/**
+ * Published when a questions deliverable for a project was reviewed and contained user-visible
+ * status changes or feedback
+ */
+data class QuestionsDeliverableReviewedEvent(
+    val deliverableId: DeliverableId,
+    val projectId: ProjectId,
+    val currentWorkflows: Map<VariableId, ExistingVariableWorkflowHistoryModel>,
```
sgrimm, line 15: It looks to me like this map is being used to compare against the latest data to detect if anything has changed.

If that's the case, then it might be more efficient to replace this with a workflow history ID, and then the "has anything new happened" check goes from a map fetch and comparison to an "exists" query that checks whether there are greater workflow history IDs for variables with the same project + deliverable.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/documentproducer/event/QuestionsDeliverableReviewedEvent.kt
tommylau523: The tricky bit is that we don't want to notify users if it is only an internal comment update. So the payload is all added to allow for the check of if status or feedback have changed down the line.

tommylau523: But in addition to that, I suspect that we probably want some sort of changelogs for email notifications/in app screen (still need designs according to PRD). Maybe a better approach is to keep track of a "last-notified", and batch a bunch of history changes together and notify that way.

sgrimm: Good point about internal comment updates. In that case this approach is reasonable for now.

src/main/kotlin/com/terraformation/backend/accelerator/SubmissionNotifier.kt
tommylau523: Added a fetch method for this!

sgrimm (approved):

tommylau523:
### Merge activity

* **Jul 10, 6:56 PM EDT**: @tommylau523 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2137).
* **Jul 10, 6:57 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2137).

PR #2146, opened 2024-07-11
SW-5639 Log report validation failure reason

sgrimm:
When a user tried to submit a report and the report validation failed, we weren't
logging the failure reason which made the failure hard to diagnose after the fact.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
@@ -298,32 +298,37 @@ class ReportStore(
   fun submit(reportId: ReportId) {
     requirePermissions { updateReport(reportId) }
 
-    ifLocked(reportId) {
-      val body =
-          dslContext
-              .select(REPORTS.BODY)
-              .from(REPORTS)
-              .where(REPORTS.ID.eq(reportId))
-              .fetchOne(REPORTS.BODY)!!
-              .let { objectMapper.readValue<ReportBodyModel>(it.data()) }
-              .toLatestVersion()
+    try {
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#2146** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/report/db/ReportStore.kt
```diff
@@ -298,32 +298,37 @@ class ReportStore(
...
-      saveSeedBankInfo(body)
-      saveNurseryInfo(body)
+        saveSeedBankInfo(body)
+        saveNurseryInfo(body)
 
-      eventPublisher.publishEvent(ReportSubmittedEvent(reportId, body))
+        eventPublisher.publishEvent(ReportSubmittedEvent(reportId, body))
+      }
+    } catch (e: Exception) {
+      log.info("Report $reportId cannot be submitted: ${e.message}")
```
tommylau523, line 330: Not part of this PR: Wondering if we should do an application-wide catch/log/throw.

sgrimm: Exceptions that bubble up past the controller method are already logged (without stack traces, though, which maybe we should change). But this one was being caught by a try/catch in the controller method so the default logging didn't happen.

Auditing the controller methods and make sure any exceptions they catch get logged somewhere would probably be a good idea.

PR #2158, opened 2024-07-15
Fix the default value population for new documents

nickgraz:
- The `hasValues` query from before seemed to ignore the passed in manifest ID (due to a recent change moving document ID -> project ID), so it would produce false positives if a value has ever been saved to another document or a deliverable
- Now that values are saved against a project, and used in documents or deliverables, we need to also check against the manifest ID to determine if we can populate the values


nickgraz:
* **#2158** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):

PR #2193, opened 2024-07-23
Remember order of modules in spreadsheet

sgrimm:
Add a `position` column to the modules table to record the module's ordinal
position in the modules spreadsheet. This will allow us to list application
modules in the same order they appear on the sheet.


sgrimm:
* **#2193** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ModulesImporter.kt
```diff
@@ -9,6 +9,7 @@ import com.terraformation.backend.importer.processCsvFile
 import jakarta.inject.Named
 import java.io.InputStream
 import java.time.InstantSource
+import org.hsqldb.Tokens.POSITION
```
nickgraz: I think this import should be removed, perhaps it was imported before jOOQ updated the `MODULES` table generated class?

sgrimm: You are exactly right. Thanks for catching.

nickgraz (approved):

PR #2201, opened 2024-07-26
Fix default value population for subsequent documents

nickgraz:
- Populate default values for documents that are created, even if there are other documents in the system
- The condition wasn't checking the manifest ID, so it was failing for all documents after at least one exists in the system for the project

PR #2202, opened 2024-07-26
Fix document upgrade for section variable injections

nickgraz:
- When upgrading a document, variable fragments for `SectionVariable` s where not being carried forward correctly because it was looking for the existing variable ID within the context of the variable manifest, but non Section variables are not intrinsically tied to variable manifests anymore.
- Load all variables during the document upgrade `init` so we can determine if a section fragment for a variable injection needs to be carried forward.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
@@ -214,7 +214,7 @@ class DocumentUpgradeCalculator(
               is SectionValueText -> sectionValue.value
               is SectionValueVariable -> {
                 val existingUsedVariableId = sectionValue.value.usedVariableId
-                // This needs to come from all variables, since variables now
+                // This needs to come from all variables, since variables can
                 // exist outside the context of a manifest
                 val newUsedVariableId =
                     if (existingUsedVariableId in allVariables) {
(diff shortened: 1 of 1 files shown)


nickgraz:
* **#2202** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2201** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
```diff
@@ -210,8 +214,10 @@ class DocumentUpgradeCalculator(
               is SectionValueText -> sectionValue.value
               is SectionValueVariable -> {
                 val existingUsedVariableId = sectionValue.value.usedVariableId
+                // This needs to come from all variables, since variables now
```
sgrimm: Super minor quibble: replace "now" with "can" since someone reading this comment down the road will probably only care about why the code currently behaves the way it does, not how it differs from what used to happen in the document producer.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
nickgraz: Excellent feedback, thanks

PR #2204, opened 2024-07-30
Exclude springdoc-openapi from Renovate group

sgrimm:
We can't upgrade springdoc-openapi until a particular bug is fixed. Renovate
is supposed to not suggest upgrading dependencies to versions it has previously
suggested, but that only works for dependencies that aren't grouped together.

Add an exclusion rule to stop springdoc-openapi from being grouped with the
other Spring dependencies so it won't continue to be included when other
dependencies in the group have new versions.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -33,6 +33,8 @@
         {
             "description": "Group Spring dependencies",
             "matchPackagePatterns": ["spring"],
+            // Until https://github.com/springdoc/springdoc-openapi/issues/2639 is fixed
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#2204** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2205, opened 2024-07-30
SW-5686 When answer is updated for a Variable the status doesn't go back to In Review

constanzauanini:


src/main/kotlin/com/terraformation/backend/accelerator/VariableValueStatusNotifier.kt
```diff
@@ -0,0 +1,20 @@
+package com.terraformation.backend.accelerator
+
+import com.terraformation.backend.accelerator.event.VariableValueUpdatedEvent
+import com.terraformation.backend.db.docprod.VariableWorkflowStatus
+import com.terraformation.backend.documentproducer.db.VariableWorkflowStore
+import jakarta.inject.Named
+import org.springframework.context.event.EventListener
+
+@Named
+class VariableValueStatusNotifier(
```
nickgraz: Perhaps "updater" (instead of "notifier") would make more sense here. This class does not notify when it observes the event.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt
```diff
@@ -881,4 +883,19 @@ class VariableValueStore(
...
+
+  private fun updateStatus(
+      projectId: ProjectId,
+      values: List<ExistingValue>,
+  ) {
+    val variableRows =
+        values
+            .map { it.variableId }
+            .toSet()
+            .associateWith { variablesDao.fetchOneById(it) ?: throw VariableNotFoundException(it) }
```
nickgraz, line 896: I don't think this line is needed, you should be able to just pass the variable ID along in the event without fetching the full variable.

constanzauanini: I kept that line to check if the variable has a deliverableId. I think we don't need to change the status if it doesn't have a deliverable associated

nickgraz: Oh I see! Makes sense

nickgraz (commented):
This is looking good to me, can you add a test? [Here's](https://github.com/terraware/terraware-server/blob/97c68a85cc2e229409b098f1e1e64f9f12d1a2e3/src/test/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStoreTest.kt#L367-L387) an example of a test that is asserting the expected events are published.

nickgraz (approved):

PR #2208, opened 2024-07-30
Fix variable fetch for "replaced" variables

nickgraz:
When a variable is imported, and it has been "replaced" before (new variable due to change in all variables import csv), importing sections that reference those variables in their default text were failing to import. 

- This fixes that by ensuring we are only grabbing the most recent version of a variable when fetching it by stable ID.
- Remove redundant code in the manifest importer where the variable is loaded by stable ID.


nickgraz:
* **#2208** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):

PR #2210, opened 2024-07-30
SW-5727 Calculate weighted project score totals

sgrimm:
Treat all the "project lead" score categories as a single combined value when
calculating totalproject scores, rather than treating each of them as a
separate input into the overall average.

For phase 0, update the project score calculation so that it uses a weighted
average of a subset of the categories, and so that it only calculates the
value once all categories have been scored.


sgrimm:
* **#2210** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2210?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Looks good to me, thanks for all the comments

PR #2219, opened 2024-07-31
SW-5741 Always set application internal name

sgrimm:
When an application is created, set its internal name using "XXX" as the country
code, updating it the first time a country can be determined from a boundary.

The internal name continues to be nullable in the API because it isn't supposed
to be visible to external users. Unfortunately, we weren't actually enforcing that
visibility rule; start doing so.


sgrimm:
* **#2219** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

PR #2224, opened 2024-08-01
SW-5749 Added modified time to submission model and deliverable controller

tommylau523:


tommylau523:
* **#2224** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2224?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -1059,6 +1059,8 @@ abstract class DatabaseBackedTest {
       createdTime: Instant = Instant.EPOCH,
       deliverableId: Any? = inserted.deliverableId,
       feedback: String? = null,
+      modifiedBy: UserId = currentUser().userId,
```
sgrimm: This is what's causing `PermissionTest` to fail. In that test, because we need to manipulate which user's permissions are being checked, there is no "current user" and we pass the user ID in explicitly to all the insert functions.

The right fix here is probably to change this line to `modifiedBy: UserId = createdBy,`.

Could also do the same with the modified time (default it to the created time) so it doesn't end up earlier than the created time if the caller only passes in a created time.

tommylau523:
### Merge activity

* **Aug 1, 5:49 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2224).

PR #2231, opened 2024-08-02
SW-5605 Stop using hardwired facility IDs in tests

sgrimm:
As a step toward being able to run tests in parallel, eliminate nearly all uses of
hardwired facility IDs in tests, instead letting the database allocate IDs
dynamically.

The "insert" methods in `DatabaseBackedTest` whose facility ID arguments used to
default to the hardwired ID now use the last inserted ID instead; this eliminates
the need to pass in facility IDs explicitly in a number of tests.

The exception is `PermissionTest` where the numeric values of the IDs affect the
test logic; that one will require more extensive reworking.


sgrimm:
* **#2231** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2231?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Great work, thanks for trudging through this work

PR #2240, opened 2024-08-05
SW-5763 Include project name with list document responses, delete unused function in DocumentStore

nickgraz:
- Add nullable `projectName` field to `ExistingDocumentModel` and change the two fetch methods in the `DocumentStore` that are used in the "list documents" API call to use the DSL Context instead of DAO, and plumb in the Project Name field
- Removes an unused function in the `DocumentStore` - `fetchProjectId` (only usage removed [here](https://github.com/terraware/terraware-server/commit/aed4f6dff473c5df8797456a5b0a48f990661f13#diff-4635e395b44f0b9a82689f3909db748d1e303e6af7388c5a870b3158e72c2597L114))

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
@@ -38,7 +38,6 @@ data class ExistingDocumentModel(
       name = row.name!!,
       ownedBy = row.ownedBy!!,
       projectId = row.projectId!!,
-      projectName = null,
       status = row.statusId!!,
       variableManifestId = row.variableManifestId!!,
   )
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
@@ -2,12 +2,15 @@ package com.terraformation.backend.documentproducer.model
 
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.default_schema.tables.references.PROJECTS
 import com.terraformation.backend.db.docprod.DocumentId
 import com.terraformation.backend.db.docprod.DocumentStatus
 import com.terraformation.backend.db.docprod.DocumentTemplateId
 import com.terraformation.backend.db.docprod.VariableManifestId
 import com.terraformation.backend.db.docprod.tables.pojos.DocumentsRow
+import com.terraformation.backend.db.docprod.tables.references.DOCUMENTS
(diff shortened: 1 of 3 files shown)


nickgraz:
* **#2242** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2240** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2240?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
```diff
@@ -19,6 +22,7 @@ data class ExistingDocumentModel(
     val name: String,
     val ownedBy: UserId,
     val projectId: ProjectId,
+    val projectName: String? = null,
```
sgrimm: Is it ever actually valid for this to be null, or is this just because the `DocumentsRow` constructor doesn't have a value for the project name?

If that's the only reason, maybe it would make sense to have the `DocumentsRow` constructor take the project name as a second argument, or alternately, replace all the uses of that constructor with the new `of()` function and make sure callers pass in a record that includes the project name.

sgrimm (approved):
This is good to merge as-is. I had one comment that you can think about or ignore as you see fit.

src/main/kotlin/com/terraformation/backend/documentproducer/model/ExistingDocumentModel.kt
nickgraz: It is only really used by the "list" calls, so maybe I should just make a separate "existing model". I looked at all the usages, and decided against refactoring everything using the model since it is used in a few different places. 

To answer your question, no, it technically isn't valid for it to be null, but based on where it is being called from, it may be null because it consumers don't care about it. Saying that out loud, perhaps it should be changed up. I have a subsequent PR that adds yet another field to this payload, which, again, is only there under certain circumstances. It seems that a new, slightly different model would make sense here.

nickgraz: I noticed there was another ticket that needs the project name, for the "get single document" call. So I ended up replacing all uses of the Row constructor with the Record `of()` function.

sgrimm (approved):

nickgraz:
### Merge activity

* **Aug 6, 5:09 PM EDT**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2240).
* **Aug 6, 5:10 PM EDT**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2240).

PR #2242, opened 2024-08-05
SW-5764 Add latestSavedVersionId to the "get documents" API response payload

nickgraz:
- This is used in the table view in the frontend, only the ID is needed.


nickgraz:
* **#2242** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2240** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/db/DocumentStore.kt
```diff
@@ -229,14 +229,25 @@ class DocumentStore(
...
-          .map { ExistingDocumentModel.of(it) }
+  private fun fetchByCondition(condition: Condition? = null): List<ExistingDocumentModel> {
+    val lastSavedVersionIdField =
+        with(DOCUMENT_SAVED_VERSIONS) {
+          DSL.field(
+              DSL.select(ID)
+                  .from(DOCUMENT_SAVED_VERSIONS)
+                  .where(DOCUMENT_ID.eq(DOCUMENTS.ID))
+                  .orderBy(CREATED_TIME.desc())
+                  .limit(1))
```
sgrimm: How much do we care about this being sorted by created time specifically? An alternate approach that will probably give the same result nearly all the time would be
```kotlin
DSL.select(DSL.max(ID))
    .from(DOCUMENT_SAVED_VERSIONS)
    .where(DOCUMENT_ID.eq(DOCUMENTS.ID))
```
If you wanted to get super fancy, this could also be done with a window function, but that's a bit much for this query, I think.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/db/DocumentStore.kt
nickgraz: Created time is just a proxy for "the latest one", which is probably served better by the ID as you point out, thanks Steve!

src/main/kotlin/com/terraformation/backend/documentproducer/db/DocumentStore.kt
```diff
@@ -101,7 +101,7 @@ class DocumentStore(
     return ExistingSavedVersionModel(versionsRow)
   }
 
-  fun findAll(): List<ExistingDocumentModel> = fetchByCondition()
+  fun fetchAll(): List<ExistingDocumentModel> = fetchByCondition()
```
nickgraz, line 104: This was renamed to be consistent with other functions that are fetching things from the database using the DSL context as opposed to the DAO classes.

sgrimm:
Is the stacking correct here? Looks like this now includes adding project name too.

nickgraz:
> Is the stacking correct here? Looks like this now includes adding project name too.

This is fixed

sgrimm (approved):

nickgraz:
### Merge activity

* **Aug 6, 5:09 PM EDT**: @nickgraz started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2242).
* **Aug 6, 5:12 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2242) rebased this pull request as part of a merge.
* **Aug 6, 5:21 PM EDT**: @nickgraz merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2242).

PR #2248, opened 2024-08-07
Add DocumentsTable to Search API

nickgraz:
- Also changed the field type for `idWrapperField` so we can use fields created with `DSL.field` as well.
- Added as sublist to `projects` search table so we can access through `projects.documents` prefix.
- Also added a `DocumentTemplatesTable` as a single value sub list accessible through the documents table

Sample query:
```
{
    "prefix": "projects.documents",
    "fields":
    [
        "createdTime",
        "documentTemplate_id",
        "documentTemplate_name",
        "id",
        "lastSavedVersionId",
        "modifiedTime",
        "name",
        "project_id",
        "project_name",
        "status"
    ],
    "search":
    {
        "operation": "and",
        "children": []
    },
    "sortOrder":
    [
        {
            "field": "name",
            "direction": "Ascending"
        }
    ],
    "count": 1000
}

### CODE REVIEW

**Comment by nickgraz** [comment]
* **#2248** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

File: src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt [hunk 1]
```diff
@@ -152,7 +154,9 @@ class SearchController(
 
   private fun resolvePrefix(prefix: String?): SearchFieldPrefix {
     val table =
-        if (prefix != null) {
+        if (nonOrganizationTables.containsKey(prefix)) {
+          nonOrganizationTables[prefix]!!
```
nickgraz: The original error I was seeing before this change was "Sublist documents not found" ([from here](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/search/SearchTable.kt#L219))

src/main/kotlin/com/terraformation/backend/search/table/DocumentsTable.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jooq.impl.DSL
+import org.jooq.impl.SQLDataType
+
+class DocumentsTable(tables: SearchTables) : SearchTable() {
+  override val primaryKey: TableField<out Record, out Any?>
+    get() = DOCUMENTS.ID
+
+  override val sublists: List<SublistField> by lazy {
+    with(tables) {
+      listOf(projects.asMultiValueSublist("projects", PROJECTS.ID.eq(DOCUMENTS.PROJECT_ID)))
```
sgrimm: I think this should be a single-value sublist since a document can only have one project. (And it should be named `project` instead of `projects` to reflect that.)

Also, the projects table probably ought to have a multi-value sublist pointing to this table. Which would also eliminate the need to add special handling of prefixes; you'd be able to address this table with a prefix of `projects.documents`.

src/main/kotlin/com/terraformation/backend/search/table/DocumentsTable.kt
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          timestampField("createdTime", DOCUMENTS.CREATED_TIME, nullable = false),
+          idWrapperField("documentTemplateId", DOCUMENTS.DOCUMENT_TEMPLATE_ID) {
+            DocumentTemplateId(it)
+          },
+          idWrapperField("id", DOCUMENTS.ID) { DocumentId(it) },
+          integerField(
```
sgrimm: Should this be an `idWrapperField` instead? Then you don't need to cast to an integer.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/search/table/DocumentsTable.kt
nickgraz: I did try that but ended up with type errors.

sgrimm: Ah, okay, let me look at that and see if I can make it not require a `TableField`.

sgrimm: Looks like there's no reason to require a `TableField`. To make this work you can edit `IdWrapper.kt` to change `TableField<*, T?>` to `Field<T?>` (make sure you import the `Field` class from the jOOQ library, not the one from the Java standard library).

nickgraz: I was able to make it work in my latest commit

nickgraz: Sweet, that's what I went for!

src/main/kotlin/com/terraformation/backend/search/table/DocumentsTable.kt
```diff
@@ -0,0 +1,62 @@
...
+          idWrapperField("id", DOCUMENTS.ID) { DocumentId(it) },
+          idWrapperField(
+              "lastSavedVersionId",
+              with(DOCUMENT_SAVED_VERSIONS) {
+                DSL.field(
+                    DSL.select(ID)
+                        .from(DOCUMENT_SAVED_VERSIONS)
+                        .where(DOCUMENT_ID.eq(DOCUMENTS.ID))
+                        .orderBy(ID.desc())
+                        .limit(1))
```
sgrimm, line 45: Did the `DSL.max()` approach not work with the ID wrapper field?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/table/DocumentsTable.kt
nickgraz: Unfortunately it didn't work, it started complaining that the other fields needed to use an aggregate function.

PR #2254, opened 2024-08-08
SW-5795 Use dynamic organization IDs in tests

sgrimm:
As a step toward being able to run tests in parallel, eliminate all the
hardwired organization IDs from tests, with the exception of `PermissionTest`
which will be updated separately.

This is a somewhat large change because a lot of tests used to use the default
organization ID from `DatabaseBackedTest`, which no longer exists. All the
insert methods that take organization ID arguments now default to using the
most recently inserted organization ID instead; in some places this means we
can get rid of explicit organization ID arguments at call sites.

The organization ID arguments of the various insert methods are now required
to be of type `OrganizationId` since there's no longer a valid reason to use
numeric literals.


sgrimm:
* **#2254** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2254?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
🙌

PR #2265, opened 2024-08-12
SW-5563 Update email notification for Deliverable updates to display more details

constanzauanini:


src/main/resources/templates/email/deliverable/readyForReview/body.txt.ftl
```diff
@@ -4,6 +4,7 @@ Deliverable was submitted by ${participantName}
 A deliverable was submitted by ${participantName} and is ready for review for approval.
 
 ${deliverableUrl}
+${deliverable}
```
sgrimm: This should probably include the project, category, and deliverable name too; `${deliverable}` will render the whole object as a string.

sgrimm (changes requested):
This generally looks good! There's just one minor thing that should be fixed before merging.

src/main/resources/templates/email/deliverable/readyForReview/body.txt.ftl
constanzauanini: Oh, ok! Will update it, but also, what is the purpose of rendering the strings in this file? I wan't sure if I needed to include all the new text here

sgrimm: It's probably a bit of overengineering, but the idea is that if someone has HTML email disabled or they're getting an excerpt of the email over a channel that doesn't support HTML (phone notification, etc.), they'll see the plain text version instead.

src/test/kotlin/com/terraformation/backend/accelerator/SubmissionNotifierTest.kt
```diff
@@ -108,10 +109,12 @@ class SubmissionNotifierTest : DatabaseTest(), RunsAsUser {
       insertSubmissionDocument()
       val documentId = insertSubmissionDocument()
 
+      val deliverable = mockDeliverable()
```
constanzauanini: Is this the correct way to get the deliverable @sgrimm @nickgraz? I'm not sure how to do it here

nickgraz: Since the deliverable is added to the test DB on line 87, you can either pull it from the database ([like this](https://github.com/terraware/terraware-server/blob/main/src/test/kotlin/com/terraformation/backend/accelerator/SpeciesNotifierTest.kt#L102)) or simply create an "expected deliverable" where you need it ([like this](https://github.com/terraware/terraware-server/blob/main/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableCompleterTest.kt#L206)).

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -634,11 +639,13 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
 
     insertUserDeliverableCategory(DeliverableCategory.GIS, user.userId)
     val deliverableId = insertDeliverable(deliverableCategoryId = DeliverableCategory.GIS)
+    val deliverable =
```
constanzauanini, line 643: I'm getting "No permission to read all deliverables" in these tests, but there is a     `every { user.canReadAllDeliverables() } returns true ` in the setUp function, not sure what I'm doing wrong

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -609,12 +611,15 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
     val participantId = insertParticipant(name = "participant1", cohortId = cohortId)
     val projectId = insertProject(participantId = participantId)
     insertModule()
+    insertCohortModule()
```
sgrimm, line 614: The problem here was that `fetchDeliverableSubmissions` needs to get information about the cohort's module (for due dates), and the module wasn't associated with the cohort. So the SQL query returned no rows.

src/test/kotlin/com/terraformation/backend/email/EmailNotificationServiceTest.kt
```diff
@@ -244,6 +247,31 @@ internal class EmailNotificationServiceTest {
...
+          organizationName = organization.name,
+          participantId = participant.id,
+          participantName = participant.name,
+          projectId = project.id,
+          projectName = project.name,
+          status = SubmissionStatus.Completed,
+          submissionId = SubmissionId(1),
+          templateUrl = null,
+          type = DeliverableType.Questions,
+      )
```
sgrimm, line 274: The problem here was that the code was trying to access information from `deliverable` (including its ID) but it was a mock object that wasn't configured to return any values. `DeliverableSubmissionModel` is just a bundle of data, not a class with any real behavior of its own, so constructing a real instance of it is fine. This instance is returned by `deliverableStore.fetchDeliverableSubmissions()` due to the `every {}` call you added, and then the code can access the fields here as needed.

src/test/kotlin/com/terraformation/backend/customer/AppNotificationServiceTest.kt
```diff
@@ -263,7 +263,6 @@ internal class AppNotificationServiceTest : DatabaseTest(), RunsAsUser {
     every { user.canReadSpecies(any()) } returns true
     every { user.locale } returns Locale.ENGLISH
     every { user.organizationRoles } returns mapOf(organizationId to Role.Admin)
-    every { deliverableStore.fetchDeliverableSubmissions() } returns listOf(deliverable)
```
sgrimm, line 266: In this test, unlike the email one, we're inserting real values into the database, so `deliverableStore` is a real instance of `DeliverableStore` rather than a mock.

src/test/kotlin/com/terraformation/backend/email/EmailNotificationServiceTest.kt
constanzauanini: Nice, thanks for explaining this! I was trying to create an instance using this example https://github.com/terraware/terraware-server/blob/main/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableCompleterTest.kt#L206 but I was getting errors in other parts of the code when I changed the DeliverableSubmissionModel, and I didn't try creating it here

sgrimm (approved):

PR #2287, opened 2024-08-13
SW-5841 Run variable workflow update as system user

nickgraz:
Since a regular org user can update a variable value, which consequently updates the variable's workflow details (by setting it to "In Review"), run the update as a system user with elevated permissions.

In order to ensure we know who updated a variable's status, the user ID can optionally be passed in to the update function, otherwise it defaults to the current user.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/VariableValueStatusUpdater.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/VariableValueStatusUpdater.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.accelerator
 
 import com.terraformation.backend.accelerator.event.VariableValueUpdatedEvent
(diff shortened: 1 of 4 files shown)


nickgraz:
* **#2287** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2287?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):
This approach means the "created by" value in the history entries that are triggered by value updates will all show up with the system user's user ID rather than the user ID of the person who edited the value. Maybe that's okay (in which case this change is good to go as-is) but if not, then we might need to pass the user ID in to the update function or something.

nickgraz:
@sgrimm This was updated to allow passing the user ID in.

sgrimm (approved):

PR #2292, opened 2024-08-14
SW-5553 Return HTTP 507 on failed document uploads

sgrimm:
The client needs to be able to indicate to end users that the system has already
notified customer support about a failed document upload. Make the server return
HTTP 507 (which is named "insufficient storage" but is actually a catch-all WebDAV
code for the server being unable to store a file) in cases where it has attempted
to generate a support ticket.


sgrimm:
* **#2292** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2304, opened 2024-08-20
Added country code column to applications

tommylau523:


tommylau523:
* **#2304** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2304?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (changes requested):
The country detector is very CPU-intensive for complex site boundaries. It would be better to calculate it once and store it in the application data.

One possibility would be to create the project accelerator details earlier. Right now it gets created (with the country code populated from the country detector) when pre-screening succeeds, but we could instead do it either at application creation time or when a valid boundary is set.

src/main/resources/db/migration/0250/V298__ApplicationCountryCode.sql
```diff
@@ -0,0 +1,9 @@
+-- Add country_code column
+ALTER TABLE accelerator.applications ADD country_code TEXT REFERENCES countries (code);
+
+-- Migrate existing columns by looking up first 3 of internal name
+UPDATE accelerator.applications
+SET country_code = countries.code
+FROM countries
+WHERE SUBSTR(accelerator.applications.internal_name, 1, 3) = countries.code_alpha3
+  AND SUBSTR(accelerator.applications.internal_name, 1, 3) != 'XXX';
```
sgrimm: Out of curiosity, what happens if you leave off the XXX condition? I think it would end up setting the value to null since the query wouldn't find a countries row.

Having the condition is fine for clarity, of course.

src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationStore.kt
```diff
@@ -503,7 +503,11 @@ class ApplicationStore(
    *
    * @return The name that was actually used, possibly including a suffix.
    */
-  private fun updateInternalName(applicationId: ApplicationId, internalName: String): String {
+  private fun updateInternalName(
```
sgrimm: This function name feels a little misleading now. I wonder if it'd be better to split the country code update out into its own helper function. Then you don't have to pass `null` to this at one of the call sites, which kind of felt like a bit of a code smell to me.

sgrimm (approved):
This can be merged as-is but I have one suggestion you might want to think about.

PR #2309, opened 2024-08-20
SW-5856 Remove "get variables by manifest" API and implement "get variables by document" API

nickgraz:
Remove the query parameter usage of the "get variables" API for getting manifest variables, and replace it with one that gets all variables for a document. This includes the "manifest variables" for the document's variable manifest ID and the version-specific variables (as opposed to the "current" version of variables) that are injected into any of the document's section texts. 

This allows the frontend to display out of date variables that may have been updated, and the user has not yet "upgraded" their document's reference to the updated variable.


nickgraz:
* **#2309** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2309?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
```diff
@@ -0,0 +1,44 @@
...
+  fun fetchDocumentVariables(documentId: DocumentId): List<Variable> {
+    requirePermissions { readDocument(documentId) }
+
+    val document = documentStore.fetchDocumentById(documentId)
+    val manifestVariables = variableStore.fetchManifestVariables(document.variableManifestId!!)
+
+    val sectionVariableIds =
+        manifestVariables.flatMap { parent ->
+          when (parent) {
+            is SectionVariable -> parent.children.map { child -> child.id }
```
sgrimm: This won't handle more than one level of section nesting, I think.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+    val injectedSectionValues =
+        variableValueStore
+            .listValues(projectId = document.projectId!!, variableIds = sectionVariableIds)
+            .map { it.value }
+            .filterIsInstance<SectionValueVariable>()
+
+    val injectedSectionValueVariables =
+        injectedSectionValues.map { variableStore.fetchVariable(it.usedVariableId) }
```
sgrimm: This is going to repeatedly fetch the same variable if it's referenced multiple times in a document. Deduplicating the list of used variable IDs would help.

sgrimm: I wonder if it makes sense to do a custom fetch method to get the referenced variables. This process is going to fetch a bunch of data it doesn't end up needing, and do more round trips to the database than may be needed.

Maybe something like
```kotlin
  fun fetchUsedVariableIds(documentId: DocumentId): Set<VariableId> {
    return dslContext
        .select(
            VARIABLE_VALUES.VARIABLE_ID,
            VARIABLE_VALUES.LIST_POSITION,
            VARIABLE_SECTION_VALUES.USED_VARIABLE_ID)
        .distinctOn(
            VARIABLE_VALUES.VARIABLE_ID,
            VARIABLE_VALUES.LIST_POSITION)
        .from(VARIABLE_VALUES)
        .join(VARIABLE_SECTION_VALUES)
        .on(VARIABLE_VALUES.ID.eq(VARIABLE_SECTION_VALUES.VARIABLE_VALUE_ID))
        .where(VARIABLE_SECTION_VALUES.USED_VARIABLE_ID.isNotNull)
        .orderBy(
            VARIABLE_VALUES.VARIABLE_ID,
            VARIABLE_VALUES.LIST_POSITION,
            VARIABLE_VALUES.ID.desc())
        .fetchSet(VARIABLE_SECTION_VALUES.USED_VARIABLE_ID.asNonNullable())
  }
```
to fetch the referenced IDs, and then a new `VariableStore.fetchVariables()` that takes a collection of variable IDs and fetches them all at once?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
sgrimm: I forgot `VariableStore` caches variable data, so fetching multiple times isn't that big a deal. But deduplicating would still be good so the return value doesn't list the same variable multiple times.

nickgraz: I will give that a go, thanks.

nickgraz: I had to change it up a bit to include the document ID in my latest commit.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
nickgraz: This is all changed up now that I am using a custom fetch query, and I've included a test that has values injected into a "grandchild" section.

sgrimm (approved):

PR #2313, opened 2024-08-22
Fix intermittent test failures

sgrimm:
`BatchStoreWithdrawTest` and `ModuleStoreTest` were assuming that unordered 
database queries would return results in a certain order, which was usually the
case but not 100% of the time. Change them to ignore the order.


sgrimm:
* **#2313** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

sgrimm:
The irony! Looks like there is another intermittent test failure of the same variety here. Will fix it too.

PR #2316, opened 2024-08-22
SW-5821 Write feedback as HTML

tommylau523:


tommylau523:
* **#2316** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2316?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationStore.kt
```diff
@@ -428,7 +428,8 @@ class ApplicationStore(
   private fun updatePrescreenFeedback(applicationId: ApplicationId, feedback: List<String>) {
     val feedbackField =
         if (feedback.isNotEmpty()) {
-          feedback.joinToString(separator = "\n")
+          feedback.joinToString(
```
sgrimm, line 424: We should probably HTML-escape the feedback strings here just so we don't get bitten if at some point down the road we end up adding a problem message that includes a user-entered string value, like "Project name $name is too short" or something.

Spring has a utility function for this, `HtmlUtils.htmlEscape()`.

sgrimm (approved):

tommylau523:
### Merge activity

* **Aug 22, 12:46 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2316).

PR #2322, opened 2024-08-26
SW-5900 Stop requiring map area to match land use totals

sgrimm:
The original requirement for pre-screening was to check that the area of the map
boundary was approximately equal to the total of the hectare counts for different
land usage model types, but that turned out to be too finicky. Instead, make sure
both the boundary area and the total of the land usage model type hectares fall
within the minimum and maximum for the site's country.


sgrimm:
* **#2322** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2323, opened 2024-08-27
SW-5745 Add created/modified info to CohortPayload

sgrimm:
The web app needs to show when cohorts were created and modified, and by whom;
return that information with the other data about a cohort.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
@@ -501,6 +501,10 @@ data class IndividualUser(
 
   override fun canReadUser(userId: UserId) = isAcceleratorAdmin()
 
+
+
+
+  
   override fun canReadUserDeliverableCategories(userId: UserId) = isAcceleratorAdmin()
 
   override fun canReadViabilityTest(viabilityTestId: ViabilityTestId) =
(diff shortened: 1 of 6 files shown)


sgrimm:
* **#2323** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2323?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -501,6 +501,10 @@ data class IndividualUser(
 
   override fun canReadUser(userId: UserId) = isAcceleratorAdmin()
 
+
+
+
+
```
nickgraz: I think these should go

sgrimm: Yeah, I was testing a pre-push hook that failed to prevent this from getting pushed! Sorry you had to see it.

nickgraz (approved):
Looks good, thanks for the addition of the test assertion messages.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
nickgraz: Ah, no problem!

PR #2339, opened 2024-08-29
WIP - Pass project ID along with deliverable ID so we can return outdated variables with values for deliverables.

nickgraz:
Since a variable can be updated, which creates a new variable, a questionnaire deliverable can have a new variable with no values, even though the user has already filled in an answer for the variable. In order to fix this, whenever the frontend requests the variables that belong to a deliverable, they must pass in the project ID. This allows us to return back variables that have outdated values, so the frontend can show the outdated values, force the UI to revalidate them, and ultimately resubmit the old values for the new variables, "upgrading" the variables in the process.


nickgraz:
* **#2339** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm:
Closing this since we decided to take a different approach to the problem of variable upgrades.

PR #2342, opened 2024-08-30
Remove Spring Boot R2DBC workaround

sgrimm:
At one point there was a problem with Spring Boot's jOOQ autoconfiguration where
it would try to set up a database library we don't use; as a workaround we added
configuration to explicitly disable that library. The problem has since been
fixed in Spring Boot, so the workaround isn't needed any more.


sgrimm:
* **#2342** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

PR #2347, opened 2024-08-30
Added ApplicationSubmittedEvent

tommylau523:
Added an event for when an application is fully submitted. This will later be used for sending notifications

PR #2348, opened 2024-08-30
Added application recipient store

tommylau523:
Implemented a whitelist of users to receive application submitted notifications

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationRecipientsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationRecipientsStore.kt
@@ -1,0 +1,62 @@
+package com.terraformation.backend.accelerator.db
+
+import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.accelerator.tables.daos.ApplicationRecipientsDao
+import com.terraformation.backend.db.accelerator.tables.references.APPLICATION_RECIPIENTS
+import com.terraformation.backend.db.asNonNullable
+import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.default_schema.tables.references.USERS
+import jakarta.inject.Named
+import java.time.InstantSource
+import org.jooq.Condition
+import org.jooq.DSLContext
+import org.jooq.impl.DSL
+
+@Named
+class ApplicationRecipientsStore(
+    private val applicationRecipientsDao: ApplicationRecipientsDao,
+    private val clock: InstantSource,
+    private val dslContext: DSLContext,
+) {
+  fun fetch(): Set<UserId> {
+    requirePermissions { readApplicationRecipients() }
+
+    return with(APPLICATION_RECIPIENTS) {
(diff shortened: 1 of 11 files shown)


tommylau523:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2348?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2350** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2349** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2348** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2348?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2347** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2347?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2358](https://github.com/terraware/terraware-server/pull/2358) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2358?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

sgrimm (changes requested):
I understand the argument for not using deliverable categories for this, but "internal people get specific notification types related to their job" feels like something we're better off generalizing.

I just looked ahead at the other PRs in the stack and it looks like a more unified, general approach is what the application code is going to be doing anyway, but with special cases to map this table to its generalized implementation. That feels more complicated than it needs to be to me.

How about this instead? Table names are examples; maybe you can find better ones.

1. Add an enum table for internal people's interest areas, with values for all the deliverable categories plus `Sourcing`. I'll call it `internal_interests`.
2. Add a new table for which users have which interest areas, e.g., `user_internal_interests`. Basically identical to `user_deliverable_categories`, just pointing to the new table instead.
3. Copy all the existing `user_deliverable_categories` data to `user_internal_interests`, mapping the deliverable category IDs to interest IDs. (See below for a thought about mapping between these enums.)
4. Update all the code that interacts with `user_deliverable_categories` to use `user_internal_interests` instead. Map deliverable categories to interests as needed.
5. The new application notification is then just a notification with the `Sourcing` interest and doesn't need any special-case handling.
6. Drop `user_deliverable_categories`.

That would avoid the need for a second store class and also avoid the need for special handling in the UI and separate support in the API.

An approach for mapping that doesn't require keeping the IDs in sync or writing out all the cases in a `when` block in the application code: add an `internal_interest_id` column to `deliverable_categories` to hold the mapping, such that in the application code you just need to reference the field on the enum. We do something similar with the `notification_types` enum table, which refers to `notification_criticalities`. One minor quirk of doing it this way is that you'd need to put `internal_interests` before `deliverable_categories` in `R__TypeCodes.sql`.

tommylau523:
ahh that does make a lot more sense. I was trying to not disturb the existing `user_deliverable_categories` but I like the approach you outlined a lot more.

PR #2367, opened 2024-09-04
SW-5706 Implemented endpoint to return application boundary as a GeoJson

tommylau523:


tommylau523:
* **#2367** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2367?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/model/ApplicationModel.kt
```diff
@@ -25,6 +30,38 @@ data class ExistingApplicationModel(
     val projectName: String,
     val status: ApplicationStatus,
 ) {
+  fun toGeoFeature(): SimpleFeature {
+    if (boundary == null) {
+      throw IllegalStateException("Application has no boundary")
+    }
+
+    val featureTypeBuilder = SimpleFeatureTypeBuilder()
```
tommylau523, line 38: We did not use our types, and instead Primitive types here because I don't know how to pass our own JSON serializer to  `FeatureJSON`

src/main/kotlin/com/terraformation/backend/api/Extensions.kt
```diff
@@ -2,7 +2,10 @@ package com.terraformation.backend.api
 
 import com.terraformation.backend.file.SizedInputStream
 import jakarta.ws.rs.NotSupportedException
+import java.io.ByteArrayOutputStream
 import org.apache.tika.mime.MimeTypes
+import org.geotools.api.feature.simple.SimpleFeature
+import org.geotools.geojson.feature.FeatureJSON
```
tommylau523, line 8: This is a handy tool to output to GeoJSON. It doesn't seem to support using a custom serializer. At some point we can build our own GeoJSON builder.

src/main/kotlin/com/terraformation/backend/accelerator/api/ApplicationsController.kt
```diff
@@ -88,6 +93,19 @@ class ApplicationsController(
         deliverables.map { ApplicationDeliverablePayload(it) })
   }
 
+  @GetMapping("/{applicationId}/geoJson", produces = ["application/geo+json"])
+  @Operation(summary = "Get GeoJSON for an application")
+  fun getApplicationGeoJson(
+      @PathVariable applicationId: ApplicationId
+  ): ResponseEntity<InputStreamResource> {
```
sgrimm: If at some point we want to support exporting in other formats, we can maybe consider making this a less GeoJSON-specific API the way, e.g., the search API looks at the `Accept` HTTP header to tell if it should return a CSV file.

Not saying we should do that now, but maybe we should make the URL suffix something like `boundary` or `export` instead of `geoJson`.

src/main/kotlin/com/terraformation/backend/accelerator/model/ApplicationModel.kt
```diff
@@ -25,6 +30,38 @@ data class ExistingApplicationModel(
...
+    val featureBuilder = SimpleFeatureBuilder(featureType)
+    featureBuilder.set("the_geom", boundary.toMultiPolygon())
+    featureBuilder.set("applicationId", id.value)
+    countryCode?.let { featureBuilder.set("countryCode", it) }
+    featureBuilder.set("internalName", internalName)
+    featureBuilder.set("organizationId", organizationId.value)
+    featureBuilder.set("organizationName", organizationName)
+    featureBuilder.set("projectId", projectId.value)
+    featureBuilder.set("projectName", projectName)
+    featureBuilder.set("status", status.jsonValue)
```
sgrimm, line 61: Should this map to external statuses if the user isn't an accelerator admin?

src/main/kotlin/com/terraformation/backend/accelerator/model/ApplicationModel.kt
```diff
@@ -25,6 +30,38 @@ data class ExistingApplicationModel(
...
+    featureTypeBuilder.add("projectName", String::class.java)
+    featureTypeBuilder.add("status", String::class.java)
+
+    val featureType = featureTypeBuilder.buildFeatureType()
+
+    val featureBuilder = SimpleFeatureBuilder(featureType)
+    featureBuilder.set("the_geom", boundary.toMultiPolygon())
+    featureBuilder.set("applicationId", id.value)
+    countryCode?.let { featureBuilder.set("countryCode", it) }
+    featureBuilder.set("internalName", internalName)
```
sgrimm, line 56: Won't this expose the internal name to end users?

sgrimm (changes requested):
The PR could use a more precise title and a description. When I first saw it pop up in my inbox, I thought, "Wait, don't basically _all_ our endpoints return objects in JSON format already?"

src/main/kotlin/com/terraformation/backend/accelerator/model/ApplicationModel.kt
tommylau523: This is only an internal endpoint. But I think I probably should hide that behind a store and permission.

sgrimm: Ah, I assumed it was meant to be end-user accessible since nothing was checking for permissions beyond the ability to read the application.

src/main/kotlin/com/terraformation/backend/accelerator/api/ApplicationsController.kt
tommylau523: Agreed! I did capture this in the JIRA ticket I think. But let me make sure to have it in the backlog.

sgrimm (approved):

tommylau523:
### Merge activity

* **Sep 5, 7:19 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2367).

PR #2369, opened 2024-09-05
SW-5928 Image variable as a part of questionnaire deliverable never marked as submitted

constanzauanini:
I think this is happening because it is never marked as in review


tommylau523 (commented):
Could you add a test for this?

sgrimm (approved):

PR #2387, opened 2024-09-06
Added TFContact User to acelerator organization payload

tommylau523:


tommylau523:
* **#2387** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2387?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -360,12 +362,14 @@ data class OrganizationPayload(
     val timeZone: ZoneId?,
     @Schema(
         description = "The total number of users in the organization, including the current user.")
+    val tfContactUserId: UserId?,
```
sgrimm: This should go above the `@Schema` since the description there is for `totalUsers`.

sgrimm (changes requested):
I think we can add a permission check here without too much pain so that we're not exposing this user ID to users who don't have permission to see which users are in the organization.

Easiest (and reasonable, I think, in terms of product behavior) would be to make `fetchTerraformationContact` return null if the user lacks "can list organization users" permission, and then modify that permission check so it returns true for internal people if the organization is an accelerator-tagged one, which there's a helper function for. Something like,
```
  override fun canListOrganizationUsers(organizationId: OrganizationId) =
      isManagerOrHigher(organizationId) || isGlobalReader(organizationId)
```

src/test/kotlin/com/terraformation/backend/customer/db/OrganizationStoreTest.kt
```diff
@@ -597,6 +597,13 @@ internal class OrganizationStoreTest : DatabaseTest(), RunsAsUser {
         "Should find a Terraformation Contact")
   }
 
+  @Test
+  fun `fetchTerraformationContact throws exception if no permission ot list organization users`() {
```
sgrimm: Nit: "ot" instead of "to"

sgrimm (changes requested):
This needs to update `IndividualUser.canListOrganizationUsers` (see my previous comment on the PR) or it will return a permission error for accelerator admins who aren't admins of the org in question.

sgrimm:
Test failures seem legitimate here.

src/test/kotlin/com/terraformation/backend/customer/model/PermissionTest.kt
```diff
@@ -1621,6 +1621,8 @@ internal class PermissionTest : DatabaseTest() {
         createReport = true,
         readOrganization = true,
         readOrganizationDeliverables = true,
+        readOrganizationUser = true,
+        listOrganizationUsers = true,
```
sgrimm: Alphabetize these for consistency with the other expect functions.

sgrimm (approved):
PR title could use an update to reflect the fact that this moved to a different payload, but looks good otherwise with a couple nits (one from an earlier review, one new one).

tommylau523:
### Merge activity

* **Sep 9, 5:15 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2387) couldn't merge this PR because it failed for an unknown reason.

PR #2395, opened 2024-09-08
SW-5969 Allow CRUD of internal variables and values by global role permissions

tommylau523:
Fetch internal variables only by permission

Updated permission for allowing list users


sgrimm (commented):
A high-level thought about the approach here: our permission system usually exposes "you don't have permission to read object X" as "object X not found" and I think we should do the same here.

Modeling it that way would mean we could filter out internal-only variables in the database queries if the user doesn't have permission to see them. Filtering in the queries would remove the need for `safeFetchVariable()`.

tommylau523:
* **#2395** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -324,11 +326,30 @@ class VariableStore(
...
+    return variable
+  }
+
   /**
    * Logic for recursively fetching a variable and the variables it's related to. Variable fetching
    * is stateful because we want to detect cycles.
    */
-  private inner class FetchContext(private val manifestId: VariableManifestId?) {
+  private inner class FetchContext(
+      private val user: TerrawareUser,
```
sgrimm: It doesn't look like this is ever anything other than `currentUser()` -- does it need to be a new parameter, or could we call `currentUser()` in this class? What would it mean to use a different user's permissions here?

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -324,11 +326,30 @@ class VariableStore(
     variables.clear()
   }
 
+  /** Returns variable if found * */
```
sgrimm: Nit: extra asterisk at the end here.

sgrimm (approved):
Oops, commented on this earlier and forgot to submit the review. Looks generally good, though I had one question about whether a new parameter is really needed.

tommylau523:
### Merge activity

* **Sep 11, 3:02 PM EDT**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2395).

PR #2418, opened 2024-09-16
Add convenience methods for asserting table contents

sgrimm:
A common pattern in our tests is to call some application code and then verify
that the database was updated correctly. This usually looks something like

    assertEquals(
        listOf(
            SomeTableRow(col1 = "foo", col2 = "bar"),
        ),
        someTableDao.findAll().map { it.id = null })

This pattern is an occasional source of spurious test failures when we forget
that `findAll()` doesn't guarantee any particular order of results.

Introduce a new set of methods in `DatabaseBackedTest` that are more succinct to
call, offer more flexibility, and avoid the possibility of test failures due to
assumptions about result order.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
@@ -20,7 +20,6 @@ import com.terraformation.backend.gis.CountryDetector
 import com.terraformation.backend.i18n.Messages
 import com.terraformation.backend.mockUser
 import io.mockk.every
-import org.junit.jupiter.api.Assertions.*
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 
@@ -105,7 +104,7 @@ class DeliverableServiceTest : DatabaseTest(), RunsAsUser {
   @Test
   fun `updates existing submission to Complete status`() {
     insertSubmission(deliverableId = deliverable1, submissionStatus = SubmissionStatus.NotNeeded)
-    val existing = dslContext.fetchOne(SUBMISSIONS, SUBMISSIONS.ID.eq(inserted.submissionId))!!
+    val existing = dslContext.fetchSingle(SUBMISSIONS)
     service.setDeliverableCompletion(deliverable1, inserted.projectId, true)
     assertTableEquals(existing.apply { submissionStatusId = SubmissionStatus.Completed })
   }
@@ -113,11 +112,9 @@ class DeliverableServiceTest : DatabaseTest(), RunsAsUser {
   @Test
   fun `updates existing submission to NotSubmitted status if isComplete is false`() {
     insertSubmission(deliverableId = deliverable1, submissionStatus = SubmissionStatus.NotNeeded)
-    val existing = submissionsDao.fetchOneById(inserted.submissionId)!!
+    val existing = dslContext.fetchSingle(SUBMISSIONS)
     service.setDeliverableCompletion(deliverable1, inserted.projectId, false)
-    assertEquals(
-        existing.copy(submissionStatusId = SubmissionStatus.NotSubmitted),
-        submissionsDao.fetchOneById(inserted.submissionId))
+    assertTableEquals(existing.apply { submissionStatusId = SubmissionStatus.NotSubmitted })
   }
 
   @Test
--- a/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
@@ -13,8 +13,9 @@ import com.terraformation.backend.db.accelerator.ApplicationModuleStatus
 import com.terraformation.backend.db.accelerator.CohortPhase
 import com.terraformation.backend.db.accelerator.DeliverableId
 import com.terraformation.backend.db.accelerator.SubmissionStatus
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#2418** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3047,6 +3051,128 @@ abstract class DatabaseBackedTest {
...
+   * @param transform Function to apply to records from the database before comparing them to
+   *   [expected]. Can be used to clear or hardwire specific fields that aren't relevant to the
+   *   behavior being tested.
+   */
+  protected fun <R : UpdatableRecord<R>> assertTableEquals(
+      table: Table<R>,
+      expected: Set<R>,
+      message: String = table.name,
+      where: Condition? = null,
+      clearPrimaryKeys: Boolean = false,
```
sgrimm: Is this name too verbose? I also considered `nullIds` but that seemed less clear to me (not least because the lower-case "L" looks too similar to the upper-case "I" in some fonts.)

sgrimm (commented):
DatabaseBackedTest is a good place to start looking at this. I also converted a few tests to use the new methods so people can see what it looks like in action.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3047,6 +3051,128 @@ abstract class DatabaseBackedTest {
...
+   * @param where Optional query condition to assert on a subset of a table's contents. Only rows
+   *   matching the condition will be considered.
+   * @param clearPrimaryKeys If true, the primary key field(s) of the records from the database will
+   *   be cleared before comparing against [expected]. This can be used to ignore database-generated
+   *   IDs.
+   * @param transform Function to apply to records from the database before comparing them to
+   *   [expected]. Can be used to clear or hardwire specific fields that aren't relevant to the
+   *   behavior being tested.
+   */
+  protected fun <R : UpdatableRecord<R>> assertTableEquals(
```
sgrimm, line 3072: Callers should rarely need to use this variant of `assertTableEquals`. The variants below this require fewer arguments. It's possible I should make this `private` to drive that point home.

src/test/kotlin/com/terraformation/backend/accelerator/ParticipantProjectSpeciesServiceTest.kt
```diff
@@ -166,19 +167,18 @@ class ParticipantProjectSpeciesServiceTest : DatabaseTest(), RunsAsUser {
...
+              createdBy = userId,
+              createdTime = now,
+              deliverableId = deliverableId,
+              feedback = "So far so good",
+              id = submissionId,
+              modifiedBy = userId,
+              modifiedTime = now,
+              projectId = projectId,
+              submissionStatusId = SubmissionStatus.NotSubmitted),
+          where = SUBMISSIONS.DELIVERABLE_ID.eq(deliverableId))
```
sgrimm, line 181: This isn't strictly needed here since there's only one row in the table anyway, but it demonstrates the `where` argument and also maintains the behavior of the old version that filtered by deliverable ID.

src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
```diff
@@ -99,18 +99,15 @@ class DeliverableServiceTest : DatabaseTest(), RunsAsUser {
...
   @Test
   fun `updates existing submission to Complete status`() {
     insertSubmission(deliverableId = deliverable1, submissionStatus = SubmissionStatus.NotNeeded)
-    val existing = submissionsDao.fetchOneById(inserted.submissionId)!!
+    val existing = dslContext.fetchOne(SUBMISSIONS, SUBMISSIONS.ID.eq(inserted.submissionId))!!
     service.setDeliverableCompletion(deliverable1, inserted.projectId, true)
-    assertEquals(
-        existing.copy(submissionStatusId = SubmissionStatus.Completed),
-        submissionsDao.fetchOneById(inserted.submissionId))
+    assertTableEquals(existing.apply { submissionStatusId = SubmissionStatus.Completed })
```
sgrimm: Not totally happy with this pattern; it feels like there may be more room for improvement here.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
tommylau523: perhaps we can use `includePrimaryKeys = true`? I think it is clear enough either way!

tommylau523 (approved):

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
sgrimm: Good call. "Include primary keys" is a bit more intuitive.

src/test/kotlin/com/terraformation/backend/accelerator/DeliverableServiceTest.kt
sgrimm: Improved this a little bit by using the jOOQ `fetchSingle` method. Still don't love the `apply {}` part of this since it mutates the existing record but I don't see a better alternative.

PR #2419, opened 2024-09-17
SW-6029 Use application cohort phases for projects

sgrimm:
If a project isn't in a cohort but has an accelerator application, it is
considered to be in one of the two cohort phases related to applications, but the
API wasn't returning the phase as part of the project data.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
@@ -65,15 +65,16 @@ class ProjectStore(
         .fetchOne { (cohortId, cohortPhase, applicationStatus) ->
           if (cohortId != null && cohortPhase != null) {
             ProjectCohortData(cohortId, cohortPhase)
-          } else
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#2419** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -47,17 +50,32 @@ class ProjectStore(
   }
 
   fun fetchCohortData(projectId: ProjectId): ProjectCohortData? {
+    requirePermissions { readProject(projectId) }
+
     return dslContext
```
tommylau523, line 55: Nice query! This may be a valuable pattern for the ModuleStore/DeliverableStore as well so applications can use the same endpoints to fetch "non-cohort" modules.

PR #2421, opened 2024-09-17
SW-6034 "View Application" email links invalid

constanzauanini:


src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
```diff
@@ -103,7 +103,7 @@ class WebAppUrls(
   }
 
   fun acceleratorConsoleApplication(applicationId: ApplicationId): URI {
```
sgrimm: Can you rename this to `fullAcceleratorConsoleApplication`? We try to use the `full` prefix on methods that return full URLs rather than just paths.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/email/WebAppUrls.kt
constanzauanini: Sure! I will create a new method for this, because I think that for the notification it's ok to use just the path

sgrimm (approved):

PR #2422, opened 2024-09-17
SW-6038 Set country code and internal name during pre-screen submissions

tommylau523:
This fixes an issue where internal name and the country columns are not set for an application. This happens when a user uploads a shapefile boundary instead of drawing one. 

This implementation will read the country variable at pre-screen submission time, and set both columns.


tommylau523:
* **#2422** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2422?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (commented):
SW-5038 seems like an unrelated Jira task and there's no description on this PR.

Am I correct in thinking that this is fixing a problem with applications where the person uploaded a shapefile instead of drawing a map, and we thus can't automatically figure out the country, so we need to copy it from a variable?

And also, judging from the test, maybe fixing a problem with country codes disappearing on name collisions?

tommylau523:
Updated the ticket number and description! 

I am not quite aware of the problem with country codes disappearing. I wanted to simplify some computing of internal names since it was doing the same operations in a few areas.

sgrimm (approved):
Thanks!

PR #2426, opened 2024-09-18
SW-6034 Fix "View Application" console email link

constanzauanini:


sgrimm (changes requested):
Change seems fine, but needs a quick `./gradlew spotlessApply` run to fix the formatting (line length issue).

sgrimm (approved):

PR #2427, opened 2024-09-18
Fix UnusedSquareFinder edge-case failure

sgrimm:
`PlantingZoneModelTest` was occasionally failing because there was an edge case
where `UnusedSquareFinder` would fail to find a monitoring plot in one of the
test's randomly-generated sparse site maps. The edge case was a literal one: as
the code ran its search by subdividing the site into quadrants, if the only
available plot area fell across the edge of one of the quadrants in just the
right way, the code would fail to find it.

Fix the problem by making the quadrants overlap. This means we potentially end up
searching the overlapping areas twice, but ensures we'll find the full set of
possible results.

Mitigate the extra cost of searching the overlapping areas twice by only
subdividing a search area along the X or Y axis if its size along that axis is
larger than the grid interval; previously, if search areas were tall and skinny,
the code would keep dividing them along the X axis anyway.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/model/UnusedSquareFinder.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/model/UnusedSquareFinder.kt
@@ -135,14 +135,10 @@ class UnusedSquareFinder(
     val regionWidth = northeast.x - southwest.x
     val regionHeight = northeast.y - southwest.y
 
-    // If all possible points in the region will round to the same point on the grid, or if the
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#2427** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

PR #2433, opened 2024-09-19
Skip internal-only variables for unprivileged users

sgrimm:
If a user without permission to read internal-only variables requested the variable
list for a questionnaire deliverable that included internal-only variables, the request
would fail because the code was trying to sort them but couldn't access them.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -132,7 +132,7 @@ class VariableStore(
                   .where(replacementVariables.REPLACES_VARIABLE_ID.eq(ID)))
           .groupBy(STABLE_ID)
           .fetch()
-          .map { fetchVariable(it[DSL.max(ID)]!!) }
+          .mapNotNull { fetchVariable(it[DSL.max(ID)]!!) }
           .sortedBy { it.deliverablePosition }
     }
   }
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2433** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -132,7 +132,7 @@ class VariableStore(
                   .where(replacementVariables.REPLACES_VARIABLE_ID.eq(ID)))
           .groupBy(STABLE_ID)
           .fetch()
-          .map { fetchVariable(it[DSL.max(ID)]!!) }
+          .mapNotNull { fetchVariable(it[DSL.max(ID)]!!) }
           .sortedBy { it.deliverablePosition }
```
tommylau523, line 136: I am surprised that Kotlin didn't throw an error here previously, seeing that `it` would be a `Variable?` here.

sgrimm: This puzzled me too. Turns out it's because the `map()` being called here isn't the Kotlin one, but rather a method on the jOOQ `Result` class, and since it's a Java function without an explicit nullability annotation, the return value is `List<Variable!>` rather than `List<Variable?>` which disables the nullability check on the next line.

PR #2434, opened 2024-09-19
Fix caching of internal-only variables

sgrimm:
We were using a single cache for all variables, and the check for permission to
read internal-only variables was only happening on cache miss. As soon as a
privileged user fetched deliverable variables, the internal-only variables would
be in the cache and would start being returned to all users regardless of
privilege.

Since there can be internal-only variables nested inside public variables, the fix
isn't as simple as just filtering internal-only variables from the "list
variables" API responses for unprivileged users; the same variable could be
visible to both privileged and unprivileged users but with a different set of
child variables.

Fix the problem by using a separate cache for unprivileged users. It will only
contain public variables, both at the top level and nested in parent variables.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -71,14 +71,25 @@ class VariableStore(
     private val variableTextsDao: VariableTextsDao,
 ) {
   /**
-   * Cache of information about variables. This just holds variable definitions, not values! There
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2434** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523:
Instead of having two caches, would it be easier to add a condition to check the variable permission when retrieving from the one cache?

tommylau523 (approved):

PR #2436, opened 2024-09-20
SW-6052 Support getting specific variables

sgrimm:
Currently, the APIs for fetching variables and variable values operate in bulk:
they return all the variables for a given deliverable or document.

To support cross-deliverable variable dependencies, where the client needs to
reference the value of a variable that isn't part of the deliverable it's focused
on, add optional `variableId` query parameters to both the variable and variable
value GET endpoints to access variables regardless of which deliverable or
document they're in. Multiple variables may be fetched in one request by
specifying the `variableId` parameter more than once in the URL.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/api/ValuesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/api/ValuesController.kt
@@ -62,6 +62,12 @@ class ValuesController(
                   "saved document versions.")
       @RequestParam
       maxValueId: VariableValueId? = null,
+      @Parameter(
+          description =
+              "If specified, return the value of this variable. May be specified more than once " +
+                  "to return values for multiple variables.")
+      @RequestParam
+      variableId: List<VariableId>? = null,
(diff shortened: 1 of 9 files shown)


sgrimm:
* **#2436** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/api/ValuesController.kt
```diff
@@ -62,6 +62,12 @@ class ValuesController(
                   "saved document versions.")
       @RequestParam
       maxValueId: VariableValueId? = null,
+      @Parameter(
+          description =
+              "If specified, return the value of this variable. May be specified more than once " +
+                  "to return values for multiple variables.")
+      @RequestParam
+      variableId: List<VariableId>? = null,
```
tommylau523, line 70: should this be plural?

PR #2458, opened 2024-09-26
SW-6061 Allow accelerator admins to not trigger workflows after updating variable values

tommylau523:
Added a flag in our update values endpoint to bypass running update statuses. This is intended for use by accelerator admins while reviewing variable values.

1) We reused the existing `triggerWorkflow` flag that is added for importing and upgrading variables
2) Since we are exposing the flag, we added a permission for by passing the workflow. We reused `updateInternalWorkflowDetails`. (since it is choosing not to update here)


tommylau523:
* **#2458** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2458?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

constanzauanini (approved):
Not sure if it's necessary or not, but maybe adding a test where the new updateStatuses flag is false and after calling updateValues check that the status is the same?

tommylau523:
I am going to skip for now because our controller tests don't actually check for workflow history.

PR #2461, opened 2024-09-27
SW-6066 Added events and updater to handle file naming updates

tommylau523:
This feature keeps our project naming/application internal name in sync with the actual folder name in our Google Drive.

It depends on an already existing Google drive folder, and attempts to rename the folder according to the following order
1) Project naming if it is set
2) Application internal name (which should be always set)

This uses an event publisher/listener, because it removes a problem with stale data and race conditions. It can technically still happen if one process reads before another process updates it, but the update should fire a new event and will overwrite it basically.


tommylau523:
* **#2462** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2462?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2461** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2461?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

constanzauanini (approved):
Lgtm

PR #2482, opened 2024-10-07
SW-6099 Added feature to specify grid origin when importing planting sites

tommylau523:
Our plot size detection uses a grid system that is aligned to the bounding box of the planting site boundary. 

For a small bounding site, this can cause issues because the grid alignment may lead to a lack of solutions.

This temporary fix adds the ability for the user to specify a grid origin point (easiest way is to use a SW corner of a valid permanent plot). Screenshot belows show the lat-long to make this work for this site.

In longer term, a more complete solution is to think about ways to automatically compute a grid system that will not split up potential solution (likely some topography/convex hull problem so maybe this is not worth the effort at all).

But given that this is a very small edge case for very small planting sites. This should be okay for now.


tommylau523:
* **#2482** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

constanzauanini (approved):

PR #2485, opened 2024-10-10
SW-6134 Implemented project accelerator variables

tommylau523:
Added a project accelerator variables service as part of the plan to replace project accelerator details columns with variables.

Top-down design
1) Variable services will be accessed mostly by service/store classes to fetch variable values for building the data models. They should not be accessed by controllers directly
2) Users are still expected to use the data model abstractions via services/stores. The CRUD operations underneath can call these variable services methods. This will make sure we have data in one place, instead of doing a lot of work duplicating them into different places. 

As part of this changes we also 
1) Added a `variables` package under accelerator, to sort out dependencies. The `variables` package should really only depend on `documentproducer` stores so this should avoid circular dependencies
2) Refactored out stable IDs and common read/write variable functions into common classes

Next steps:
1) Migration step to write all of existing project accelerator details into these variables
2) Update existing `projectAccleratorDetails` to reference this new service class
3) After migration, delete the now unused columns
4) Do the same for `applications`, which only has the country variables but there are areas to add more references to variables.


tommylau523:
* **#2488** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2488?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2486** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2486?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2485** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2485?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/variables/VariableValuesHelper.kt
```diff
@@ -0,0 +1,125 @@
...
+import com.terraformation.backend.documentproducer.model.NewTextValue
+import com.terraformation.backend.documentproducer.model.NumberVariable
+import com.terraformation.backend.documentproducer.model.SelectVariable
+import com.terraformation.backend.documentproducer.model.TextVariable
+import com.terraformation.backend.documentproducer.model.UpdateValueOperation
+import com.terraformation.backend.documentproducer.model.ValueOperation
+import com.terraformation.backend.documentproducer.model.Variable
+import java.math.BigDecimal
+
+fun getNumberValue(values: Map<String, ExistingValue>, stableId: String): BigDecimal? {
```
sgrimm: Maybe we should encapsulate the `Map<String, ExistingValue>` in a "collection of values by stable ID" class and turn these functions into methods on that class. Otherwise it's not clear just from looking at the function signature that the `values` argument should use stable IDs as keys.

tommylau523: In the latest revision, this uses StableID enum as an input. But I agree with your comment, I can refactor this set of helpers into a class.

src/main/kotlin/com/terraformation/backend/accelerator/variables/StableId.kt
```diff
@@ -0,0 +1,50 @@
+package com.terraformation.backend.accelerator.variables
+
+import com.terraformation.backend.db.default_schema.LandUseModelType
+
+enum class StableId(val value: String) {
```
sgrimm: I'm a little torn on making this an enum as opposed to an `object` with a bunch of `val` properties. The enum gives us type safety, but it means that the `StableId` isn't the representation of a "stable ID" in the general sense, just of the specific ones we care about. What would this look like if `StableId` were a wrapper around an arbitrary string value, the way our other ID classes like `UserId` wrap an underlying value? And then these constants could either be declared in its companion object or in a separate `StableIds` object, e.g.,
```kotlin
object StableIds {
  val Country = StableId("1")
}
```
With this approach, we could use the `StableId` type everywhere we need to represent any stable ID, whether or not it's one of the ones we care about in applications.

sgrimm (changes requested):
Setting this to "changes requested" since it sounds like we're going to add a wrapper class for the stable-ID-mapped variable functions, and potentially also replace the enum though that can be discussed further.

src/main/kotlin/com/terraformation/backend/accelerator/variables/VariableValuesHelper.kt
tommylau523: In the latest revision, I left the helper function as is, but updated the variables to be named valuesByStableId, which should help with the clarity issues for now.

In the later PR, I want to take a bit of time to work through if we can simplified the "Read/Write typed variable values by stable ID" operations. It seems like we may be able to determine the type of values by the actual variable data type, instead of relying on us to pick the right helper method here.

src/main/kotlin/com/terraformation/backend/accelerator/ApplicationCountryUpdater.kt
```diff
@@ -22,7 +24,7 @@ class ApplicationCountryUpdater(
     systemUser.run {
       val variable = variableStore.fetchOneVariable(event.variableId)
 
-      if (variable.stableId == STABLE_ID_COUNTRY) {
+      if (StableId(variable.stableId) == StableIds.country) {
```
sgrimm, line 27: It will probably make sense to change `Variable` so that the `stableId` field is of type `StableId` rather than type `String`, since we're introducing a type wrapper for stable IDs, so we aren't having to wrap and unwrap the values regularly. But that can be its own separate PR.

src/main/kotlin/com/terraformation/backend/accelerator/variables/AcceleratorProjectVariableValuesService.kt
```diff
@@ -0,0 +1,340 @@
...
+import com.terraformation.backend.documentproducer.model.NumberVariable
+import com.terraformation.backend.documentproducer.model.SelectVariable
+import com.terraformation.backend.documentproducer.model.TextVariable
+import com.terraformation.backend.documentproducer.model.ValueOperation
+import com.terraformation.backend.documentproducer.model.Variable
+import com.terraformation.backend.log.perClassLogger
+import jakarta.inject.Named
+
+@Named
+class AcceleratorProjectVariableValuesService(
```
sgrimm, line 25: Minor naming thing: this class's name uses the term "accelerator project variable" whereas the model class's name uses "project accelerator variable." Any reason those shouldn't be consistent?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/variables/AcceleratorProjectVariableValuesService.kt
tommylau523: standardized all to "projectAccelerator"

tommylau523:
### Merge activity

* **Oct 15, 5:45 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2485).
* **Oct 15, 5:45 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2485).

PR #2494, opened 2024-10-15
Added problems details for importing modules/deliverables CSV

tommylau523:


tommylau523:
* **#2494** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -362,6 +370,17 @@ data class DeliverablePayload(
   )
 }
 
+data class ImportDeliverableProblemElement(
+    val row: Int,
+    val problem: String,
+) : SuccessResponsePayload
```
sgrimm: Why is this extending `SuccessResponsePayload`? (Ditto with the other similar element in the other controller.) Will anything read the `status` field in each element?

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -362,6 +370,17 @@ data class DeliverablePayload(
...
+
+data class ImportDeliverableResponsePayload(
+    val success: Boolean,
+    val problems: List<ImportDeliverableProblemElement> = emptyList(),
+    val message: String? = null,
+) : SuccessResponsePayload
```
sgrimm: Could make this extend `ResponsePayload` instead, and then use the `status` field to indicate success or error. As it stands now, this will give you a payload like
```
{
  "status": "ok",
  "success": false,
  ...
}
```
which, while it's not invalid JSON or anything, seems self-contradictory to me. Of course, there's always the argument that this means "the request was successfully processed but the action didn't succeed" but it's no harder to use `status` for this than to add a second field.

That said, if doing it the way I'm suggesting would cause headaches on the FE, ignore this suggestion.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/DeliverablesController.kt
```diff
@@ -362,6 +372,17 @@ data class DeliverablePayload(
   )
 }
 
+data class ImportDeliverableProblemElement(
+    val row: Int,
+    val problem: String,
+) : SuccessResponsePayload
```
sgrimm: I'm still unclear on why this needs to be a `SuccessResponsePayload`. It's never returned as a top-level response, so effectively this just adds a hardwired `"status": "ok"` to every element of the `problems` array.

tommylau523: Sorry I missed this one. Removing it.

sgrimm (approved):

PR #2498, opened 2024-10-16
SW-6070 Added export to GPX feature

tommylau523:


tommylau523:
* **#2498** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2498?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/api/Gpx.kt
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Constructs an HTTP response in GPX format.
+ *
+ * @param filename Filename that will be used by default if the user saves the GPX file. If null,
+ *   response will not include a filename.
+ * @param columnNames Contents of the first row of the CSV; this should be a list of the names of
+ *   the columns, typically in human-readable form.
+ * @param writeRows Callback function that writes the data rows to the CSV stream. This should call
+ *   [CSVWriter.writeNext] for each data row.
```
sgrimm: There don't appear to be `columnNames` or `writeRows` parameters.

src/main/kotlin/com/terraformation/backend/api/Gpx.kt
```diff
@@ -0,0 +1,78 @@
...
+    writer.writeStartDocument("utf-8", "1.0")
+    writer.writeStartElement("gpx")
+    waypoints.forEach { writeWaypoint(it) }
+
+    // TODO: Write track, or routes if applicable
+
+    writer.writeEndElement()
+    writer.writeEndDocument()
+
+    writer.flush()
```
sgrimm: Is the `flush()` needed? I'd think `close()` would auto-flush.

src/main/kotlin/com/terraformation/backend/tracking/model/AssignedPlotDetails.kt
```diff
@@ -15,4 +16,33 @@ data class AssignedPlotDetails(
...
     val plantingSubzoneName: String,
     val plotName: String,
-)
+) {
+  fun gpxWaypoints(): List<GpxWaypoint> {
+    return listOf(
+        GpxWaypoint(
+            boundary.coordinates[SOUTHWEST].y,
+            boundary.coordinates[SOUTHWEST].x,
+            "$plotName Southwest Corner"),
```
sgrimm: Is it acceptable to have waypoint names in English if the user is using the app in Spanish? (Maybe a product question; I'm not sure if the names are even user-visible at all.)

src/main/kotlin/com/terraformation/backend/tracking/model/AssignedPlotDetails.kt
```diff
@@ -15,4 +16,33 @@ data class AssignedPlotDetails(
...
+            "$plotName Southeast Corner"),
+    )
+  }
+
+  companion object {
+    // 0 Index vertices for monitoring plot corners.
+    const val SOUTHWEST = 0
+    const val NORTHWEST = 1
+    const val NORTHEAST = 2
+    const val SOUTHEAST = 3
```
sgrimm: The polygons for monitoring plots don't have their vertices in this order -- they're ordered counterclockwise rather than clockwise. So it should be SW, SE, NE, NW. Obviously it's fine to export them in a different order in the GPX file if we want them to be clockwise there.

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -187,6 +190,24 @@ class ObservationsController(
...
+              Content(
+                  mediaType = "application/gpx+xml",
+                  schema = Schema(type = "string", format = "binary"))])
+  @GetMapping("/{observationId}/plots", produces = ["application/gpx+xml"])
+  @Operation(summary = "Exports monitoring plots assigned to an observation as a GPX file.")
+  fun exportAssignedPlots(@PathVariable observationId: ObservationId): ResponseEntity<ByteArray> {
+    val observations = observationStore.fetchObservationPlotDetails(observationId)
+    val waypoints = observations.flatMap { it.gpxWaypoints() }
+
+    val filename = "observation.gpx"
```
sgrimm: Would it make sense to include the observation ID or date or something in the filename to avoid filename collisions if people do this for multiple observations over time?

src/main/kotlin/com/terraformation/backend/api/Gpx.kt
```diff
@@ -0,0 +1,78 @@
...
+data class GpxWaypoint(val latitude: Double, val longitude: Double, val name: String?)
+
+class GpxWriter(output: OutputStream, private val waypoints: List<GpxWaypoint>) {
+  private val writer: XMLStreamWriter =
+      XMLOutputFactory.newInstance().createXMLStreamWriter(output, StandardCharsets.UTF_8.name)
+
+  private fun writeWaypoint(waypoint: GpxWaypoint) {
+    writer.writeStartElement("wpt")
+    writer.writeAttribute("lat", "${waypoint.latitude}")
+    writer.writeAttribute("lon", "${waypoint.longitude}")
```
sgrimm: Does the GPX format allow all the string representations of floating-point numbers the JVM can produce? I'm thinking of stuff like a site near the equator ending up with a latitude value of `5.0E-9` using the default string rendering for `Double`s.

If not, we might need to use a `NumberFormat` to get precise control over number of decimal places and such.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/tracking/model/AssignedPlotDetails.kt
tommylau523: They will be user-visible in a GPS device. I think the shorthanded direction (SW, NW etc....) is likely a lot easier but let me double check with product.

src/main/kotlin/com/terraformation/backend/api/Gpx.kt
tommylau523: I opted to use BigDecimal set scale, because we don't actually want locale formatting.

src/main/kotlin/com/terraformation/backend/tracking/model/AssignedPlotDetails.kt
sgrimm: In that case they'll need to be localizable, I'm betting. The French word for "west" is "ouest," so the French abbreviation for "northwest" is "NO" rather than "NW."

tommylau523: Yes that is correct! Using localized messages now!

src/main/kotlin/com/terraformation/backend/api/Gpx.kt
```diff
@@ -0,0 +1,91 @@
...
+  private val MAX_DECIMAL_PLACE = 8 // precision to about 1mm for latitude longitude
+
+  private fun writeWaypoint(waypoint: GpxWaypoint) {
+    writer.writeStartElement("wpt")
+
+    val roundedLatitude = waypoint.latitude.setScale(MAX_DECIMAL_PLACE, RoundingMode.HALF_UP)
+    val roundedLongitude = waypoint.longitude.setScale(MAX_DECIMAL_PLACE, RoundingMode.HALF_UP)
+
+    writer.writeAttribute("lat", "$roundedLatitude")
+    writer.writeAttribute("lon", "$roundedLongitude")
```
sgrimm: This won't help with the scientific notation issue. Admittedly it's a corner case, but
```kotlin
println("${BigDecimal("0.0000005")}")
```
prints `5E-7`. You'd want `roundedLatitude.toPlainString()` instead of `"$roundedLatitude"`.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -116,6 +116,11 @@ manifestCsvTopLevelNameNotUnique=Top Level Variable Name must be unique
 manifestCsvVariableNameNotUniqueWithinParent=No two children of the same parent can have the same name
 manifestCsvVariableParentDoesNotExist=Parent for variable does not exist
 manifestCsvWrongDataTypeForChild=Only tables and sections may be listed as parents
+#{0} is plot name
+monitoringPlotsNortheastCorner={0} Northeast corner
+monitoringPlotsNorthwestCorner={0} Northwest corner
+monitoringPlotsSoutheastCorner={0} Southeast corner
+monitoringPlotsSouthwestCorner={0} Southwest corner
```
sgrimm: IMO we should either capitalize both the direction and the word "corner" or neither of them.

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -147,6 +147,18 @@ class Messages {
 
   fun batchCsvQuantityInvalid() = getMessage("batchCsvQuantityInvalid")
 
+  fun monitoringPlotNortheastCorner(plotName: String) =
+      getMessage("monitoringPlotsNortheastCorner", plotName)
```
sgrimm: The string key uses "plots" while the function name uses "plot" -- probably better to make them the same.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -116,6 +116,11 @@ manifestCsvTopLevelNameNotUnique=Top Level Variable Name must be unique
 manifestCsvVariableNameNotUniqueWithinParent=No two children of the same parent can have the same name
 manifestCsvVariableParentDoesNotExist=Parent for variable does not exist
 manifestCsvWrongDataTypeForChild=Only tables and sections may be listed as parents
+#{0} is plot name
```
sgrimm: FYI, this comment would only show up for `monitoringPlotsNortheastCorner` for translators, not for the other three strings. Phrase treats each comment in a properties file as the description of the string right after the comment.

sgrimm (changes requested):

src/main/resources/i18n/Messages_en.properties
tommylau523: This was done to follow naming of search columns. I am guessing we should update those to stay consistent too.

tommylau523: I guess actually, since it follows the plot name, I will keep them lowercase here.

sgrimm (approved):
Looks good!

PR #2546, opened 2024-10-29
When variables are upgraded, also replace injections/references to the old variable within section text for documents

nickgraz:
When a variable is upgraded (because its definition is changed within the all variables CSV and uploaded), there may still be old injections/references to this variable within the section text for documents. This causes a few issues, the most obvious being that the frontend receives multiple versions of a variable. The frontend receives the old version of the variable, because it exists within the value of a variable present (the section variable), and it receives the new variable because that is the current version of a variable that can be injected/referenced within the document.

This PR fixes the issue by adding an additional step to the "Upgrade Old Variable Values" which goes through and upgrades old injections/references to "old" variables.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
@@ -33,6 +33,7 @@ class VariableService(
           }
 
           variableValueStore.upgradeSectionDefaultValues(result.replacements)
+          variableValueStore.upgradeSectionValueVariables(result.replacements)
         }
       }
 
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
@@ -75,6 +75,7 @@ class VariableService(
               }
 
               variableValueStore.upgradeSectionDefaultValues(replacements)
+              variableValueStore.upgradeSectionValueVariables(replacements)
             }
       }
     }
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableUpgradeCalculator.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableUpgradeCalculator.kt
@@ -76,8 +76,8 @@ class VariableUpgradeCalculator(
   }
 
   private fun variableOperations(newVariable: Variable): List<ValueOperation> {
-    val oldVariableId = oldVariableIds[newVariable.id]
-    val valuesOfOldVariable = oldVariableId?.let { valuesOfOldVariables[it] } ?: return emptyList()
+    val valuesOfOldVariable =
+        oldVariableIds[newVariable.id]?.let { valuesOfOldVariables[it] } ?: return emptyList()
 
     val oldVariable = variableStore.fetchOneVariable(valuesOfOldVariable.first().variableId)
 
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableUpgradeCalculator.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableUpgradeCalculator.kt
@@ -76,8 +76,8 @@ class VariableUpgradeCalculator(
   }
 
   private fun variableOperations(newVariable: Variable): List<ValueOperation> {
-    val valuesOfOldVariable =
-        oldVariableIds[newVariable.id]?.let { valuesOfOldVariables[it] } ?: return emptyList()
+    val oldVariableId = oldVariableIds[newVariable.id]
+    val valuesOfOldVariable = oldVariableId?.let { valuesOfOldVariables[it] } ?: return emptyList()
 
     val oldVariable = variableStore.fetchOneVariable(valuesOfOldVariable.first().variableId)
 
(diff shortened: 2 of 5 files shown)


nickgraz:
* **#2546** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2546?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm:
Just pushed a commit with a few changes.

- Made the database query for references to a variable only return current results, which eliminates the need for all the logic to check whether the reference still exists. I didn’t add a test case to verify this, but it might be useful to have one.
- Used UpdateValueOperation, which can replace a single value in a list (ReplaceValueOperation replaces the entire list at once). This eliminates the need to fetch the entire list of values.
- Moved the test from VariableUpgradeCalculatorTest to VariableValueStoreTest and made it check for the end result rather than the list of update operations. This new upgrade logic doesn’t live in VariableUpgradeCalculator at all, so we’d never see these new operations in that class’s output.
- Removed all the references to rowValueId which is not relevant for section variables since they can’t be nested inside tables.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableUpgradeCalculator.kt
```diff
@@ -76,8 +76,8 @@ class VariableUpgradeCalculator(
   }
 
   private fun variableOperations(newVariable: Variable): List<ValueOperation> {
-    val valuesOfOldVariable =
-        oldVariableIds[newVariable.id]?.let { valuesOfOldVariables[it] } ?: return emptyList()
+    val oldVariableId = oldVariableIds[newVariable.id]
+    val valuesOfOldVariable = oldVariableId?.let { valuesOfOldVariables[it] } ?: return emptyList()
```
sgrimm: Guess this change isn't needed.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt
```diff
@@ -226,13 +225,21 @@ class VariableValueStore(
...
-  fun fetchSectionValueVariable(
-      variableId: VariableId
-  ): List<VariableValue<VariableValueId, SectionValueVariable>> {
-    val conditions = listOf(VARIABLE_SECTION_VALUES.USED_VARIABLE_ID.eq(variableId))
+  private fun fetchSectionValuesUsingVariable(variableId: VariableId): List<ExistingSectionValue> {
+    val variableValues2 = VARIABLE_VALUES.`as`("variable_values_2")
+    val conditions =
+        listOf(
+            VARIABLE_SECTION_VALUES.USED_VARIABLE_ID.eq(variableId),
+            // Don't return values that were later superseded by other values
```
nickgraz, line 233: 🙏

nickgraz:
Thanks for the help with this one @sgrimm.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt
sgrimm: Can you add a test case to verify that this works as intended and that we don't erroneously upgrade variables from previous versions of a section?

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStoreTest.kt
```diff
@@ -564,6 +570,106 @@ class VariableValueStoreTest : DatabaseTest(), RunsAsUser {
...
+      // A project value is added to the new variable
+      insertValue(textValue = "referenced text", variableId = newVariableId)
+
+      every { user.canUpdateInternalVariableWorkflowDetails(projectId) } returns true
+
+      store.upgradeSectionValueVariables(mapOf(oldVariableId to newVariableId))
+
+      // There should not be any values updated since the up-to-date version of the section value
+      // is already referencing the upgraded variable
+      verify(exactly = 1) { store.updateValues(emptyList(), false) }
```
sgrimm: This is testing an implementation detail; if we add code to skip the `updateValues` call if the list is empty, this test will break even though the end result of the upgrade method didn't change at all.

A better approach, I think, would be to fetch the value of the section variable before and after the upgrade call and assert that the two are equal, since the behavior we want in this scenario is that the value was left as-is.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStoreTest.kt
nickgraz: Makes sense!

src/main/kotlin/com/terraformation/backend/documentproducer/VariableService.kt
```diff
@@ -75,6 +75,7 @@ class VariableService(
               }
 
               variableValueStore.upgradeSectionDefaultValues(replacements)
+              variableValueStore.upgradeSectionValueVariables(replacements)
```
sgrimm, line 79: We should call this in `importAllVariables` too, I'm pretty sure. `upgradeAllVariables` is called when you hit the "Upgrade Old Variable Values" button in the admin UI, not when you upload a new CSV.

sgrimm (approved):
Looks good, but I noticed one more thing that should be fixed before this is merged.

PR #2567, opened 2024-11-04
Improve assertTableEquals usability

sgrimm:
Make a few improvements to the assertTableEquals functions:

* The "include primary keys" flag is now inferred from the list of expected
  records; callers no longer have to specify it explicitly.

* The expected records can now be any kind of collection, not just a Set.

* The lists of expected and actual records are now sorted before they're passed
  to assertEquals, rather than passed as sets. This should make it easier to
  spot the differences in assertion failure messages, especially when the test
  is run in an IDE that can diff the expected and actual values.


sgrimm:
* **#2567** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2567?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Nice improvement

PR #2574, opened 2024-11-06
SW-6223 Fix permission error on variable value update

sgrimm:
When a user edits the value of a deliverable variable, we reset the variable's
workflow status in the project to "In Review." Commit 0a442949 moved the logic
for that from an event handler to a method of `VariableValueService`, but didn't
include the `systemUser.run` from the event handler. Variable updates started
failing because non-admin, non-system users don't have permission to update
variable workflow statuses.

Fix it by changing the permission check to only require admin privileges for
statuses that are set via the accelerator console; regular users can set the
status to "In Review" as long as they have write access to the project.


sgrimm:
* **#2574** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2576, opened 2024-11-06
Fix issue with section variable upgrade due to multiple projects existing with values to upgraded variables

nickgraz:
Since there can be multiple projects with values for variables that can be upgraded, we need to separate the operations by project ID since there [is a check](https://github.com/terraware/terraware-server/blob/main/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt#L505-L507) to ensure that we only update a set of operations for one project at a time.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableValueStore.kt
@@ -478,7 +478,9 @@ class VariableValueStore(
           }
         }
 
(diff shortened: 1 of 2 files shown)


nickgraz:
* **#2576** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2576?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

sgrimm (approved):

PR #2577, opened 2024-11-06
Expand assertTableEquals table support; improve assertion

sgrimm:
Previously, assertTableEquals and assertTableEmpty only worked on tables with
primary keys that jOOQ could generate code for, which excluded tables with
compound primary keys where we didn't configure a synthetic primary key type
in `Config.kt`. This was because we needed to empty out the primary key fields
of the records from the query result.

Modify the code to only empty the primary key fields if the records actually
have them.

Also make assertion failure messages look nicer by using the jOOQ Result class
instead of generic Kotlin collections. Result has logic to pretty-print lists
of records in a compact, consistently-aligned style.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
@@ -360,6 +360,7 @@ import org.jooq.Configuration
 import org.jooq.DSLContext
 import org.jooq.JSONB
 import org.jooq.Table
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2577** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2577?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz:
Nice improvement

nickgraz (approved):

PR #2579, opened 2024-11-07
SW-6228 Allow TF Expert users to set TF contacts

sgrimm:
Previously, modifying the Terraformation contact for an organization required
super-admin privileges, but now we want to treat the contact as part of the
organization's accelerator profile which means it needs to be editable by the
same user types that can edit other profile data, namely TF Expert or higher.
Change the permission check accprdingly.

We had four permissions for modifying the TF contact, but the differences
among them were kind of fuzzy. Remove two of the permissions to make the
intent of the remaining two clearer.


sgrimm:
* **#2579** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2583, opened 2024-11-08
SW-6230 Allow fetching of application submissions

sgrimm:
When the web app saves changes to fields such as the internal comment on an
application deliverable, it tries to refresh the deliverable data using
`GET /api/v1/accelerator/deliverables/{id}/submissions/{projectId}`. Currently,
that endpoint only supports non-application deliverables (application deliverables
are exposed via an alternate endpoint) because the underlying query code needs
to avoid returning lists of application deliverables on the Deliverables page
in the app.

This doesn't appear to cause any user-visible problems, but it does cause an
error message to be logged by the server.

Since the client is requesting a specific deliverable in this case, there's no
danger of returning a list that mixes application and non-application
deliverables. Update the query to only exclude application deliverables if the
caller didn't supply a specific deliverable ID.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/DeliverableStore.kt
@@ -120,6 +120,18 @@ class DeliverableStore(
     val deliverableIdField = DELIVERABLES.ID.`as`("deliverable_id")
     val projectIdField = PROJECTS.ID.`as`("project_id")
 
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2583** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
Looks good to me

PR #2588, opened 2024-11-11
Support running tests as concrete users

sgrimm:
Introduce a new `RunsAsDatabaseUser` interface for tests, which can be used 
instead of `RunsAsUser` to indicate that the test should run as a concrete user
from the users table rather than as a user that's constructed using MockK.

Tests that implement this new interface and want to verify that permissions are 
checked correctly will no longer be able to call the MockK `every` function; 
instead they should adjust the current user's global and organization roles so 
that the user looks like one that should or shouldn't be able to perform the
action being tested.

To allow tests to change the current user's permissions, `DatabaseBackedTest`
now has `deleteUserGlobalRole` and `deleteOrganizationUser` functions, and the
existing `insertOrganizationUser` function will now update the user's role in
the organization if they're already a member. All these functions, as well as
`insertUserGlobalRole`, will clear out the current user's permissions cache so
that tests won't use stale data to do permission checks.

`ProjectStoreTest` and `ObservationServiceTest` are updated to use the new
interface as a demonstration.


sgrimm:
* **#2589** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2588** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2588?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):
This looks great, the user/permissions setup for the tests you modified looks a lot more like you'd expect.

PR #2589, opened 2024-11-11
Detect permission inversions

sgrimm:
We've had a number of bugs that are caused by permission inversions: situations
where we require a particular permission to perform a function, but then the
function ends up needing to do something that requires a higher level of
permissions, either directly or via an event handler. We sometimes fail to find
these bugs in testing if we happen to test with a user who passes both the
permission checks.

This change adds logic to track which permission checks are guarded by which other
checks. If it detects an inversion while handling a user request, it logs a
warning message. If it detects an inversion in a unit test, it fails the test.
In either case, the output includes abbreviated stack traces of the two checks
to make it easier to fix the problem.

Only tests that implement `RunsAsDatabaseUser` interface added in commit 137c89a2
will have their permission checks tracked and verified.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ParentStore.kt
@@ -182,6 +182,9 @@ class ParentStore(private val dslContext: DSLContext) {
       fetchFieldById(
           monitoringPlotId, MONITORING_PLOTS.ID, MONITORING_PLOTS.plantingSubzones.PLANTING_SITE_ID)
 
+  fun getPlantingSiteId(observationId: ObservationId): PlantingSiteId? =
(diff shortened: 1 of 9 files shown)


sgrimm:
* **#2589** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2589?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#2588** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

sgrimm:
Set the GitHub diff viewer to hide whitespace and the diff of `PermissionRequirements.kt` will get a lot easier to read/skim.

nickgraz (approved):
This looks great! The checks done on the trace seem pretty sturdy.

nickgraz (approved):
Nice work!

PR #2590, opened 2024-11-12
Add TF Contact info to organization list API

nickgraz:
The FE requires the TF Contact information for organizations, including those which do not include a project that is part of an accelerator participant. This PR pulls TF Contact, if it exists and is allowed for the current user, for each organization that is returned by the API in the same way as for the `AcceleratorOrganizationsController`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.customer.api
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.terraformation.backend.accelerator.api.TerraformationContactUserPayload
 import com.terraformation.backend.api.ApiResponse404
 import com.terraformation.backend.api.ApiResponse409
 import com.terraformation.backend.api.ApiResponseSimpleSuccess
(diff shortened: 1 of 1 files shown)


nickgraz:
* **#2590** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
```diff
@@ -66,8 +70,10 @@ class OrganizationsController(
   ): ListOrganizationsResponsePayload {
     val elements =
         organizationStore.fetchAll(depth).map { model ->
-          OrganizationPayload(model, getRole(model))
+          OrganizationPayload(
+              model, getRole(model), userStore.getTerraformationContactUser(model.id))
```
sgrimm: This will throw an exception if the user doesn't have permission to see the contact for an organization, which they won't if they have the Contributor role.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/customer/api/OrganizationsController.kt
nickgraz: Since this is only necessary to include for the API call that fires within the accelerator console side of Terraware, is it acceptable to swallow the error? Or should I do a more explicit "get tf contact user if user `canListOrganizationUsers`" check?

nickgraz: I went with the latter

sgrimm (approved):

PR #2591, opened 2024-11-12
SW-6244 Don't log error if GBIF species not found

sgrimm:
If the user searches for a species that's not in the GBIF database, we shouldn't
consider it a server error, just a normal outcome of the request.

Update the GBIF lookup endpoint to throw an explicit web-server "not found"
exception rather than an application-defined exception, and remove the error
logging for that kind of exception since when we throw it, it will always mean
the code is deliberately returning a response to the user.


sgrimm:
* **#2591** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2591?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

nickgraz (approved):

PR #2597, opened 2024-11-14
Check for permission inversions in controller tests

sgrimm:
Make the controller integration tests implement `RunsAsDatabaseUser` so they check
for permission inversions during request handling.

This requires a little bit of finagling because a fresh copy of the user is loaded
from the database as part of the request handling code. We need to grab that copy
of the user in order to see what permission checks it recorded.


sgrimm:
* **#2597** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2597?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2597?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

sgrimm:
Tested this by introducing an out-of-order permission check in one of the controllers and confirming that it caused that controller's tests to fail.

The flow of control is getting a little convoluted here and I want to simplify it, but cleaning it up would have made this change a lot bigger and messier; I'll save it for a followup PR.

src/test/kotlin/com/terraformation/backend/api/ControllerIntegrationTest.kt
```diff
@@ -71,9 +79,10 @@ abstract class ControllerIntegrationTest : DatabaseBackedTest(), RunsAsUser {
    * specified contents. The expected JSON string is generated by a function and may be constructed
    * dynamically (e.g., if it needs to include dynamically-allocated ID values).
    *
-   * @param strict If true, require the content of the response to exactly match [json]: keys that
-   *   are present in the response but not in [json] are treated as assertion failures. By default,
-   *   extra keys are ignored. Whitespace is ignored even if this is true.
+   * @param strict If true, require the content of the response to exactly match the return value of
+   *   [jsonFunc]: keys that are present in the response but not in the return value of [jsonFunc]
+   *   are treated as assertion failures. By default, extra keys are ignored. Whitespace is ignored
+   *   even if this is true.
```
sgrimm, line 85: This comment fix didn't seem worth splitting out into its own PR.

nickgraz (approved):
Seems reasonable, the finagling seems minimal

PR #2599, opened 2024-11-15
SW-6257 Allow internal users to update projects

sgrimm:
Commit d6d4e226 fixed a problem that caused end users to be unable to update
variables that would cause their workflow statuses to be reset to "In Review,"
but introduced a new bug: accelerator admins could no longer set variables to
that status.

Convert `VariableWorkflowStoreTest` to use `RunsAsDatabaseUser`, add a test
case to reproduce the problem, and fix the problem by updating the permission
logic for project updates to allow accelerator-related projects to be updated
by internal users with TF Expert or higher global roles.


nickgraz (approved):
Nice fix

sgrimm:
### Merge activity

* **Nov 15, 12:15 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2599).

PR #2612, opened 2024-11-19
SW-6162 Fix issue with disappearing variable value references / injections when a document is updated to a new variable manifest version

nickgraz:
When a document template has a new variable manifest uploaded for it, section variable IDs may be updated, for example if the section name goes from "Project Details" to "The Project Details". When this occurs, a user may "upgrade" their document (which points at the old version of the variable manifest) to the new version. 

During this upgrade, for project values within each section, "update" and "delete" operations were created to determine the upgrade path for the user-entered data. Now that the non-section variables exist outside of the manifest, and any out-dated references to them are updated during their own upgrade, there is no longer a need to create the "delete" operations which were responsible for cleaning up variable references that no longer existed. 

This solves a bug that was deleting references to variables within section text that belonged to sections that were part of an upgraded variable manifest.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
@@ -155,14 +155,9 @@ class DocumentUpgradeCalculator(
   private fun sectionOperations(variable: SectionVariable): List<ValueOperation> {
     return if (variable.id in existingValues) {
       // The section wasn't replaced, so existing text values are fine, as are existing
-      // references to variables that weren't replaced or removed; we just need to update or
-      // delete any references to old variables. We do this in reverse list position order
-      // since deleting will cause new value IDs to be generated for items with higher list
-      // positions (at which point our in-memory list of existing values wouldn't be usable).
-      val sectionValues =
-          existingValues[variable.id]!!
-              .filterIsInstance<ExistingSectionValue>()
-              .sortedByDescending { it.listPosition }
+      // references to variables that weren't replaced or removed; we just need to update
+      // any references to old variables.
+      val sectionValues = existingValues[variable.id]!!.filterIsInstance<ExistingSectionValue>()
 
       val updateOperations =
           sectionValues.mapNotNull { sectionValue ->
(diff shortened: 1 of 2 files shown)


nickgraz:
* **#2612** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculatorTest.kt
```diff
@@ -224,44 +225,64 @@ class DocumentUpgradeCalculatorTest : DatabaseTest(), RunsAsUser {
...
-        insertVariableManifestEntry(insertNumberVariable(), manifestId = oldManifestId)
-    val obsoleteVariableId =
-        insertVariableManifestEntry(insertTextVariable(), manifestId = oldManifestId)
-    val sectionVariableId = insertVariableManifestEntry(insertSectionVariable())
-    val newManifestId = insertVariableManifest()
-    val updatedVariableId =
-        insertVariableManifestEntry(
-            insertTextVariable(
-                insertVariable(type = VariableType.Text, replacesVariableId = outdatedVariableId)),
-            manifestId = newManifestId)
```
nickgraz, line 240: There are a few things in this test that were inconsistent with the way things work now, with separated document template manifests and variables. We no longer have non-section type variables directly attached to variable manifests.

This entire test was rewritten into a more representative flow of how things occur in the system, and one that represents the flow that produced the bug that caused this investigation.

src/main/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculator.kt
```diff
@@ -190,18 +185,7 @@ class DocumentUpgradeCalculator(
...
-                sectionValue.value.usedVariableId !in newManifestVariables &&
-                sectionValue.value.usedVariableId !in replacementVariableIds) {
-              DeleteValueOperation(projectId, sectionValue.id)
-            } else {
-              null
-            }
-          }
-
-      updateOperations + deleteOperations
+      updateOperations
```
sgrimm: I guess this variable can go away and the `mapNotNull` call can be the value of the `if` block.

src/test/kotlin/com/terraformation/backend/documentproducer/DocumentUpgradeCalculatorTest.kt
```diff
@@ -224,44 +225,64 @@ class DocumentUpgradeCalculatorTest : DatabaseTest(), RunsAsUser {
...
-    insertVariableManifestEntry(unmodifiedVariableId, manifestId = newManifestId)
-    insertVariableManifestEntry(sectionVariableId, manifestId = newManifestId)
+    // A few variables are uploaded to the system
```
sgrimm, line 228: Test name seems like it's no longer accurate.

sgrimm (approved):

PR #2615, opened 2024-11-19
SW-6179 Removed notifying for all categories if none is selected 

tommylau523:
The requirement has updated to be not notifying unless a category is selected. Terraware will now expect Global users to opt-into every category if they expect to receive every notificaiton

### DIFF
--- a/src/main/resources/db/migration/R__TypeCodes.sql
+++ b/src/main/resources/db/migration/R__TypeCodes.sql
@@ -98,7 +98,6 @@ VALUES (1, 'Compliance'),
        (6, 'Proposed Restoration Activities'),
        (7, 'Verra Non-Permanence Risk Tool (NPRT)'),
        (8, 'Supplemental Files'),
-       (100, 'None'),
        (101, 'Sourcing')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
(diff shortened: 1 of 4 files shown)


tommylau523:
* **#2615** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -98,6 +98,7 @@ VALUES (1, 'Compliance'),
        (6, 'Proposed Restoration Activities'),
        (7, 'Verra Non-Permanence Risk Tool (NPRT)'),
        (8, 'Supplemental Files'),
+       (100, 'None'),
```
tommylau523: IMO this seems very clumsy to me but is exactly the product requirement. 

I do think the better approach, is to remove the "notify user for all category, if none is selected" behavior. Assuming that the number of global users that want to see all notifications is small, we can expect them to opt into each one instead. But we need product clarity for that.

sgrimm: To guess at the original reasoning: if you forget to configure the categories for a user, getting all the notifications means it'll be obvious you forgot, whereas a lack of any notifications might take a lot longer to notice.

sgrimm (approved):
Agreed this seems clumsy and we should wait for product clarity, but if this is the way we want to go, the change here seems fine.

sgrimm (approved):
Great, this is much cleaner IMO.

tommylau523:
### Merge activity

* **Nov 20, 6:05 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2615).

PR #2617, opened 2024-11-19
Separated cohortModules and modules endpoints

tommylau523:


tommylau523:
* **#2617** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/CohortsController.kt
```diff
@@ -100,6 +106,34 @@ class CohortsController(
     return getCohort(cohortId)
   }
 
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{cohortId}/modules")
```
tommylau523, line 111: This removes some flexible querying for cohort modules, but FE always has cohort ID anyways so I think we prefer to keep API usage consistent instead.

src/main/kotlin/com/terraformation/backend/accelerator/api/ModulesController.kt
```diff
@@ -42,47 +34,28 @@ import org.springframework.web.multipart.MultipartFile
 @RequestMapping("/api/v1/accelerator/modules")
 @RestController
 class ModulesController(
```
tommylau523, line 36: Updated modules to return only modules data. Moved cohort modules information to under cohorts controller.

tommylau523:
There are type changes required in the FE to match these changes.

src/main/kotlin/com/terraformation/backend/accelerator/api/CohortsController.kt
```diff
@@ -100,6 +106,34 @@ class CohortsController(
...
+  @Operation(summary = "List modules.")
+  fun listModules(
```
sgrimm: A minor thing: rename this to something like `listCohortModules` and the other method to `getCohortModule`. By default, the OpenAPI schema generator uses the controller method name as the name of the operation, and operation names have to be unique in a schema. So it ends up appending suffixes like `_1` to the duplicate operation names to make them unique, which is fine initially, but then if we ever get rid of or rename the existing `listModules` method, it has the side effect of renaming this endpoint's operation (because the suffix is no longer needed).

If you look at the output of the ["Diff OpenAPI docs"](https://github.com/terraware/terraware-server/actions/runs/11923050776/job/33230594392#step:14:1) step of the CI build, you'll see what I'm talking about.

src/main/kotlin/com/terraformation/backend/accelerator/api/CohortsController.kt
```diff
@@ -100,6 +106,34 @@ class CohortsController(
     return getCohort(cohortId)
   }
 
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{cohortId}/modules")
+  @Operation(summary = "List modules.")
```
sgrimm: This is the same summary as the existing `listModules` endpoint; it'd be nice if the summaries said what they do differently.

sgrimm (approved):
A couple small things regarding the OpenAPI schema, but looks good otherwise.

PR #2623, opened 2024-11-21
SW-6282 Add table for ad hoc plot observations and related tables

nickgraz:
Adds `ad_hoc_plot_observations`, `ad_hoc_observation_types`, and `ad_hoc_plot_observation_photos` tables, used to store information and photos for ad hoc plot observations.


nickgraz:
* **#2623** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2623?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE tracking.ad_hoc_observation_types (
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+CREATE TABLE tracking.ad_hoc_plot_observations (
+    ad_hoc_observation_type_id INTEGER NOT NULL REFERENCES tracking.ad_hoc_observation_types,
+    boundary GEOMETRY(MultiPolygon) NOT NULL,
```
nickgraz, line 8: @sgrimm since this is an auto-sized plot, created by the SW corner, do you have a preference if that is calculated server side or client side? I believe @joshbuchea is planning on doing it client side but I want to make sure that is not going to conflict with the (possibly) standardized way we are doing this already.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE tracking.ad_hoc_observation_types (
```
nickgraz, line 1: Should these "observation types" only be scoped to the ad hoc plot observations? Or should they just exist at `tracking.observation_types`?

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,27 @@
...
+    ad_hoc_observation_type_id INTEGER NOT NULL REFERENCES tracking.ad_hoc_observation_types,
+    boundary GEOMETRY(MultiPolygon) NOT NULL,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    elevation_meters INTEGER,
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    name TEXT,
+    size_hectares NUMERIC NOT NULL
```
sgrimm, line 16: Maybe call this `area_ha` for consistency with the hectares columns in other tables.

Also, should we have a `size_meters` column like we do for regular monitoring plots?

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE TABLE tracking.ad_hoc_plot_observations (
+    ad_hoc_observation_type_id INTEGER NOT NULL REFERENCES tracking.ad_hoc_observation_types,
+    boundary GEOMETRY(MultiPolygon) NOT NULL,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    elevation_meters INTEGER,
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    name TEXT,
```
sgrimm, line 15: Should the name be non-nullable? There's always going to be one even if it's auto-generated by the client, right?

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
sgrimm: I think client-side is probably the way to go so that it's guaranteed to be identical to what the user saw in the mobile app when they created the plot.

sgrimm (commented):
The high-level approach here is different than what I'd been preparing for with my recent changes to the existing tracking tables, and I want to make sure we're aligned on where we're going with this. I am not saying my way is the "right" way!

What I'd been anticipating was that we'd essentially model an ad-hoc plot as a temporary monitoring plot with a flag to indicate that it was ad-hoc. It'd go in the `monitoring_plots` table where it would be linked to the planting site. Similarly, we'd model the observation using the existing `observations` table with a new column or two to indicate ad-hocness and observation type. We'd reuse the existing tables for things like photos and recorded plants.

Treating ad-hoc observations as their own completely separate little corner of the world does have some advantages. There's no chance we'll accidentally mix them up with assigned observations when we're showing history or calculating statistics, and there's no need to add logic to exclude them in situations like figuring out which plots to include in an assigned observation. When I was thinking about this initially, I considered that approach.

Two main things pushed me in the other direction. First is that for ad-hoc _monitoring_ observations, as far as I know, the actual data we'll be collecting is identical in format to what we're collecting for assigned observations, so reusing the existing code and tables seemed like a way to reduce the scope of the code change.

Second is that in the various product discussions, people have been saying things like, "_For now_, ad-hoc observations aren't included in any site statistics," which leads me to think we're likely to change that decision later. At that point, if the data is already sitting in the same tables as assigned observations, it'll be easy to combine them in our calculations.

I do have a couple minor comments on the actual PR here, but I think the high-level discussion is probably worth having first.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
nickgraz: I believe you're right.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
nickgraz: Good catch on both counts

nickgraz:
> The high-level approach here is different than what I'd been preparing for with my recent changes to the existing tracking tables, and I want to make sure we're aligned on where we're going with this. I am not saying my way is the "right" way!
> 
> What I'd been anticipating was that we'd essentially model an ad-hoc plot as a temporary monitoring plot with a flag to indicate that it was ad-hoc. It'd go in the `monitoring_plots` table where it would be linked to the planting site. Similarly, we'd model the observation using the existing `observations` table with a new column or two to indicate ad-hocness and observation type. We'd reuse the existing tables for things like photos and recorded plants.
> 
> Treating ad-hoc observations as their own completely separate little corner of the world does have some advantages. There's no chance we'll accidentally mix them up with assigned observations when we're showing history or calculating statistics, and there's no need to add logic to exclude them in situations like figuring out which plots to include in an assigned observation. When I was thinking about this initially, I considered that approach.
> 
> Two main things pushed me in the other direction. First is that for ad-hoc _monitoring_ observations, as far as I know, the actual data we'll be collecting is identical in format to what we're collecting for assigned observations, so reusing the existing code and tables seemed like a way to reduce the scope of the code change.
> 
> Second is that in the various product discussions, people have been saying things like, "_For now_, ad-hoc observations aren't included in any site statistics," which leads me to think we're likely to change that decision later. At that point, if the data is already sitting in the same tables as assigned observations, it'll be easy to combine them in our calculations.
> 
> I do have a couple minor comments on the actual PR here, but I think the high-level discussion is probably worth having first.

I see. I suppose I should have reached out to you earlier. After talking with Alex, it seemed like the consensus was that this _would_ be its own little corner of the world. But I agree with you that, considering how this will be fully utilized in the future, it does seem to make more sense to use the pre-existing plot and observation setup. I will submit another pull request with that in mind.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
joshbuchea: Does anyone know the correct method to calculate a plot boundary starting from the southwest corner? 

Here’s my current (likely incorrect) attempt:

```ts
export const getPlotBoundaryPolygon = (swCorner: Position) => {
  const southwest = point(swCorner);
  const northwest = destination(southwest, 30, 0, { units: 'meters' }); // 30m north
  const northeast = destination(northwest, 30, 90, { units: 'meters' }); // 30m east
  const southeast = destination(southwest, 30, 90, { units: 'meters' }); // 30m east

  const coordinates = [
    southwest.geometry.coordinates,
    northwest.geometry.coordinates,
    northeast.geometry.coordinates,
    southeast.geometry.coordinates,
    southwest.geometry.coordinates,
  ];

  return turfPolygon([coordinates]);
};
```

Also, without providing the NW corner explicitly, wouldn’t multiple boundaries be possible from the same SW corner depending on how other points are determined?

joshbuchea: Is that fine?

joshbuchea:
> The high-level approach here is different than what I'd been preparing for with my recent changes to the existing tracking tables, and I want to make sure we're aligned on where we're going with this. I am not saying my way is the "right" way!
> 
> What I'd been anticipating was that we'd essentially model an ad-hoc plot as a temporary monitoring plot with a flag to indicate that it was ad-hoc. It'd go in the `monitoring_plots` table where it would be linked to the planting site. Similarly, we'd model the observation using the existing `observations` table with a new column or two to indicate ad-hocness and observation type. We'd reuse the existing tables for things like photos and recorded plants.
> 
> Treating ad-hoc observations as their own completely separate little corner of the world does have some advantages. There's no chance we'll accidentally mix them up with assigned observations when we're showing history or calculating statistics, and there's no need to add logic to exclude them in situations like figuring out which plots to include in an assigned observation. When I was thinking about this initially, I considered that approach.
> 
> Two main things pushed me in the other direction. First is that for ad-hoc _monitoring_ observations, as far as I know, the actual data we'll be collecting is identical in format to what we're collecting for assigned observations, so reusing the existing code and tables seemed like a way to reduce the scope of the code change.
> 
> Second is that in the various product discussions, people have been saying things like, "_For now_, ad-hoc observations aren't included in any site statistics," which leads me to think we're likely to change that decision later. At that point, if the data is already sitting in the same tables as assigned observations, it'll be easy to combine them in our calculations.
> 
> I do have a couple minor comments on the actual PR here, but I think the high-level discussion is probably worth having first.

If ad-hoc observations are stored in the same table as assigned ones, will they be returned by the `/api/v1/tracking/observations` endpoint? 

If so, this could cause issues in current mobile app versions by mixing ad-hoc and assigned observations. If we wanted to use the same endpoint, we could make ad-hoc observation data opt-in using a query param to avoid the issue.

nickgraz:
> > The high-level approach here is different than what I'd been preparing for with my recent changes to the existing tracking tables, and I want to make sure we're aligned on where we're going with this. I am not saying my way is the "right" way!
> > What I'd been anticipating was that we'd essentially model an ad-hoc plot as a temporary monitoring plot with a flag to indicate that it was ad-hoc. It'd go in the `monitoring_plots` table where it would be linked to the planting site. Similarly, we'd model the observation using the existing `observations` table with a new column or two to indicate ad-hocness and observation type. We'd reuse the existing tables for things like photos and recorded plants.
> > Treating ad-hoc observations as their own completely separate little corner of the world does have some advantages. There's no chance we'll accidentally mix them up with assigned observations when we're showing history or calculating statistics, and there's no need to add logic to exclude them in situations like figuring out which plots to include in an assigned observation. When I was thinking about this initially, I considered that approach.
> > Two main things pushed me in the other direction. First is that for ad-hoc _monitoring_ observations, as far as I know, the actual data we'll be collecting is identical in format to what we're collecting for assigned observations, so reusing the existing code and tables seemed like a way to reduce the scope of the code change.
> > Second is that in the various product discussions, people have been saying things like, "_For now_, ad-hoc observations aren't included in any site statistics," which leads me to think we're likely to change that decision later. At that point, if the data is already sitting in the same tables as assigned observations, it'll be easy to combine them in our calculations.
> > I do have a couple minor comments on the actual PR here, but I think the high-level discussion is probably worth having first.
> 
> If ad-hoc observations are stored in the same table as assigned ones, will they be returned by the `/api/v1/tracking/observations` endpoint?
> 
> If so, this could cause issues in current mobile app versions by mixing ad-hoc and assigned observations. If we wanted to use the same endpoint, we could make ad-hoc observation data opt-in using a query param to avoid the issue.

I will make sure that the ad hoc ones are only supplied via a query param.

PR #2627, opened 2024-11-22
SW-6282 Add is_ad_hoc and observation_type to observations table

nickgraz:
In order to support ad hoc observations, add `is_ad_hoc` and `observation_type` to `observations` table.


nickgraz:
* **#2642** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2642?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2641** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2641?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2627** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

nickgraz:
@sgrimm as I've started working on the actual creation of the plot and observation, I am starting to think that I also need to modify the `observation_plots` table. If I understand correctly, a monitoring plot exists, outside the context of any observation, but as soon as the plot is observed, an observation plot is created. Am I on the right path in thinking that the creation of the ad hoc plot observation will create an entry in all 3 tables?

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,11 @@
...
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+ALTER TABLE tracking.observations ADD COLUMN is_ad_hoc BOOLEAN DEFAULT FALSE NOT NULL;
+ALTER TABLE tracking.observations ADD COLUMN observation_type_id INTEGER REFERENCES tracking.observation_types;
+ALTER TABLE tracking.observations ADD CONSTRAINT ad_hoc_observation_type CHECK (
+    (is_ad_hoc = FALSE AND observation_type_id IS NULL)
+    OR (is_ad_hoc = TRUE AND observation_type_id IS NOT NULL)
+);
```
sgrimm: What would you think about making observation type non-nullable and setting it to "Monitoring" for existing (and new non-ad-hoc) observations?

My sense is that observation type and ad-hocness are independent variables. We've already talked about possibly having assigned biomass observations, for example.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE tracking.observation_types (
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+ALTER TABLE tracking.observations ADD COLUMN is_ad_hoc BOOLEAN DEFAULT FALSE NOT NULL;
```
sgrimm: Opinions differ on this, but I prefer not using `DEFAULT` for this kind of thing because it eliminates a little bit of automated bug detection: if, somewhere in the server code, we accidentally forget to set the `is_ad_hoc` value when we're inserting a new row, the insertion will succeed and use the default, rather than failing because we didn't specify a value for a mandatory column.

That's why a lot of my migrations look like "add nullable column, populate existing rows, set it to non-nullable."

That said, `DEFAULT` is part of SQL for a reason. If we want "forgot to include the column when inserting" cases to make a best-effort attempt to insert the row rather than abort the whole transaction, this is how to do it.

sgrimm:
> Am I on the right path in thinking that the creation of the ad hoc plot observation will create an entry in all 3 tables?

Yes, I think that's right.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
nickgraz: Makes sense. I was doing this as a means to save on modification of the pre-existing observations code, but I see the benefit in explicitly migrating the field.

src/main/resources/db/migration/0300/V320__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE tracking.observations ADD COLUMN observation_type_id INTEGER REFERENCES tracking.observation_types;
+
+UPDATE tracking.observations
+SET observation_type_id = 1
+WHERE observation_type_id IS NULL;
+
+ALTER TABLE tracking.observations ALTER COLUMN observation_type_id SET NOT NULL;
+
+ALTER TABLE tracking.monitoring_plots ALTER COLUMN planting_subzone_id DROP NOT NULL;
+ALTER TABLE tracking.observation_plots ALTER COLUMN monitoring_plot_history_id DROP NOT NULL;
```
nickgraz: I had to add these and plumb their nullability through the pre-existing code. There isn't any mention of subzones in the PRD, and it seems like these plots need to be able to be created before any subzones exist. I also opted to allow the monitoring plot history to be optional, since creation of ad hoc observations will be a single event.

src/main/resources/db/migration/0300/V321__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE tracking.observation_types (
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+INSERT INTO tracking.observation_types (id, name)
+VALUES (1, 'Monitoring'),
+       (2, 'Biomass Measurements')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
```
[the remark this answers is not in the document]
sgrimm: We _want_ this to bomb out if there's a duplicate name.

src/main/resources/db/migration/0300/V321__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE tracking.observations ADD COLUMN observation_type_id INTEGER REFERENCES tracking.observation_types;
+
+UPDATE tracking.observations
+SET observation_type_id = 1
+WHERE observation_type_id IS NULL;
+
+ALTER TABLE tracking.observations ALTER COLUMN observation_type_id SET NOT NULL;
+
+ALTER TABLE tracking.monitoring_plots ALTER COLUMN planting_subzone_id DROP NOT NULL;
+ALTER TABLE tracking.observation_plots ALTER COLUMN monitoring_plot_history_id DROP NOT NULL;
```
sgrimm, line 28: Dropping the not-null from monitoring plot subzone IDs is definitely needed both for this and for the flexible site editing changes.

I'm nervous about making plot history IDs optional. I think it's likely to force us to write more complicated queries for things like viewing historical ad-hoc observations on a site map since we can no longer count on observation plots being tied to particular map versions via history IDs (meaning we could get the correct map geometry with simple joins on the history IDs) and we'll instead need to calculate the right map to show by, I guess, comparing the observation times against ranges of modification times on the site history.

Also, I think it's plausible we'll allow users to rename or otherwise edit ad-hoc plots after the fact since we're letting users name them initially, which means they would have more than one history row.

Was this just about avoiding storing data that didn't seem necessary, or will it cause problems to keep history for these plots?

src/main/resources/db/migration/0300/V321__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,28 @@
...
+
+ALTER TABLE tracking.observations ADD COLUMN observation_type_id INTEGER REFERENCES tracking.observation_types;
+
+UPDATE tracking.observations
+SET observation_type_id = 1
+WHERE observation_type_id IS NULL;
+
+ALTER TABLE tracking.observations ALTER COLUMN observation_type_id SET NOT NULL;
+
+ALTER TABLE tracking.monitoring_plots ALTER COLUMN planting_subzone_id DROP NOT NULL;
```
sgrimm, line 27: FYI, I included this (dropping the not-null constraint on monitoring plot subzone IDs) in #2660 since I'm going to need it for planting site editing.

src/main/resources/db/migration/0300/V321__AdHocPlotObservations.sql
```diff
@@ -0,0 +1,28 @@
...
+);
+
+INSERT INTO tracking.observation_types (id, name)
+VALUES (1, 'Monitoring'),
+       (2, 'Biomass Measurements')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
+
+ALTER TABLE tracking.observations ADD COLUMN is_ad_hoc BOOLEAN;
+
+UPDATE tracking.observations
```
tommylau523, line 13: Would it make more sense to put `is_ad_hoc` in the monitoring plots table? 

We can set up a constraint (via SQL or store functions) so that if observationType is set to Biomass Measurements, it needs to be a adHoc monitoring plot.

I am looking at the requirements that require adhoc plots to be on the map outside the extent of planting zones or site.

sgrimm: My guess is we'll end up having it in both places.

We definitely want it on monitoring plots so we can efficiently ignore existing ad-hoc plots when we're figuring out the plots to use for an assigned observation.

We may also want it on observations so we can easily filter the observations list based on ad-hocness.

As for plots being outside the site, I've already made the subzone ID nullable on monitoring plots and added a planting site ID to them. See #2660 .

sgrimm:
Can probably close this in favor of #2681

PR #2635, opened 2024-11-25
SW-6293 Fix seed withdrawals of weight-based accessions

sgrimm:
If a user withdraws a seed quantity from a weight-based accession, we use the
subset weight and count to calculate how much weight to subtract from the
accession. This can be off by a bit due to the limited precision of subset weight
measurements, so we had a fudge factor to try to detect if the user was trying to
withdraw all the remaining seeds and set the remaining weight to zero if so.

Unfortunately, the fudge factor wasn't enough for accessions measured in
kilograms or pounds where the weight of a single seed can round down to zero.

Replace the fudge factor with a more direct approach: see if the withdrawal
quantity is the same as the estimated seed count.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
@@ -4,6 +4,7 @@ import com.terraformation.backend.db.seedbank.AccessionState
 import com.terraformation.backend.db.seedbank.SeedQuantityUnits
 import com.terraformation.backend.db.seedbank.WithdrawalPurpose
 import com.terraformation.backend.seedbank.grams
+import com.terraformation.backend.seedbank.kilograms
 import com.terraformation.backend.seedbank.milligrams
 import com.terraformation.backend.seedbank.model.AccessionModel
 import com.terraformation.backend.seedbank.model.SeedQuantityModel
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#2635** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
```diff
@@ -104,29 +105,19 @@ internal class AccessionModelCalculationsTest : AccessionModelTest() {
...
-      val accession =
-          accession(remaining = grams(8), subsetCount = 1, subsetWeight = grams(3))
+          accession(
+                  remaining = kilograms(BigDecimal("0.005324")),
+                  subsetCount = 100,
+                  subsetWeight = kilograms(BigDecimal("0.000364")))
               .withCalculatedValues()
 
-      assertEquals(3, accession.estimatedSeedCount)
+      assertEquals(1463, accession.estimatedSeedCount)
```
nickgraz, line 118: Is there a reason these numbers are so specific? Would something like this be sufficient? 

```
val accession =
    accession(
            remaining = kilograms(BigDecimal("0.09")),
            subsetCount = 100,
            subsetWeight = kilograms(BigDecimal("0.03")))
        .withCalculatedValues()
assertEquals(300, accession.estimatedSeedCount)
```

nickgraz (approved):
Looks good to me.

src/test/kotlin/com/terraformation/backend/seedbank/model/accession/AccessionModelCalculationsTest.kt
sgrimm: Different numbers would definitely work. The numbers here are from the real accession in production that resulted in the bug report, so my thinking is that this test serves as a regression test to verify that the real accession is now fixed. Added a comment to that effect.

nickgraz: Got it, thanks

PR #2639, opened 2024-11-26
SW-6253 Implemented project overall scores table and endpoints

tommylau523:
This will replace project scores once we deprecate the V1 endpoints.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
@@ -44,11 +44,7 @@ class ProjectScoresControllerV2(
   @ApiResponse200
   @ApiResponse404
   @PutMapping
-  @Operation(
-      summary = "Upserts score selections for a single project.",
-      description =
-          "Update the scores for the project phase. If the (project, phase, category) does not " +
-              "exist, a new entry is created. Setting a `score` to `null` removes the score.")
+  @Operation(summary = "Updates overall score for a single project.")
   fun upsertProjectScores(
       @PathVariable("projectId") projectId: ProjectId,
       @RequestBody @Valid payload: UpdateProjectOverallScoreRequestPayload,
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
@@ -34,11 +34,7 @@ class ProjectScoresControllerV2(
       @PathVariable("projectId") projectId: ProjectId,
   ): GetProjectOverallScoreResponsePayload {
     val model = projectOverallScoreStore.fetch(projectId)
-    return GetProjectOverallScoreResponsePayload(
-        ProjectOverallScorePayload(model),
-        model.modifiedBy,
-        model.modifiedTime,
-    )
+    return GetProjectOverallScoreResponsePayload(ProjectOverallScorePayload(model))
   }
 
   @ApiResponse200
@@ -55,19 +51,11 @@ class ProjectScoresControllerV2(
   }
 }
 
-data class ProjectOverallScorePayload(
+data class UpdateProjectOverallScorePayload(
     val detailsUrl: URI?,
     val overallScore: Double?,
     val summary: String?,
 ) {
-  constructor(
-      model: ProjectOverallScoreModel
-  ) : this(
-      model.detailsUrl,
-      model.overallScore,
-      model.summary,
-  )
-
   fun toModel(projectId: ProjectId) =
       ProjectOverallScoreModel(
           detailsUrl,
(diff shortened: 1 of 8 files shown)


tommylau523:
* **#2639** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.time.Instant
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@AcceleratorEndpoint
+@RequestMapping("/api/v2/accelerator/projects/{projectId}/scores")
```
sgrimm, line 24: Nice! We haven't added a v2 endpoint since we overhauled the seed banking API.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectScoresControllerV2.kt
```diff
@@ -0,0 +1,88 @@
...
+
+data class UpdateProjectOverallScoreRequestPayload(
+    val score: ProjectOverallScorePayload,
+)
+
+data class GetProjectOverallScoreResponsePayload(
+    val score: ProjectOverallScorePayload,
+    val modifiedBy: UserId?,
+    val modifiedTime: Instant?,
+) : SuccessResponsePayload
```
sgrimm: Might make more sense to split `ProjectOverallScorePayload` into two versions, one for reading and one for writing, and put the `modifiedBy` and `modifiedTime` fields in the read-side version. That way those fields get included for each project if we eventually add an endpoint to list overall scores for multiple projects, and also the actual data in the response is all at the same level of the JSON tree.

src/main/resources/db/migration/0300/V320__ProjectOverallScores.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE accelerator.project_overall_scores (
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    overall_score FLOAT,
```
sgrimm, line 3: `FLOAT` is totally fine, but just noting that elsewhere in the data model we use `NUMERIC` which gets mapped to `BigDecimal` in the Kotlin code. The main reason is to eliminate the chance of floating-point errors where we do math and end up with 3.99999999998 instead of 4, that kind of thing. But if we're not doing math and we're just storing user-entered values without a lot of decimal places, `FLOAT` should work.

src/test/kotlin/com/terraformation/backend/accelerator/db/ProjectOverallScoreTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import java.net.URI
+import java.time.Instant
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.springframework.security.access.AccessDeniedException
+
+class ProjectOverallScoreTest : DatabaseTest(), RunsAsUser {
```
sgrimm: Can you make this use `RunsAsDatabaseUser` instead of `RunsAsUser`? I added that while you were on vacation, so the quick version is:

1. Use `override lateinit var user: TerrawareUser` instead of `override val user = mockUser()`
2. Instead of the `every {}` blocks to change permissions, use `insertOrganizationUser` and/or `insertUserGlobalRole`. There are also now `deleteOrganizationUser` and `deleteUserGlobalRole` helper methods if you want to remove permissions.

sgrimm (changes requested):
Looks good! I had a couple minor comments you can take or leave, and one requested change regarding the test.

src/main/resources/db/migration/0300/V320__ProjectOverallScores.sql
```diff
@@ -0,0 +1,11 @@
...
+    summary TEXT,
+    details_url TEXT,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    PRIMARY KEY (project_id)
```
sgrimm: If you want, you can put this on line 2, `project_id BIGINT PRIMARY KEY REFERENCES projects ON DELETE CASCADE`. This is fine too.

sgrimm (approved):

tommylau523:
### Merge activity

* **Nov 26, 3:49 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2639).

PR #2656, opened 2024-12-02
Stop using soon-to-be-deprecated Spring Security API, try 2

sgrimm:
Spring Security 6.4's Kotlin DSL deprecates the `authorizeRequests` block and
suggests using `authorizeHttpRequests` instead. Make that change so that the
upgrade will succeed rather than failing to compile with a deprecation warning.

The upgrade requires a workaround for a missing feature in the Kotlin DSL; when
they've addressed that, we can remove the workaround.

This is the same change as commit 6e662da4, but with a rule added to allow load
balancers to access the health check endpoint; with the deprecated API, the
health check endpoint was accessible by default.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
@@ -9,6 +9,7 @@ import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#2656** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

tommylau523 (approved):

PR #2665, opened 2024-12-05
Added deal name to deliverable list endpoint

tommylau523:

PR #2664, opened 2024-12-05
Added list participants API and added deal name to projects if visible

tommylau523:

PR #2666, opened 2024-12-05
Added project deal name to documents payload

tommylau523:
As part of the PR, we simplified setting up adding a series of variables for testing.


tommylau523:
* **#2666** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2666?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2665** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2665?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2664** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2664?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/variables/StableId.kt
```diff
@@ -1,3 +1,8 @@
 package com.terraformation.backend.accelerator.variables
 
-data class StableId(val value: String)
+import com.terraformation.backend.db.docprod.VariableType
+
+data class StableId(
+    val value: String,
+    val variableType: VariableType,
+)
```
sgrimm: I see what this change is for on the test side, but this change means that `StableId` is no longer a type-safe wrapper for an underlying string value which was the reason we added it.

The semantics change a bit: `StableId("123", VariableType.Text)` and `StableId("123", VariableType.Number)` are now both valid values but they aren't equal to each other despite having the same ID value. This could be more than a theoretical concern in contexts like importing the all-variables CSV, where the type of a variable can change and we compare stable IDs to match the old and new versions.

I think a better approach here, since `variableType` is never used anywhere in non-test code, would be to add a `Map<StableId, VariableType>` constant in the test code to keep the type information. Then the application code can continue to treat `StableId` as a simple type-safe ID value, which will mean this change needs to touch much less code.

sgrimm (changes requested):

sgrimm (approved):

PR #2680, opened 2024-12-10
SW-6285 Reduce number of permanent clusters

sgrimm:
Previously, we used a statistical formula to determine the number of permanent
clusters to assign in an observation, and then we created 1/4 as many temporary
plots as the number of permanent clusters.

We want to change that so that the statistical formula determines the _total_
number of plots, including both permanent and temporary ones, with a 3:1 ratio
between the number of permanent and temporary plots and a minimum of 1 temporary
plot. In effect, we want to subtract temporary plots from the total rather than
adding them to the total, except in cases where the math would leave us with no
temporary plots at all.

The only place we do this math on the fly is in the site editing form in the admin
UI. Update it to use the new calculations. For new planting zones, update the
hardwired default number of permanent clusters.

In addition, update the permanent cluster counts of existing planting zones to
reflect the new formula by reducing them by 25%. The number of temporary plots
stays the same since it was already based on the result of the statistical
formula.


sgrimm:
* **#2680** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2680?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2680?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

tommylau523 (approved):

PR #2692, opened 2024-12-13
SW-6351 Admin page for sending test email, localized email for documents update

nickgraz:
- Adds the "Terraformation documents update" email, localized with language-specific links (TOS link is the same for all 3 languages).

- Adds an admin page for sending test emails
To use this page locally, you will need to include some email related configuration:

[Here's](https://mailtrap.io/blog/gmail-smtp/) a page that helps you set up your Gmail account for using as an SMTP server. You will need to use the "app password" since the "less secure apps" option is no longer available.

SMTP config
```
spring:
  mail:
    host: smtp.gmail.com
    port: 587
    username: youremail@terraformation.com
    password: your app password
```

Email config
```
terraware:
  email:
    enabled: true
    alwaysSendToOverrideAddress: false
    senderAddress: youremail@terraformation.com
```


nickgraz:
* **#2692** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -0,0 +1,70 @@
...
+    return "/admin/email"
+  }
+
+  @PostMapping("/sendEmail")
+  fun sendTestEmail(
+      @NotBlank @RequestParam emailName: String,
+      @NotBlank @RequestParam recipient: String,
+      redirectAttributes: RedirectAttributes,
+  ): String {
+    val user = userStore.fetchByEmail(recipient) as IndividualUser
```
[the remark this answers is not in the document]
sgrimm: Not needed; nobody will be sending email to a device manager

src/main/resources/i18n/Messages_en.properties
```diff
@@ -173,6 +173,15 @@ notification.document.completedSectionVariableUpdated.email.body.2=You are recei
 notification.document.completedSectionVariableUpdated.email.buttonLabel=View Document
 notification.document.completedSectionVariableUpdated.email.subject=Variable edited in a "Completed" document section
 notification.document.completedSectionVariableUpdated.email.title=Variable edited in a "Completed" document section
+notification.documentsUpdate.email.body.1=Dear Terraware User,
+notification.documentsUpdate.email.body.2=We are writing to notify you that Terraformation recently updated its [End User License Agreement (EULA)], [Privacy Policy] and [Terms of Use]. These updates, which you can view via the links above, went into effect in the Fall of 2024. By continuing to use our software, you agree to our updated terms.
```
sgrimm: Nit: "The links above" seems like an odd word choice considering some of the links might be on the same line as that phrase depending on the user's window width. (I know this text was part of the spec.)

src/main/resources/templates/admin/email.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE HTML>
+<html xmlns:th="http://www.thymeleaf.org">
+<body>
+
+<span th:replace="~{/admin/header :: top}"/>
+
+<a href="/admin/">Home</a> -
```
sgrimm: Remove the hyphen at the end since this is the only element in the list of parent pages.

src/main/resources/templates/admin/email.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE HTML>
```
sgrimm, line 1: Is this page linked from anywhere?

src/main/resources/templates/admin/email.html
```diff
@@ -0,0 +1,27 @@
...
+    <input type="email" name="recipient" id="recipient" required />
+    <br/>
+    <label for="emailName">Email to Send</label>
+    <select name="emailName" id="emailName">
+        <option th:each="emailName : ${emailNames}"
+                th:text="${emailName}"
+                th:value="${emailName}">
+            Email Name
+        </option>
+    </select>
```
sgrimm: This might be simpler as straight HTML since the list of options from the controller is just a list of string literals (as opposed to something like an enum that would give us guaranteed consistency with the valid values on the POST endpoint). Using plain HTML also means it's easy to give the options more human-friendly labels.

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -0,0 +1,70 @@
...
+          user,
+          when (emailName) {
+            "DocumentsUpdate" -> DocumentsUpdate(config)
+            else -> throw IllegalArgumentException("Invalid test email name $emailName")
+          },
+          false,
+      )
+
+      redirectAttributes.successMessage =
+          if (config.email.enabled) {
```
sgrimm, line 58: Might also be nice to check this on the GET endpoint and include a message to warn people. Not that email will be disabled in staging or production, of course.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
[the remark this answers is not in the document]
nickgraz: I decided to just throw an error if a user isn't found with the email.

src/main/resources/templates/admin/email.html
nickgraz: It is now linked from the main admin page.

src/main/resources/templates/admin/email.html
nickgraz: Good callout, the longer option text is much better.

src/main/resources/i18n/Messages_en.properties
nickgraz: @alexTerraformation what do you think about changing this from "view via the links above" to "view via these links" or "view via the aforementioned links"?

alexTerraformation: I think either of those alternatives is fine.  @constanzauanini

sgrimm (approved):

PR #2688, opened 2024-12-13
API to fetch observations by ad-hoc flag

tommylau523:

PR #2697, opened 2024-12-13
Combined functionality to schedule and complete ad-hoc observation into one endpoint

tommylau523:


tommylau523:
* **#2688** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2697** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2697?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -408,38 +412,56 @@ class ObservationService(
   }
 
   /**
-   * Schedule an ad-hoc observation. This creates an ad-hoc observation, creates an ad-hoc
-   * monitoring plot, and adds the plot to the observation.
+   * Record an ad-hoc observation. This creates an ad-hoc observation, creates an ad-hoc monitoring
```
sgrimm: Nit: KDoc convention, which was inherited from Javadoc, is to use singular present tense. So this would be "Records" instead of "Record" (and "completes" instead of "complete" on the next line).

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -408,38 +412,56 @@ class ObservationService(
...
-  fun scheduleAdHocObservation(
-      endDate: LocalDate,
+  fun completeAdHocObservation(
+      conditions: Set<ObservableCondition>,
+      notes: String?,
+      observedTime: Instant,
       observationType: ObservationType,
       plantingSiteId: PlantingSiteId,
+      plants: Collection<RecordedPlantsRow>,
       plotName: String,
```
sgrimm, line 428: Just noting that plot names are going away shortly, so we shouldn't build new client code that expects to be able to pass in a user-supplied name. (But removing this parameter isn't appropriate yet since names still exist for now.)

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -408,38 +412,56 @@ class ObservationService(
...
-      startDate: LocalDate,
       swCorner: Point,
   ): Pair<ObservationId, MonitoringPlotId> {
     requirePermissions { scheduleAdHocObservation(plantingSiteId) }
 
-    validateSchedule(plantingSiteId, startDate, endDate)
+    if (observedTime.isAfter(clock.instant())) {
```
sgrimm: It's good to reject wildly wrong timestamps, but I'm concerned this check will be sensitive to clock skew and will reject legitimate observations where the user's phone clock is ahead of the server's clock by a small amount.

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -527,9 +549,9 @@ class ObservationService(
   }
 
   /**
-   * Validation rules:
-   * 1. start date can be up to one year from today and not earlier than today
-   * 2. end date should be after the start date but no more than 2 months from the start date
+   * Validation rules: 1a. for non-ad-hoc, start date can be up to one year from today and not
+   * earlier than today. 1b. for ad-hoc, start date can be on or before today.
+   * 2. end date should be after the start date but no more than 2 months from the start date.
```
sgrimm: I think the "1a" and "1b" are confusing the formatter because they're not valid Markdown list item numbers.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1374,13 +1374,17 @@ class ObservationStore(
...
+        updateSpeciesTotalsTable(
+            OBSERVED_ZONE_SPECIES_TOTALS.PLANTING_ZONE_ID,
+            observationId,
+            plantingZoneId,
+            isPermanent,
+            plantCountsBySpecies,
+        )
+      }
+
       updateSpeciesTotalsTable(
```
sgrimm: Verifying the intent here: we want to update the site-level species totals based on ad-hoc observations?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
tommylau523: I will add it a buffer margin!

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
tommylau523: This is mostly due to how observation result store works. Since it relies on a `plantingSiteSpeciesMultiset` to get a lot of the data. 

I was thinking that, since ad-hoc plots are always temporary, it will not have permanent data so it should not be so consequential. But perhaps it is easier to write new queries to deal with ad-hoc plots on the result side.

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -938,6 +940,21 @@ data class PlantingSiteObservationSummaryPayload(
...
 }
 
+data class CompleteAdHocObservationRequestPayload(
+    val conditions: Set<ObservableCondition>,
+    val notes: String?,
+    @Schema(description = "Date and time the observation was performed in the field.")
+    val observedTime: Instant,
+    @Schema(description = "Observation type for this observation.")
+    val observationType: ObservationType,
+    @Schema(description = "The plot name for the ad-hoc plot.") val plotName: String,
```
sgrimm: If you put a trailing `//` after the annotation, ktfmt will preserve the line break so the `val` lines up with the other ones. Right now, visually, at first glance it looks like this is an annotation on the `plants` property.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: Observations of non-permanent plots don't count toward mortality rate calculations but they do still count toward the plant totals. If someone does an ad-hoc observation outside the site boundary, for example, does it make sense to include those plants in the site totals? The PRD says ad-hoc observations shouldn't be included in the plants dashboard, which suggests to me the answer is no.

sgrimm (changes requested):
Requesting changes because I believe the product behavior here won't be correct per the PRD, but if the product team clarifies that what's here is what we want, then I'll switch to "Approve" since the code itself seems fine.

tommylau523:
Updated to not include totals

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -798,18 +800,25 @@ class ObservationStore(
...
 
   fun removePlotFromTotals(monitoringPlotId: MonitoringPlotId) {
-    val (plantingSiteId, plantingZoneId) =
+    val (plantingSiteId, plantingZoneId, isAdHoc) =
         dslContext
             .select(
-                PLANTING_ZONES.PLANTING_SITE_ID.asNonNullable(), PLANTING_ZONES.ID.asNonNullable())
+                PLANTING_ZONES.PLANTING_SITE_ID.asNonNullable(),
+                PLANTING_ZONES.ID.asNonNullable(),
+                MONITORING_PLOTS.IS_AD_HOC.asNonNullable(),
```
sgrimm, line 808: The `asNonNullable` on the zone and site IDs can go away since the values are actually nullable now.

sgrimm (approved):

tommylau523:
### Merge activity

* **Dec 17, 4:28 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2697).
* **Dec 17, 4:28 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2697).

PR #2704, opened 2024-12-17
SW-6368 Allow withdrawals from backdated batches

sgrimm:
If a user wants to record a historical nursery batch, they create it with an added
date in the past and a withdrawal date in the past. The system wasn't treating the
withdrawal as a quantity change in that case because it assumed that the seedling
quantities entered by the user when they created the batch were the quantities at
the time the batch was being created, and thus that older withdrawals shouldn't
affect the more recent quantity values supplied by the user.

Change batch creation such that if a batch's added date is in the past (based on
the nursery's time zone), we treat its quantities as having been observed at the
start of that day. Backdated withdrawals will thus be properly deducted from the
batch's quantities.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/BatchStore.kt
@@ -70,6 +70,7 @@ import com.terraformation.backend.seedbank.event.AccessionSpeciesChangedEvent
 import jakarta.inject.Named
 import java.time.Clock
 import java.time.LocalDate
+import java.time.ZonedDateTime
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#2704** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #2710, opened 2024-12-19
SW-6351 Send email from admin ui to all terraware users

constanzauanini:
Added a checkbox to send email to all users
Added sendAllUsersNotification function

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
@@ -40,7 +40,7 @@ class AdminEmailController(
   fun sendEmail(
       @NotBlank @RequestParam emailName: String,
       @RequestParam recipient: String?,
-      @RequestParam sendToAll: Boolean,
+      @RequestParam sendToAll: Boolean?,
       redirectAttributes: RedirectAttributes,
   ): String {
 
@@ -51,7 +51,7 @@ class AdminEmailController(
         }
 
     try {
-      if (sendToAll) {
+      if (sendToAll == true) {
         emailService.sendAllUsersNotification(
             emailNameResult,
             false,
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
@@ -38,31 +38,53 @@ class AdminEmailController(
 
   @PostMapping("/sendEmail")
   fun sendTestEmail(
-      @NotBlank @RequestParam emailName: String,
+      @RequestParam emailName: String,
       @NotBlank @RequestParam recipient: String,
+      @RequestParam sendToAll: Boolean,
       redirectAttributes: RedirectAttributes,
   ): String {
-    val user = userStore.fetchByEmail(recipient) ?: throw UserNotFoundForEmailException(recipient)
+    if (sendToAll) {
+      try {
+        emailService.sendAllUsersNotification(
+            when (emailName) {
+              "DocumentsUpdate" -> DocumentsUpdate(config)
+              else -> throw IllegalArgumentException("Invalid test email name $emailName")
+            },
+            false,
+        )
 
-    try {
-      emailService.sendUserNotification(
-          user,
-          when (emailName) {
-            "DocumentsUpdate" -> DocumentsUpdate(config)
-            else -> throw IllegalArgumentException("Invalid test email name $emailName")
-          },
-          false,
-      )
+        redirectAttributes.successMessage =
+            if (config.email.enabled) {
+              "Test email sent."
+            } else {
+              "Email sending is currently disabled."
+            }
+      } catch (e: Exception) {
+        log.error("Failed to send alert", e)
+        redirectAttributes.failureMessage = "Failed to send test email."
+      }
+    } else {
+      val user = userStore.fetchByEmail(recipient) ?: throw UserNotFoundForEmailException(recipient)
+      try {
+        emailService.sendUserNotification(
+            user,
+            when (emailName) {
+              "DocumentsUpdate" -> DocumentsUpdate(config)
+              else -> throw IllegalArgumentException("Invalid test email name $emailName")
+            },
+            false,
+        )
 
-      redirectAttributes.successMessage =
-          if (config.email.enabled) {
-            "Test email sent."
-          } else {
-            "Email sending is currently disabled."
-          }
-    } catch (e: Exception) {
-      log.error("Failed to send alert", e)
-      redirectAttributes.failureMessage = "Failed to send test email."
+        redirectAttributes.successMessage =
+            if (config.email.enabled) {
+              "Test email sent."
+            } else {
+              "Email sending is currently disabled."
+            }
+      } catch (e: Exception) {
+        log.error("Failed to send alert", e)
+        redirectAttributes.failureMessage = "Failed to send test email."
+      }
     }
 
     return "redirect:/admin/email"
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
@@ -37,54 +37,44 @@ class AdminEmailController(
   }
 
   @PostMapping("/sendEmail")
-  fun sendTestEmail(
-      @RequestParam emailName: String,
-      @NotBlank @RequestParam recipient: String,
+  fun sendEmail(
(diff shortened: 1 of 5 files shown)


src/main/resources/templates/admin/email.html
```diff
@@ -1,16 +1,29 @@
...
-
-<span th:replace="~{/admin/header :: top}"/>
+<head th:replace="~{/admin/header :: head}">
+    <script th:fragment="additionalScript">
+        const sendToAllCheckbox = document.getElementById('sendToAll');
+        const recipientInput = document.getElementById('recipient');
 
+        sendToAllCheckbox.addEventListener('change', function() {
+            if (sendToAllCheckbox.checked) {
+                recipientInput.removeAttribute('required');
```
sgrimm: Would it make sense to disable the input too?

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -38,31 +38,53 @@ class AdminEmailController(
 
   @PostMapping("/sendEmail")
   fun sendTestEmail(
-      @NotBlank @RequestParam emailName: String,
+      @RequestParam emailName: String,
       @NotBlank @RequestParam recipient: String,
```
sgrimm: Should `recipient` be `String?` without the `@NotBlank` annotation? That way if `sendToAll` is true, you're not required to enter a recipient address too.

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -38,31 +38,53 @@ class AdminEmailController(
 
   @PostMapping("/sendEmail")
   fun sendTestEmail(
```
sgrimm: Should probably rename this to `sendEmail` since it's not just for testing any more.

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -38,31 +38,53 @@ class AdminEmailController(
...
       redirectAttributes: RedirectAttributes,
   ): String {
-    val user = userStore.fetchByEmail(recipient) ?: throw UserNotFoundForEmailException(recipient)
+    if (sendToAll) {
+      try {
+        emailService.sendAllUsersNotification(
+            when (emailName) {
+              "DocumentsUpdate" -> DocumentsUpdate(config)
+              else -> throw IllegalArgumentException("Invalid test email name $emailName")
+            },
```
sgrimm: This `when` result can be a local variable that's set above the `if` statement so it can be used in both the send-all and send-one cases.

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -38,31 +38,53 @@ class AdminEmailController(
...
+        redirectAttributes.successMessage =
+            if (config.email.enabled) {
+              "Test email sent."
+            } else {
+              "Email sending is currently disabled."
+            }
+      } catch (e: Exception) {
+        log.error("Failed to send alert", e)
+        redirectAttributes.failureMessage = "Failed to send test email."
+      }
```
sgrimm: Moving the try/catch outside the `if` would reduce duplication since the exception handling is the same in both cases.

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -109,6 +109,29 @@ class EmailService(
...
+   * @param [requireOptIn] If false, send the notification even if the user has not opted into email
+   *   notifications. The default is to obey the user's notification preference, which is the
+   *   correct thing to do in the majority of cases.
+   */
+  fun sendAllUsersNotification(
+      model: EmailTemplateModel,
+      requireOptIn: Boolean = true,
+  ) {
+    val allUsers = userStore.fetchUsers(requireOptIn)
+    allUsers.forEach { usr ->
```
sgrimm: Why is this `usr` instead of `user`?

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
```diff
@@ -38,31 +38,53 @@ class AdminEmailController(
...
-      )
+        redirectAttributes.successMessage =
+            if (config.email.enabled) {
+              "Test email sent."
+            } else {
+              "Email sending is currently disabled."
+            }
+      } catch (e: Exception) {
+        log.error("Failed to send alert", e)
+        redirectAttributes.failureMessage = "Failed to send test email."
```
sgrimm: Might be good to include `${e.message}` in the failure message so the admin gets more detail about why it failed.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/admin/AdminEmailController.kt
constanzauanini: Oh, I removed the `@NotBlank` to the emailName instead of the recipient. Thanks

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
constanzauanini: Oh, I think I was using `user` for other variable at the beginning. Will update it

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -109,6 +109,30 @@ class EmailService(
...
+      requireOptIn: Boolean = true,
+  ) {
+    val allUsers = userStore.fetchUsers(requireOptIn)
+    allUsers.forEach { user ->
+      if (requireOptIn && !user.emailNotificationsEnabled) {
+        log.info(
+            "Skipping email notification for user ${user.userId} because they didn't enable it")
+      } else {
+        val locale = user.locale ?: Locale.ENGLISH
+        locale.use { send(model, listOf(user.email)) }
```
constanzauanini, line 131: @sgrimm locally, my first user email is "system" so this is failing with error "invalid email", and looks like it doesn't continue with the others users. How can I make that continue ?

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -216,6 +216,21 @@ class UserStore(
...
+   * Returns all users.
+   *
+   * @param requireOptIn Only return the users who are opted into email notifications.
+   */
+  fun fetchUsers(requireOptIn: Boolean = true): List<IndividualUser> {
+
+    return dslContext
+        .select(USERS.asterisk())
+        .from(USERS)
+        .apply { if (requireOptIn) where(USERS.EMAIL_NOTIFICATIONS_ENABLED.isTrue) }
```
sgrimm: Missed this the first time around, but this needs a couple more conditions:

* `.where(USERS.USER_TYPE_ID.eq(UserType.Individual))` to filter out system and device manager users
* `.and(USERS.DELETED_TIME.isNull)` to filter out deleted users

constanzauanini: Thanks, that was why I was getting the system user!

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -109,6 +109,34 @@ class EmailService(
...
+  fun sendAllUsersNotification(
+      model: EmailTemplateModel,
+      requireOptIn: Boolean = true,
+  ) {
+    val allUsers = userStore.fetchUsers(requireOptIn)
+    allUsers.forEach { user ->
+      try {
+        if (requireOptIn && !user.emailNotificationsEnabled) {
+          log.info(
+              "Skipping email notification for user ${user.userId} because they didn't enable it")
```
sgrimm: This would be a sign of a bug, right? `fetchUsers()` should already be filtering these users out.

constanzauanini: Yes, I will remove this condition, it's not necessary

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
sgrimm: This will end up putting two `where`s on the query which will be a syntax error. Probably easiest to change this to an `and` and put it below the other conditions.

sgrimm (changes requested):

sgrimm (approved):

PR #2738, opened 2025-01-06
SW-6409 Added biomass tables and database test methods

tommylau523:


tommylau523:
* **#2738** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
+CREATE TABLE tracking.biomass_forest_types (
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+-- This is also in R__TypeCodes.sql, but is repeated here to ensure migration is successful
```
sgrimm: These values aren't being referenced by the migration as far as I can see; are they really needed here?

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -265,6 +270,11 @@ VALUES (1, 'SeedBank'),
        (3, 'PlantingSite')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
+INSERT INTO tracking.mangrove_tides (id, name)
+VALUES (1, 'High'),
+       (2, 'Low')
```
sgrimm: Possibly silly nitpick, but would it be better if "low" had a lower ID than "high" rather than the other way around?

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+VALUES (1, 'High'),
+       (2, 'Low')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
+
+CREATE TABLE tracking.observation_biomass_details (
+    observation_id BIGINT NOT NULL REFERENCES tracking.observations ON DELETE CASCADE,
+    monitoring_plot_id BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    description TEXT,
+    forest_type_id INTEGER NOT NULL REFERENCES tracking.biomass_forest_types,
+    small_trees_count_low INTEGER NOT NULL CHECK ( 0 <= small_trees_count_low),
```
sgrimm: I think this check constraint would be slightly easier to read as `small_trees_count_low >= 0`.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+    small_trees_count_high INTEGER NOT NULL CHECK ( small_trees_count_low <= small_trees_count_high),
+    herbaceous_cover_percent FLOAT NOT NULL CHECK ( 0 <= herbaceous_cover_percent AND herbaceous_cover_percent <= 100),
```
sgrimm: Could use `herbaceous_cover_percent BETWEEN 0 AND 100`

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+                AND tide_id IS NOT NULL
+                AND tide_time IS NOT NULL)
+            ),
+
+    PRIMARY KEY (observation_id, monitoring_plot_id),
+    FOREIGN KEY (observation_id, monitoring_plot_id)
+        REFERENCES tracking.observation_plots (observation_id, monitoring_plot_id)
+);
+
+CREATE INDEX ON tracking.observation_biomass_details(observation_id);
```
sgrimm: The primary key index will also act as an index on `observation_id`, so we don't need a separate one with just that column.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+    description TEXT,
+
+    PRIMARY KEY (observation_id, monitoring_plot_id, position_id),
+    FOREIGN KEY (observation_id, monitoring_plot_id)
+        REFERENCES tracking.observation_biomass_details (observation_id, monitoring_plot_id)
+        ON DELETE CASCADE
+);
+
+CREATE UNIQUE INDEX ON tracking.observation_biomass_quadrant_details
+    (observation_id, monitoring_plot_id, position_id);
```
sgrimm: The primary key already gives us this unique index.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+
+CREATE INDEX ON tracking.observation_biomass_quadrant_details(observation_id);
```
sgrimm: The primary key index will cover this too since `observation_id` is its first column. (This is true for later tables too; I won't repeat this comment for each one.)

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+CREATE INDEX ON tracking.observation_biomass_quadrant_details(monitoring_plot_id);
+
+CREATE TABLE tracking.observation_biomass_quadrant_species (
```
sgrimm: Isn't it "quadrat" rather than "quadrant"?

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,212 @@
...
+    observation_id BIGINT NOT NULL REFERENCES tracking.observations ON DELETE CASCADE,
+    monitoring_plot_id BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    position_id INTEGER NOT NULL REFERENCES tracking.observation_plot_positions,
+    species_id BIGINT REFERENCES species,
+    species_name TEXT,
+    is_invasive BOOLEAN NOT NULL,
+    is_threatened BOOLEAN NOT NULL,
+    abundance_percent FLOAT NOT NULL CHECK ( 0 <= abundance_percent AND abundance_percent <= 100),
```
sgrimm: Is there a reason this is `FLOAT` instead of `NUMERIC`? We usually use `NUMERIC` because it doesn't introduce precision errors.

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -510,27 +514,16 @@ VALUES (1, 'Not Submitted'),
...
-                               description = excluded.description;
-
 INSERT INTO timeseries_types (id, name)
 VALUES (1, 'Numeric'),
        (2, 'Text')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
+INSERT INTO tracking.tree_growth_forms (id, name)
+VALUES (1, 'Trees'),
+       (2, 'Shrubs')
```
sgrimm: Should these be singular instead of plural?

sgrimm (commented):
Will look at this more later, but here are some initial comments on the SQL part.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,178 @@
...
+);
+
+CREATE TABLE tracking.mangrove_tides(
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+CREATE TABLE tracking.observation_biomass_details (
+    observation_id BIGINT NOT NULL REFERENCES tracking.observations ON DELETE CASCADE,
+    monitoring_plot_id BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
```
sgrimm, line 13: We'll want an index on the monitoring plot ID here and on the other tables.

Observation ID is covered by the primary key index here, but monitoring plot ID isn't the first column in that index so it needs to be indexed separately.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,178 @@
...
+    branch_number BIGINT NOT NULL CHECK ( branch_number >= 1 ),
+    diameter_at_breast_height_cm NUMERIC NOT NULL CHECK ( diameter_at_breast_height_cm >= 0 ),
+    point_of_measurement_m NUMERIC NOT NULL CHECK ( point_of_measurement_m >= 0 ),
+    is_dead BOOLEAN NOT NULL,
+    description TEXT,
+
+    UNIQUE (tree_id, branch_number)
+);
+
+CREATE INDEX ON tracking.recorded_branches(tree_id);
```
sgrimm: No need for this since the index for the unique constraint starts with this column.

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -488,10 +514,17 @@ COMMENT ON COLUMN tracking.plantings.species_id IS 'Which species was planted.';
 
 COMMENT ON TABLE tracking.recorded_plant_statuses IS '(Enum) Possible statuses of a plant recorded during observation of a monitoring plot.';
 
+COMMENT ON TABLE tracking.recorded_branches IS 'Recorded branches of a tree in a biomass observation.';
+COMMENT ON COLUMN tracking.recorded_branches.tree_id IS 'Foreign key ID of the tree the branch belongs to. The tree must have growth form "Tree".';
+COMMENT ON COLUMN tracking.recorded_branches.branch_number IS 'A system-assigned incremental number starting at 1for accounting branches.';
```
sgrimm: Nit: missing space between "1" and "for"

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -587,21 +622,9 @@ COMMENT ON TABLE accelerator.submission_statuses IS '(Enum) Statuses of submissi
...
-
-COMMENT ON TABLE docprod.document_statuses IS '(Enum) Current stage of a document''s lifecycle.';
-
-COMMENT ON TABLE docprod.document_templates IS 'Templates for the different types of documents this system can produce.';
-
-COMMENT ON TABLE docprod.documents IS 'Top-level information about documents.';
+COMMENT ON TABLE tracking.tree_growth_forms IS '(Enum) Growth form of each species in a biomass observation.';
 
-COMMENT ON FUNCTION docprod.reject_delete_value() IS 'Trigger function that rejects deletion of `variable_values` rows unless the entire document is being deleted.';
+COMMENT ON TABLE accelerator.user_internal_interests IS 'Which internal interest categories are assigned to which internal users. This affects things like which accelerator admins are notified.';
```
sgrimm, line 617: I'm not sure I understand some of the changes here. Are we getting rid of the separate sections for the accelerator and docprod schemas in this file? Can we maybe split that reorganization out into its own PR?

src/main/resources/i18n/Enums_en.properties
```diff
@@ -303,9 +303,17 @@ seedbank.WithdrawalPurpose.Nursery=Nursery
 seedbank.WithdrawalPurpose.Other=Other
 seedbank.WithdrawalPurpose.OutPlanting=Out-planting
 seedbank.WithdrawalPurpose.ViabilityTesting=Viability Testing
+tracking.BiomassForestType.Mangrove=Mangrove
+tracking.BiomassForestType.Terrestrial=Terrestrials
```
sgrimm: Should the string here be "Terrestrial"?

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -2322,6 +2367,130 @@ abstract class DatabaseBackedTest {
...
+  fun insertObservationBiomassAdditionalSpecies(
+      row: ObservationBiomassAdditionalSpeciesRow = ObservationBiomassAdditionalSpeciesRow(),
+      observationId: ObservationId = row.observationId ?: inserted.observationId,
+      monitoringPlotId: MonitoringPlotId = row.monitoringPlotId ?: inserted.monitoringPlotId,
+      speciesId: SpeciesId? = row.speciesId,
+      speciesName: String? = row.speciesName,
+      isInvasive: Boolean = row.isInvasive ?: false,
+      isThreatened: Boolean = row.isThreatened ?: false,
+  ) {
+    val rowWithDefaults =
```
sgrimm: Since this function isn't using the DAO to insert the row, I'm not sure we need this object.

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationBiomassDatabaseTest.kt
```diff
@@ -0,0 +1,1145 @@
...
+import com.terraformation.backend.db.tracking.tables.records.RecordedTreesRecord
+import java.math.BigDecimal
+import java.time.Instant
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.springframework.dao.DataIntegrityViolationException
+
+class ObservationBiomassDatabaseTest : DatabaseTest(), RunsAsDatabaseUser {
```
sgrimm: Am I correct in thinking that this test is essentially guarding against us messing up the integrity constraints on these tables at some point in the future?

If so, I'm not convinced it's worth writing low-level database tests like this. To be clear: this is more of a gut feeling than a hard-and-fast rule I can back up with solid empirical evidence. But it kind of boils down to cost/benefit. The cost is that we'll have to maintain this test and wait for it to build and run from now on. Not a very high cost, but a nonzero one. The chance of us dropping or altering a constraint by mistake is also nonzero, but it's not the kind of thing we've really had a problem with historically, and we'll also have code review and the Store class tests to catch some of these cases.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationBiomassDatabaseTest.kt
tommylau523: I am happy to remove them. For the most part I had these to help me validate constraints I put in. I will remove them.

src/main/resources/db/migration/R__Comments.sql
tommylau523: Reverted these for now. I think I am confused about the section organization. Since we used both alphabetically organizing by entity names, as well as separating by sections.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,187 @@
...
+
+    FOREIGN KEY (observation_id, monitoring_plot_id)
+        REFERENCES tracking.observation_biomass_details (observation_id, monitoring_plot_id)
+        ON DELETE CASCADE,
+
+    UNIQUE (observation_id, tree_number)
+);
+
+CREATE INDEX ON tracking.recorded_trees(observation_id);
+CREATE INDEX ON tracking.recorded_trees(monitoring_plot_id);
```
sgrimm: Don't need the index on observation ID since it's covered by the unique index on `(observation_id, tree_number)`.

sgrimm (approved):
One minor index tweak, but this looks good otherwise.

src/main/resources/db/migration/0300/V333__ObservationPlotBiomassDetails.sql
```diff
@@ -0,0 +1,186 @@
...
+    name TEXT NOT NULL UNIQUE
+);
+
+CREATE TABLE tracking.observation_biomass_details (
+    observation_id BIGINT NOT NULL REFERENCES tracking.observations ON DELETE CASCADE,
+    monitoring_plot_id BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    description TEXT,
+    forest_type_id INTEGER NOT NULL REFERENCES tracking.biomass_forest_types,
+    small_trees_count_low INTEGER NOT NULL CHECK ( small_trees_count_low >= 0),
+    small_trees_count_high INTEGER NOT NULL CHECK ( small_trees_count_low <= small_trees_count_high),
```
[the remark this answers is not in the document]
tommylau523: Not circular.

tommylau523:
### Merge activity

* **Jan 13, 2:55 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2738).

PR #2744, opened 2025-01-13
SW-6416 Read zone settings from shapefiles

sgrimm:
We want to allow setting the inputs to the plot count calculations at the same
time the site map is updated so that any adjustments to plots will take both
the map changes and the plot count changes into account.

Update the shapefile importer to require the error margin and variance values
to be supplied as planting subzone properties. The Student's t value may also
be overridden in a property but isn't required. As before, the shapefile can
also directly set the permanent and temporary plot counts, but the calculation
inputs are still required even if the counts are overridden.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteImporter.kt
@@ -31,6 +31,11 @@ class PlantingSiteImporter(
     val targetPlantingDensityProperties = setOf("plan_dens", "density")
     val zoneNameProperties = setOf("planting_z", "zone")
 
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#2752** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2752?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2744** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #2745, opened 2025-01-14
SW-6419 Implemented API to record Biomass Observation

tommylau523:


tommylau523:
* **#2751** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2745** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2745?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2745?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -437,6 +439,14 @@ class ObservationService(
       throw IllegalArgumentException("Observed time is in the future")
     }
 
+    if (observationType == ObservationType.BiomassMeasurements && plants.isNotEmpty()) {
+      throw IllegalArgumentException("Biomass observations must not contain plants")
+    }
+
+    if (observationType == ObservationType.BiomassMeasurements && biomassDetails == null) {
+      throw IllegalArgumentException("Biomass observations must contain biomass details")
+    }
```
sgrimm: Might make sense to validate that `biomassDetails` is null for monitoring observations. Maybe something more like
```kotlin
when (observationType) {
  ObservationType.BiomassMeasurements -> {
    if (plants.isNotEmpty()) { throw ... }
    if (biomassDetails == null) { throw ... }
  }
  ObservationType.Monitoring -> {
    if (biomassDetails != null) { throw ... }
  }
}
```

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -636,6 +646,101 @@ data class AssignedPlotPayload(
...
+  fun toModel(): BiomassAdditionalSpeciesModel {
+    return BiomassAdditionalSpeciesModel(isInvasive, isThreatened, speciesId, speciesName)
+  }
+}
+
+data class BiomassMeasurementPayload(
+    val additionalSpecies: List<BiomassAdditionalSpeciesPayload>,
+    val description: String?,
+    val forestType: BiomassForestType,
+    @Schema(minimum = "0", maximum = "100") val herbaceousCoverPercent: BigDecimal,
```
sgrimm: Could put a trailing `//` after the annotation to make all the `val` keywords line up. Same with the other annotations in this file that aren't long enough to make ktfmt put the `val` on a new line.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -809,11 +821,134 @@ class ObservationStore(
...
+                it.isDead,
+                it.isTrunk,
+                it.diameterAtBreastHeightCm,
+                it.pointOfMeasurementM,
+                it.heightM,
+                it.shrubDiameterCm,
+                it.description,
+            )
+          }
+      recordedTreesDao.insert(recordedTreesRows)
```
sgrimm: Would `dslContext.batchInsert()` of `RecordedTreesRecord` objects work here too, for consistency with the earlier code?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -809,11 +821,134 @@ class ObservationStore(
...
+            details.species.map {
+              ObservationBiomassQuadratSpeciesRecord(
+                  observationId,
+                  plotId,
+                  position,
+                  it.speciesId,
+                  it.speciesName,
+                  it.isInvasive,
+                  it.isThreatened,
+                  it.abundancePercent,
```
sgrimm: Named parameters would add a little extra safety here since there are, e.g., two boolean arguments in a row and we might not notice if they got swapped by accident. Same with the other `Record` class constructor calls in the function.

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
```diff
@@ -0,0 +1,108 @@
...
+typealias NewRecordedBranchModel = RecordedBranchModel<Nothing?>
+
+data class RecordedTreeModel<TreeId : RecordedTreeId?, BranchId : RecordedBranchId?>(
+    val id: TreeId,
+    val branches: List<RecordedBranchModel<BranchId>> = emptyList(),
+    val description: String? = null,
+    val diameterAtBreastHeightCm: BigDecimal? = null,
+    val heightM: BigDecimal? = null,
+    val isDead: Boolean,
+    val isTrunk: Boolean? = null,
```
sgrimm, line 55: What does a null value mean here? Don't we always know if it's a trunk or not?

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
```diff
@@ -0,0 +1,108 @@
...
+    val smallTreeCountRange: Pair<Int, Int>,
+    val soilAssessment: String,
+    val plotId: PlotId,
+    val tide: MangroveTide? = null,
+    val tideTime: Instant? = null,
+    val trees: List<RecordedTreeModel<TreeId, BranchId>>,
+    val waterDepthCm: BigDecimal? = null,
+) {
+  fun validate() {
+    // TODO convert database validation errors to human-readable error messages
```
sgrimm: There's an argument to be made for a single source of truth for validation rules, but I wonder if trying to reverse-engineer database constraint failures would end up being more complex and/or brittle than doing validation in the application code (aside from foreign key constraints) where it's trivial to generate precise error messages, and treating the database rules as protection against bugs in application-level validation.

src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
```diff
@@ -2046,6 +2052,121 @@ class ObservationServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
+              null,
+              emptyMap(),
+              null,
+              0 to 0,
+              "Basic soil assessment",
+              null,
+              null,
+              null,
+              emptyList(),
+              null)
```
sgrimm: Named parameters would make this more readable.

src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
```diff
@@ -2046,6 +2052,121 @@ class ObservationServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
+              emptySet(),
+              "Notes",
+              observedTime,
+              ObservationType.BiomassMeasurements,
+              plantingSiteId,
+              emptySet(),
+              point(1),
+          )
+
+      assertEquals(
```
sgrimm, line 2082: Does `assertTableEquals` not work for this assertion?

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationStoreTest.kt
```diff
@@ -2058,6 +2080,72 @@ class ObservationStoreTest : DatabaseTest(), RunsAsUser {
...
+                  ObservationPlotId(inserted.observationId, inserted.monitoringPlotId))
+              .single()
+              .copy(
+                  notes = "Notes",
+                  observedTime = observedTime,
+                  statusId = ObservationPlotStatus.Completed,
+              )
+
+      assertTableEquals(ObservationPlotsRecord(newRow), "Updated observation plot entry")
+      assertTableEquals(expectedConditions, "Inserted observation plot conditions ")
```
sgrimm: Nit: Trailing space on the assertion message.

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationStoreTest.kt
```diff
@@ -2058,6 +2080,72 @@ class ObservationStoreTest : DatabaseTest(), RunsAsUser {
...
+      assertTableEmpty(RECORDED_PLANTS, "No plants recorded")
+
+      assertTableEmpty(OBSERVED_PLOT_SPECIES_TOTALS, "Observed plot species should be empty")
+      assertTableEmpty(OBSERVED_SUBZONE_SPECIES_TOTALS, "Observed subzone species should be empty")
+      assertTableEmpty(OBSERVED_ZONE_SPECIES_TOTALS, "Observed zone species should be empty")
+      assertTableEmpty(OBSERVED_SITE_SPECIES_TOTALS, "Observed site species should be empty")
+
+      assertEquals(
```
sgrimm: Can this and the assertions below it be `assertTableEquals`?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
tommylau523: The requirement specifies that `isTrunk` is only applicable for `Tree` growth form. I supposed we can default it to false in that case. But for now, in the database, this is a nullable type. 

We can remedy this in a future PR once I get clarity from product. There are a few type changes + numeric precisions I have to implement regardless in the followup.

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
tommylau523: Sorry I did mean that we will do validations according to product requirements. That being said some lower/upper bounds are being finalized so I will implement this in a follow up.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
tommylau523: I updated to use records, but records do not populate with ID by default. I have added a select statement to get the new IDs.

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
sgrimm: Ah, okay, forgot that this wasn't just for trees. Makes sense to keep this nullable if it doesn't apply at all to some recorded "trees."

sgrimm (approved):

PR #2750, opened 2025-01-14
Fixed planting site summaries null-pointer exception

tommylau523:
I was not successful in reproducing this issue, but I suspect that a completed observation can somehow contain a subzone without a completed time. This does not solve the root cause but can alleviate the symptoms for now. 

Logs of error:
```
java.lang.NullPointerException: null
	at com.terraformation.backend.tracking.db.ObservationResultsStore.plantingSiteSummary(ObservationResultsStore.kt:148)
	at com.terraformation.backend.tracking.db.ObservationResultsStore.fetchSummariesForPlantingSite(ObservationResultsStore.kt:119)
	at com.terraformation.backend.tracking.db.ObservationResultsStore.fetchSummariesForPlantingSite$default(ObservationResultsStore.kt:103)
	at com.terraformation.backend.tracking.api.ObservationsController.listObservationSummaries(ObservationsController.kt:189)
	at com.terraformation.backend.tracking.api.ObservationsController.listObservationSummaries$default(ObservationsController.kt:178)
```

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
@@ -111,7 +111,7 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
         }
 
     val numObservations = completedObservations.size
-    val depth = max(0, limit?.let { min(it, numObservations) } ?: numObservations)
+    val depth = max(1, limit?.let { min(it, numObservations) } ?: numObservations)
 
     // Remove observations one at a time, to build a historical summary
     return List(depth) {
@@ -141,7 +141,9 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
 
     val observationsBySubzone =
         completedObservations
-            .flatMap { observation -> observation.plantingZones.flatMap { it.plantingSubzones } }
+            .flatMap { it.plantingZones }
+            .flatMap { it.plantingSubzones }
+            .filter { it.completedTime != null }
             .groupBy { it.plantingSubzoneId }
 
     val latestPerSubzone =


tommylau523:
* **#2750** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm (changes requested):
Maybe the changes here are fine, but none of this seems to be related to null safety so I'm not sure why this would fix any NPEs. A PR description saying what the problem was would help.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -141,7 +141,9 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
 
     val observationsBySubzone =
         completedObservations
-            .flatMap { observation -> observation.plantingZones.flatMap { it.plantingSubzones } }
+            .flatMap { it.plantingZones }
+            .flatMap { it.plantingSubzones }
+            .filter { it.completedTime != null }
             .groupBy { it.plantingSubzoneId }
 
     val latestPerSubzone =
```
tommylau523, line 149: <i class='graphite__hidden'>[Re: line +150]<br/><br/></i>

`it.completedTime!!` here was leading to a NPE. This should not be possible since observations were already filtered.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2750?utm_source=unchanged-line-comment">Graphite</a>.</span>

sgrimm: What about the change to the depth calculation?

tommylau523:
Added some details. I am unsure either since I was not able to consistently reproduce this issue.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
tommylau523: That is just a logical mistake previously. It never made sense to return a depth of 0 because that means that we were looking for 0 entries.

sgrimm:
> a completed observation can somehow contain a subzone without a completed time

That's an expected state, I think. Users can mark subzones as incomplete in the web app (Seedlings -> Withdrawals -> Planting Progress).

Do we need a deeper change here to account for people clearing the "planting complete" status of a subzone after an observation is done?

tommylau523:
Sorry I meant a completed observation, as opposed to planting completion. And tracing back, subzones/zones/sites should all have completedTime set iff all assocaited monitoringPlots had a completedTime as well. 

Perhaps this is a relic of an "Abandoned" observation. I am still investigating.

tommylau523:
But even if a user can not complete part of an observation, we probably still want this filter, since we want to take accounts of latest completed observation per subzone. The filter should be logically correct still, but I think it is worth doing a deeper dive to understand how an observation can end up in this state.

sgrimm:
That's fair. Could this maybe be related to requesting a plot replacement in an observation?

sgrimm (approved):

PR #2766, opened 2025-01-22
SW-6462 Populate default text for unmodified sections

sgrimm:
If a section variable's definition hasn't changed since the previous manifest
for the same document template, we reuse the existing variable ID. However, the
default section text is per-manifest data (changing it doesn't prevent existing
variables from being reused) so it needs to be inserted into the database with
the new variable manifest ID even if it's the same as in the previous manifest.

This was causing newly-created documents to be missing default text if they used
a template whose manifest had been uploaded more than once.


sgrimm:
* **#2766** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #2769, opened 2025-01-23
Make PlantingSiteBuilder usage more concise

sgrimm:
The `zone()` function in `PlantingSiteBuilder` didn't allow callers to set the
plot counts because initially only a handful of tests needed to change them.
But the number of such tests has grown over time and will shortly grow even more.
Add optional parameters to allow the counts to be set more concisely, and update
the existing test code to use them.

Some unnecessary calls to `ZoneBuilder.subzone()` are removed here as well; the
default behavior is to create a single subzone with the same boundary as its zone
so there's no need to do it explicitly.

Finally, it is now possible to set the plot number when creating a permanent
cluster via `SubzoneBuilder.cluster()` (previously this required calling `plot()`
which has more required parameters) and the plot number is correctly used as
the plot ID.


sgrimm:
* **#2769** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2769?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2769?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #2770, opened 2025-01-23
SW-6455 Implemented Biomass Species tables

tommylau523:
Herbaceous and tree species are required to have `isThreatened` and `isInvasive` attributes within a biomass observation consistently. 

This PR introduces two tables, to track these attributes, and sets up foreign keys so that any referenced herbaceous and tree species must be in those tables. 

The API and store payloads are updated accordingly.


tommylau523:
* **#2770** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2770?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2770?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,58 @@
...
+
+    CONSTRAINT species_identifier
+        CHECK ((species_id IS NOT NULL AND scientific_name IS NULL)
+            OR (species_id IS NULL AND scientific_name IS NOT NULL)),
+
+    FOREIGN KEY (observation_id, monitoring_plot_id)
+        REFERENCES tracking.observation_biomass_details (observation_id, monitoring_plot_id)
+        ON DELETE CASCADE,
+
+    UNIQUE (observation_id, monitoring_plot_id, species_id, scientific_name)
```
sgrimm: Unique keys and nullable columns don't mix well in the version of PostgreSQL we're using in production, because null always fails equality tests. You can try it in psql:
```
create temporary table foo ( 
  col1 integer,
  col2 integer,
  unique (col1, col2)
);
CREATE TABLE

insert into foo values (1, 1);
INSERT 0 1

-- This fails as expected
insert into foo values (1, 1);
ERROR:  duplicate key value violates unique constraint "foo_col1_col2_key"
DETAIL:  Key (col1, col2)=(1, 1) already exists.

insert into foo values (1, null);
INSERT 0 1

-- This doesn't fail because null isn't equal to anything including itself
insert into foo values (1, null);
INSERT 0 1
```
More recent PostgreSQL versions have a way to treat null as a distinct value in unique indexes, but we still need to upgrade our databases, so we can't use that.

In the meantime, the solution is to use partial indexes, like this:
```
CREATE UNIQUE INDEX ON mytable (a, b) WHERE c IS NULL;
CREATE UNIQUE INDEX ON mytable (a, c) WHERE b IS NULL;
```

sgrimm (changes requested):
Will look at the rest of this tomorrow, but one comment on the unique constraint in the meantime.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
sgrimm: This can get a little confusing because null in SQL doesn't behave like null in most programming languages. In SQL I like to think of null as "the absence of any value at all" rather than as a value that happens to be a magic constant. From that point of view, the behavior here makes a kind of sense: X can't be equal to Y if X and Y don't exist at all.

tommylau523: This makes sense. I updated the uniqueness clause to be two clause, one for ID and one for the name. The UNIQUE tag for it to be used as a foreign key later. Coupled with the species_identifier constraints, this is functionally correct, even though it doesn't quite read right.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,64 @@
...
+        ON DELETE CASCADE,
+
+    UNIQUE(observation_id, monitoring_plot_id, species_id),
+    UNIQUE(observation_id, monitoring_plot_id, scientific_name)
+);
+
+CREATE INDEX ON tracking.observation_biomass_herbaceous_species(observation_id);
+CREATE INDEX ON tracking.observation_biomass_herbaceous_species(monitoring_plot_id);
+
+CREATE TABLE tracking.observation_biomass_tree_species (
```
sgrimm: Ditto the above comment about an ID column.

Having two completely identical tables like this feels a little weird to me. I get that it's because we don't want to allow the quadrat species or the recorded trees tables to be able to point to the wrong kind of species. But it means we're going to have to either duplicate a bunch of code or do dynamic references to the columns by name.

Alternate idea, cribbing from the variable values tables:
```
CREATE TABLE tracking.observation_biomass_species (
    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    is_tree BOOLEAN NOT NULL,
    -- Other columns and constraints...

    -- This is needed for the foreign-key reference in the next part
    UNIQUE (id, is_tree)
);

ALTER TABLE tracking.observation_biomass_quadrat_species
    ADD COLUMN observation_biomass_species_id BIGINT NOT NULL,
    ADD COLUMN is_tree BOOLEAN NOT NULL DEFAULT FALSE,
    ADD CHECK (is_tree = FALSE),
    ADD FOREIGN KEY (observation_biomass_species_id, is_tree)
        REFERENCES tracking.observation_biomass_species (id, is_tree);
```

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,64 @@
+CREATE TABLE tracking.observation_biomass_herbaceous_species (
```
sgrimm: Adding an ID column to this table would make foreign-key references to it simpler.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,64 @@
...
+-- This table is now redundant
+DROP TABLE tracking.observation_biomass_additional_species;
+
+ALTER TABLE tracking.observation_biomass_quadrat_species
+    DROP COLUMN is_invasive,
+    DROP COLUMN is_threatened,
+    ADD FOREIGN KEY (observation_id, monitoring_plot_id, species_id)
+        REFERENCES tracking.observation_biomass_herbaceous_species(observation_id, monitoring_plot_id, species_id),
+    ADD FOREIGN KEY (observation_id, monitoring_plot_id, species_name)
+        REFERENCES tracking.observation_biomass_herbaceous_species(observation_id, monitoring_plot_id, scientific_name);
```
sgrimm: If the species table had an ID column, this table could refer to that column and wouldn't need the species ID and name columns.

sgrimm (changes requested):
The code here seems straightforward, but I'm not sure two separate tables is the right model here. See inline for more.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
tommylau523: Part of it right now is that I haven't gotten confirmation whether a species can be both herbaceous and tree. Although I think we can effectively combine these tables for now, and not add the `is_tree` boolean. The only reason we need them to be separated is to keep track of "additional" herbaceous species, which are herbaceous species not found in either quadrats.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -223,28 +225,27 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
         DSL.multiset(
                 DSL.select(
                         POSITION_ID,
-                        SPECIES_ID,
-                        SPECIES_NAME,
-                        IS_INVASIVE,
-                        IS_THREATENED,
                         ABUNDANCE_PERCENT,
+                        OBSERVATION_BIOMASS_SPECIES.SPECIES_ID,
```
tommylau523, line 229: I was hoping to use the jooq implicit join field here but the field ends up being something like `observationBiomassQuadratSpeciesBiomassSpeciesFk...`

Is there a config to make it less verbose. We can of course keep it as an inner join like we have now.

sgrimm: Maybe try giving the foreign key constraint a name. Not sure that'll work, but that looks kind of like a PostgreSQL-generated constraint name.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,62 @@
...
+
+    CONSTRAINT species_identifier
+        CHECK ((species_id IS NOT NULL AND scientific_name IS NULL)
+            OR (species_id IS NULL AND scientific_name IS NOT NULL)),
+
+    FOREIGN KEY (observation_id, monitoring_plot_id)
+        REFERENCES tracking.observation_biomass_details (observation_id, monitoring_plot_id)
+        ON DELETE CASCADE,
+
+    UNIQUE(id, observation_id, monitoring_plot_id)
```
sgrimm, line 19: This won't do anything. ID is already always unique.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
```diff
@@ -0,0 +1,62 @@
...
+    monitoring_plot_id BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    position_id INTEGER NOT NULL REFERENCES tracking.observation_plot_positions,
+    biomass_species_id BIGINT NOT NULL REFERENCES tracking.observation_biomass_species ON DELETE CASCADE,
+    abundance_percent INTEGER NOT NULL CHECK ( abundance_percent BETWEEN 0 AND 100),
+
+    FOREIGN KEY (observation_id, monitoring_plot_id, biomass_species_id)
+        REFERENCES tracking.observation_biomass_species (observation_id, monitoring_plot_id, id)
+        ON DELETE CASCADE,
+
+    UNIQUE (observation_id, monitoring_plot_id, position_id, biomass_species_id)
```
sgrimm: Could make this `PRIMARY KEY` instead of `UNIQUE` to slightly clarify the intent.

src/main/kotlin/com/terraformation/backend/tracking/model/BiomassDetailsModel.kt
```diff
@@ -12,24 +12,28 @@ import com.terraformation.backend.db.tracking.TreeGrowthForm
 import java.math.BigDecimal
 import java.time.Instant
 
-data class BiomassAdditionalSpeciesModel(
+data class BiomassSpeciesKey(
+    val speciesId: SpeciesId? = null,
+    val scientificName: String? = null,
+)
```
sgrimm, line 18: This is totally fine, but if you like, you could define this in the `EMBEDDABLES` list in `Config.kt` and then jOOQ will include it as a field in the generated classes so you don't have to construct it manually from the ID and name fields.

sgrimm (approved):
This seems nicer. There's one unnecessary constraint it'd be good to get rid of before merging, and I had a couple other comments you can take or leave as you see fit.

src/main/resources/db/migration/0300/V338__BiomassSpecies.sql
tommylau523: Mostly wanted to use this for foreign key constraint later.

sgrimm: Ah, right, ignore my comment, then.

tommylau523:
### Merge activity

* **Jan 23, 6:17 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2770).

PR #2792, opened 2025-01-31
Updated mortality rates to be null if there is no plants in permanent plots

tommylau523:
Previously, if there is not recorded plants observed in a permanent plot, we defaulted mortality rate to zero. Set to null instead to better reflect the distinction.


tommylau523:
* **#2792** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2792?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -1070,7 +1070,7 @@ data class ObservationPlantingSubzoneResultsPayload(
             "Percentage of plants of all species that were dead in this subzone's permanent " +
                 "monitoring plots.")
     val monitoringPlots: List<ObservationMonitoringPlotResultsPayload>,
-    val mortalityRate: Int,
+    val mortalityRate: Int?,
```
sgrimm, line 1073: Should the standard deviation also be null if the mortality rate is null?

sgrimm (approved):
Nit: "permanent" is misspelled in the title, but since "permanent plants" aren't really a concept, I think it'd be clearer as, "no plants in permanent plots."

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
tommylau523: I can't think of an example where this is not the case. Because the mortality rate will be non-null, if any permanent plot has at least one recorded permanent live or dead plant. Then, we will also have a standard deviation value.

But this leads me to think, if our variance calculation for mortality rate is correct. One plot with many more plants would get weighted the same as another plot with a few plants.

PR #2797, opened 2025-02-01
Used weighted standard deviation for mortality rate

tommylau523:
Our previous mortality rate calculations treat each plot mortality rate as one data point. This is incorrect because our mortality rate in larger sampling sizes (subzones, zones, sites) actually calculates the value, by adding up all the dead and live plants found in the plots. This is equivalent to a weighted mean. 

To correct the calculations, we need to use the variance of mortality rate weighted by the number of plants found in that plot instead. So a plot with X number of plants has a weight of X when calculating this result.

Another way to think of this is, if a plot has X number of plants and Y mortality rate, we are counting this as X number of data points, each with a value Y. So the sum of difference squared and the variance calculation will now correctly factored in the number of plants within a plot.

https://docs.google.com/spreadsheets/d/17-9VlElRgnXU7rnCPsesw6bM5c1Du7Z_ZIvSq8yJPbI/edit?usp=sharing


tommylau523:
* **#2797** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2797?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2797?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/resources/tracking/observation/ObservationsSummary/PlotStats.csv
```diff
@@ -1,20 +1,20 @@
...
-312,,,,,,,,158,3,32%,20,43,95,688,158,3,32%,20,43,95,688
-313,,,,,,,,188,4,18%,17,76,95,1216,188,4,18%,17,76,95,1216
-321,,,,,,,,186,4,22%,20,71,95,1136,186,4,22%,20,71,95,1136
-322,,,,,,,,203,4,21%,21,77,105,1232,203,4,21%,21,77,105,1232
-323,,,,,,,,245,4,22%,31,112,102,1792,245,4,22%,31,112,102,1792
-324,,,,,,,,243,4,12%,18,131,94,2096,243,4,12%,18,131,94,2096
-325,,,,,,,,236,4,16%,21,110,105,1760,236,4,16%,21,110,105,1760
\ No newline at end of file
+Observation ->,1,1,1,1,1,1,1,1,2,2,2,2,2,2,2,2,3,3,3,3,3,3,3,3
+Plot,# Plants,# Species,Mortality Rate,Cumulative Dead Plants,Live Plants,# Live + Dead,Existing Plants,Planting Density,# Plants,# Species,Mortality Rate,Cumulative Dead Plants,Live Plants,# Live + Dead,Existing Plants,Planting Density,# Plants,# Species,Mortality Rate,Cumulative Dead Plants,Live Plants,# Live + Dead,Existing Plants,Planting Density
```
tommylau523, line 2: In this file, we mostly changed total dead to cumulative dead, and added live + dead, since live + dead is used for weights of std dev.

src/test/resources/tracking/observation/ObservationsSummary/SiteStats.csv
```diff
@@ -1,3 +1,3 @@
 1,1,1,1,1,1,2,2,2,2,2,2,3,3,3,3,3,3
 Planting Density,Planting Density Std Dev,Estimated # Plants,# Species,Mortality Rate,Mortality Rate Std Dev,Planting Density,Planting Density Std Dev,Estimated # Plants,# Species,Mortality Rate,Mortality Rate Std Dev,Planting Density,Planting Density Std Dev,Estimated # Plants,# Species,Mortality Rate,Mortality Rate Std Dev
-762,309,,4,25%,12%,1053,496,2220000,4,22%,10%,1614,859,4571000,4,23%,8%
```
tommylau523, line 3: In this and subsequent files, the std dev values are updated since they are now weighted for number of plants correctly

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -684,7 +685,17 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
 
               val mortalityRate = species.calculateMortalityRate()
               val mortalityRateStdDev =
-                  monitoringPlots.mapNotNull { it.mortalityRate }.calculateStandardDeviation()
+                  monitoringPlots
+                      .mapNotNull { plot ->
+                        plot.mortalityRate?.let {
```
sgrimm: This would be slightly clearer with a named lambda variable so the last line is more like `mortalityRate to permanentPlots`.

sgrimm (approved):
One small clarity suggestion, but looks reasonable other than that.

tommylau523:
### Merge activity

* **Feb 3, 2:12 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2797).

PR #2807, opened 2025-02-03
SW-6197 Add Likely Intermediate values to Seed Storage Behavior field

ice1080:
- Adds "Likely Intermediate" values to database and code
- (also adds some IntelliJ notes to `CONVENTIONS.md`)


docs/CONVENTIONS.md
```diff
@@ -10,7 +10,8 @@ Currently, the build targets the Java 23 JVM. We use the Amazon Corretto JVM and
 
 All Kotlin code must be formatted with [ktfmt](https://github.com/facebookincubator/ktfmt). You can run ktfmt in a couple different ways:
 
-* Install the [ktfmt IntelliJ plugin](https://plugins.jetbrains.com/plugin/14912-ktfmt) (recommended). This will cause IntelliJ's "Reformat Code" action to use ktfmt.
+* Install the [ktfmt IntelliJ plugin](https://plugins.jetbrains.com/plugin/14912-ktfmt) (recommended). This will cause IntelliJ's "Reformat Code" action to use ktfmt (Default Code Style is fine).
+  * You can also enable this to run automatically on saving of the file if you want. To do this: Preferences -> Tools -> Actions on Save, enable Reformat Code (optionally limit file type) -> OK.
```
sgrimm: This should ideally be a separate little PR since it has nothing to do with adding seed storage behaviors.

We don't have a strict rule about combining unrelated changes in PRs. A general rule of thumb we try to follow is that it's okay to include minor tidying of nearby code that's at least somewhat related to the topic of the PR, but changes to completely unrelated code/docs should be split off to their own PRs.

sgrimm (changes requested):
The main change here seems good. A few minor things about the PR title, just for consistency with our existing commit message style:

* No need for a hyphen between the Jira task number and the rest of the title.
* Capitalize the first word of the title after the Jira task.
* Use either present-tense imperative or past tense. Imperative is kind of the de facto "official" git style (e.g., it's used in the Linux kernel repo and is the style that's generated by git itself for merge commits) and it's what I try to use. But past tense is common too; Tommy prefers that. This isn't a huge deal, but IMO it's preferable to not introduce a third style to our history.

docs/CONVENTIONS.md
ice1080: Good to know. 
I'm used to having so much process around PRs that it was almost always preferable to combine unrelated changes when they wouldn't cause any issues. Also wasn't sure if a flood of tiny change PRs was ok with the team

ice1080: Removed from this PR

sgrimm (approved):
Looks good.

PR #2813, opened 2025-02-04

terraware-phrase:


sgrimm:
@tommylau523 Are the report-related strings correct here? The PR is adding some English ones which makes me think the enum names are out of date and the strings should be removed using the Phrase UI.

tommylau523:
Interesting. I found these strings on top of main. Not sure why they are showing up as diffs. But the strings are all correct as of now.

sgrimm (approved):

PR #2815, opened 2025-02-06
SW-6538 Reworked Biomass tables and APIs

tommylau523:
A few important changes to our data:
1) Eliminated `isTrunk` as an attribute. Instead, used a new growth form `Trunk`. It shares the same attributes as growth form `Tree`, except that height is now required only for trees
2) Branches are now obsolete/renamed to trunks. Split the Tree payload into three types of payloads.
3) Trunk number is added to track multi-trunk tree (Trunk growth form). For other growth form, trunk must be 1, because it should only have one trunk. This is handled by the system, and inferred by list index.

Other quality-of-life changes:
1) Throws exception for incorrect/missing values. This is intended for quick development in the FE for now. Future work will be done to return a list of problems as strings, including localized messages. Test cases for validations will be added then. 
2) "Excess" values are now omitted instead of resulting in errors. Test cases covered these.


tommylau523:
* **#2815** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2815?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

joshbuchea:
@tommylau523 some of these BE updates add unnecessary FE complexity:

- A user-supplied tree number isn't needed—it can be inferred/generated in the BE based on current requirements
- Removing the `branches` property, introducing a new type of tree, and introducing Tree & Trunk IDs forces the FE to manually associate trees and trunks via user-supplied IDs, increasing complexity and error risk

Suggested approach:

- Infer tree & trunk IDs in the BE (if needed)
- Replace `branches` with `trunks` to keep related data grouped by tree

src/main/kotlin/com/terraformation/backend/tracking/api/BiomassPayloads.kt
```diff
@@ -0,0 +1,319 @@
...
+    @Schema(minimum = "0", maximum = "100") //
+    val herbaceousCoverPercent: Int,
+    @Schema(description = "Required for Mangrove forest.", minimum = "0", maximum = "14")
+    val ph: BigDecimal?,
+    val quadrats: List<BiomassQuadratPayload>,
+    @Schema(description = "Measured in ppt. Required for Mangrove forest.", minimum = "0")
+    val salinity: BigDecimal?,
+    @Schema(minimum = "0") //
+    val smallTreeCountLow: Int,
+    @Schema(minimum = "smallTreeCountHigh") //
```
sgrimm: This isn't a valid minimum.

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
```diff
@@ -566,46 +558,38 @@ class ObservationResultsStoreTest : DatabaseTest(), RunsAsUser {
...
                           diameterAtBreastHeightCm = BigDecimal.TEN,
                           pointOfMeasurementM = BigDecimal.valueOf(1.5),
-                          heightM = BigDecimal.TEN,
                           isDead = false,
-                          isTrunk = true,
                           speciesId = treeSpeciesId2,
-                          treeGrowthForm = TreeGrowthForm.Tree,
+                          treeGrowthForm = TreeGrowthForm.Trunk,
+                          treeNumber = 3,
+                          trunkNumber = 1,
```
sgrimm, line 582: This doesn't seem to match the requirements/designs, if I'm reading them right. My interpretation is that for trees, there's always a "tree" entity that has a height measurement, and then in addition to that, there can optionally also be "trunk" entities without height measurements. So the "trunk" growth form would always have a trunk number of 2 or higher, and trunk number 1 would always be the "tree" or "shrub" growth form. Example from the mobile app designs:

Screenshot 2025-02-06 at 10 45 42 AM

sgrimm (changes requested):
The code looks fine mechanically, but I don't think it's quite the right data model given the requirements.

tommylau523:
@joshbuchea I made some changes based on what you suggested. It is difficult to find a middle ground because we were making a lot of assumptions about the data models based on how they were collected, but the storage and usages don't quite reflect how they will be queried and used in the web app.

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
tommylau523: Chudo and Jennifer have made that distinction. A multi-trunk tree will never require a height. A single-trunk tree will always require height.

sgrimm: Fair enough. I didn't see that in the requirements/designs.

sgrimm (approved):

PR #2818, opened 2025-02-07
SW-4476 Stop stubbing out string lookups in AppNotificationServiceTest

ice1080:


sgrimm (approved):
Looks good. One nit on the PR title: no need for the hyphen between the Jira task and the rest of the title. Not super important but it gives us a couple more characters before the text gets wrapped or truncated when viewing the commit history.

ice1080:
> Looks good. One nit on the PR title: no need for the hyphen between the Jira task and the rest of the title. Not super important but it gives us a couple more characters before the text gets wrapped or truncated when viewing the commit history.

Thanks for the context. I'm happy to change things for nits any time, but it's even more helpful understanding context.
I've got a few years of muscle memory built up for adding that hyphen in so might still include it sometimes, but I'm happy to remove it whenever I forget

PR #2820, opened 2025-02-07
SW-4477 Check Event Listeners in AppNotificationServiceTest

ice1080:
- Check Event Listeners in AppNotificationServiceTest
- Add missing EventListener to AppNotificationService and test


sgrimm (approved):
Looks good!

PR #2822, opened 2025-02-08
SW-6533 Include all subzones in mortality rates

sgrimm:
Fix two problems with the mortality rate calculations for observations that don't
cover the entire planting site.

1. The cumulative dead plant count for each monitoring plot is supposed to carry
   forward from the previous observation of that plot, but we were only copying
   the value if the same plot was observed in two consecutive observations of the
   same planting site. With partial observations, it's possible for the most
   recent past observation to not include a given plot because a different set of
   subzones were observed.

   This was causing mortality rates to reset each time a different set of subzones
   was observed because past dead plants were being forgotten.

   The fix is to copy the cumulative dead counts from the most recent previous
   observation of each plot, not from the most recent observation of the site as
   a whole.

2. The plot- and subzone-level mortality rates are based on the live plants
   recorded in the current observation. But if the current observation only
   covers a subset of all the subzones in the planting site, the zone- and
   site-level rates need to also include plants that were recorded in the other
   subzones' most recent observations.

   The fix is to recalculate the zone- and site-level mortality rates when an
   observation finishes, with the recalculation taking data from the whole site
   into account, not just the parts that were included in the current observation.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -70,6 +70,7 @@ import com.terraformation.backend.tracking.model.NewObservedPlotCoordinatesModel
 import com.terraformation.backend.tracking.model.ObservationModel
 import com.terraformation.backend.tracking.model.ObservationPlotCounts
 import com.terraformation.backend.tracking.model.ObservationPlotModel
+import com.terraformation.backend.util.eqOrIsNull
 import jakarta.inject.Named
 import java.time.Instant
 import java.time.InstantSource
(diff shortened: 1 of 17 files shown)


sgrimm:
* **#2829** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2829?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2822** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2822?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2822?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2827** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2827?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sgrimm:
The test failures here are expected because those tests are based on the numbers from the plant tracking math spreadsheet which needs to be updated to reflect these behavior changes. Putting this PR out mostly so people can sanity-check the new test scenario I added.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1389,6 +1390,174 @@ class ObservationStore(
...
+        with(OBSERVED_SUBZONE_SPECIES_TOTALS) {
+          val observationIdForSubzoneField =
+              DSL.select(OBSERVATIONS.ID)
+                  .from(OBSERVATIONS)
+                  .join(OBSERVATION_REQUESTED_SUBZONES)
+                  .on(OBSERVATIONS.ID.eq(OBSERVATION_REQUESTED_SUBZONES.OBSERVATION_ID))
+                  .where(
+                      OBSERVATION_REQUESTED_SUBZONES.PLANTING_SUBZONE_ID.eq(PLANTING_SUBZONES.ID))
+                  .and(
+                      OBSERVATIONS.COMPLETED_TIME.le(
```
tommylau523, line 1440: I think we may need `IS_AD_HOC.isFalse()` to a lot of these queries. There are probably other places that need this too.

tommylau523 (approved):
The math looks good! I am just wondering if we need to do this at a plot level.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: Good call. I don't think we need it in the `populateCumulativeDead` query because that's only looking at plots that were marked as permanent in their observations and ad-hoc observations don't have permanent plots.

sgrimm:
> I am just wondering if we need to do this at a plot level.

I asked this too in the requirements discussion. Turns out we want the subzone-level totals to only consider historical data from the plots in the current observation, even if there were additional permanent plots in past observations. But we want the zone- and site-level totals to count _subzones_ that aren't in the current observation. If we did this at the plot level too, then we'd end up counting the cumulative dead plants from plots where the current observation won't ever have a live plant count, which would skew the subzone-level mortality rate.

sgrimm:
### Merge activity

* **Feb 11, 4:48 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2822).
* **Feb 11, 4:48 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2822).

PR #2824, opened 2025-02-08
SW-4478 Extract common logic of AppNotificationServiceTest

ice1080:


sgrimm (commented):
This is looking good so far.

sgrimm (approved):
Great! This should make it easier to test more notifications going forward.

PR #2846, opened 2025-02-18
SW-6556 Set completedTime to latest plot completion time instead of time of action

tommylau523:
A separate migration PR to fix previous observation completedTime to follow.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -575,12 +575,22 @@ class ObservationStore(
     withLockedObservation(observationId) { observation ->
       observation.validateStateTransition(newState)
 
+      val maxCompletedTime =
+          if (newState == ObservationState.Completed || newState == ObservationState.Abandoned) {
+            dslContext
+                .select(DSL.max(OBSERVATION_PLOTS.COMPLETED_TIME))
+                .from(OBSERVATION_PLOTS)
+                .where(OBSERVATION_PLOTS.OBSERVATION_ID.eq(observationId))
+                .and(OBSERVATION_PLOTS.STATUS_ID.eq(ObservationPlotStatus.Completed))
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#2846** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2846?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2846?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -575,12 +575,22 @@ class ObservationStore(
...
 
+      val maxCompletedTime =
+          if (newState == ObservationState.Completed || newState == ObservationState.Abandoned) {
+            dslContext
+                .select(DSL.max(OBSERVATION_PLOTS.COMPLETED_TIME))
+                .from(OBSERVATION_PLOTS)
+                .where(OBSERVATION_PLOTS.OBSERVATION_ID.eq(observationId))
+                .and(OBSERVATION_PLOTS.STATUS_ID.eq(ObservationPlotStatus.Completed))
+                .fetchOneInto(Instant::class.java)
+                ?: throw IllegalStateException("Observation $observationId has no completed plots")
```
sgrimm, line 586: To make sure I understand: this doesn't break the "abandon a brand-new observation" case because observations without any completed plots are deleted rather than having their states updated, so this code will never be called.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
tommylau523: Yes this is correct. An observation with no completed plots will simply be deleted. 

The observation state change assumes one completed plot.

PR #2849, opened 2025-02-20
SW-6550 Update subsections of new parent section

sgrimm:
The manifest importer wasn't detecting when child sections moved to new parents,
either because their parent names were updated or because the parent section's
stable ID was changed. Handle both of those cases.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
@@ -140,8 +140,8 @@ class ManifestImporter(
               .fetchManifestVariables(existingManifestId)
               .filterIsInstance<SectionVariable>()
 
-      // Child variables such as subsections are not in the top-level list, so we
-      // need to recursively process them.
+      // Child variables such as subsections are not in the top-level list, so we need to
+      // recursively process them.
       fun processVariables(variables: List<SectionVariable>) {
         variables.forEach { variable ->
           val csvVariable = csvVariableByStableId[variable.stableId]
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#2849** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/db/manifest/ManifestImporter.kt
```diff
@@ -168,6 +170,24 @@ class ManifestImporter(
...
+    /**
+     * If a section is new but its children already exist, mark the children as non-reusable. Apply
+     * the same logic to each child section's children, walking the section tree to find
+     * transplanted branches.
+     */
+    private fun markMovedSubsectionsNonReusable(variable: SectionVariable) {
+      if (csvVariableByStableId[variable.stableId] != null) {
+        variable.children.forEach { markMovedSubsectionsNonReusable(it) }
+      } else {
+        fun findExistingSubsectionsStillInCsv(variable: SectionVariable): List<CsvSectionVariable> {
```
tommylau523, line 182: Is there ever a chance of this not terminating? I know it should never have cycles. Should we filter by children that have not been marked not reusable yet?

sgrimm: I don't think it's possible to end up with circular parent-child relationships given that a section's parent has to be a section that was already listed in the CSV. You'd need some way to do a forward reference. I did think about this a bit, but even with the ability to move sections to new parents, I wasn't able to come up with a test scenario where I'd end up with a cycle.

PR #2866, opened 2025-02-26
SW-6624 Use Int columns to store reference for metrics

tommylau523:
To avoid sorting issues and complications, product team was able to confirmed that this will always be numbers, and up to 3. So we updated the columns to reflect that.

### DIFF
--- a/src/main/resources/db/migration/0300/V347__ReportMetricReference.sql
+++ b/src/main/resources/db/migration/0300/V347__ReportMetricReference.sql
@@ -1,0 +1,11 @@
+ALTER TABLE accelerator.standard_metrics
+    DROP COLUMN reference,
+    ADD COLUMN reference INTEGER NOT NULL CHECK (reference >= 0),
+    ADD COLUMN sub_reference INTEGER CHECK (sub_reference >= 0),
+    ADD COLUMN sub_sub_reference INTEGER CHECK (sub_sub_reference >= 0);
+
+ALTER TABLE accelerator.project_metrics
+    DROP COLUMN reference,
+    ADD COLUMN reference INTEGER NOT NULL CHECK (reference >= 0),
+    ADD COLUMN sub_reference INTEGER CHECK (sub_reference >= 0),
+    ADD COLUMN sub_sub_reference INTEGER CHECK (sub_sub_reference >= 0);
--- a/src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
@@ -111,7 +111,8 @@ class ReportStoreTest : DatabaseTest(), RunsAsDatabaseUser {
               component = MetricComponent.Climate,
               description = "Climate standard metric description",
               name = "Climate Standard Metric",
-              reference = "3.0",
+              reference = 3,
+              subReference = 0,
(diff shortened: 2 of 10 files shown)


tommylau523:
* **#2866** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2866?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0300/V347__ReportMetricReference.sql
```diff
@@ -0,0 +1,11 @@
...
+    DROP COLUMN reference,
+    ADD COLUMN reference INTEGER NOT NULL CHECK (reference >= 0),
+    ADD COLUMN sub_reference INTEGER CHECK (sub_reference >= 0),
+    ADD COLUMN sub_sub_reference INTEGER CHECK (sub_sub_reference >= 0);
+
+ALTER TABLE accelerator.project_metrics
+    DROP COLUMN reference,
+    ADD COLUMN reference INTEGER NOT NULL CHECK (reference >= 0),
+    ADD COLUMN sub_reference INTEGER CHECK (sub_reference >= 0),
+    ADD COLUMN sub_sub_reference INTEGER CHECK (sub_sub_reference >= 0);
```
tommylau523, line 11: <p class='graphite__hidden'><i>[Re: line +12]</i></p>
No migration here for existing entries since we have no actual data yet. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2866?utm_source=unchanged-line-comment">Graphite</a>.</p>

src/main/resources/db/migration/0300/V347__ReportMetricReference.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE accelerator.standard_metrics
+    DROP COLUMN reference,
+    ADD COLUMN reference INTEGER NOT NULL CHECK (reference >= 0),
+    ADD COLUMN sub_reference INTEGER CHECK (sub_reference >= 0),
+    ADD COLUMN sub_sub_reference INTEGER CHECK (sub_sub_reference >= 0);
```
ice1080, line 5: Even though product confirmed only 3, do we think it could ever become 4? E.g. `2.1.4.7`?

src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
ice1080, line 1: Should we add tests to this file for subSubReference also? Or are we comfortable enough with the few tests of it in `ReportMetricStoreTest.kt`?

tommylau523:
I may opt to delay this until some frontend work has begun. I gave it another thought, and the FE is expected to combine and sort a series of metrics anyways.

sgrimm:
Is this so that, e.g., `10.1.2` comes before `10.1.15`? If so, there are other ways to do that, like using a collation that's number-aware. We're already doing that for subzone and monitoring plot names. See `V183__MonitoringPlotBoundary.sql`.

tommylau523:
Yes that is correct. Let me take a look at it!

PR #2868, opened 2025-02-26
SW-6620 Create Tables for Funding Entities and Funders

ice1080:
Data Model reference - https://www.figma.com/board/zjrmUHFkxKhcEndw6GROlZ/Funding-Entity-Data-Model?node-id=0-1&p=f&t=rTRRng6vfD32b01d-0

- Add `user_types` row for Funder
- Create tables for Funding Entities
- Add corresponding Kotlin code for these
- Update .editorconfig to fix markdown issue


src/main/resources/db/migration/0300/V347_FundingEntities.sql
```diff
@@ -0,0 +1,29 @@
+INSERT INTO user_types (id, name)
```
tommylau523: This should be added to `R__TypeCodes.sql` instead.

tommylau523:
This migration may need to be renamed, depending if the other PR on report reference gets merged first.

src/main/resources/db/migration/0300/V348__FundingEntities.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE accelerator.funding_entities
```
sgrimm: Maybe bikeshedding, but I wonder if we should create a new schema for these tables rather than putting them under `accelerator`. Functionally there's no difference, of course; it's just about organizing things. My assumption is that we'll be adding lots more funder-related tables over time and it might be nice to have them in a separate namespace from the tables that hold project data.

src/main/resources/db/migration/0300/V348__FundingEntities.sql
```diff
@@ -0,0 +1,26 @@
...
+(
+    id   BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    name TEXT NOT NULL,
```
sgrimm: Do funding entity names need to be unique?

src/main/resources/db/migration/0300/V348__FundingEntities.sql
```diff
@@ -0,0 +1,26 @@
...
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL
+);
+
+CREATE TABLE accelerator.funding_entity_projects
+(
+    funding_entity_id BIGINT NOT NULL REFERENCES accelerator.funding_entities,
+    project_id        BIGINT NOT NULL REFERENCES projects
```
sgrimm: Should these have `ON DELETE CASCADE`?

src/main/resources/db/migration/0300/V348__FundingEntities.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE INDEX ON accelerator.funding_entity_projects(project_id);
+
+
+CREATE TABLE accelerator.funding_entity_users
+(
+    funding_entity_id BIGINT NOT NULL REFERENCES accelerator.funding_entities,
+    user_id           BIGINT NOT NULL REFERENCES users
+);
+CREATE INDEX ON accelerator.funding_entity_users(funding_entity_id);
+CREATE INDEX ON accelerator.funding_entity_users(user_id);
```
sgrimm: Could make the user ID index `UNIQUE` to enforce the "user can only be in one funding entity" constraint. Or even make user ID the primary key of this table.

src/main/resources/db/migration/0300/V348__FundingEntities.sql
ice1080: Indeed! I meant to add that and then got distracted.

src/main/resources/db/migration/0300/V348__FundingEntities.sql
ice1080: I like that idea and this is probably the best time to do it. Using schema `funder`

src/main/resources/db/migration/0300/V348__FundingEntities.sql
ice1080: I like the primary key idea, will add that!

src/main/resources/db/migration/0300/V348__FundingEntities.sql
ice1080: Good question, have a question out to product on this one

ice1080: Jennifer just confirmed they need to be unique. Updating this

tommylau523:
Another number used!

ice1080:
No problem, I'm seeing the PR's coming in and am still going through the kotlin code for these anyway 👍

ice1080:
Unsure if I needed to update jooq/Config.kt as well. I read through the comments but may need a quick call to understand when this is needed

(update) - after reading through a few more times, I don't _believe_ that any changes are needed to that file?

.editorconfig
```diff
@@ -67,7 +67,7 @@ ij_javascript_array_initializer_wrap = off
 ij_javascript_assignment_wrap = off
 ij_javascript_binary_operation_sign_on_next_line = false
 ij_javascript_binary_operation_wrap = off
-ij_javascript_blacklist_imports = rxjs/Rx,node_modules/**,**/node_modules/**,@angular/material,@angular/material/typings/**
+ij_javascript_blacklist_imports = rxjs/Rx, node_modules/**, **/node_modules/**, @angular/material, @angular/material/typings/**
```
sgrimm: Changes to this file should probably be a separate PR (I assume IntelliJ did this on its own).

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
```diff
@@ -0,0 +1,78 @@
...
+    val lastName: String?,
+    val cookiesConsented: Boolean?,
+    val cookiesConsentedTime: Instant?,
+) : TerrawareUser, UserDetails {
+  companion object {
+    private val log = perClassLogger()
+  }
+
+  override val timeZone: ZoneId
+    get() = ZoneOffset.UTC
```
sgrimm: Is this right? I'd think we'd want to let funder users set their time zones like we do for regular users, but maybe not.

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
```diff
@@ -0,0 +1,78 @@
...
+
+  override val userType: UserType
+    get() = UserType.Funder
+
+  /*
+   * Funders are not tied to organizations or facilities so do not have roles related to them
+   */
+  override val organizationRoles: Map<OrganizationId, Role>
+    get() {
+      throw NotImplementedError("System user does not support enumerating roles")
```
sgrimm: Update this to say "Funder user" instead of "System user"

src/main/resources/db/migration/0300/V349__FundingEntities.sql
```diff
@@ -0,0 +1,28 @@
...
+    name TEXT NOT NULL UNIQUE,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL
+);
+
+CREATE TABLE funder.funding_entity_projects
+(
+    funding_entity_id BIGINT NOT NULL REFERENCES funder.funding_entities,
```
sgrimm: Do we want ON DELETE CASCADE for funding entities too?

sgrimm (changes requested):
You'll need to edit `Config.kt` to define a type-safe wrapper for funding entity IDs; otherwise they'll be of type `Long`. The wrappers are mostly to protect against passing the wrong kind of ID somewhere by mistake.

src/main/resources/db/migration/0300/V349__FundingEntities.sql
ice1080: ~Not quite following this one. What would be getting deleted that would cascade and delete a funding entity?~

.editorconfig
ice1080: I had to make changes to .editorconfig so that index.md passes the spotless check since they were inconsistent with each other. When I did this, IntelliJ did auto-change this file because it didn't match the configs defined in itself. I'm happy to move the whitespace changes to a different PR, but the changes to the markdown section seem to be related here... unless you want me to move all of that to a different PR and get it merged to main and then merged into this PR before this PR can be merged. I'm good with whichever

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
ice1080: Will check with product and work on understanding more about how current users set their timezones.

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
ice1080: Thanks, I thought I updated both but must have missed this one.

src/main/resources/db/migration/0300/V349__FundingEntities.sql
ice1080: Ohh sorry was thinking of the wrong table. Ya it does probably need to be added for both. Will update

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
ice1080: Product confirmed adding all the same settings to Funders. Michael will likely update the design to reflect so I'll get this file updated to reflect as well. Thanks for calling this out!

.editorconfig
ice1080: Created PR for this - https://github.com/terraware/terraware-server/pull/2873

ice1080: This file should no longer show as modified once the other PR gets merged to main and I merge it into this one

ice1080: No longer in this PR!

sgrimm (approved):

PR #2870, opened 2025-02-26
SW-6561 Fix migration PostgreSQL 17 incompatibility

sgrimm:
In PostgreSQL 17, `system_user` is now a reserved keyword, so we can't use it
unquoted as a subquery or CTE name any more.

We were using it in two migrations. One of them is a numbered migration, and
switching to a different CTE name means editing the migration script. By default,
Flyway flags edits to numbered migrations as validation errors. So we also need
to tell Flyway to not treat this specific edit as an error, but to still flag
edits to other numbered migrations.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/db/FlywayConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/db/FlywayConfig.kt
@@ -1,0 +1,48 @@
+package com.terraformation.backend.db
+
+import org.flywaydb.core.Flyway
+import org.flywaydb.core.api.CoreErrorCode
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#2870** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2870?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2870?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #2887, opened 2025-03-03
SW-6607 CRUD Funding Entities

ice1080:
A user with the correct permissions can:
- select a funding entity (includes projectIds)
- create a funding entity with at least a name, and optionally a set of projectIds
- update a funding entity's name, and/or add/remove projects by id
- delete a funding entity
  - cascades in the db to delete `funding_entity_projects` rows and `funding_entity_users` rows

Additionally if a user is deleted, the corresponding `funding_entity_users` row is removed if one exists for the user.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
@@ -111,17 +111,18 @@ class FundingEntityService(
     }
   }
 
-  private fun removeProjectsFromEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+  private fun removeProjectsFromEntity(
+      fundingEntityId: FundingEntityId,
+      projectIds: Set<ProjectId>
+  ) {
     requirePermissions { updateFundingEntityProjects() }
 
-    for (projectId in projects) {
-      with(FUNDING_ENTITY_PROJECTS) {
-        dslContext
-            .deleteFrom(FUNDING_ENTITY_PROJECTS)
-            .where(FUNDING_ENTITY_ID.eq(fundingEntityId))
-            .and(PROJECT_ID.eq(projectId))
-            .execute()
-      }
+    with(FUNDING_ENTITY_PROJECTS) {
+      dslContext
+          .deleteFrom(FUNDING_ENTITY_PROJECTS)
+          .where(FUNDING_ENTITY_ID.eq(fundingEntityId))
+          .and(PROJECT_ID.`in`(projectIds))
+          .execute()
     }
   }
 
--- a/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
@@ -46,23 +46,17 @@ class FundingEntityService(
                 .fetchOne(ID) ?: throw FundingEntityExistsException(name)
           }
 
-      for (projectId in projects.orEmpty()) {
-        with(FUNDING_ENTITY_PROJECTS) {
-          dslContext
-              .insertInto(FUNDING_ENTITY_PROJECTS)
-              .set(FUNDING_ENTITY_ID, fundingEntityId)
-              .set(PROJECT_ID, projectId)
-              .onConflict()
-              .doNothing()
-              .execute()
-        }
-      }
+      addProjectsToEntity(fundingEntityId, projects.orEmpty())
 
       fundingEntityStore.fetchOneById(fundingEntityId)
     }
   }
 
-  fun update(row: FundingEntitiesRow) {
+  fun update(
+      row: FundingEntitiesRow,
+      addProjects: Set<ProjectId>? = null,
+      removeProjects: Set<ProjectId>? = null,
+  ) {
     val fundingEntityId = row.id ?: throw IllegalArgumentException("Funding Entity ID must be set")
 
     requirePermissions { manageFundingEntities() }
@@ -70,18 +64,23 @@ class FundingEntityService(
     val userId = currentUser().userId
     val now = clock.instant()
 
-    try {
-      with(FUNDING_ENTITIES) {
-        dslContext
-            .update(FUNDING_ENTITIES)
-            .set(NAME, row.name)
-            .set(MODIFIED_BY, userId)
-            .set(MODIFIED_TIME, now)
-            .where(ID.eq(fundingEntityId))
-            .execute()
+    dslContext.transaction { _ ->
+      try {
+        with(FUNDING_ENTITIES) {
+          dslContext
+              .update(FUNDING_ENTITIES)
+              .set(NAME, row.name)
+              .set(MODIFIED_BY, userId)
+              .set(MODIFIED_TIME, now)
+              .where(ID.eq(fundingEntityId))
+              .execute()
+        }
+      } catch (e: DuplicateKeyException) {
+        throw FundingEntityExistsException(row.name!!)
       }
-    } catch (e: DuplicateKeyException) {
-      throw FundingEntityExistsException(row.name!!)
+
+      removeProjectsFromEntity(fundingEntityId, removeProjects.orEmpty())
+      addProjectsToEntity(fundingEntityId, addProjects.orEmpty())
     }
   }
 
@@ -95,6 +94,32 @@ class FundingEntityService(
           .deleteFrom(FUNDING_ENTITIES)
           .where(FUNDING_ENTITIES.ID.eq(fundingEntityId))
           .execute()
+    }
+  }
+
+  private fun addProjectsToEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+    for (projectId in projects) {
+      with(FUNDING_ENTITY_PROJECTS) {
+        dslContext
+            .insertInto(FUNDING_ENTITY_PROJECTS)
+            .set(FUNDING_ENTITY_ID, fundingEntityId)
+            .set(PROJECT_ID, projectId)
+            .onConflict()
+            .doNothing()
+            .execute()
+      }
+    }
+  }
+
+  private fun removeProjectsFromEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+    for (projectId in projects) {
+      with(FUNDING_ENTITY_PROJECTS) {
+        dslContext
+            .deleteFrom(FUNDING_ENTITY_PROJECTS)
+            .where(FUNDING_ENTITY_ID.eq(fundingEntityId))
+            .and(PROJECT_ID.eq(projectId))
+            .execute()
+      }
     }
   }
 }
--- a/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
@@ -25,7 +25,7 @@ class FundingEntityService(
   private val log = perClassLogger()
 
   fun create(name: String, projects: Set<ProjectId>? = null): FundingEntityModel {
-    requirePermissions { manageFundingEntities() }
+    requirePermissions { createFundingEntity() }
 
     val userId = currentUser().userId
     val now = clock.instant()
@@ -59,7 +59,7 @@ class FundingEntityService(
   ) {
     val fundingEntityId = row.id ?: throw IllegalArgumentException("Funding Entity ID must be set")
 
-    requirePermissions { manageFundingEntities() }
+    requirePermissions { updateFundingEntities() }
 
     val userId = currentUser().userId
     val now = clock.instant()
@@ -84,20 +84,9 @@ class FundingEntityService(
     }
   }
 
-  fun deleteFundingEntity(fundingEntityId: FundingEntityId) {
-    requirePermissions { manageFundingEntities() }
+  fun addProjectsToEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+    requirePermissions { updateFundingEntityProjects() }
 
-    log.info("Deleting funding entity $fundingEntityId")
-
-    dslContext.transaction { _ ->
-      dslContext
-          .deleteFrom(FUNDING_ENTITIES)
-          .where(FUNDING_ENTITIES.ID.eq(fundingEntityId))
-          .execute()
-    }
-  }
-
-  private fun addProjectsToEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
     for (projectId in projects) {
       with(FUNDING_ENTITY_PROJECTS) {
         dslContext
@@ -111,7 +100,9 @@ class FundingEntityService(
     }
   }
 
-  private fun removeProjectsFromEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+  fun removeProjectsFromEntity(fundingEntityId: FundingEntityId, projects: Set<ProjectId>) {
+    requirePermissions { updateFundingEntityProjects() }
+
     for (projectId in projects) {
       with(FUNDING_ENTITY_PROJECTS) {
         dslContext
@@ -120,6 +111,21 @@ class FundingEntityService(
             .and(PROJECT_ID.eq(projectId))
             .execute()
       }
+    }
+  }
+
+  fun deleteFundingEntity(fundingEntityId: FundingEntityId) {
+    requirePermissions { deleteFundingEntity() }
+
+    log.info("Deleting funding entity $fundingEntityId")
+
+    dslContext.transaction { _ ->
+      dslContext
+          .deleteFrom(FUNDING_ENTITIES)
+          .where(FUNDING_ENTITIES.ID.eq(fundingEntityId))
+          .execute()
+
+      // todo delete associated Funders here
     }
   }
 }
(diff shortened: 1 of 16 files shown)


src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
```diff
@@ -0,0 +1,160 @@
...
+  fun on(event: UserDeletionStartedEvent) {
+    val fundingEntityId = fundingEntityUserStore.getFundingEntityId(event.userId)
+
+    if (fundingEntityId != null) {
+      dslContext
+          .deleteFrom(FUNDING_ENTITY_USERS)
+          .where(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID.eq(fundingEntityId))
+          .and(FUNDING_ENTITY_USERS.USER_ID.eq(event.userId))
+          .execute()
+    }
```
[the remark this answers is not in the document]
ice1080: `funding_entity_users` has a primary key of `user_id` so this comment is inaccurate, since users cannot be associated with multiple funding entities.

src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    for (projectId in projects) {
+      with(FUNDING_ENTITY_PROJECTS) {
+        dslContext
+            .deleteFrom(FUNDING_ENTITY_PROJECTS)
+            .where(FUNDING_ENTITY_ID.eq(fundingEntityId))
+            .and(PROJECT_ID.eq(projectId))
+            .execute()
+      }
+    }
```
sgrimm: This could use ``.where(PROJECT_ID.`in`(projects))`` instead of looping over the projects in the application, unless the looping is also going to be used for something else later on.

src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
```diff
@@ -0,0 +1,103 @@
...
+  @ApiResponse200
+  @ApiResponse404
+  @GetMapping("/{fundingEntityId}")
+  @Operation(summary = "Gets information about a funding entity")
+  fun getFundingEntity(
+      @PathVariable fundingEntityId: FundingEntityId,
+  ): GetFundingEntityResponsePayload {
+    val model = fundingEntityStore.fetchOneById(fundingEntityId)
+    return GetFundingEntityResponsePayload(
+        FundingEntityPayload(model),
```
sgrimm: Style nit: No need for the trailing comma here since there are unlikely to be additional arguments added in the future.

src/main/kotlin/com/terraformation/backend/funder/model/FundingEntityModel.kt
```diff
@@ -0,0 +1,27 @@
...
+import java.time.Instant
+import org.jooq.Field
+import org.jooq.Record
+
+data class FundingEntityModel(
+    val id: FundingEntityId,
+    val name: String,
+    val createdTime: Instant,
+    val modifiedTime: Instant,
+    val projects: List<ProjectId>? = emptyList(),
```
sgrimm: Does this need to be nullable? Is there a difference in meaning between null and an empty list?

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3759,6 +3769,43 @@ abstract class DatabaseBackedTest {
     }
   }
 
+  protected fun insertFundingEntity(
+      name: String = "TestFundingEntity",
```
sgrimm: Since funding entity names have to be unique, this will cause intermittent test failures if two tests run concurrently and both of them call `insertFundingEntity` with the default name. The usual approach we use for globally-unique names is to include a UUID in the default value. See `insertParticipant` for an example.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3759,6 +3769,43 @@ abstract class DatabaseBackedTest {
...
+        )
+
+    fundingEntitiesDao.insert(row)
+
+    return row.id!!.also { inserted.fundingEntitiesIds.add(it) }
+  }
+
+  protected fun insertFundingEntityProject(
+      fundingEntityId: FundingEntityId,
+      projectId: ProjectId,
```
sgrimm: These should default to `inserted.fundingEntityId` and `inserted.projectId` respectively. Then callers can do
```kotlin
insertFundingEntity()
insertProject()
insertFundingEntityProject()
```
to set up a simple scenario without having to pass in the IDs explicitly.

Same with the function below this one.

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3949,6 +3996,7 @@ abstract class DatabaseBackedTest {
     val eventIds = mutableListOf<EventId>()
     val facilityIds = mutableListOf<FacilityId>()
     val fileIds = mutableListOf<FileId>()
+    val fundingEntitiesIds = mutableListOf<FundingEntityId>()
```
sgrimm: Nit: This should be `fundingEntityIds` since it's a list of funding entity ID objects.

Also, there should be a `val` defined in this class to get the most recently inserted value so callers (and, more importantly, default values in insert methods) can reference it using `inserted.fundingEntityId`.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import org.springframework.security.access.AccessDeniedException
+
+class FundingEntityServiceTest : DatabaseTest(), RunsAsUser {
+  override val user: TerrawareUser = mockUser()
+
+  private val clock = TestClock()
+  private lateinit var fundingEntityStore: FundingEntityStore
+  private lateinit var fundingEntityUserStore: FundingEntityUserStore
+  private lateinit var fundingEntityId: FundingEntityId
+  private lateinit var service: FundingEntityService
```
sgrimm: Using `lateinit var` for this is fine and we do it in a number of tests. FYI, the idiom I've come to prefer is to use `by lazy` instead. That way the instantiation isn't in a different place in the test source than the declaration, so it reads more like a field initializer. So for example,
```kotlin
private val fundingEntityStore: FundingEntityStore by lazy { FundingEntityStore(dslContext) }
```
or more concisely using type inference,
```kotlin
private val fundingEntityStore by lazy { FundingEntityStore(dslContext) }
```

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import org.junit.jupiter.api.assertThrows
+import org.springframework.security.access.AccessDeniedException
+
+class FundingEntityServiceTest : DatabaseTest(), RunsAsUser {
+  override val user: TerrawareUser = mockUser()
+
+  private val clock = TestClock()
+  private lateinit var fundingEntityStore: FundingEntityStore
+  private lateinit var fundingEntityUserStore: FundingEntityUserStore
+  private lateinit var fundingEntityId: FundingEntityId
```
sgrimm: I think this would be slightly easier to read if the service objects were separated from the IDs so it's easier to visually skip over one or the other depending on what you're looking for. (Concretely: move `fundingEntityId` below `service` and separate the two with a blank line.)

ice1080:
* **#2887** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  @Test
+  fun `create populates created and modified fields`() {
+    val newTime = clock.instant().plusSeconds(1000)
+    clock.instant = newTime
+
+    val newUserId = insertUser()
+
+    val name = "New Funding Entity"
```
sgrimm: We'll need to be careful to not use this name in any other tests. Or throw a UUID on here to prevent name collisions with other tests running concurrently.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+    val expected =
+        row.copy(
+            createdBy = user.userId,
+            createdTime = clock.instant(),
+            id = createdModel.id,
+            modifiedBy = newUserId,
+            modifiedTime = clock.instant(),
+            name = name,
+        )
```
sgrimm: Two things here.

I'm not sure what value `row` is adding here since it's just initialized with one field and that field is overridden in the `copy()` call.  Could get rid of `row` and do `service.create(name)` instead, then turn `row.copy()` into `FundingEntitiesRow()`.

Second, this could use `FundingEntitiesRecord` instead of `FundingEntitiesRow`, at which point you can use `assertTableEquals` instead of having to explicitly fetch the actual row.

The difference between the two is that the `Record` classes are a bit more heavyweight, but part of the extra weight is that they include additional metadata such as which table they came from. We're able to use that in `assertTableEquals` to automatically query the right table.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+    val actualRow = fundingEntitiesDao.fetchOneById(createdModel.id)!!
+    val actualModel = fundingEntityStore.fetchOneById(createdModel.id)
+
+    assertEquals(expected, actualRow)
+    assertEquals(0, actualModel.projects!!.size)
+  }
+
+  @Test
+  fun `create rejects duplicates by name`() {
+    assertThrows<FundingEntityExistsException> { service.create("TestFundingEntity") }
```
sgrimm: To be concurrency-safe, this will either need to fetch the funding entity that's inserted in `setUp` (presuming that `insertFundingEntity` is changed to use randomized names) or use its own randomized name and try to create it twice so it can verify the second creation fails.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+    val projectId1 = insertProject()
+    val projectId2 = insertProject()
+    val projectSet = setOf(projectId1, projectId2)
+
+    val inserted = service.create("New Funding Entity with Projects", projectSet)
+
+    val actual = fundingEntityStore.fetchOneById(inserted.id)
+
+    assertEquals(2, actual.projects!!.size)
+    assertEquals(projectSet, actual.projects!!.toSet())
```
sgrimm: The size assertion isn't needed here; the set comparison will fail if the sizes aren't equal, and the assertion failure message will be more useful in that case since you'll be able to see which specific values were missing or added.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+    val actual = fundingEntitiesDao.fetchOneById(fundingEntityId)!!
+    assertEquals(expected, actual)
+  }
+
+  @Test
+  fun `update correctly adds new projects`() {
+    val row = FundingEntitiesRow(id = fundingEntityId, name = "TestFundingEntity")
+    val projectId1 = insertProject()
+
+    assertEquals(0, fundingEntityProjectsDao.fetchByFundingEntityId(fundingEntityId).size)
```
sgrimm: `DatabaseBackedTest` has a helper method for this: you can use `assertTableEmpty(FUNDING_ENTITY_PROJECTS)`. It will give a more informative error message on failure.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+
+    service.update(row, setOf(projectId1))
+
+    assertEquals(1, fundingEntityProjectsDao.fetchByFundingEntityId(fundingEntityId).size)
```
sgrimm: Could use `assertTableEquals` here.

Also, though this isn't as much of an issue if you use `assertTableEquals`, when I have multiple calls to the same `assert` method in a test, especially if the arguments are of the same type, I like to make sure each one has a description so it's easier to tell at a glance which one failed.

sgrimm (changes requested):
This seems generally good. I have some specific feedback, most of it related to tests.

A few suggestions like using `assertTableEquals` apply to a bunch of test methods; I didn't go through each test and write a comment on it because I thought that'd probably be kind of annoying, but hopefully I commented on enough to make the patterns clear. Happy to do more if it's helpful.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
ice1080: Won't this potentially cause concurrency issues if other tests also insert into that table?

sgrimm: No, that'll be fine. Each test runs in its own transaction that gets rolled back afterwards. So you don't see the data from other tests when you do `SELECT`.

That might seem like it contradicts what I'm saying about name collisions. The difference is that unique constraints need to be enforced globally, across transactions, even in cases where the colliding row isn't visible in the current transaction. Otherwise we'd get situations where an `INSERT` would claim to have succeeded even though it violated the constraint.

ice1080: Ahh that explanation is helpful. Thanks!

src/main/kotlin/com/terraformation/backend/funder/model/FundingEntityModel.kt
ice1080: Good question. I was initially thinking that it makes the api simpler by not requiring projects to be specified, but the default value should accomplish that anyway. Will remove the nullable

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
ice1080: It seems that to use `assertTableEquals`, the table has to be empty otherwise, so to use that I'd have to also get the record for the entity that is inserted in the `@BeforeEach` setup. Is that preferred? Seems like the same number of queries either way 🤔

ice1080: Sorry, I forgot I had everything I need to create the record for the initial entity; will get this updated

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityUserStore.kt
```diff
@@ -0,0 +1,19 @@
...
+
+@Named
+class FundingEntityUserStore(private val dslContext: DSLContext) {
+  fun getFundingEntityId(userId: UserId): FundingEntityId? {
+    return dslContext
+        .select(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID)
+        .from(FUNDING_ENTITY_USERS)
+        .where(FUNDING_ENTITY_USERS.USER_ID.eq(userId))
+        .fetchOne()
+        ?.getValue(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID)
```
sgrimm: Could do `.fetchOne(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID)`.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -0,0 +1,306 @@
...
+  fun `create populates created and modified fields`() {
+    val initialRecord =
+        FundingEntitiesRecord(
+            createdBy = currentUser().userId,
+            createdTime = clock.instant(),
+            id = fundingEntityId,
+            modifiedBy = currentUser().userId,
+            modifiedTime = clock.instant(),
+            name = fundingEntityName,
+        )
```
sgrimm: A few other options here.

1. If you make `fundingEntityName` lazy too, then you don't need `initialRecord` because this test doesn't refer to the lazy-inserted initial row so it just won't get inserted at all.
2. It seems like a decent percentage of the tests in this file don't use the initial record. Maybe it shouldn't be a test-class-level fixture at all, but rather a thing each test inserts if needed. In that case you don't need `initialRecord` either.
3. If you _do_ want the initial row to always be present, then rather than constructing its record like this, you could do something like `dslContext.selectFrom(FUNDING_ENTITIES).fetchOne()`.

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
ice1080: Good thought, I went with option 2.

PR #2894, opened 2025-03-05
SW-6615 Send Funder Invite Email from Funding Entity

ice1080:
- Added `inviteFunder` endpoint to create and invite funders to funding entities
- Added email template and notification for funder invitations

Email screenshot:
image

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
@@ -10,6 +10,7 @@ import com.terraformation.backend.customer.event.UserDeletionStartedEvent
 import com.terraformation.backend.customer.model.DeviceManagerUser
 import com.terraformation.backend.customer.model.FunderUser
 import com.terraformation.backend.customer.model.IndividualUser
+import com.terraformation.backend.customer.model.SystemUser
 import com.terraformation.backend.customer.model.TerrawareUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.KeycloakRequestFailedException
@@ -186,7 +187,7 @@ class UserStore(
    *   be folded to lower case.
    * @return null if no user has the requested email address, or if users that do have been deleted.
    */
-  fun fetchUserRowByEmail(email: String): UsersRow? {
+  fun fetchTerrawareUserByEmail(email: String): TerrawareUser? {
     val user = usersDao.fetchByEmail(email.lowercase()).firstOrNull()
 
     // Deleted users have invalid email addresses that should never match legitimate calls to this
@@ -196,7 +197,7 @@ class UserStore(
       return null
     }
 
-    return user
+    return user?.let { rowToModel(user) }
   }
 
   /**
@@ -645,8 +646,8 @@ class UserStore(
     requirePermissions { deleteUsers() }
 
     val user = fetchOneById(userId)
-    if (user !is IndividualUser || user.userType != UserType.Individual) {
-      throw AccessDeniedException("Not allowed to delete non-individual users")
+    if (user.userType != UserType.Individual && user.userType != UserType.Funder) {
+      throw AccessDeniedException("Not allowed to delete non-individual and non-funder users")
     }
 
     deleteUser(user)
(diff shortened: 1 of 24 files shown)


ice1080:
* **#2894** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2894?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#2900](https://github.com/terraware/terraware-server/pull/2900) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2900?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#2908](https://github.com/terraware/terraware-server/pull/2908) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2908?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2894?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
```diff
@@ -178,6 +179,26 @@ class UserStore(
...
   }
 
+  /**
+   * Returns the row for the user with a given email address.
+   *
+   * @param email Email address to search for. If the address has upper-case characters, they will
+   *   be folded to lower case.
+   * @return null if no user has the requested email address, or if users that do have been deleted.
+   */
+  fun fetchUserRowByEmail(email: String): UsersRow? {
```
sgrimm: Nit: This should probably be `fetchUsersRowByEmail` (plural "Users") to match the type name.

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
```diff
@@ -17,16 +17,16 @@ import org.springframework.security.core.userdetails.UserDetails
...
     val email: String,
-    val emailNotificationsEnabled: Boolean,
-    val firstName: String?,
-    val lastName: String?,
-    val countryCode: String?,
-    val cookiesConsented: Boolean?,
-    val cookiesConsentedTime: Instant?,
-    override val locale: Locale?,
-    override val timeZone: ZoneId?,
+    val emailNotificationsEnabled: Boolean? = null,
```
sgrimm: Does this need to be nullable, or could we default it to `true` instead since notifications are enabled for new users? That way we don't have to use null-safe operations every time we want to check it.

src/main/kotlin/com/terraformation/backend/email/EmailService.kt
```diff
@@ -138,9 +136,23 @@ class EmailService(
...
+  /**
+   * Sends an email notification to a specific list of emails, using the specified Locale.
+   *
+   * @param [recipients] Email addresses to send the message to.
+   * @param [model] Model object containing values that can be referenced by the template.
+   * @param [initialLocale] Optional Locale object; uses ENGLISH if unspecified.
+   */
+  fun sendLocaleEmails(
+      recipients: List<String>,
+      model: EmailTemplateModel,
```
sgrimm: Maybe flip the order of `recipients` and `model` so it matches the order of `send()`'s arguments.

src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
```diff
@@ -27,6 +33,8 @@ class FundingEntityService(
     private val dslContext: DSLContext,
     private val fundingEntityStore: FundingEntityStore,
     private val fundingEntityUserStore: FundingEntityUserStore,
+    private val userStore: UserStore,
+    private val publisher: ApplicationEventPublisher,
```
sgrimm, line 36: If these two new arguments are flipped, the list of dependencies will remain in alphabetical order.

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
sgrimm: Would it make sense to make this return `TerrawareUser?` instead? That is, call `rowToModel()` on the row before returning it? That way callers wouldn't have to do null-safe field accesses on things like the user ID, and also it'd keep the public API of this class a little more consistent in that all the methods would return instances of `TerrawareUser` or one of its subclasses.

src/test/kotlin/com/terraformation/backend/customer/db/UserStoreTest.kt
```diff
@@ -945,4 +970,19 @@ internal class UserStoreTest : DatabaseTest(), RunsAsUser {
...
+    @Test
+    fun `createFunderUser happy path`() {
+      val email = "testFunderUser@example.com"
+
+      val funderUser = userStore.createFunderUser(email)
+
+      assertNotNull(funderUser.userId)
+      assertEquals(email.lowercase(), funderUser.email)
+      assertEquals(Instant.EPOCH, funderUser.createdTime)
+      assertEquals(UserType.Funder, funderUser.userType)
```
sgrimm: A couple things here.

First, I think messages would be useful here, if only for the not-null assertion. Otherwise an assertion failure message will look something like, "Expected null, got 3" which isn't really self-explanatory; you can't tell that it's a user ID just from the message.

But second, maybe instead of these separate assertions, this would be better as a single `assertEquals()` on the entire user object. If you change `FunderUser` to be a `data class`, Kotlin will generate an `equals` method that should work for this test.

src/test/kotlin/com/terraformation/backend/funder/FundingEntityServiceTest.kt
```diff
@@ -299,4 +339,87 @@ class FundingEntityServiceTest : DatabaseTest(), RunsAsUser {
...
+    @Test
+    fun `inviteFunder creates the funder, adds to entity, and sends event`() {
+      val email = "email_${UUID.randomUUID()}@example.com"
+      val fundingEntityId = insertFundingEntity()
+
+      service.inviteFunder(fundingEntityId, email)
+
+      val userRecord = dslContext.selectFrom(USERS).where(USERS.EMAIL.eq(email)).fetchOne()!!
+
+      assertNull(userRecord.authId)
```
sgrimm: This could use an assertion message.

sgrimm (changes requested):
This looks pretty good in general; I have some small changes to suggest.

One high-level comment is that this could have been split up into a series of smaller PRs that each focused on a single thing. As I reviewed this I was having to scroll past a lot of email-related code to refer back to the user-fetching code and vice versa.

src/main/kotlin/com/terraformation/backend/customer/db/UserStore.kt
ice1080: It was mainly to avoid having to implement `rowToSystemUser` as well as avoid having to update the method anytime we add a new user type, but I'm fine if it's preferred

ice1080: Thanks. I'm renaming it now to match the return type per your later comment

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
ice1080: Good thought, will update. I wasn't sure on the defaults for all of those since they're not specified in `IndividualUser` (and are nullable there too)

sgrimm (approved):

PR #2908, opened 2025-03-11
Refactor Terraware User to contain more from Individual

ice1080:
Refactor Terraware User to contain more from Individual

Update docs


ice1080:
* **#2908** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2908?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2908?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2894** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2894?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#2900](https://github.com/terraware/terraware-server/pull/2900) <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2900?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -130,39 +108,24 @@ data class IndividualUser(
   private val _globalRoles = ResettableLazy { permissionStore.fetchGlobalRoles(userId) }
   override val globalRoles: Set<GlobalRole> by _globalRoles
 
-  /** History of permission checks performed in the current request or job. */
-  val permissionChecks: MutableList<PermissionCheck> = mutableListOf()
-
```
ice1080, line 135: Moved `permissionChecks` to bottom of file to be closer to its relevant code

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -1,10 +1,10 @@
 package com.terraformation.backend.customer.model
 
-import com.terraformation.backend.auth.CurrentUserHolder
 import com.terraformation.backend.auth.SuperAdminAuthority
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.db.PermissionStore
+import com.terraformation.backend.customer.model.TerrawareUser.Companion.log
```
sgrimm: This will cause log messages to say they come from `TerrawareUser` rather than `IndividualUser`.

But also, are we logging anything in this class any more? Maybe this isn't needed at all.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -130,39 +108,24 @@ data class IndividualUser(
...
     _globalRoles.reset()
   }
 
   val fullName: String?
-    get() = makeFullName(firstName, lastName)
-
-  override fun <T> run(func: () -> T): T {
-    return CurrentUserHolder.runAs(this, func, authorities)
-  }
+    get() = TerrawareUser.makeFullName(firstName, lastName)
```
sgrimm: Could move this to `TerrawareUser` too since I assume we'll want this for funder users.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -130,39 +108,24 @@ data class IndividualUser(
...
 
   /** Returns true if the user is an admin, owner or Terraformation Contact of any organizations. */
   override fun hasAnyAdminRole() =
       organizationRoles.values.any {
         it == Role.Owner || it == Role.Admin || it == Role.TerraformationContact
       }
 
-  override fun isAdminOrHigher(organizationId: OrganizationId?): Boolean {
-    return organizationId?.let {
-      recordPermissionCheck(RolePermissionCheck(Role.Admin, organizationId))
```
ice1080, line 157: Moved `isAdminOrHigher` to bottom to be closer to its relevant code

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -130,39 +108,24 @@ data class IndividualUser(
...
-        Role.Admin,
-        Role.Owner,
-        Role.TerraformationContact -> true
-        else -> false
-      }
-    } ?: false
-  }
+  override fun getName(): String = authId ?: throw IllegalStateException("User is unregistered")
+
+  override fun getUsername(): String = authId ?: throw IllegalStateException("User is unregistered")
```
sgrimm: These would probably make sense to move to `TerrawareUser` since they're the same for funder users and device managers (though not for the system user, so they can stay overridden in `SystemUser`.)

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
```diff
@@ -734,6 +682,13 @@ data class IndividualUser(
 
   override fun canUploadPhoto(accessionId: AccessionId) = canReadAccession(accessionId)
 
+  // When adding new permissions, put them in alphabetical order.
```
sgrimm: I had this at the bottom of the class so that people wouldn't be tempted to just put new permission methods at the bottom. I think it makes sense to either keep it there or remove it; its intended audience probably won't notice it in the middle like this.

sgrimm (changes requested):
A few minor suggestions, but looks pretty good in general.

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
ice1080: Ah ya I had intentionally moved it up to the bottom of the permissions section but you're probably right that it makes more sense at the bottom of the file; will move it up. I wonder if something like this should get added to `CONVENTIONS.md` because of the amount of places we prefer things to be alphabetical

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
ice1080: To answer your question of if we're logging anything anymore in TerrawareUser, I left the log inside of `getPassword`. 
I'll create a separate companion object inside `IndividualUser` so it doesn't say `TerrawareUser`, but as the code is now, both `IndividualUser` and `TerrawareUser` need it which is why I left it in

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
ice1080: Ya I guess the DeviceManager user will always have its `authId`. I didn't move it there originally because the implementation above throws an error if it doesn't exist instead of just returning null

src/main/kotlin/com/terraformation/backend/customer/model/IndividualUser.kt
sgrimm: Yeah, adding it to `CONVENTIONS.md` seems reasonable. I suppose once we've figured out what the rule actually is, it'd be nice to go through the code and make sure we're following it consistently. (That may be the kind of thing an AI coding assistant would be good at.)

sgrimm (approved):

PR #2911, opened 2025-03-11
SW-6621 Make UsersController general and finalizations for home page

ice1080:
- Update `UsersController` to work with multiple user types
  - not sure if there's a better way to do this in kotlin
- Update `UserStore` to be able to update any TerrawareUser type
- Update FunderUser to be able to load home page without server errors
- Update registration url for funders to match frontend

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
@@ -9,12 +9,14 @@ import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessResponsePayload
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.UserStore
+import com.terraformation.backend.customer.model.FunderUser
 import com.terraformation.backend.customer.model.IndividualUser
-import com.terraformation.backend.db.UserNotFoundException
+import com.terraformation.backend.customer.model.TerrawareUser
 import com.terraformation.backend.db.UserNotFoundForEmailException
 import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.db.default_schema.UserType
 import io.swagger.v3.oas.annotations.Operation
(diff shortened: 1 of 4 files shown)


ice1080:
* **#2911** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2911?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2911?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
```diff
@@ -64,6 +62,19 @@ class UsersController(private val clock: InstantSource, private val userStore: U
...
+      val model =
+          user.copy(
+              countryCode = payload.countryCode,
+              emailNotificationsEnabled =
+                  payload.emailNotificationsEnabled ?: user.emailNotificationsEnabled,
+              firstName = payload.firstName,
+              lastName = payload.lastName,
+              locale = payload.locale?.let { Locale.forLanguageTag(it) },
+              timeZone = payload.timeZone,
+          )
```
sgrimm, line 75: I assume this duplicate block (and the cookies-consented one) is what the "not sure if there's a better way" comment in the PR description was referring to.

I think this is fine as-is, but if you wanted to reduce duplication here, one way to do it would be to make your own "copy this object" method rather than using the auto-generated one (you'll need to call it something other than `copy`) and declare it in `TerrawareUser`.

Something like
```kotlin
interface Foo {
  val field1: String
  val field2: String
  fun duplicate(field1: String = this.field1, field2: String = this.field2): Foo
}

data class Bar(override val field1: String, override val field2: String) : Foo {
  override fun duplicate(field1: String, field2: String): Bar = Bar(field1, field2)
}
```

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/customer/api/UsersController.kt
ice1080: That makes sense, and that's what I had seen from AI as well; just wanted to confirm in case there was some other way. I'll leave it for now and potentially add that in a future PR since I'll likely be needing to do the same more in the future.

PR #2912, opened 2025-03-12
SW-6641 Create fetchAll entities endpoint

ice1080:
# Add API endpoint to list all funding entities

This PR adds a new endpoint to list all funding entities, along with their associated projects. It includes:

- A new `SimpleProjectModel` to represent projects in a lightweight way
- A `fetchAll` method in `FundingEntityStore` to retrieve all funding entities
- Refactoring of `FundingEntityModel` to include project details rather than just IDs
- A new GET endpoint in `FundingEntitiesController` to expose the list functionality

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
@@ -56,8 +56,3 @@ data class ProjectModel<ID : ProjectId?>(
 typealias NewProjectModel = ProjectModel<Nothing?>
 
 typealias ExistingProjectModel = ProjectModel<ProjectId>
-
-data class SimpleProjectModel(
-    val id: ProjectId,
-    val name: String,
-)
--- a/src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
@@ -56,3 +56,8 @@ data class ProjectModel<ID : ProjectId?>(
 typealias NewProjectModel = ProjectModel<Nothing?>
 
 typealias ExistingProjectModel = ProjectModel<ProjectId>
+
+data class SimpleProjectModel(
+    val id: ProjectId,
+    val name: String,
+)
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
@@ -15,7 +15,7 @@ import com.terraformation.backend.funder.FundingEntityService
 import com.terraformation.backend.funder.db.FundingEntityNotFoundException
 import com.terraformation.backend.funder.db.FundingEntityStore
 import com.terraformation.backend.funder.db.FundingEntityUserStore
-import com.terraformation.backend.funder.model.FundingEntityWithProjectsModel
+import com.terraformation.backend.funder.model.FundingEntityModel
 import io.swagger.v3.oas.annotations.Operation
 import jakarta.validation.Valid
 import jakarta.ws.rs.BadRequestException
@@ -121,7 +121,7 @@ data class FundingEntityPayload(
     val id: FundingEntityId,
     val name: String,
 ) {
-  constructor(model: FundingEntityWithProjectsModel) : this(id = model.id, name = model.name)
+  constructor(model: FundingEntityModel) : this(id = model.id, name = model.name)
 }
 
 data class GetFundingEntityResponsePayload(val fundingEntity: FundingEntityPayload) :
@@ -152,7 +152,7 @@ data class FundingEntityWithProjectsPayload(
     val projects: List<SimpleProjectModel>,
 ) {
   constructor(
-      model: FundingEntityWithProjectsModel
+      model: FundingEntityModel
   ) : this(id = model.id, name = model.name, projects = model.projects)
 }
 
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
@@ -6,6 +6,7 @@ import com.terraformation.backend.api.ApiResponseSimpleSuccess
 import com.terraformation.backend.api.FunderEndpoint
 import com.terraformation.backend.api.SimpleSuccessResponsePayload
 import com.terraformation.backend.api.SuccessResponsePayload
+import com.terraformation.backend.customer.model.SimpleProjectModel
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.db.funder.FundingEntityId
@@ -15,6 +16,7 @@ import com.terraformation.backend.funder.db.FundingEntityNotFoundException
 import com.terraformation.backend.funder.db.FundingEntityStore
 import com.terraformation.backend.funder.db.FundingEntityUserStore
 import com.terraformation.backend.funder.model.FundingEntityModel
+import com.terraformation.backend.funder.model.FundingEntityWithProjectsModel
 import io.swagger.v3.oas.annotations.Operation
 import jakarta.validation.Valid
 import jakarta.ws.rs.BadRequestException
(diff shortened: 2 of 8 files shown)


ice1080:
* **#2912** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2912?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
```diff
@@ -135,3 +145,16 @@ data class UpdateFundingEntityRequestPayload(
 data class InviteFundingEntityFunderRequestPayload(val email: String)
 
 data class InviteFundingEntityFunderResponsePayload(val email: String) : SuccessResponsePayload
+
+data class FundingEntityWithProjectsPayload(
+    val id: FundingEntityId,
+    val name: String,
+    val projects: List<SimpleProjectModel>,
```
sgrimm: A few thoughts here.

In general, we don't expose our internal model classes directly in the API (with the exception of enums) but instead create payload classes for them. That way we can evolve the API and the internal implementation separately if needed. So if we stay with this structure, this should define `SimpleProjectPayload` and use that instead.

However, that said, this feels like it introduces an unnecessary inconsistency in the API. It seems odd to me that fetching a single funding entity gives you back _less_ information about it than fetching the list of all entities.

Obviously it's good to send less superfluous data, but it's also good to reduce the number of one-off variants in the API so clients need less special-case code.

With that in mind, I'd suggest adding the `projects` field to `FundingEntityPayload` and making it a list of `ProjectPayload` rather than defining a new payload class for stripped-down project information. That way the client can reuse its existing `ProjectPayload`-related code here, and it can treat the elements of the "list funding entities" response the same way it treats the single element in the "get funding entity" response.

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
```diff
@@ -56,3 +56,8 @@ data class ProjectModel<ID : ProjectId?>(
 typealias NewProjectModel = ProjectModel<Nothing?>
 
 typealias ExistingProjectModel = ProjectModel<ProjectId>
+
+data class SimpleProjectModel(
```
tommylau523: Good idea to have a lighter weight representation! Although I think in this case it doesn't hurt too much to pass around a bigger object in the BE, it doesn't add a huge amount of data usage. And we could end up with a lot of these smaller models that are a subset of the fuller model and that can become a lot of extra code.

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
```diff
@@ -56,3 +56,8 @@ data class ProjectModel<ID : ProjectId?>(
...
+    val id: ProjectId,
+    val name: String,
```
tommylau523: This is likely a product question, but we often use deal names in place of project names for a lot of accelerator features. For reference, deal names are basically the internally-used project names for accelerator projects.

ice1080: Good thought, I'll check with product on this!

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
ice1080: Ya I wondered about this since there there might be no end to the list of smaller models. My main reason for not using the primary model was because we don't need organizationId which is non-nullable, but I suppose it won't hurt to pass that data to the frontend too. I'll update to using `ProjectModel` and get rid of this one 👍

src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
sgrimm: Obviously this suggestion will have ripple effects, e.g., you won't need `SimpleProjectModel` any more.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
ice1080: That makes sense and that's helpful to know about using a separate payload instead of the model. 

Regarding the inconsistency in the API, my plan was actually to eventually combine the payloads later on to support the need for the projects of a single entity in the frontend. I chose not to do that in this PR due to scope creep but if that's the preferred way for things like this I can make the update now instead.

And sounds good on the SimpleProjectModel. I actually just finished removing it locally per Tommy's suggestion as well

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
tommylau523: Passing smaller objects to the FE makes a lot of sense! I have used smaller "ControllerPayload" classes to accomplish this too. But looks like Steve has this comment covered.

src/main/kotlin/com/terraformation/backend/funder/api/FundingEntitiesController.kt
ice1080: Combined!

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
ice1080: Combined!

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityStore.kt
```diff
@@ -63,8 +64,11 @@ class FundingEntityStore(
                   groupRecords
                       .filter { it[PROJECTS.ID] != null }
                       .map { record ->
-                        SimpleProjectModel(
-                            id = record[PROJECTS.ID]!!, name = record[PROJECTS.NAME]!!)
+                        ProjectModel(
+                            id = record[PROJECTS.ID]!!,
+                            name = record[PROJECTS.NAME]!!,
+                            organizationId = record[PROJECTS.ORGANIZATION_ID]!!,
+                        )
```
sgrimm: Could use `ProjectModel.of(record)`.

src/main/kotlin/com/terraformation/backend/funder/model/FundingEntityModel.kt
```diff
@@ -11,7 +12,7 @@ data class FundingEntityModel(
     val name: String,
     val createdTime: Instant,
     val modifiedTime: Instant,
-    val projects: List<SimpleProjectModel> = emptyList(),
+    val projects: List<ProjectModel<ProjectId>> = emptyList(),
```
sgrimm: Maybe slightly clearer to use the `ExistingProjectModel` type alias instead of `ProjectModel<ProjectId>` here.

src/test/kotlin/com/terraformation/backend/funder/db/FundingEntityStoreTest.kt
```diff
@@ -47,7 +49,7 @@ class FundingEntityStoreTest : DatabaseTest(), RunsAsUser {
 
   @Test
   fun `fetchOneById retrieves entity with projects`() {
-    insertOrganization()
+    organizationId // ensures that organization exists prior to projects
```
sgrimm: An alternate way to do this would be to do `insertProject(name = "whatever", organizationId = organizationId)`.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityStore.kt
sgrimm: Scratch this -- that'll throw an exception.

ice1080: Just to confirm my understanding, that would throw an exception because my query is only requesting id, name and orgId, and the `of` method requires more data to be in the record?

src/main/kotlin/com/terraformation/backend/customer/model/ProjectModel.kt
ice1080: After discussion in round table, we decided to stick with project name for the moment until we decide and implement a different way to handle and manage deal names (as well as potentially making them required)

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityStore.kt
sgrimm: Yeah, I had it in my head that accessing a nonexistent field on a jOOQ Record would return null just like accessing a nonexistent key in a `Map` (in which case `of` would work fine) but then I double-checked and no, it throws an exception if the field doesn't exist on the record.

It'd be possible to modify `of` to do something other than a simple `record[FIELD_NAME]` but calling the constructor like you're doing here is probably better.

sgrimm (approved):

PR #2913, opened 2025-03-13
SW-6687 Added feature to update report config dates

tommylau523:


tommylau523:
* **#2913** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
```diff
@@ -206,6 +206,24 @@ class ProjectReportsController(
...
   }
 
+  @ApiResponse200
+  @ApiResponse400
+  @ApiResponse404
+  @PostMapping("/configs/{configId}")
+  @Operation(summary = "Update accelerator report configuration.")
+  fun updateAcceleratorReportConfig(
+      @PathVariable configId: ProjectReportConfigId,
+      @RequestBody payload: UpdateAcceleratorReportConfigRequestPayload,
```
ice1080, line 216: Does this need projectId in the PathVariables because of the base controller path?

tommylau523: ProjectID will not be used.

ice1080: That's what I was thinking seeing the code in the store. Do we need to handle a case where a client passes in a `projectId` that doesn't match the project for the `configId` passed in? Or do we just assume that the `configId` is the final source of truth?

tommylau523: I think we can assume the latter in this case. As far as I can tell there is not really a way to abuse this. Perhaps the downside here is not a great UX for the API but we can live with that.

ice1080 (approved):
Approving though wasn't real sure about the case that graphite mentioned of either list being empty since the break only happens if they both have at least one item that exists

tommylau523:
In theory the cases should never ever happen. But I added the extra case handling anyways to make the function complete.

tommylau523:
### Merge activity

* **Mar 13, 1:31 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2913).

PR #2914, opened 2025-03-13
Added API endpoint for reviewing system metrics and refreshing the data

tommylau523:
This PR updates the endpoints to include system metrics.

PR #2915, opened 2025-03-13
SW-6625 Support annual reporting targets

tommylau523:
This is likely a temporary change for MVP. This includes a few changes to our features. The MVP specifies that a project should only have one config for now, although this will change (and this PR will be reverted) when Annual or other cadences of reports are supported

Key Changes:
1) Updated PUT /configs endpoint to not require frequency, and instead create both a quarterly and annual report. The quarterly report will be visible and publishable. The annual report will exist only for setting targets. The expectation is that it will still be functionally possible to edit and submit an annual report.
2) Added a new endpoint to update all report config dates for a project. This is so that every project config is kept in sync. This is a temporary endpoint to enable keeping annual report and quarterly reports dates matched up. In the future, when multiple report types are allowed, this endpoint will likely be deprecated.
3) Make report frequency easily visible to the FE. This will be helpful to determine which to show as full reports, and which to show as targets only.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
@@ -186,7 +186,10 @@ class ProjectReportsController(
       @PathVariable projectId: ProjectId,
       @RequestBody payload: CreateAcceleratorReportConfigRequestPayload,
   ): SimpleSuccessResponsePayload {
-    reportStore.insertProjectReportConfig(payload.config.toModel(projectId))
+    reportStore.insertProjectReportConfig(
+        payload.config.toModel(projectId, ReportFrequency.Quarterly))
+    reportStore.insertProjectReportConfig(payload.config.toModel(projectId, ReportFrequency.Annual))
+
     return SimpleSuccessResponsePayload()
   }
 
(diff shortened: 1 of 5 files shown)


tommylau523:
* **#2915** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2914** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2914?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
```diff
@@ -203,6 +206,20 @@ class ProjectReportsController(
         configs.map { ExistingAcceleratorReportConfigPayload(it) })
   }
 
+  @ApiResponse200
+  @ApiResponse400
+  @ApiResponse404
+  @PostMapping("/configs")
+  @Operation(summary = "Update accelerator report configuration.")
```
ice1080: This summary matches the one below it even though they have slightly different parameters (projectId vs configId). Might make sense to explain why they're different, if nothing else then for devs coming back to this code after a long time

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
```diff
@@ -186,7 +186,10 @@ class ProjectReportsController(
       @PathVariable projectId: ProjectId,
       @RequestBody payload: CreateAcceleratorReportConfigRequestPayload,
   ): SimpleSuccessResponsePayload {
-    reportStore.insertProjectReportConfig(payload.config.toModel(projectId))
+    reportStore.insertProjectReportConfig(
+        payload.config.toModel(projectId, ReportFrequency.Quarterly))
+    reportStore.insertProjectReportConfig(payload.config.toModel(projectId, ReportFrequency.Annual))
+
```
ice1080, line 192: Is there any possibility that this could cause issues with the frontend when it just passes in a  `projectId` since that project will now have 2 configs associated with it? It seems like `updateProjectReportConfig` would end up modifying the start and end dates of an annual report as well as a quarterly report

src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
```diff
@@ -2201,6 +2231,50 @@ class ReportStoreTest : DatabaseTest(), RunsAsDatabaseUser {
               ),
           ))
     }
+
+    @Test
+    fun `updates all configs by projectId`() {
```
sgrimm, line 2236: I'm guessing this won't be the behavior we want once we support multiple configs. Would it perhaps be better to treat multiple configs as an error condition for now, if we're only supporting one config for MVP?

tommylau523: I was thinking that too. But I would likely put that as a database constraint, such that each project can only have one report config of each frequenxy.

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
tommylau523: This is the intended behavior. That said, once annual reports are used for more than just targets, we would remove this update by project function.

ice1080 (approved):

tommylau523:
### Merge activity

* **Mar 14, 4:28 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2915).

PR #2916, opened 2025-03-13
SW-6561 Use custom PostgreSQL 17 image for build/test

sgrimm:
Make builds and tests use a custom version of PostgreSQL 17 that has the
extensions required by Terraware and is available for both ARM and x86.
This replaces the x86-only PostgreSQL 13 image we were using.

For now, we will continue to stick with PostgreSQL 13-compatible SQL since
the production environment hasn't been upgraded to 17 yet.

The build/test configuration in `gradle.properties` uses a specific version
tag while the demo configuration in `docker/docker-compose.yml` uses a
generic `release` tag. The assumption is that the latter is used for quick
one-off runs of the server where you don't need fine control over the
exact version, whereas for builds, we want repeatability which means
we need to be specific about dependency versions.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,7 +21,7 @@ postgresJdbcVersion=42.7.5
 springDocVersion=2.8.5
 
 # For code generation, we run database migrations against an ephemeral database in a Docker
-# container using this image. If you change this, make sure you also change DatabaseTest.kt.
+# container using this image.
 
-postgresDockerRepository=postgis/postgis
-postgresDockerTag=13-3.0
+postgresDockerRepository=terraware/postgres
+postgresDockerTag=v20250227.1
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -16,7 +16,7 @@ version: "3.8"
 
 services:
   postgres:
-    image: "postgis/postgis:13-3.0"
+    image: "terraware/postgres:release"
     ports:
       - "5432:5432"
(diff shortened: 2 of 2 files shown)


sgrimm:
* **#2916** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

gradle.properties
```diff
@@ -21,7 +21,7 @@ postgresJdbcVersion=42.7.5
 springDocVersion=2.8.5
 
 # For code generation, we run database migrations against an ephemeral database in a Docker
-# container using this image. If you change this, make sure you also change DatabaseTest.kt.
+# container using this image.
 
-postgresDockerRepository=postgis/postgis
-postgresDockerTag=13-3.0
+postgresDockerRepository=terraware/postgres
+postgresDockerTag=v20250227.1
```
ice1080, line 27: I'm assuming you're fine with this tag not matching docker-compose?

ice1080 (approved):

gradle.properties
sgrimm: Yes. I'll clarify that in the PR description.

PR #2917, opened 2025-03-13
Fix NPE when fetching system user by endpoint

ice1080:
- Made email non-nullable on `TerrawareUser` interface
- Add fake emails for `SystemUser` and `DeviceManagerUser` to handle cases where they may be queried from the frontend because of `createdBy` or `modifiedBy` fields

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
@@ -10,7 +10,6 @@ import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.Role
 import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.db.default_schema.UserType
-import com.terraformation.backend.log.perClassLogger
 import java.time.ZoneId
 import java.time.ZoneOffset
 import org.springframework.security.core.GrantedAuthority
@@ -31,14 +30,17 @@ data class DeviceManagerUser(
     private val parentStore: ParentStore,
     private val permissionStore: PermissionStore,
 ) : TerrawareUser {
+  companion object {
+    private const val EMAIL = "deviceManagerUser@terraformation.com"
+  }
+
   override val timeZone: ZoneId
     get() = ZoneOffset.UTC
 
   override val userType: UserType
     get() = UserType.DeviceManager
 
-  override val email: String?
-    get() = null
+  override val email = EMAIL
 
   override val organizationRoles: Map<OrganizationId, Role> by lazy {
     mapOf(organizationId to Role.Contributor)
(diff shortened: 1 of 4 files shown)


ice1080:
* **#2917** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2917?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2917?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
```diff
@@ -31,14 +30,17 @@ data class DeviceManagerUser(
...
+
   override val timeZone: ZoneId
     get() = ZoneOffset.UTC
 
   override val userType: UserType
     get() = UserType.DeviceManager
 
-  override val email: String?
-    get() = null
+  override val email = EMAIL
```
sgrimm: Very slight nit: If you use `get() =` here, the generated bytecode will have a getter method that returns the singleton string constant. If you use an initializer like this, every instance of the class will have its own `email` field in memory, though other than using reflection tricks, there'll be no way to set it to a different value later on.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
ice1080: Ah. I had understood getters to be not storing the value in memory which is why I intentionally avoided it, but I'm guessing that's only the case when the value isn't stored in the companion object like we're doing.

sgrimm (approved):

ice1080:
### Merge activity

* **Mar 13, 7:24 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2917).

src/main/kotlin/com/terraformation/backend/customer/model/DeviceManagerUser.kt
sgrimm: That's it exactly. If you're doing something like `get() = BigDecimal(123)` then it would construct a new `BigDecimal` on each getter call.

String literals are kind of a special case, though. If you do `get() = "foo"` it won't construct a new `String` each time, but rather return the same object from the string pool each time.

sgrimm: Hmm, I didn't mean to imply this only works because `EMAIL` is a string literal. Your comment about the companion object is correct and would apply to any kind of object, not just strings.

ice1080: Right, makes sense. That's how I interpreted your message

PR #2928, opened 2025-03-18
SW-6587 Create Ask Terraware tables

sgrimm:
Create the database tables to support the experimental Ask Terraware feature,
which uses Spring AI to talk to the OpenAI chat API.

The big change here is the introduction of a table to store embeddings extracted
from uploaded documents and project details. It uses the pgvector extension to
support similarity searches on embeddings.

PR #2927, opened 2025-03-18
SW-6587 Include location in SubmissionDocumentModel

sgrimm:
Generating embeddings for uploaded documents requires that we be able to fetch
them, which means we need to know where they are on their file stores. Include
the file location in the model class for uploaded document metadata.

PR #2929, opened 2025-03-18
SW-6587 Add Spring AI libraries

sgrimm:
The experimental Ask Terraware feature will use Spring AI to talk to OpenAI's API.
Add the dependencies we'll need.

By default, Spring AI's autoconfiguration requires an OpenAI API key to be
present if the OpenAI client library is available. Override that behavior so that
developers who aren't working on Ask Terraware don't have to configure API keys.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -90,6 +90,11 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.session:spring-session-jdbc")
 
+  implementation(platform("org.springframework.ai:spring-ai-bom:1.0.0-M6"))
+  implementation("org.springframework.ai:spring-ai-openai-spring-boot-starter")
+  implementation("org.springframework.ai:spring-ai-pgvector-store-spring-boot-starter")
+  implementation("org.springframework.ai:spring-ai-tika-document-reader")
+
   implementation("ch.qos.logback.access:logback-access-tomcat:2.0.6")
   implementation("com.drewnoakes:metadata-extractor:2.19.0")
   implementation("com.dropbox.core:dropbox-core-sdk:7.0.0")
--- a/src/main/kotlin/com/terraformation/backend/ask/ConditionalOnSpringAi.kt
+++ b/src/main/kotlin/com/terraformation/backend/ask/ConditionalOnSpringAi.kt
@@ -1,0 +1,6 @@
+package com.terraformation.backend.ask
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+
+/** Annotation for beans that should only be instantiated if Spring AI is configured. */
+@ConditionalOnProperty("spring.ai.openai.api-key") annotation class ConditionalOnSpringAi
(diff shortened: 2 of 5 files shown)


sgrimm:
* **#2929** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2929?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2929?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2928** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2927** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2927?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

build.gradle.kts
```diff
@@ -90,6 +90,11 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.session:spring-session-jdbc")
 
+  implementation(platform("org.springframework.ai:spring-ai-bom:1.0.0-M6"))
+  implementation("org.springframework.ai:spring-ai-openai-spring-boot-starter")
+  implementation("org.springframework.ai:spring-ai-pgvector-store-spring-boot-starter")
+  implementation("org.springframework.ai:spring-ai-tika-document-reader")
+
```
ice1080, line 97: Would we want to make these dependencies also optional on the `openai.api-key` property?

src/main/kotlin/com/terraformation/backend/ask/ConditionalOnOpenAiKey.kt
```diff
@@ -0,0 +1,5 @@
+package com.terraformation.backend.ask
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+
+@ConditionalOnProperty("spring.ai.openai.api-key") annotation class ConditionalOnOpenAiKey
```
ice1080: Would it make sense to exclude specific vendor name (openai) in case we ever decide to change the backend LLM to a different vendor?

ice1080 (approved):

build.gradle.kts
sgrimm: The property is set at runtime, whereas these are build-time dependencies. If you're running terraware-server from a Docker image (like in the terraware-web environment) the libraries need to be in the image so that you can set the API key in the environment, or not, and get the right behavior.

ice1080: Ah interesting. Perhaps that's a difference with kotlin, but I know I've used conditional dependencies in build.gradle files before. Perhaps it only works with groovy

sgrimm: Ah, I think I see what you're getting at. Sorry it took me a moment. You mean we could add the ability to produce two different terraware-server builds, one with Spring AI and one without, and you pick which one you want with a build-time setting.

I'll admit I didn't think of that! It's doable, but off the top of my head, it seems like it'd require more than just excluding these libraries. We'd need to either make sure to skip compiling all the classes that reference Spring AI classes or conditionally configure these libraries as compile-time but not runtime dependencies. And we'd either have to publish two Docker images (so you could choose to run the AI-less one in terraware-web) or support running the AI-capable one without an API key, at which point the benefit of having two images seems marginal to me.

Or am I misunderstanding?

ice1080: I think in my head I was more thinking about the developer use case, so that it's not downloading dependencies that we're not using. But it's a low priority question since it's not really gonna hurt anyone to have dependencies they're not using

PR #2932, opened 2025-03-19
SW-6587 Ask Terraware implementation

sgrimm:
This experimental change adds integration with the OpenAI API to Terraware.

It is exposed via a page in the admin UI, `/admin/ask/`. Currently that page isn't
linked from elsewhere in the admin UI; we want to roll this feature out to a small
handful of specific internal users initially but we don't want to forbid others
from using it.

The chat UI uses the "htmx" library to do lightweight dynamic updates rather than
full page reloads.

Currently, projects must be "prepared" before you can ask the chat UI about them.
The "prepare" process generates embeddings for all the project's variable values
and uploaded documents as well as for some non-variable metadata like the project
accelerator details.

While we'll deploy this change to production for limited testing, it's still a
prototype; there are no tests and the technical documentation is incomplete.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/ask/ChatService.kt
+++ b/src/main/kotlin/com/terraformation/backend/ask/ChatService.kt
@@ -1,0 +1,126 @@
+package com.terraformation.backend.ask
+
+import com.terraformation.backend.ask.db.TerrawareChatMemory
+import com.terraformation.backend.db.default_schema.ProjectId
+import jakarta.inject.Named
+import org.springframework.ai.chat.client.ChatClient
+import org.springframework.ai.chat.client.advisor.AbstractChatMemoryAdvisor
+import org.springframework.ai.chat.client.advisor.MessageChatMemoryAdvisor
+import org.springframework.ai.chat.client.advisor.QuestionAnswerAdvisor
+import org.springframework.ai.chat.client.advisor.SimpleLoggerAdvisor
+import org.springframework.ai.vectorstore.SearchRequest
+import org.springframework.ai.vectorstore.VectorStore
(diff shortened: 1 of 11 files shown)


sgrimm:
* **#2932** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2932?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/ask/ChatService.kt
```diff
@@ -0,0 +1,126 @@
...
+                  if (projectId != null) {
+                    filterExpression(FilterExpressionBuilder().eq("projectId", projectId).build())
+                  }
+                }
+                .topK(contextItemsToInclude)
+                .build(),
+            template)
+
+    return chatClientBuilder
+        .build()
```
ice1080, line 109: Would we eventually want this to not be built for every question? Or does it need to be to get all the context for the new question?

ice1080 (approved):
Nothing seems too crazy. Just one minor question

src/main/kotlin/com/terraformation/backend/ask/ChatService.kt
sgrimm: It's possible we could share an instance, yeah. When I wrote this, I didn't want to have to worry about whether or not it's stateful in a way that would cause problems if there are multiple conversations going on concurrently. But maybe that's not an issue.

PR #2937, opened 2025-03-20
SW-6655 Delete Funders when deleting entity

ice1080:
Delete associated funders when deleting an entity.

Add permission check for deleting funders (accelerator admin).

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
@@ -110,9 +110,7 @@ class FundingEntityService(
 
     dslContext.transaction { _ ->
       val funders = userStore.fetchByFundingEntityId(fundingEntityId)
-      for (funder in funders) {
-        userStore.deleteFunderById(funder.userId)
-      }
+      funders.forEach { userStore.deleteFunderById(it.userId) }
 
       // no need to remove users from funding_entity_users table because funding_entities performs
       // cascade deletes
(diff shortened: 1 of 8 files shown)


ice1080:
* **#2937** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2937?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2937?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
```diff
@@ -109,6 +109,14 @@ class FundingEntityService(
     log.info("Deleting funding entity $fundingEntityId")
 
     dslContext.transaction { _ ->
+      val funders = userStore.fetchByFundingEntityId(fundingEntityId)
+      for (funder in funders) {
+        userStore.deleteFunderById(funder.userId)
+      }
```
sgrimm: This is totally valid, but my impression from discussions in Kotlin land is that if you're not doing break/continue, it's considered slightly more idiomatic to prefer `forEach`, like
```kotlin
funders.forEach { userStore.deleteFunderById(it) }
```
There's no performance difference, FYI, despite `forEach` taking a lambda function as its argument. It is declared `inline` and it compiles to exactly the same bytecode as `for/in`.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/funder/FundingEntityService.kt
ice1080: Makes sense; I do prefer to follow idioms whenever possible. Updated!

PR #2949, opened 2025-03-25
SW-6643 Updated workflow service to not copy unchanged workflow row

tommylau523:
This prevents a race condition from happening when a user updates the variable value as well as the variable workflow. This is common when reviewing documents/questionnaire deliverables when an accelerator admin may adjust a value but also change the status of the variable. This was causing an issue when the value update API call, takes the workflow row that is not the latest, and creates a new workflow history row with it. 

The result is something like this:
Newest:
Row 3, Old Status, New value ID (Created by the update value API call)
Row 2, New Status, Old value ID (Created by the update workflow API call)
Row 1, Old Status, Old value ID (Existing)

But we want a guarantee of this:
Row 3, New Status, New value ID

The update order doesn't actually matter that much.

Key changes:
~~1) Moved the fetch existing instead a transaction, and behind a select for update.~~
2) Change the update function to take an update function, instead of individual values. This is to simplify the call from `VariableValuesService` since it is also doing a fetch workflow there. 
3) The actual simplification part in the service class of step 2.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.documentproducer
 
+import com.terraformation.backend.customer.model.SystemUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.accelerator.tables.references.DELIVERABLE_VARIABLES
 import com.terraformation.backend.db.asNonNullable
@@ -23,9 +24,10 @@ import org.jooq.DSLContext
 @Named
 class VariableValueService(
     private val dslContext: DSLContext,
+    private val systemUser: SystemUser,
     private val variableStore: VariableStore,
     private val variableValueStore: VariableValueStore,
-    private val variableWorkflowStore: VariableWorkflowStore
+    private val variableWorkflowStore: VariableWorkflowStore,
 ) {
   /**
    * Service method to validate, write values, and update workflows.
@@ -37,7 +39,9 @@ class VariableValueService(
   fun updateValues(operations: List<ValueOperation>, triggerWorkflows: Boolean = true) {
     validate(operations)
     val values = variableValueStore.updateValues(operations, triggerWorkflows)
-    updateWorkflowHistory(values, triggerWorkflows)
+    if (triggerWorkflows) {
+      systemUser.run { updateWorkflowHistory(values) }
+    }
   }
 
   /**
@@ -46,7 +50,7 @@ class VariableValueService(
    * @param values list of updated values
    * @param updateStatus if set to true, will set status to In Review, and feedback to null
    */
-  private fun updateWorkflowHistory(values: List<ExistingValue>, updateStatus: Boolean = true) {
+  private fun updateWorkflowHistory(values: List<ExistingValue>) {
     val projectId = values.firstOrNull()?.projectId ?: return
 
     val variableIdsInDeliverables: Set<VariableId> =
@@ -62,25 +66,10 @@ class VariableValueService(
         values.filter { it.variableId in variableIdsInDeliverables }.groupBy { it.variableId }
 
     valuesByVariables.keys.forEach { variableId ->
-      variableWorkflowStore.update(
-          projectId,
-          variableId,
-      ) {
-        val status =
-            if (updateStatus) {
-              VariableWorkflowStatus.InReview
-            } else {
-              it.status
-            }
-        val feedback =
-            if (updateStatus) {
-              null
-            } else {
-              it.feedback
-            }
+      variableWorkflowStore.update(projectId, variableId) {
         it.copy(
-            status = status,
-            feedback = feedback,
+            status = VariableWorkflowStatus.InReview,
+            feedback = null,
         )
       }
     }
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
@@ -60,31 +60,29 @@ class VariableValueService(
 
     val valuesByVariables =
         values.filter { it.variableId in variableIdsInDeliverables }.groupBy { it.variableId }
-    val currentWorkflow = variableWorkflowStore.fetchCurrentForProject(projectId)
 
     valuesByVariables.keys.forEach { variableId ->
-      val existing = currentWorkflow[variableId]
-      val status =
-          if (updateStatus || existing == null) {
-            VariableWorkflowStatus.InReview
-          } else {
-            existing.status
-          }
-
-      val feedback =
-          if (updateStatus) {
-            null
-          } else {
-            existing?.feedback
-          }
-
       variableWorkflowStore.update(
           projectId,
           variableId,
-          status,
-          feedback,
-          existing?.internalComment,
-      )
+      ) {
+        val status =
+            if (updateStatus) {
+              VariableWorkflowStatus.InReview
+            } else {
+              it.status
+            }
+        val feedback =
+            if (updateStatus) {
+              null
+            } else {
+              it.feedback
+            }
+        it.copy(
+            status = status,
+            feedback = feedback,
+        )
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


tommylau523:
* **#2949** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
```diff
@@ -60,31 +60,29 @@ class VariableValueService(
```
tommylau523, line 64: <p class='graphite__hidden'><i>[Re: line +40]</i></p>
I am wondering if we could simply not call this function, if we don't have `triggerWorkflows` set to `true`. I think the original rationale was to make sure that every value update has at least one equivalent (meaning created at the same time, referencing the same `maxValueId`) workflow history row. But I can see the argument of, not having an equivalent workflow history row can simply mean the last workflow history row is still valid. 
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2949?utm_source=unchanged-line-comment">Graphite</a>.</p>

ice1080 (commented):
Seems to make sense to me, but I'll let @sgrimm do the approval since this might need more understanding of the overall workflow here

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
```diff
@@ -60,31 +60,29 @@ class VariableValueService(
...
+            } else {
+              it.status
+            }
+        val feedback =
+            if (updateStatus) {
+              null
+            } else {
+              it.feedback
+            }
+        it.copy(
```
sgrimm, line 70: Can we use a named lambda parameter here? This is a bit of a long code block to be using `it` IMO (generally `it` is used when the body of the lambda is 1-2 lines long).

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
```diff
@@ -101,35 +92,61 @@ class VariableWorkflowStore(
...
+      // Block all workflow rows regarding this project variables. This is to prevent overwriting
+      // with stale results when multiple updates are called.
+      with(VARIABLE_WORKFLOW_HISTORY) {
+        dslContext
+            .select(asterisk())
+            .from(this)
+            .where(PROJECT_ID.eq(projectId))
+            .and(VARIABLE_ID.eq(variableId))
+            .forUpdate()
+            .execute()
```
sgrimm: A few things here:

This won't lock anything on the first write since there won't be any rows to lock. Maybe that's okay, but it means the first update still has the race condition. If we want to protect the first write too, we'll need to lock something else instead.

It will acquire locks on _all_ the history rows if there are multiple of them. That's not necessarily a problem, but it's unnecessary. Locking just the first row (with `.orderBy(ID).limit(1)`) would be enough here. You'd want to lock the first row rather than the last one just because you don't want to lock the wrong row if someone else is inserting at the same time.

And finally, this is fetching the values of the history rows then throwing the data away. You do have to select _something_, but Instead of `select(asterisk())`, it'll work to do `selectOne()`.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
```diff
@@ -79,19 +80,9 @@ class VariableWorkflowStore(
...
-    requirePermissions {
-      if (status == VariableWorkflowStatus.InReview) {
-        // Non-admins editing a variable causes its status to reset to In Review.
-        updateProject(projectId)
-      } else {
-        // Other statuses can only be set by admins.
-        updateInternalVariableWorkflowDetails(projectId)
-      }
-    }
+    requirePermissions { updateInternalVariableWorkflowDetails(projectId) }
```
sgrimm, line 85: To confirm: we're no longer updating variable statuses to `In Review` when values are edited by non-admin users? If we still are, won't that fail because those users will fail this permission check?

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
sgrimm: I didn't write this code, so I'm not 100% sure if that was the original intent, but it seems reasonable to me on the face of it. I guess one thing to make sure about is that there's no possibility of ending up with no workflow history at all because we've skipped the call every time. I don't _think_ that should be possible but best to double-check.

sgrimm (changes requested):
The read-then-update logic here was definitely prone to race conditions, and locking is a reasonable approach. So I think the high-level goal of the change here is fine even if we end up skipping the call as suggested.

sgrimm:
Having now looked at the Jira task for this, I wonder if the real fix for the problem is to either (a) change the API so that two separate calls aren't needed in the first place, or (b) do the calls sequentially rather than in parallel? Unless I'm missing something, it seems like there's still a race if we're doing both calls in parallel. Even with this locking change, there's no guarantee which of the two calls will grab the lock first, right? Depending on which one wins the race, we could still end up overwriting the status.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
tommylau523: No workflow history seems to be defaulted to status = "Not Submitted" and feedback and internalComment = null. So I am not too worried about this scenario either.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
tommylau523: I think perhaps this automatic setting of status to `In Review` can be handled by the SystemUser. Since the non-admin user doesn't explicitly set the status to `In Review` either. I think when an external user calls the update workflow API, and set it to `In Review`, it should still be Forbidden

tommylau523:
I think the two calls should affect completely different tables. Because in the review screen, an accelerator admin should never trigger the workflow to automatically set the status. So one API writes to the values table, another API writes to the workflow table. 

I think by removing the copying of existing status to a new row behavior, this race condition is already eliminated. So I will opt for that, instead of the locking mechanism.

There can still be a race condition, if an accelerator admin is updating values/workflow, the same time an external user is filling out a variable in a deliverable. But that is more like a real-life race condition that we can't avoid.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
tommylau523: If we are removing the call to create a new row with the existing workflow details, I think I can remove this whole select for update part. The reason being, we are never really updating depending on the existing row, other than that one behavior. So this should simplify a lot of things.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
```diff
@@ -37,7 +39,9 @@ class VariableValueService(
   fun updateValues(operations: List<ValueOperation>, triggerWorkflows: Boolean = true) {
     validate(operations)
     val values = variableValueStore.updateValues(operations, triggerWorkflows)
-    updateWorkflowHistory(values, triggerWorkflows)
+    if (triggerWorkflows) {
+      systemUser.run { updateWorkflowHistory(values) }
```
sgrimm, line 43: This will cause `created_by` to always be the system user's user ID in `variable_workflow_history`. Maybe that's okay.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
```diff
@@ -79,19 +80,9 @@ class VariableWorkflowStore(
   fun update(
       projectId: ProjectId,
       variableId: VariableId,
-      status: VariableWorkflowStatus,
-      feedback: String?,
-      internalComment: String?,
+      updateFunc: (VariableWorkflowModel) -> VariableWorkflowModel,
```
sgrimm, line 83: Is this change to the function signature needed if the locking isn't there any more?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableWorkflowStore.kt
tommylau523: I supposed not but I do think it will simplify a read operation in some use cases, since the update is already fetching the existing.

src/main/kotlin/com/terraformation/backend/documentproducer/VariableValueService.kt
tommylau523: I think the behavior going forward will be there, the workflow is either set to `InReview` by the system, or reviewed by the accelerator user. I will make that clear to product team.

PR #2951, opened 2025-03-26
SW-6617 Added permission tests for Funder User, and list funders endpoint

tommylau523:
Key changes:
1) New permissions for `readFundingEntity` and `listFundingEntityUsers`
2) New test methods for testing funders permissions
3) New method and models for `fetchFundersForFundingEntity`. New model is to expose only a subset of user data from `FunderUser` similar to `OrganizationUserModel`.
4) Permission store to look up the Funder Entity that a `FunderUser` belongs to.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
@@ -11,6 +11,8 @@ import com.terraformation.backend.db.default_schema.tables.references.FACILITIES
 import com.terraformation.backend.db.default_schema.tables.references.ORGANIZATION_USERS
 import com.terraformation.backend.db.default_schema.tables.references.USERS
 import com.terraformation.backend.db.default_schema.tables.references.USER_GLOBAL_ROLES
+import com.terraformation.backend.db.funder.FundingEntityId
+import com.terraformation.backend.db.funder.tables.references.FUNDING_ENTITY_USERS
 import jakarta.inject.Named
 import org.jooq.DSLContext
 
@@ -38,6 +40,15 @@ class PermissionStore(private val dslContext: DSLContext) {
         .on(ORGANIZATION_USERS.ORGANIZATION_ID.eq(FACILITIES.ORGANIZATION_ID))
         .where(ORGANIZATION_USERS.USER_ID.eq(userId))
         .fetchMap(FACILITIES.ID.asNonNullable(), ORGANIZATION_USERS.ROLE_ID.asNonNullable())
+  }
+
+  /** Returns a user's funding entity */
+  fun fetchFundingEntity(userId: UserId): FundingEntityId? {
+    return dslContext
+        .select(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID)
+        .from(FUNDING_ENTITY_USERS)
+        .where(FUNDING_ENTITY_USERS.USER_ID.eq(userId))
+        .fetchOne { it[FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID] }
   }
 
   /** Returns a user's global roles. These roles are not tied to organizations. */
--- a/src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
@@ -25,8 +25,8 @@ data class FunderUser(
     override val timeZone: ZoneId? = null,
     private val permissionStore: PermissionStore,
 ) : TerrawareUser {
-  private val _fundingEntity = ResettableLazy { permissionStore.fetchFundingEntity(userId) }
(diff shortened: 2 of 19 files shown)


tommylau523:
* **#2951** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2951?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
```diff
@@ -40,6 +42,15 @@ class PermissionStore(private val dslContext: DSLContext) {
...
   }
 
+  /** Returns a user's funding entity */
+  fun fetchFundingEntity(userId: UserId): FundingEntityId? {
+    return dslContext
+        .select(FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID)
+        .from(FUNDING_ENTITY_USERS)
+        .where(FUNDING_ENTITY_USERS.USER_ID.eq(userId))
+        .fetchOne { it[FUNDING_ENTITY_USERS.FUNDING_ENTITY_ID] }
+  }
```
ice1080, line 52: I think I'm confused as to why this is in `PermissionStore`. The similar types of lazy fetches in `IndividualUser` are all related to permissions and roles so those make sense to me to be in this class. But wouldn't we want to use a different store for this? Or is it just to reduce the number of classes that need to be added to create a `FunderUser`?

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
```diff
@@ -20,7 +23,10 @@ data class FunderUser(
     override val cookiesConsentedTime: Instant? = null,
     override val locale: Locale? = null,
     override val timeZone: ZoneId? = null,
+    private val permissionStore: PermissionStore,
 ) : TerrawareUser {
+  private val _fundingEntity = ResettableLazy { permissionStore.fetchFundingEntity(userId) }
+  override val fundingEntity: FundingEntityId? by _fundingEntity
```
ice1080: Should these be named `fundingEntityId` to avoid confusion?

src/test/kotlin/com/terraformation/backend/funder/db/FundingEntityStoreTest.kt
```diff
@@ -17,23 +17,25 @@ import org.junit.jupiter.api.assertThrows
...
 class FundingEntityStoreTest : DatabaseTest(), RunsAsUser {
-  override val user: TerrawareUser = mockUser()
+  override val user = mockUser()
 
   private val store by lazy { FundingEntityStore(dslContext) }
 
-  private val fundingEntityId by lazy { insertFundingEntity() }
-  private val organizationId by lazy { insertOrganization() }
+  private lateinit var fundingEntityId: FundingEntityId
+  private lateinit var organizationId: OrganizationId
```
ice1080: Perhaps you and @sgrimm can talk through this one. He told me that he prefers `by lazy` over `lateinit var` so I've been using the former

src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
tommylau523: Permission store is only used for checking permission purposes. I think `FunderUser` is similar to `IndividualUser` for that.

src/main/kotlin/com/terraformation/backend/customer/model/FunderUser.kt
tommylau523: Good idea!

src/test/kotlin/com/terraformation/backend/funder/db/FundingEntityStoreTest.kt
tommylau523: Let me make the change.

tommylau523: The lazy funding entity insertion was a bit troublesome for me. The test results end up depending on whether the `fundingEntityId` field is accessed or not. I think I was trying to get around this issue by making it a `lateinit var` and making it initialized in `@BeforeEach`.

To ensure consistency, I moved the funding entity creation under each test method. This will guarantee that it gets inserted every test method.

src/main/kotlin/com/terraformation/backend/customer/db/PermissionStore.kt
ice1080: Ah that makes sense. Because of the method name, I was thinking you were using this to retrieve the full Funding Entity for the user, but I should have looked closer to see you're just getting the id for permissions purposes

ice1080 (approved):

PR #2955, opened 2025-03-26
SW-6613 Add deal name to entity projects

ice1080:


ice1080:
* **#2955** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2955?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2955?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityStore.kt
```diff
@@ -64,10 +65,12 @@ class FundingEntityStore(
...
-                            name = record[PROJECTS.NAME]!!,
-                            organizationId = record[PROJECTS.ORGANIZATION_ID]!!,
+                        val projectId = record[PROJECTS.ID]!!
+                        val variableValues =
+                            acceleratorProjectVariableValuesService.fetchValues(projectId)
+                        ProjectAcceleratorDetailsModel(
+                            projectId = projectId,
+                            dealName = variableValues.dealName,
                         )
                       })
```
ice1080, line 75: I don't love fetching values inside the map function, so I'm open to suggestions here. 
In case context changes anything here, most funding entities will only have one project (though each project is possible and likely to be associated with multiple entities), and entities that have more than one project will still likely have a small number.

sgrimm: I think the other problem with this is that it's fetching a bunch of variable values unnecessarily. It's definitely possible to fold "get the latest value of the 'deal name' variable for this project" into the query, perhaps in the form of a subquery. It'd break encapsulation a bit but we're already joining with non-funding-entity-related tables so maybe that's not too bad.

That said, we're still planning to move/copy the deal name to the project accelerator details table, right? If so, then this change will be short-lived (I guess it'd never even be executed by actual end users) and presumably we'd instead join with that table in the query here to get the value.

src/main/kotlin/com/terraformation/backend/funder/db/FundingEntityStore.kt
```diff
@@ -39,8 +41,7 @@ class FundingEntityStore(
                 FUNDING_ENTITIES.CREATED_TIME,
                 FUNDING_ENTITIES.MODIFIED_TIME,
                 PROJECTS.ID,
-                PROJECTS.NAME,
-                PROJECTS.ORGANIZATION_ID)
+            )
```
sgrimm, line 44: Do we still need the join with `projects` here? The ID can come from `funding_entity_projects`.

This query is using the project name in its `ORDER BY` but that seems wrong since the project name could sort differently than the deal name. If we want to sort by the user-facing name, then we'll either need to have a subquery or a CTE to fetch the deal name as part of the query so we can sort on it, or we'll need to sort in the application code instead of in SQL. Either way, the join with `projects` isn't needed.

ice1080: Ah thanks for catching, I had meant to remove the PROJECTS join, and had forgotten that I was ordering by deal name. I might have to update the FE to order by deal name, though if what Tommy is talking about will actually happen then this PR might be deleted anyway

ice1080:
Closing in favor of new `dealName` column in `project_accelerator_details`

PR #2956, opened 2025-03-27
Added deal name column and update the column with variable values

tommylau523:


tommylau523:
* **#2957** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2957?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2956** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2956?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2956?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/AcceleratorVariablesUpdater.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val variableStore: VariableStore,
+) {
+
+  @EventListener
+  fun on(event: VariableValueUpdatedEvent) {
+    systemUser.run {
+      val variable = variableStore.fetchOneVariable(event.variableId)
+      val values = acceleratorProjectVariableValuesService.fetchValues(event.projectId)
+
+      if (StableId(variable.stableId) == StableIds.country) {
```
sgrimm: May as well avoid creating the extra wrapper object here since we're comparing a string to a wrapped string anyway. That is, this could be `if (variable.stableId == StableIds.country.value)`.

Or, perhaps better, we should change `BaseVariableProperties.stableId` from `String` to `StableId` so it's type-safe like our numeric IDs. I realize that's out of scope for this PR so we could do it as a followup if we decide it's what we want.

src/main/kotlin/com/terraformation/backend/accelerator/variables/AcceleratorProjectVariableValuesService.kt
```diff
@@ -55,10 +55,7 @@ class AcceleratorProjectVariableValuesService(
 
   private val variablesById: Map<VariableId, Variable> by lazy {
     projectAcceleratorVariablesStableIds
-        .map {
-          variableStore.fetchByStableId(it.value)
-              ?: throw IllegalStateException("No variable with stable ID ${it.value}")
-        }
+        .mapNotNull { variableStore.fetchByStableId(it.value) }
```
sgrimm: Curious what motivated this change. The original version was designed to fail fast if one of the expected variables wasn't defined. With this change, you can no longer be certain that `variablesById` has an ID for every one of the hardwired stable IDs.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/AcceleratorVariablesUpdater.kt
tommylau523: I think converting the `BaseVariableProperties.stableId` route. Let me go with that.

src/main/kotlin/com/terraformation/backend/accelerator/variables/AcceleratorProjectVariableValuesService.kt
tommylau523: I am thinking whether we want our software to fail, when the variable list is not updated. 

I think returning `null` and logging the problem is likely a better solution for now.

tommylau523: But I am also happy to revert this if failing loudly is the preferred approach.

tommylau523:
@sgrimm with the `StableId` refactored into `BaseVariableProperties`, this is turning out to touch a lot more lines. Although most are one line changes.

sgrimm (approved):
Ah, sorry to make you turn this into a massive PR. What I'd had in mind with my comment about the wrappers was that that would be a separate change we could do later.

tommylau523:
No worries at all! Thank you for reviewing!

PR #2958, opened 2025-03-28
Added quarter to reports

tommylau523:
For reports, the client ends up having to determine the quarter of the report by month. 

This solves the problem by explicitly saving the report quarter, to avoid this repeated calculations later. This adds the bonus benefit of checking that a quarterly report dates are valid.


tommylau523:
* **#2958** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2958?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2958?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0350/V361__ReportQuarters.sql
```diff
@@ -0,0 +1,48 @@
...
+    name TEXT NOT NULL UNIQUE
+);
+
+-- Also in R__TypeCodes.sql
+INSERT INTO accelerator.report_quarters (id, name)
+VALUES (1, 'Q1'),
+       (2, 'Q2'),
+       (3, 'Q3'),
+       (4, 'Q4')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
```
ice1080: Will this need the annual reports in here too?

src/main/resources/db/migration/0350/V361__ReportQuarters.sql
```diff
@@ -0,0 +1,48 @@
...
+    FROM accelerator.project_report_configs AS c
+    WHERE c.id = config_id
+    LIMIT 1
+);
+
+-- Add foreign key to check that config/frequency pair matches up
+ALTER TABLE accelerator.reports
+    ALTER COLUMN report_frequency_id SET NOT NULL,
+    ADD FOREIGN KEY (config_id, report_frequency_id)
+        REFERENCES accelerator.project_report_configs(id, report_frequency_id);
```
ice1080: Does this mean that a given report's frequency is now stored in both `accelerator.reports` and in that report's `project_report_configs` row?

sgrimm: My assumption, but maybe I'm wrong, is that the "frequency" here has a slightly different meaning. Like, the config says we generate a report once a quarter, but the frequency here says, "This is a quarterly report."

If I'm right, you'd expect a project to have multiple configs for different frequencies (one for annual, one for quarterly) but each individual report would only have one frequency to say what kind of report it is.

That said, it took me a minute to figure out what this probably meant. An explanation in `R__Comments.sql` would be helpful.

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -440,6 +440,13 @@ VALUES (1, 'Achieved'),
        (3, 'Unlikely')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
+INSERT INTO accelerator.report_quarters (id, name)
+VALUES (1, 'Q1'),
+       (2, 'Q2'),
+       (3, 'Q3'),
+       (4, 'Q4')
+ON CONFLICT (id) DO UPDATE SET name = excluded.name;
```
sgrimm, line 449: My first reaction to this was, "Can't we use an integer that's constrained from 1 to 4?" but on reflection, I think this is a nice bit of added type safety.

src/main/resources/db/migration/0350/V361__ReportQuarters.sql
```diff
@@ -0,0 +1,48 @@
...
+    WHERE report_frequency_id = 1;
+
+-- Add constraint that report quarters must line up with dates, and present for quarterly reports
+-- This also has the benefit of checking that both start and end date are in the same quarter
+ALTER TABLE accelerator.reports
+    ADD CONSTRAINT quarterly_report_has_quarter
+    CHECK ((report_frequency_id = 1
+                AND report_quarter_id IS NOT NULL
+                AND report_quarter_id = EXTRACT(QUARTER FROM start_date)
+                AND report_quarter_id = EXTRACT(QUARTER FROM end_date)) OR
```
sgrimm: This means the numeric IDs for the `report_quarters` table have meanings, which isn't how most of our enums work. Which is fine, but it might be worth a brief comment in `R__TypeCodes.sql` or `R__Comments.sql`.

sgrimm (approved):

src/main/resources/db/migration/0350/V361__ReportQuarters.sql
ice1080: Ah got it, I hadn't quite understood that `project_report_configs` is tied to the project instead of each report.

tommylau523: I think an easier way to think about it is, `frequency` is just a report type, that also happens to determine the quantity and how often the reports are. But I will make a comment.

tommylau523: In the future, `frequency` will be used to determine metrics/data inclusion too when annual reports get introduced with a different set of data.

tommylau523:
### Merge activity

* **Mar 28, 2:29 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2958).

PR #2971, opened 2025-03-31
Updated refresh system metrics to set override value to null, and fill columns only when reports are submitted

tommylau523:
This is to change the function of resetting system metrics based on requirements.

The new requirement specifies that overwrites are always erased when refreshed. And this functionality will be used for when reports are not submitted as well.

Based on this requirement, the function now supports both behaviors, but correctly leaves the columns empty if a report has not yet been submitted.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
@@ -165,21 +165,12 @@ class ReportStore(
   fun refreshSystemMetricValues(reportId: ReportId, metrics: Collection<SystemMetric>) {
     requirePermissions { reviewReports() }
 
-    val report =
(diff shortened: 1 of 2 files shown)


tommylau523:
* **#2971** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2971?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2971?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm (approved):
The code change looks safe enough, though I have no idea why we need it; an explanation in the PR description or a link to a Jira task in the title (if this is fixing a user-reported bug) would be helpful for context.

PR #2973, opened 2025-03-31
Added generated enum converter

tommylau523:
Added a converted that will use our existing JSON serializer to correctly convert request params into our Enum. Methods as follow:

1) Added a Spring converter/formatter that targets  ~~`ParameterizedEnums::java.class`~~ `EnumFromReferenceTable.class`, so when a string such as "Seeds Collected" is passed into the API, with the expected target type of one of the enums implementing the enums, it calls our `ObjectMapper` to resolve it. 
2) Register the `EnumFromReferenceTableConverter` to our `SpringMvcConfig`


tommylau523:
* **#2973** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2973?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2973?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/api/ParameterizedEnumConverter.kt
```diff
@@ -0,0 +1,38 @@
...
+      sourceType: TypeDescriptor,
+      targetType: TypeDescriptor
+  ): Any? {
+    if (source == null) {
+      return null
+    }
+    try {
+      return objectMapper.readValue("\"" + source + "\"", targetType.getType())
+    } catch (e: IOException) {
+      return null
```
tommylau523: This will still correctly throws `MethodArgumentTypeMismatchException`, although it may not do so for a nullable type. I am currently testing this out.

src/main/kotlin/com/terraformation/backend/api/ParameterizedEnumConverter.kt
```diff
@@ -0,0 +1,38 @@
...
+  override fun convert(
+      @Nullable source: Any?,
+      sourceType: TypeDescriptor,
+      targetType: TypeDescriptor
+  ): Any? {
+    if (source == null) {
+      return null
+    }
+    try {
+      return objectMapper.readValue("\"" + source + "\"", targetType.getType())
```
sgrimm: `objectMapper.convertValue()` will do this without needing to wrap the value in double quotes.

src/main/kotlin/com/terraformation/backend/api/ParameterizedEnumConverter.kt
tommylau523: Looks like for a nullable type, an incorrect value will still result in 400 Bad Request! Resolving this now!

src/main/kotlin/com/terraformation/backend/api/ParameterizedEnumConverter.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.core.convert.TypeDescriptor
+import org.springframework.core.convert.converter.GenericConverter
+import org.springframework.lang.Nullable
+import org.springframework.stereotype.Component
+
+/**
+ * Customizes how web request parameters are converted to an [ParameterizedEnum], matching our JSON
+ * serializer.
+ */
+@Component
```
sgrimm: We normally use `@Named` since it's not Spring-specific.

src/main/kotlin/com/terraformation/backend/api/ParameterizedEnumConverter.kt
```diff
@@ -0,0 +1,38 @@
...
+class ParameterizedEnumConverter(private val objectMapper: ObjectMapper) : GenericConverter {
+
+  // Specify the conversion pair: String to Enum that implements MyEnumInterface
+  override fun getConvertibleTypes(): Set<GenericConverter.ConvertiblePair> {
+    return setOf(
+        GenericConverter.ConvertiblePair(String::class.java, ParameterizedEnum::class.java))
+  }
+
+  override fun convert(
+      @Nullable source: Any?,
```
sgrimm: Kotlin code doesn't need nullability annotations like this.

src/main/kotlin/com/terraformation/backend/api/SpringMvcConfig.kt
```diff
@@ -27,4 +31,11 @@ class SpringMvcConfig(private val globalRoleInterceptor: GlobalRoleInterceptor)
   override fun addInterceptors(registry: InterceptorRegistry) {
     registry.addInterceptor(globalRoleInterceptor)
   }
+
+  // Register the enum converter
+  override fun addFormatters(registry: org.springframework.format.FormatterRegistry) {
```
sgrimm: Why the fully-qualified class name here instead of an import?

src/main/kotlin/com/terraformation/backend/db/ParameterizedEnum.kt
```diff
@@ -0,0 +1,5 @@
+package com.terraformation.backend.db
+
+interface ParameterizedEnum {
```
sgrimm: A comment explaining what this interface is for would be helpful.

jooq/src/main/kotlin/com/terraformation/backend/jooq/TerrawareGenerator.kt
```diff
@@ -178,7 +181,7 @@ class TerrawareGenerator : KotlinGenerator() {
         """
       enum class $enumName(
           $properties
-      ) : EnumFromReferenceTable<$idType, $enumName>$localizableExtendsSnippet {
+      ) : EnumFromReferenceTable<$idType, $enumName>$localizableExtendsSnippet$parameterizedExtendSnippet {
```
sgrimm: I'm not sure the variable is adding anything here since its value is a constant and we're already referencing an interface of ours directly here anyway. This could be something like
```
) : EnumFromReferenceTable<$idType, $enumName>, ParameterizedEnum$localizableExtendsSnippet
```
Alternately, make `EnumFromReferenceTable` extend the new interface and then it doesn't need to be explicitly listed here at all.

sgrimm (changes requested):
The name "parameterized" doesn't seem super clear to me; there's no parameter. Maybe something like `JsonValuedEnum` instead?

src/main/kotlin/com/terraformation/backend/db/ParameterizedEnum.kt
sgrimm: Thinking about this a bit more, I'm not convinced about adding a new interface instead of using the existing `EnumFromReferenceTable` interface for this.

If at some point we want this kind of parsing for something other than our jOOQ-generated enums, it'd be trivial to introduce a new interface at that point, but if we don't ever want that, then having a separate interface would just make our inheritance hierarchy more complex.

sgrimm (approved):
Ah, that makes it all much simpler. Thanks for updating.

PR #2981, opened 2025-04-02
Fix intermittent PlantingZoneModelTest failures

sgrimm:
The assertion that was checking whether a monitoring plot was inside a specific
part of a planting zone map was too strict. We allow plots to fall _slightly_
outside the map boundaries to account for inaccuracy in the geometry calculations.

The test was failing if 0.001% or more of the plot was outside the requested
area, which was stricter than the fuzz factor we allow when searching for plot
locations. Increase the tolerance to 0.1%.

Also change the assertion to produce a more informative error message.


sgrimm:
* **#2981** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2981?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2981?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #2985, opened 2025-04-03
SW-6592 Implemented report submitted notifications

tommylau523:


tommylau523:
* **#2985** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2985?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -493,6 +498,43 @@ class AppNotificationService(
...
+          }
+
+      val reportYear = report.endDate.year
+      val reportQuarter = report.quarter?.name ?: "Quarterly"
+
+      val reportPrefix =
+          when (report.frequency) {
+            ReportFrequency.Quarterly -> "$reportYear $reportQuarter"
+            ReportFrequency.Annual -> "$reportYear Annual"
+          }
```
ice1080: Might be worth moving this logic to `ReportModel` so it can be reused

src/main/resources/i18n/Messages_en.properties
```diff
@@ -124,6 +124,15 @@ monitoringPlotNorthwestCorner={0} Northwest Corner
...
 # {0} is plot name
 monitoringPlotSouthwestCorner={0} Southwest Corner
+# {0} is the report prefix, {1} is Project Deal Name
+notification.acceleratorReport.submitted.app.title={0} Report Submitted for {1}
+# {0} is Project Deal Name, {1} is the report prefix
+notification.acceleratorReport.submitted.app.body={0} has submitted their {1} Report.
+# {0} is Project Deal Name, {1} is the report prefix
+notification.acceleratorReport.submitted.email.body={0} has submitted their {1} Report.
+notification.acceleratorReport.submitted.email.buttonLabel=View Submitted Report
+# {0} is the report prefix, {1} is Project Deal Name
```
ice1080: Not sure if we care much about accuracy of comments, but these will only be the Project Deal Name if it exists, else the Project Name

src/main/kotlin/com/terraformation/backend/accelerator/ReportNotifier.kt
```diff
@@ -0,0 +1,19 @@
...
+@Named
+class ReportNotifier(
+    private val rateLimitedEventPublisher: RateLimitedEventPublisher,
+) {
+  /** Schedules a "ready for review" notification when a report is submitted. */
+  @EventListener
+  fun on(event: AcceleratorReportSubmittedEvent) {
+    rateLimitedEventPublisher.publishEvent(
+        AcceleratorReportReadyForReviewEvent(event.reportId, event.projectId))
+  }
```
ice1080: Wouldn't we want to notify that a report is ready for review every time it gets submitted? I thought that a participant couldn't resubmit once they've submitted, so the only time we'd be receiving this event again is if a console user sent it back for changes, and the user made the changes and resubmitted. In that situation, wouldn't we want console users to receive the notification again so it doesn't get missed?

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
```diff
@@ -493,6 +498,43 @@ class AppNotificationService(
...
+
+      val renderMessage = {
+        messages.acceleratorReportSubmitted(
+            projectDealName = report.projectDealName ?: project.name, reportPrefix = reportPrefix)
+      }
+
+      insertAcceleratorNotification(
+          webAppUrls.acceleratorConsoleReport(event.reportId, event.projectId),
+          NotificationType.AcceleratorReportSubmitted,
+          project.organizationId,
```
ice1080, line 523: Should this notification type be changed to `AcceleratorReportReadyForReview`?

src/main/kotlin/com/terraformation/backend/email/model/Models.kt
```diff
@@ -496,3 +496,13 @@ class FunderAddedToFundingEntity(
   override val templateDir: String
     get() = "funder/addedToFundingEntity"
 }
+
+class AcceleratorReportSubmitted(
+    config: TerrawareServerConfig,
+    val projectDealName: String,
+    val reportPrefix: String,
```
ice1080, line 503: Could this property be confusing since it could actually be the project name?

src/main/resources/i18n/Messages_en.properties
sgrimm: I'd probably say "project name" here. These comments are written for the (mostly freelance) translators that Phrase sends our strings to. A random translator isn't going to know what a "deal name" is.

src/main/kotlin/com/terraformation/backend/accelerator/ReportNotifier.kt
tommylau523: The state would go from
Not Submitted -> Submitted -> Needs Update -> Submitted.

So it would get notified twice. But I think the rate-limit may combine the two notifications within 5 minutes.

ice1080: Right, I think I'm trying to say that a report probably needs to notify every time that it's been submitted so that console users know when to review it? E.g. if they got the notification after the first Submitted and requested updates, and the participants made the updates and submitted again, then console users wouldn't receive a notification that the updates had been made if that happened fast enough. I'm asking the question of if that's preferred behavior, or if a notification should still happen in that situation

tommylau523: The requirements are not clear on that so I will keep the rate-limits now. I will ask for clarifications. 

My assumption, a re-submit can only happen this quickly, if an accelerator admin is working closely with the project, maybe in a meeting.

src/main/kotlin/com/terraformation/backend/email/model/Models.kt
tommylau523: I think it is okay. We really don't want to use `projectName` since it is changeable by the external user. Hopefully we have a way to ensure that an internal deal name exists for a project in accelerator either by data schema or programmatically soon.

src/main/kotlin/com/terraformation/backend/customer/AppNotificationService.kt
tommylau523: The PRD describes the notification as a Submitted notification. I am going to change all the event types to match instead

src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
```diff
@@ -56,18 +58,25 @@ data class ReportModel(
...
+      }
+    }
 
   fun isEditable(): Boolean {
     return status == ReportStatus.NotSubmitted || status == ReportStatus.NeedsUpdate
   }
 
   fun validateForSubmission() {
-    if (!isSubmittable()) {
+    if (!isEditable()) {
```
tommylau523, line 79: Editable and Submittable should be the same so this is a non-functional change.

ice1080 (approved):

PR #2987, opened 2025-04-03
SW-6722 Added logframe and M&E URL to project accelerator details table 

tommylau523:


tommylau523:
* **#2987** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2987?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2987?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
```diff
@@ -269,6 +270,19 @@ class ProjectReportsController(
...
+  @ApiResponse404
+  @PostMapping("/logframe")
+  @Operation(summary = "Update project logframe URL.")
+  fun updateLogframeUrl(
+      @PathVariable projectId: ProjectId,
+      @RequestBody payload: UpdateProjectLogframeUrlRequestPayload,
+  ): SimpleSuccessResponsePayload {
+    reportStore.updateProjectLogframeUrl(projectId, payload.logframeUrl)
+    return SimpleSuccessResponsePayload()
+  }
```
ice1080, line 284: If this isn't specific to reports, should this instead go in `ProjectAcceleratorDetailsController`

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
ice1080, line 1: Same question here, should this be in a different store since the methods don't reference reports

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
tommylau523: This is tricky because the only place FE references this is the accelerator reports console. But in our table, it makes the most sense to put it with project accelerator details. 

I would say controller should definitely stay here. We could definitely make the argument to move the method to `ProjectAcceleratorDetailsStore` instead, but I also like that it is consistent with the Controller here.

ice1080: Approving, but is there no chance that a logframe URL would get added to the Edit Project page?

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
tommylau523: Not at the moment.

tommylau523:
### Merge activity

* **Apr 3, 6:18 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/2987).

PR #2997, opened 2025-04-07
SW-6755 Added funder report tables

tommylau523:
These tables are mostly just copied over from the accelerator reports, with some minor changes:

1) Metric `progress_notes` are removed since they are internal only.
2) Report config details are removed
3) Report submission time is removed
4) For system metrics, only one `value` is shown, instead of having both system and override values. Only one gets published, with priority given to `overriden_value` first.

Some intended use cases of these columns:
1) `modified_time` can be used to check when the report was last published, if the accelerator report equivalent has a later `modified_time`, we can use that to determine that data may have been updated since it was last published. However, this does not account for whether the changes are external facing yet.
2) The same `report_id` gets reused and referenced. Since accelerator reports and published reports are strictly 1:1 for now, we don't need a separate ID.

### DIFF
--- a/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
+++ b/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
@@ -50,7 +50,7 @@ val ENUM_TABLES =
                         "report_standard_metrics\\.status_id",
                         "report_system_metrics\\.status_id",
                     )),
-                EnumTable("report_quarters", listOf("reports\\.report_quarter_id")),
+                EnumTable("report_quarters"),
                 EnumTable("report_statuses", listOf("reports\\.status_id")),
                 EnumTable("score_categories", isLocalizable = false),
                 EnumTable("submission_statuses"),
@@ -252,7 +252,11 @@ val ID_WRAPPERS =
                     listOf("project_report_configs\\.id", ".*\\.config_id")),
                 IdWrapper(
                     "ReportId",
-                    listOf("accelerator\\.reports\\.id", "accelerator\\..*\\.report_id")),
+                    listOf(
+                        "accelerator\\.reports\\.id",
+                        "accelerator\\..*\\.report_id",
+                        "funder\\.funder_reports\\.id",
+                        "funder\\..*\\.report_id")),
                 IdWrapper(
                     "StandardMetricId", listOf("standard_metrics\\.id", ".*\\.standard_metric_id")),
                 IdWrapper("SubmissionDocumentId", listOf("submission_documents\\.id")),
--- a/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
+++ b/jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
@@ -255,7 +255,6 @@ val ID_WRAPPERS =
                     listOf(
                         "accelerator\\.reports\\.id",
                         "accelerator\\..*\\.report_id",
-                        "funder\\.funder_reports\\.id",
                         "funder\\..*\\.report_id")),
                 IdWrapper(
                     "StandardMetricId", listOf("standard_metrics\\.id", ".*\\.standard_metric_id")),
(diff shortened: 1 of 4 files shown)


tommylau523:
* **#2997** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2997?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/2997?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0350/V367__FunderReports.sql
ice1080, line 1: And you feel like this approach is better than some "publish_to_funders" flag? My main concern is that copying data has the danger of getting stale if something needed to get updated historically (this might just be a non concern).

tommylau523: One of the requirements is to have publish version "lag" behind until the publish button is clicked. So we will need to maintain controlled "staleness" anyways.

src/main/resources/db/migration/0350/V367__FunderReports.sql
```diff
@@ -0,0 +1,59 @@
+CREATE TABLE funder.funder_reports (
+    report_id BIGINT PRIMARY KEY REFERENCES accelerator.reports,
```
sgrimm: This doesn't have `ON DELETE CASCADE` like the `report_id` columns in the other tables here do. Intentional?

src/main/resources/db/migration/0350/V367__FunderReports.sql
```diff
@@ -0,0 +1,59 @@
...
+    project_id BIGINT NOT NULL REFERENCES projects ON DELETE CASCADE,
+    report_frequency_id INTEGER NOT NULL REFERENCES accelerator.report_frequencies,
+    report_quarter_id INTEGER REFERENCES accelerator.report_quarters,
+    start_date DATE NOT NULL,
+    end_date DATE NOT NULL CHECK (end_date > start_date),
+    highlights TEXT,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL
```
sgrimm: Would it make sense to change these to `published_by` and `published_time` to make the semantics clearer?

src/main/resources/db/migration/0350/V367__FunderReports.sql
```diff
@@ -0,0 +1,59 @@
+CREATE TABLE funder.funder_reports (
```
sgrimm: Maybe `published_reports` would make the intent of these tables clearer, since that's the terminology we're using in product discussions.

src/main/resources/db/migration/0350/V367__FunderReports.sql
tommylau523: I think so. I don't know if we want published reports to be deletable.

tommylau523: But we could programmatically maintain that first. I will add this constraint in so our DB is easier to manage.

ice1080 (commented):
Things look good to me; I like the updated table names too. 
Looks like build is failing

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -257,7 +257,11 @@ val ID_WRAPPERS =
                     listOf("project_report_configs\\.id", ".*\\.config_id")),
                 IdWrapper(
                     "ReportId",
-                    listOf("accelerator\\.reports\\.id", "accelerator\\..*\\.report_id")),
+                    listOf(
+                        "accelerator\\.reports\\.id",
+                        "accelerator\\..*\\.report_id",
+                        "funder\\.funder_reports\\.id",
```
sgrimm: This line doesn't seem necessary any more (plus it refers to the old table name).

sgrimm (approved):

PR #3007, opened 2025-04-10

terraware-phrase:


src/main/resources/i18n/Messages_en.properties
```diff
@@ -319,6 +328,11 @@ notification.plantingSeason.started.email.subject=It''s planting season\!
 notification.registration.email.buttonLabel=Create Terraware Account
 notification.registration.organization.email.buttonIntro=Please click the button below to create your Terraware account where you can view the organization.
 notification.registration.organization.email.linkIntro=Please click the link below to create your Terraware account where you can view the organization.
+notification.report.created.app.body=Your {0}-Q{1} Report is ready to be completed and submitted to Terraformation.
```
tommylau523, line 331: @alexTerraformation these need to be removed from phrase. All `notification.report.*`

sgrimm (approved):

src/main/resources/i18n/Messages_en.properties
sgrimm: @alexTerraformation FYI, I'm merging this because it also includes strings for biomass CSV export, but I haven't touched the strings @tommylau523 mentioned.

PR #3011, opened 2025-04-11
SW-6694 Convert simple planting sites to detailed

sgrimm:
Add a manual migration to the admin UI to add zones and subzones to planting
sites that don't have them. This will allow users to schedule observations of
sites they created as simple planting sites.

There is still a way to create simple planting sites in the web app; we'll need
to run the migration again once that's removed, at which point the migration can
go away.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminPlantingSitesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminPlantingSitesController.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.module.kotlin.readValue
 import com.terraformation.backend.api.RequireGlobalRole
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.OrganizationStore
+import com.terraformation.backend.customer.model.SystemUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.SRID
 import com.terraformation.backend.db.default_schema.GlobalRole
@@ -83,6 +84,7 @@ class AdminPlantingSitesController(
     private val organizationStore: OrganizationStore,
     private val plantingSiteStore: PlantingSiteStore,
     private val plantingSiteImporter: PlantingSiteImporter,
+    private val systemUser: SystemUser,
 ) {
   private val log = perClassLogger()
 
@@ -725,6 +727,22 @@ class AdminPlantingSitesController(
     } catch (e: Exception) {
       log.warn("Mortality rate recalculation failed", e)
       redirectAttributes.failureMessage = "Failed to recalculate mortality rates: ${e.message}"
+    }
+
+    return redirectToAdminHome()
+  }
+
+  @PostMapping("/migrateSimplePlantingSites")
+  @RequireGlobalRole([GlobalRole.SuperAdmin])
+  fun migrateSimplePlantingSites(redirectAttributes: RedirectAttributes): String {
+    try {
+      systemUser.run {
+        redirectAttributes.successDetails = plantingSiteStore.migrateSimplePlantingSites()
+      }
+      redirectAttributes.successMessage = "Successfully migrated simple planting sites"
+    } catch (e: Exception) {
+      log.error("Failed to migrate simple planting sites", e)
+      redirectAttributes.failureMessage = "Failed to migrate simple planting sites: ${e.message}"
     }
 
     return redirectToAdminHome()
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3011** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3011?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3011?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/admin/AdminPlantingSitesController.kt
```diff
@@ -730,6 +732,22 @@ class AdminPlantingSitesController(
...
+  @RequireGlobalRole([GlobalRole.SuperAdmin])
+  fun migrateSimplePlantingSites(redirectAttributes: RedirectAttributes): String {
+    try {
+      systemUser.run {
+        redirectAttributes.successDetails = plantingSiteStore.migrateSimplePlantingSites()
+      }
+      redirectAttributes.successMessage = "Successfully migrated simple planting sites"
+    } catch (e: Exception) {
+      log.error("Failed to migrate simple planting sites", e)
+      redirectAttributes.failureMessage = "Failed to migrate simple planting sites: ${e.message}"
```
ice1080: Would we want to stay on the migrate page if the migration failed, instead of redirecting to admin home?

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -1669,6 +1671,93 @@ class PlantingSiteStore(
...
+          val zoneId = createPlantingZone(zone, siteId)
+          val zoneHistoryId = insertPlantingZoneHistory(zone, siteHistoryId, zoneId)
+          val subzoneId = createPlantingSubzone(subzone, siteId, zoneId)
+          insertPlantingSubzoneHistory(subzone, zoneHistoryId, subzoneId)
+
+          messages.add("Migrated planting site $siteId (${existingSite.name})")
+        } catch (e: Exception) {
+          log.error("Unable to migrate planting site $siteId", e)
+          messages.add("Unable to migrate planting site $siteId: ${e.message}")
+          break
```
ice1080: We're not throwing an error here, which means that the error gets added to `messages`, which then gets set to `redirectAttributes.successDetails` in the controller, and the controller would still display `Successfully migrated simple planting sites`, which seems to be confusing since the messages below it would say `Unable to migrate planting site <bla>`

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -1669,6 +1671,93 @@ class PlantingSiteStore(
...
+                  boundary,
+                  existingSite.description,
+                  name = existingSite.name,
+                  organizationId = existingSite.organizationId,
+              )
+
+          if (newSite.areaHa == null || newSite.areaHa.equalsIgnoreScale(BigDecimal.ZERO)) {
+            messages.add("Planting site $siteId (${existingSite.name}) area too small to convert")
+            continue
+          }
```
ice1080: Confirming that we're fine to convert simple planting sites that might only be 1 hectare? Just wanting to make sure that we don't need a lower limit instead of just non-zero.

src/main/kotlin/com/terraformation/backend/admin/AdminPlantingSitesController.kt
sgrimm: There isn't a migrate page -- it's a button on the home page.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
sgrimm: The nonzero thing is just because there are some statistics that we calculate by dividing by area, and we round area to 1 decimal place. Anything greater than 0.05 hectares is fine since it won't cause divide-by-zero errors.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
sgrimm: Yeah, fair point, this is shoddy.

src/main/kotlin/com/terraformation/backend/admin/AdminPlantingSitesController.kt
ice1080: Ah good point, I should have noticed that

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
ice1080: And, if it's a one time use and you're the most likely person to use it then it might not need fixing...

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
ice1080: Got it. Then just to be nitpicky, doesn't `equalsIgnoreScale` mean that it's going to reject simple plots with an area of < 1, instead of < 0.1?

ice1080 (approved):

PR #3021, opened 2025-04-14
Reduce variable fetches

ice1080:
Reduce calls for each of `projectAcceleratorVariablesStableIds` to one call to the VariableStore for all of them together. 

Future eventual enhancement: update `fetchByStableId` and new `fetchListByStableIds`  to add variables to the cache immediately instead of passing the ID to another fetch method for every value returned.

This extra fetching was noticed as part of SW-6786, since many more stable Ids will be added for that story.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -132,7 +132,7 @@ class VariableStore(
       with(VARIABLES) {
         dslContext
             .select(ID)
-            .on(STABLE_ID)
+            .distinctOn(STABLE_ID)
             .from(VARIABLES)
             .where(STABLE_ID.`in`(stableIds))
             .orderBy(STABLE_ID, ID.desc())
--- a/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
@@ -128,6 +128,23 @@ class VariableStore(
             ?.let { fetchVariableOrNull(it) }
       }
 
+  fun fetchListByStableIds(stableIds: List<StableId>): List<Variable> =
+      with(VARIABLES) {
+        dslContext
+            .select(ID)
+            .from(VARIABLES)
+            .where(STABLE_ID.`in`(stableIds))
+            .and(
+                ID.notIn(
+                    dslContext
+                        .select(REPLACES_VARIABLE_ID)
+                        .from(VARIABLES)
+                        .where(REPLACES_VARIABLE_ID.isNotNull())))
+            .fetch(ID)
+            .filterNotNull()
+            .mapNotNull { fetchVariableOrNull(it) }
+      }
+
   fun fetchDeliverableVariables(deliverableId: DeliverableId): List<Variable> {
     val replacementVariables = VARIABLES.`as`("replacement_variables")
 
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3021** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3021?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3021?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -128,6 +128,23 @@ class VariableStore(
...
+        dslContext
+            .select(ID)
+            .from(VARIABLES)
+            .where(STABLE_ID.`in`(stableIds))
+            .and(
+                ID.notIn(
+                    dslContext
+                        .select(REPLACES_VARIABLE_ID)
+                        .from(VARIABLES)
+                        .where(REPLACES_VARIABLE_ID.isNotNull())))
```
sgrimm: This subquery is going to fetch every variable that has ever been replaced, whether or not it has one of the requested stable IDs.

Two other ways you could go about this would be to select the maximum variable ID for each stable ID (`DISTINCT ON` would help with that) or to use a correlated `NOT EXISTS` subquery.

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -128,6 +128,23 @@ class VariableStore(
...
+            .fetch(ID)
+            .filterNotNull()
```
sgrimm: I assume the `filterNotNull` is because `ID` is declared as nullable, even though in reality it can never be null in this query. That is, it's just to make the Kotlin type system happy. If so, the slightly more efficient way to do this would be to do `.fetch(ID.asNonNullable())`. The `asNonNullable` extension function is a runtime no-op that does a type cast to remove nullability from a jOOQ field.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
ice1080: Ah shoot ya I meant to include that in the subquery.
Updated to the `distinct on` approach (seems that `distinct` is actually not necessary within jooq).

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
```diff
@@ -132,16 +132,11 @@ class VariableStore(
       with(VARIABLES) {
         dslContext
             .select(ID)
+            .on(STABLE_ID)
```
sgrimm: Huh, didn't know about this. I've been using `.distinctOn()`.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/documentproducer/db/VariableStore.kt
ice1080: I may change it over to that. I was having issues with that at first, but just hadn't understood the sorting needed. The `.on` is still using a `distinct on` under the hood, so might be better to be more explicit for readability and searchability.

ice1080:
### Merge activity

* **Apr 14, 6:56 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3021).

PR #3028, opened 2025-04-15
SW-6189 Simplify plot replacement logic

sgrimm:
Remove support for multi-plot clusters from the code that handles the "request
replacement of a permanent plot" feature. Previously, it needed to tear down the
entire multi-plot cluster, but now it can operate on the requested plot in
isolation since there are no more multi-plot clusters.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -1328,66 +1328,35 @@ class PlantingSiteStore(
   /**
    * Makes a monitoring plot unavailable for inclusion in future observations.
    *
-   * The requested plot's "is available" flag is set to false.
+   * The requested plot's "is available" flag is set to false, and its permanent cluster number is
+   * cleared. If the plot was permanent, a new one with the removed cluster number will be created
+   * next time [ensurePermanentClustersExist] is called.
    *
-   * If the requested plot is part of a permanent cluster, the cluster is destroyed: the remaining
-   * plots in the cluster are updated such that they're no longer in a permanent cluster at all (but
-   * are still available for selection as temporary plots). A new cluster with the removed cluster
-   * number will be created next time [ensurePermanentClustersExist] is called.
-   *
-   * @return The plots that were modified. If the requested plot was part of a permanent cluster,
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3028** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -1328,66 +1328,35 @@ class PlantingSiteStore(
...
-        plantingSubzonesDao.fetchOneById(plotsRow.plantingSubzoneId!!)
-            ?: throw PlantingSubzoneNotFoundException(plotsRow.plantingSubzoneId!!)
-    val permanentCluster = plotsRow.permanentCluster
-    val plantingZoneId = subzonesRow.plantingZoneId!!
-
-    requirePermissions { updatePlantingSite(subzonesRow.plantingSiteId!!) }
+    requirePermissions { updatePlantingSite(plotsRow.plantingSiteId!!) }
 
-    return dslContext.transactionResult { _ ->
+    return if (plotsRow.isAvailable == true) {
```
ice1080, line 1346: NIT - could change this to just `if (plotsRow.isAvailable)` now

ice1080 (approved):
Makes sense

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
sgrimm: The `== true` is because `isAvailable` is nullable (because all the fields on the `Row` classes are nullable). The alternative would be `if (plotsRow.isAvailable!!)` but IntelliJ suggests changing that to `== true`.

ice1080: Ah, I must be thinking of JS, which treats null values as false

PR #3030, opened 2025-04-15
Use planting site time zone for report mortality rate

sgrimm:
Modify the query that fetches mortality rate data for reports so it looks at
observations whose completed times fall within the reporting period in the
planting site's time zone, not the server's system time zone.  If there is no
per-planting-site time zone, fall back to the organization's time zone and
then UTC, same as we do for date-to-time conversions elsewhere in the code.

This may potentially fix the intermittent test failures we're seeing in
`ReportStoreTest` in CI, but even if not, it should make the results more
correct.


sgrimm:
* **#3030** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3030?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3030?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
Nice

PR #3040, opened 2025-04-18
SW-6694 Backfill data from simple planting sites

sgrimm:
Commit 0f7e1c7c added an admin function to convert simple planting sites to
detailed ones, but the migration only updated the site maps. It needs to also
migrate the withdrawal data, including the detailed records of plantings and
the species totals at the zone and subzone levels.

Migrate that data as well, in such a way that it will work for sites that have
already been converted by the previous version of the migration.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -1722,7 +1722,112 @@ class PlantingSiteStore(
           break
         }
       }
+
+      migrateSimplePlantingSitePopulations()
+      migrateSimplePlantingSitePlantings()
     }
+  }
+
+  private fun migrateSimplePlantingSitePopulations() {
+    // When plants are withdrawn to a detailed planting site, we update the site, zone, and
+    // subzone populations. When they're withdrawn to a simple planting site, we only update the
+    // site populations since there are no zones or subzones.
+    //
+    // After converting a simple planting site to a detailed one, there will be values in
+    // planting_site_populations but no values for any of the site's zones or subzones in the
+    // corresponding tables. So we want to copy the site values down to those two tables.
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3040** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3040?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3040?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -1718,9 +1718,114 @@ class PlantingSiteStore(
...
+    // ID for any plantings that have null subzone IDs but where the planting site has a subzone,
+    // since those will be the sites whose zones and subzones we've just created.
+    val rowsUpdated =
+        dslContext
+            .update(PLANTINGS)
+            .set(
+                PLANTINGS.PLANTING_SUBZONE_ID,
+                DSL.select(PLANTING_SUBZONES.ID)
+                    .from(PLANTING_SUBZONES)
+                    .where(PLANTING_SUBZONES.PLANTING_SITE_ID.eq(PLANTINGS.PLANTING_SITE_ID)))
```
ice1080, line 1838: Is it possible that this could have multiple subzones, thus breaking this query? Or is that not possible because it was a simple planting site before? (I don't know much about simple planting sites)

sgrimm: That _shouldn't_ be able to happen, but I'll add a guard against it anyway; good call.

ice1080 (approved):

PR #3046, opened 2025-04-18
SW-6766 Improve zone replacement/rename handling

sgrimm:
The flexible site editing code had support for zone boundaries changing such that
a subzone moved from one zone to another. However, it wasn't handling cases where
the original zone was completely deleted. Since we match the existing and new
zones using their names when a new shapefile is uploaded, this meant that we
weren't handling cases where a planting zone is renamed.

Update the edit calculator to properly move subzones to the new zones in that
case, and don't delete the subzones as part of the deletion of their old zones.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
@@ -287,16 +287,6 @@ class PlantingSiteEditCalculatorTest {
             existingModel = existing,
             plantingZoneEdits =
                 listOf(
-                    // Zone A shrinks
-                    PlantingZoneEdit.Update(
-                        addedRegion = rectangle(0),
-                        areaHaDifference = BigDecimal("-12.5"),
-                        desiredModel = desired.plantingZones[0],
-                        existingModel = existing.plantingZones[0],
-                        monitoringPlotEdits = emptyList(),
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#3046** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/edit/PlantingSiteEditCalculatorTest.kt
```diff
@@ -829,6 +823,44 @@ class PlantingSiteEditCalculatorTest {
         desired)
   }
 
+  @Test
+  fun `updates existing subzone if original zone is deleted and new zone has a subzone with its name`() {
+    val existing =
+        existingSite(width = 1000) {
+          zone(name = "A", width = 500, numPermanent = 1) { subzone(name = "1") { cluster() } }
+          zone(name = "B", width = 500, numPermanent = 1) { subzone(name = "2") { cluster() } }
```
ice1080, line 831: I think this test will need to be updated once the other cluster PRs are merged, or vice versa (because of the use of `cluster()`

ice1080 (approved):

PR #3052, opened 2025-04-21
SW-6804 Added published metrics to report fetch

tommylau523:
Added published metrics to the funders API. This is quite a bit simpler than the accelerator side because it is read only, and omits non-readable data such as metric ID, and progress notes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ReportsMetricsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ReportsMetricsController.kt
@@ -62,7 +62,7 @@ class ReportsController(private val metricStore: ReportMetricStore) {
 
 data class SystemMetricPayload(
     val metric: SystemMetric,
-    val name: String = metric.name,
+    val name: String = metric.jsonValue,
     val description: String = metric.description,
     val component: MetricComponent = metric.componentId,
     val type: MetricType = metric.typeId,
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
@@ -4,12 +4,16 @@ import com.terraformation.backend.api.ApiResponse200
 import com.terraformation.backend.api.ApiResponse404
 import com.terraformation.backend.api.FunderEndpoint
 import com.terraformation.backend.api.SuccessResponsePayload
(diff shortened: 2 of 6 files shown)


tommylau523:
* **#3052** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ReportsMetricsController.kt
```diff
@@ -62,7 +62,7 @@ class ReportsController(private val metricStore: ReportMetricStore) {
 
 data class SystemMetricPayload(
     val metric: SystemMetric,
-    val name: String = metric.name,
+    val name: String = metric.jsonValue,
```
tommylau523, line 65: our `name` column is parsed to `jsonValue`. `name` refers to the name of the enum which is inaccurate.

sgrimm: Could this be of type `SystemMetric` instead? Then it'd get the JSON value automatically, as well as being more type-safe and giving a list of possible values in the OpenAPI spec + generated FE code.

sgrimm: Could even make it a val with a getter rather than an explicit parameter.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/api/ReportsMetricsController.kt
sgrimm: Actually, is this even needed since the metric's `jsonValue` is already going to be in the `metric` field?

tommylau523: I am happy to remove it in the future. I think our FE already use this field, so we can make it consistent for now.

tommylau523: I can remove it in a separate PR after making sure our FE does not use it.

PR #3054, opened 2025-04-22
SW-6803 Hectares Planted Metric

tommylau523:
Added hectares planted metric to accelerator reports.

Refactored some timestamp comparison for report metrics, to compare in dates instead of timestamps, to avoid edge cases.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
@@ -54,12 +54,14 @@ import com.terraformation.backend.db.nursery.tables.references.BATCHES
 import com.terraformation.backend.db.nursery.tables.references.BATCH_WITHDRAWALS
 import com.terraformation.backend.db.nursery.tables.references.WITHDRAWAL_SUMMARIES
 import com.terraformation.backend.db.seedbank.tables.references.ACCESSIONS
+import com.terraformation.backend.db.tracking.PlantingSiteId
 import com.terraformation.backend.db.tracking.RecordedSpeciesCertainty
 import com.terraformation.backend.db.tracking.tables.references.DELIVERIES
 import com.terraformation.backend.db.tracking.tables.references.OBSERVATIONS
 import com.terraformation.backend.db.tracking.tables.references.OBSERVED_SITE_SPECIES_TOTALS
 import com.terraformation.backend.db.tracking.tables.references.PLANTINGS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONES
 import jakarta.inject.Named
 import java.math.BigDecimal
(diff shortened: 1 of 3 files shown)


tommylau523:
* **#3054** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3054?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3054?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1092,6 +1094,50 @@ class ReportStore(
...
+                .join(ORGANIZATIONS)
+                .on(PLANTING_SITES.ORGANIZATION_ID.eq(ORGANIZATIONS.ID))
+                .where(PLANTING_SITES.ID.eq(plantingSiteIdField)))
+        .asNonNullable()
+  }
+
+  private val hectaresPlantedField =
+      with(PLANTING_SUBZONES) {
+        DSL.field(
+                DSL.select(DSL.sum(AREA_HA))
```
ice1080, line 1128: All projects have subzones, correct? There are no older projects that only have zones?

sgrimm: Short answer: Yes, for purposes of this PR.

Long answer: There are three styles of planting sites.

* Detailed: The kind we're used to, with a map that has zones and subzones (it's valid to have one zone and one subzone covering the whole map). Only detailed planting sites can perform observations. We require accelerator partners to have detailed planting sites.
* Simple: A planting site that has a site map but no zones or subzones. Can't do observations. This style is mostly deprecated, and nearly all the existing simple planting sites were converted to detailed thanks to the migration I added last week. But there's still a way to create simple planting sites in the web app (there's a task in the backlog to get rid of that ability but it hasn't been prioritized) and there are still a couple planting sites in production that couldn't be converted to detailed by the migration. So for now, we need to account for the existence of simple planting sites. But nobody who is submitting a report for a planting site will have one of these because we wouldn't have let them get that far in the accelerator process.
* Mapless (we don't actually have an agreed-on term for this): A planting site with no map at all, just a name. You can't create these from the web app any more, but you can do it from the admin UI or by directly calling the API, and there are still a bunch of old sites without maps in the prod and staging databases from before we made maps mandatory. We need to account for these in the guts of the code, e.g., by making the map boundary nullable, but they're irrelevant for accelerator-related code.

You can't do observations of simple or mapless sites, and if you withdraw plants to one of them, it only records the planting site ID, not the subzone ID because there isn't any subzone ID to record.

ice1080: Thanks!

sgrimm (approved):

PR #3056, opened 2025-04-22
Split ObservationStoreTest

sgrimm:
ObservationStoreTest was getting huge. It was already organized into nested test
classes, so move all the nested classes up to the top level.

No changes to the actual test logic here, just rename and move refactorings.


sgrimm:
* **#3056** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3056?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3056?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
The size of this test is probably also a sign that ObservationStore is getting too big, but one step at a time.

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreCompletePlotTest.kt
ice1080, line 1: Just a note, github marked this file as too large to auto-preview.

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreInsertBiomassDetailsTest.kt
ice1080, line 1: Just a note, github marked this file as too large to auto-preview.

ice1080 (approved):
> The size of this test is probably also a sign that ObservationStore is getting too big, but one step at a time.

Not sure how much we follow the pedantic preference of "business logic should always be in a service and not a store or controller", but if we do then this is also an indication that we've diverged from that greatly.

sgrimm:
> Not sure how much we follow the pedantic preference of "business logic should always be in a service and not a store or controller", but if we do then this is also an indication that we've diverged from that greatly.

The pattern in terraware-server is more about layering and avoiding circular dependencies: stores shouldn't depend on other stores, but a service can depend on whichever stores it wants. If an operation needs to span more than one store, it goes in a service. Event listeners usually also go in services.

PR #3057, opened 2025-04-22
SW-6769 Added published report event

tommylau523:


src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -328,7 +328,8 @@ VALUES (1, 'User Added to Organization', 1),
        (30, 'Application Submitted', 1),
        (31, 'Completed Section Variable Updated', 1),
        (32, 'Accelerator Report Submitted', 1),
-       (33, 'AcceleratorReportUpcoming', 1)
+       (33, 'Accelerator Report Upcoming', 1),
+       (34, 'Accelerator Report Published', 1)
```
sgrimm, line 332: No big deal, but this seems like it should go in what I assume will be the next PR in this sequence, since this PR doesn't add any new notifications.

sgrimm (approved):

PR #3072, opened 2025-04-28
Query deal name variable value when fetching reports

tommylau523:
This is an example PR on using subqueries to fetch variables so that the reads are efficient.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
@@ -1,6 +1,7 @@
 package com.terraformation.backend.accelerator.model
 
 import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.db.StableId
 import com.terraformation.backend.db.accelerator.ProjectMetricId
 import com.terraformation.backend.db.accelerator.ProjectReportConfigId
 import com.terraformation.backend.db.accelerator.ReportFrequency
@@ -15,7 +16,6 @@ import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.documentproducer.model.ExistingTextValue
 import com.terraformation.backend.documentproducer.model.ExistingValue
-import com.terraformation.backend.documentproducer.model.StableId
 import com.terraformation.backend.documentproducer.model.StableIds
 import java.time.Instant
 import java.time.LocalDate
--- a/src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
@@ -9,11 +9,14 @@ import com.terraformation.backend.db.accelerator.ReportQuarter
 import com.terraformation.backend.db.accelerator.ReportStatus
 import com.terraformation.backend.db.accelerator.StandardMetricId
 import com.terraformation.backend.db.accelerator.tables.pojos.ReportsRow
-import com.terraformation.backend.db.accelerator.tables.references.PROJECT_ACCELERATOR_DETAILS
 import com.terraformation.backend.db.accelerator.tables.references.REPORTS
 import com.terraformation.backend.db.accelerator.tables.references.REPORT_CHALLENGES
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
+import com.terraformation.backend.documentproducer.model.ExistingTextValue
+import com.terraformation.backend.documentproducer.model.ExistingValue
+import com.terraformation.backend.documentproducer.model.StableId
+import com.terraformation.backend.documentproducer.model.StableIds
 import java.time.Instant
 import java.time.LocalDate
 import org.jooq.Field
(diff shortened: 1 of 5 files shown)


tommylau523:
* **#3072** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3072?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/model/ReportModel.kt
```diff
@@ -147,13 +150,17 @@ data class ReportModel(
...
     ): ReportModel {
       return with(REPORTS) {
         ReportModel(
             id = record[ID]!!,
             configId = record[CONFIG_ID]!!,
             projectId = record[PROJECT_ID]!!,
-            projectDealName = record[PROJECT_ACCELERATOR_DETAILS.DEAL_NAME],
+            projectDealName =
+                variableValuesField?.let {
+                  (record[it][StableIds.dealName] as? ExistingTextValue)?.value
```
tommylau523, line 162: There is probably a way to code in the variable type itself to the StableIds class. And handle the automatic casting and reading of it.

Something like `StableIds <reified V : ExistingValue>`. 

Otherwise this casting will have to be called every read and can be a source of error.

ice1080, line 162: Wait I thought the purpose of changing to the details was so that we move away from deal name as a variable? This PR seems to be moving us back to the variable?

tommylau523: Yes that is correct. 

We moved away from variables previously because we were doing an N+1 query. This pattern should hopefully prevent that in the future.

I think the long term goal here is to remove all columns that are also variables, and instead use this pattern to retrieve variable values during read time, so we don't have to continuously sync them. Most of the offenders are in the `project_accelerator_details` table.

tommylau523: That being said, I think this should have been a draft PR so look for downstream PRs first! Hopefully they all come together more.

ice1080: Understood on that, but the whole discussion we had a few weeks ago was that deal name shouldn't be a variable? 
DB Migration V361 just added the column on 3/28

tommylau523: Yes. I think I proposed that because deal name was not used in any questionnaire deliverables or documents. But it seems like they might be used in documents going forward. Product team does want to keep deal name as a variable because of that.

I think we will have to make a lot of decisions like this for a lot of variables. Whether they should be a variable or not. 

But either way longer-term we probably don't want this middle of the road implementation where a piece of the data is both a variable and a column.

ice1080: But wasn't the whole point of moving away from it being a variable unrelated to its usage in documents, and because we needed it to be non-null in most places?

tommylau523: The current approach does not maintain non-null or uniqueness regardless. We sort of ended up with a "worst-of-both-worlds" scenario, with added work to keep them in sync.

It is true that using variables will prevent these constraints that a sql table can provide us directly.

tommylau523: Taking a step back, for about 15 of these variables, we should go through these decisions:
1) Should it be a variable, or a column (hopefully never both, but maybe deal name has to be one of these)
2) If it is a variable, remove all the table columns, and use a subquery to fetch the values
3) Migrate values and remove the duplicated columns/variables

ice1080: Ya definitely seems like a discussion with the team and product for all of this.
Even the "worst-of-both-worlds" scenario we're in was mainly because we never finished implementing the non-null/uniqueness for the column, and never got an answer from product on all of that.
Regarding # 1, what we had talked about previously was making some columns available in documents instead of always doing variables. It seems that because that was never implemented, we're now having to reverse the decision to move some of them into columns.

sgrimm: I agree it would be highly desirable to know the variable's type ahead of time. Not only to handle the casting (though that too) but also so we can make the queries more precise and efficient by not, for example, including `variable_select_option_values` in the query when we're fetching a text variable's value.

I don't think reified types alone are enough, though. Some options off the top of my head:

* Change `StableIds` to be something other than a bunch of `val whatever: StableId` properties
* Change `StableId` itself to carry optional type information along with the string value
* Make the report fetching code fetch the `Variable`s from `VariableStore` (which caches them, so there'll be no N+1 problem after the first fetch)

All three options have tradeoffs, of course. The third one appeals to me because it eliminates a source of inconsistency and toil if, say, the accelerator team decides to change a select variable to a text field.

sgrimm: Agreed we need to go through these and discuss them case by case. For deal name in particular, I am _highly_ skeptical that we will ever want to produce a document that has text like `XXX_ARCOS NETWORK_3` in the middle of a paragraph.

tommylau523: Related to deal name, it seems like we want to remove the country code portion `XXX_` from the value, and instead uses the country variable to build the country code part. 

This will be useful because product discussed using the deal name without the country code for reports. 

This will be good for keeping data independent as well. I can follow up with a separate change for that.

tommylau523: For this PR, I will keep working on effective fetching of variable values. But remove the querying of deal name part, and instead focus on the project accelerator details store, since that one is particularly affected.

tommylau523: I think the tricky part for the third option is that, the variable types will not be known until runtime, when the fetching happens, but our code/data model objects will need to know the types in compile time. 

I think there is always a layer of translation between the VariableType and the actual Java/Kotlin types we use. I think I can make that layer explicit and catch any errors loudly (for example if a `NumberVariable` is expected to be used as a `String` somewhere, which is not intended).

sgrimm:
@tommylau523 Is this PR still in progress, or should it be closed?

tommylau523:
> @tommylau523 Is this PR still in progress, or should it be closed?

I am closing it now

PR #3078, opened 2025-04-30
Added list site histories endpoint

tommylau523:


tommylau523:
* **#3078** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3078?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3078?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/db/plantingSiteStore/PlantingSiteStoreFetchSiteHistoriesTest.kt
```diff
@@ -0,0 +1,193 @@
...
+
+    val pair = expected.zip(actual)
+
+    if (!pair.all { (a, b) -> a.equals(b, 0.00001) }) {
+      assertEquals(expected, actual)
+    }
+  }
+
+  @Test
+  fun `returns empty list if site ID does not exist`() {
```
sgrimm: It'd be more consistent if this threw `PlantingSiteNotFoundException`. Otherwise you can tell whether or not a site you don't have permission to read exists: if it exists, you'll get `PlantingSiteNotFoundException` due to the permission failure, but if it doesn't exist, you'll get an empty history list.

The good news is that that's already the real behavior of `fetchSiteHistories`!

The bad news is that this test relies on the fact that the MockK double for `TerrawareUser` is configured to return true for all `canReadPlantingSite` calls. That is, the `requirePermissions { readPlantingSite(plantingSiteId) }` call in the application code will succeed in this test scenario. But with an actual database user, that call will throw an exception because `IndividualUser` won't be able to find the organization ID for the planting site, even though this test appears to "prove" that it would return an empty list in that case.

This kind of subtle test bug is an example of why I hope we can move more to using `RunsAsDatabaseUser` in our tests.

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/tracking/db/plantingSiteStore/PlantingSiteStoreFetchSiteHistoriesTest.kt
tommylau523: Let me take this chance to use `RunsAsDatabaseUser` for this test.

sgrimm: Ideally, it'd be great to separate the two. That is, make the existing test (without this change) use `RunsAsDatabaseUser`. Then this PR gets stacked on top of that change. As opposed to turning this PR into "add a new endpoint, and also revamp the test code for a bunch of other functions unrelated to that endpoint." Graphite should make that _relatively_ painless.

tommylau523: Makes sense! But in this case since this is a new test file, I will make the sweeping change in this PR. But I agree with you on our other test classes.

sgrimm: What I was getting at is that since this test is a subclass of `BasePlantingSiteStoreTest` which uses `RunsAsUser`, we'd need to change that base class to use `RunsAsDatabaseUser` which would require changing all the other subclasses too.

Another option would be to not make this test a subclass any more, I suppose, but then it's inconsistent with all the other tests for `PlantingSiteStore`.

tommylau523: Changed to not use `BasePlantingSiteStoreTest`. AdHoc functions already do not do so. I can update BasePlantingSiteStoreTest in a follow up in the future.

sgrimm (approved):

PR #3094, opened 2025-05-07
SW-6895 Prune old timeseries values

sgrimm:
The `timeseries_values` table is by far the largest one in the production
database, and most of its contents are of no interest. We only need recent history
for most of the timeseries. However, for certain timeseries, we want to retain
historical data forever, so we can't just apply a fixed retention period across
the board.

Add a column to the timeseries table to configure how many days' worth of values
to retain, and add a recurring job that deletes old values based on each
timeseries' retention setting. To avoid accidentally deleting values that we need
to retain indefinitely, this uses an opt-in model: the default is to not prune any
old values, and we'll set the retention days on the specific timeseries we want to
trim. The opt-in approach also lets us roll this out gradually, starting with just
a few timeseries to get a better understanding of the real-world performance
impact.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/device/TimeseriesPruner.kt
+++ b/src/main/kotlin/com/terraformation/backend/device/TimeseriesPruner.kt
@@ -1,0 +1,80 @@
+package com.terraformation.backend.device
+
+import com.terraformation.backend.db.asNonNullable
+import com.terraformation.backend.db.default_schema.TimeseriesId
+import com.terraformation.backend.db.default_schema.tables.references.TIMESERIES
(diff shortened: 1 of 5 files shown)


sgrimm:
* **#3094** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3094?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3094?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
Nice

PR #3096, opened 2025-05-07
SW-6897 Create endpoint to retrieve Funder-related Project Details

ice1080:
Retrieve project funder-related details.

The LandUseModelTypes were also added because there can be more types selected than there are hectares values saved, and we want to display this to the user.

Also added countryAlpha3 in case we decide to remove the country code for displaying the deal name.


ice1080:
* **#3096** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3096?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080:
@sgrimm or @tommylau523  if you're able, an initial review on this approach would be greatly helpful. I'm waiting to hear back on some `internalOnly` variables right now, but knowing if this approach is mostly correct or not will allow me to further the FE work in the interim.

tommylau523:
This looks good given how our variables are set up here. But I think we will want to move a lot of these data from the SQL table to be variables only.

ice1080:
> This looks good given how our variables are set up here. But I think we will want to move a lot of these data from the SQL table to be variables only.

@tommylau523 Not quite following? Almost all of the info returned is already from variables, retrieved from `acceleratorProjectVariableValuesService.fetchValues`. It does look like there are a few data points that are in both the `project_accelerator_details` table and the variables, but in those cases we use the variables value (see `ProjectAcceleratorDetailsModel.of()`

Unless I'm misunderstanding?

tommylau523:
> > This looks good given how our variables are set up here. But I think we will want to move a lot of these data from the SQL table to be variables only.
> 
> @tommylau523 Not quite following? Almost all of the info returned is already from variables, retrieved from `acceleratorProjectVariableValuesService.fetchValues`. It does look like there are a few data points that are in both the `project_accelerator_details` table and the variables, but in those cases we use the variables value (see `ProjectAcceleratorDetailsModel.of()`
> 
> Unless I'm misunderstanding?

Sorry I should have been more specific. Right now they live in both the `project_accelerator_details` table as well as in variables. We probably want to remove them from the `project_accelerator_details`, and fetch the values directly in the store class via a subquery. Since this only concerns reads in this PR, it should be relatively straightforward. 

Not much actionable for this PR actually. But my goal longer term to get rid of the boilerplate service class that does a fetch of the variable values and then combines it with the store class directly. Nothing has to change for this PR, but just an FYI for context!

ice1080:
> Sorry I should have been more specific. Right now they live in both the `project_accelerator_details` table as well as in variables. We probably want to remove them from the `project_accelerator_details`, and fetch the values directly in the store class via a subquery. Since this only concerns reads in this PR, it should be relatively straightforward.
> 
> Not much actionable for this PR actually. But my goal longer term to get rid of the boilerplate service class that does a fetch of the variable values and then combines it with the store class directly. Nothing has to change for this PR, but just an FYI for context!

Got it, that makes sense! Agreed it would be nice to avoid a double query for these.

tommylau523 (approved):

PR #3125, opened 2025-05-22
SW-6951 BE Data Structure for Project Profile Publishing

ice1080:
Create tables for publishing project profile data.

PR #3126, opened 2025-05-22
SW-6952 BE Publish Project Profile Data

ice1080:
Create endpoint to publish project profile data.

Create endpoint to retrieve published project profile data.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
@@ -6,7 +6,6 @@ import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.default_schema.LandUseModelType
 import com.terraformation.backend.db.default_schema.ProjectId
-import com.terraformation.backend.db.funder.tables.records.PublishedProjectDetailsRecord
 import com.terraformation.backend.db.funder.tables.references.PUBLISHED_PROJECT_CARBON_CERTS
 import com.terraformation.backend.db.funder.tables.references.PUBLISHED_PROJECT_DETAILS
 import com.terraformation.backend.db.funder.tables.references.PUBLISHED_PROJECT_LAND_USE
@@ -33,8 +32,11 @@ class PublishedProjectDetailsStore(
               .select(SDG_NUMBER.asNonNullable())
               .from(this)
               .where(PROJECT_ID.eq(projectId))
-              .fetchSet(SDG_NUMBER.asNonNullable())
-              .mapNotNull { SustainableDevelopmentGoal.bySdgNumber[it] }
+              .fetch(SDG_NUMBER.asNonNullable())
+              .map {
+                SustainableDevelopmentGoal.bySdgNumber[it]
+                    ?: throw IllegalArgumentException("Unknown goal $it")
+              }
               .toSet()
         }
     val carbonCerts =
@@ -43,7 +45,7 @@ class PublishedProjectDetailsStore(
               .select(CARBON_CERTIFICATION.asNonNullable())
               .from(this)
               .where(PROJECT_ID.eq(projectId))
-              .fetchSet(CARBON_CERTIFICATION.asNonNullable())
+              .fetch(CARBON_CERTIFICATION.asNonNullable())
               .mapNotNull { CarbonCertification.forDisplayName(it) }
               .toSet()
         }
@@ -56,12 +58,11 @@ class PublishedProjectDetailsStore(
               .fetchMap(
                   LAND_USE_MODEL_TYPE_ID.asNonNullable(), LAND_USE_MODEL_HECTARES.asNonNullable())
         }
+
     return with(PUBLISHED_PROJECT_DETAILS) {
-      dslContext
-          .selectFrom(this)
-          .where(PROJECT_ID.eq(projectId))
-          .fetchOneInto(PublishedProjectDetailsRecord::class.java)
-          ?.let { FunderProjectDetailsModel.of(it, carbonCerts, sdgList, landUseModelMap) }
+      dslContext.selectFrom(this).where(PROJECT_ID.eq(projectId)).fetchOne()?.let {
+        FunderProjectDetailsModel.of(it, carbonCerts, sdgList, landUseModelMap)
+      }
     }
   }
 
--- a/src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
@@ -1,0 +1,177 @@
+package com.terraformation.backend.funder.db
+
+import com.terraformation.backend.accelerator.model.CarbonCertification
+import com.terraformation.backend.accelerator.model.SustainableDevelopmentGoal
+import com.terraformation.backend.auth.currentUser
+import com.terraformation.backend.db.asNonNullable
+import com.terraformation.backend.db.default_schema.LandUseModelType
+import com.terraformation.backend.db.default_schema.ProjectId
+import com.terraformation.backend.db.funder.tables.records.PublishedProjectDetailsRecord
+import com.terraformation.backend.db.funder.tables.references.PUBLISHED_PROJECT_CARBON_CERTS
+import com.terraformation.backend.db.funder.tables.references.PUBLISHED_PROJECT_DETAILS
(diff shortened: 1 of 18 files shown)


ice1080:
* **#3126** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3126?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3126?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3125** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3125?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
```diff
@@ -0,0 +1,177 @@
...
+    // convert the list to a set
+    val sdgList =
+        with(PUBLISHED_PROJECT_SDG) {
+          dslContext
+              .select(SDG_NUMBER.asNonNullable())
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .fetchSet(SDG_NUMBER.asNonNullable())
+              .mapNotNull { SustainableDevelopmentGoal.bySdgNumber[it] }
+              .toSet()
```
sgrimm: Can use `.fetch` instead of `.fetchSet` since the set will be turned into a list anyway by the `.mapNotNull`.

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
```diff
@@ -0,0 +1,177 @@
...
+  fun fetchOneById(projectId: ProjectId): FunderProjectDetailsModel? {
+    // convert the list to a set
+    val sdgList =
+        with(PUBLISHED_PROJECT_SDG) {
+          dslContext
+              .select(SDG_NUMBER.asNonNullable())
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .fetchSet(SDG_NUMBER.asNonNullable())
+              .mapNotNull { SustainableDevelopmentGoal.bySdgNumber[it] }
```
sgrimm: Should we log an error, or maybe even throw an exception, if there's an unrecognized SDG number?

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
```diff
@@ -0,0 +1,177 @@
...
+              .toSet()
+        }
+    val carbonCerts =
+        with(PUBLISHED_PROJECT_CARBON_CERTS) {
+          dslContext
+              .select(CARBON_CERTIFICATION.asNonNullable())
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .fetchSet(CARBON_CERTIFICATION.asNonNullable())
+              .mapNotNull { CarbonCertification.forDisplayName(it) }
```
sgrimm, line 49: I realize this is part of the data model that came before this PR, but I'm a little unclear on the semantics here. We're storing the carbon certification as text, but then requiring it to be a valid enum value? Why aren't we storing an integer enum ID with a foreign key reference to an enum table to guarantee only valid values? What's the advantage of storing a display name here?

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
```diff
@@ -0,0 +1,177 @@
...
+              .from(this)
+              .where(PROJECT_ID.eq(projectId))
+              .fetchMap(
+                  LAND_USE_MODEL_TYPE_ID.asNonNullable(), LAND_USE_MODEL_HECTARES.asNonNullable())
+        }
+    return with(PUBLISHED_PROJECT_DETAILS) {
+      dslContext
+          .selectFrom(this)
+          .where(PROJECT_ID.eq(projectId))
+          .fetchOneInto(PublishedProjectDetailsRecord::class.java)
```
sgrimm: Is the class name needed here? I'd expect `.fetchOne()` to return a `PublishedProjectDetailsRecord`.

src/main/resources/db/migration/0350/V381__PublishedProjectDetails.sql
```diff
@@ -36,6 +35,7 @@ CREATE TABLE funder.published_project_details (
     project_highlight_photo_value_id BIGINT REFERENCES docprod.variable_values,
     project_zone_figure_value_id BIGINT REFERENCES docprod.variable_values,
     standard TEXT,
+    tf_reforestable_land NUMERIC,
```
sgrimm: Editing an existing migration will cause the server to bomb out when it starts up because the migration file's checksum won't match the checksum of the version that was already applied to the database. This change would need to be made in the previous PR before merging it.

sgrimm (changes requested):
This looks pretty good in general; requesting changes because the edit to the migration would cause problems.

src/main/resources/db/migration/0350/V381__PublishedProjectDetails.sql
ice1080: Ah whoops, I made that change in the same file before creating the other PR, but must have already staged the changes and forgotten to restage this file. Will fix

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
ice1080: Ah yes, I'll fix that. I originally had a different class in there but decided to move to the record.

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
ice1080: Good thought. I had used `fetchSet` originally but then decided to map to the enums

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
ice1080: I suppose I can throw an exception. That would indicate data corruption, since they have to be valid SDG numbers when publishing

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
ice1080: No advantage, but I'm not familiar with anywhere in the app that references the select options of a docprod variable. Carbon Certifications is a multi select variable (as are land use model type and sdg), so if we do what you're saying then we would need to reference only some rows of `docprod.variable_select_options`. Have we done something like that before?

ice1080: And regarding this:
> I realize this is part of the data model that came before this PR

I'm considering the previous PR and this one as related enough that I'll make changes to the other one if it makes more sense for the BE implementation. I won't be merging it until we're happy with this PR as well.

sgrimm: Ah, okay, I wasn't thinking of the fact that this comes from a select variable originally. In that case yeah, text is probably about the best we'll be able to do here without adding a bunch of complexity.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/funder/db/PublishedProjectDetailsStore.kt
ice1080: One option we could think of for the future. At my last company we built out a way to have our constraints be managed by data. So a specific table's column would have a check constraint on it for text that ensured the text could only be certain values, and those values were stored in another general purpose table. We could do something like this by getting the latest variable id for every stable id that is a select variable, and coming up with a way to map that to constraints on other tables (or less wide-ranging by specifying that a table's column uses a select variable stable id and it retrieves that and generates the constraint during migration)

ice1080:
### Merge activity

* **Jun 5, 3:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3126).
* **Jun 5, 3:27 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3126) rebased this pull request as part of a merge.
* **Jun 5, 3:36 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3126).

PR #3138, opened 2025-05-28
SW-7016 Optionally return plot-level "Unknown" results

sgrimm:
By default, when we're returning observation results, we don't include
"Unknown" species in the list of species-level results. But we need to be able
to include them in exported CSV files. Add an optional parameter to the
endpoint that fetches results for a single observation to allow the client to
say that it wants unknown species included.

This only affects the monitoring-plot-level species lists; the zone- and
site-level species are filtered in `ObservationResultsStore`. Including unknown
species at the site and zone levels breaks some of the mortality rate
calculations.  We only need the unknown species at the plot level in order to
generate correct CSV files, so there's no need to change the underlying
fetching logic here.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
@@ -257,11 +257,17 @@ class ObservationsController(
       summary = "Gets the results of an observation of a planting site.",
       description = "Some information is only available once all plots have been completed.")
   fun getObservationResults(
-      @PathVariable observationId: ObservationId
+      @PathVariable observationId: ObservationId,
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3138** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

sgrimm:
Abandoning this; there's a simpler approach I'll use instead.

PR #3143, opened 2025-05-29
SW-7016 Return plot-level "Unknown" species results

sgrimm:
When we're returning observation results, we don't include "Unknown" species in
the list of species-level results. But we need to be able to include them in
exported CSV files. Add a field to the plot-level results payload to hold the
plant totals for plants of unknown species.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
@@ -698,6 +698,7 @@ data class ObservationMonitoringPlotCoordinatesPayload(
   fun toModel() = NewObservedPlotCoordinatesModel(gpsCoordinates, position)
 }
 
+@Schema(description = "Information about observed plants of a particular species in a region.")
 data class ObservationSpeciesResultsPayload(
     val certainty: RecordedSpeciesCertainty,
     @Schema(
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3143** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3143?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3143?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -835,6 +838,10 @@ data class ObservationMonitoringPlotResultsPayload(
       status = model.status,
       totalPlants = model.totalPlants,
       totalSpecies = model.totalSpecies,
+      unknownSpecies =
+          model.species
+              .firstOrNull { it.certainty == RecordedSpeciesCertainty.Unknown }
```
ice1080, line 843: Is there only one species in the list that will have a certainty of unknown?

tommylau523: Yes! The table should have a constraint on that. Because it does a rollup count.

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
sgrimm: That's right. There wouldn't be any way to differentiate multiple ones since we don't have species IDs or names for them and those are the unique identifiers for observed species. Basically it's a truth table like

| species_id | species_name | certainty
| - | - | -
| non-null | non-null | (Illegal combination, enforced by check constraint)
| non-null | null | Known
| null | non-null | Other
| null | null | Unknown

and (species_id, species_name) is unique for a given observation and plot/zone/etc.

ice1080: Got it, thanks!

PR #3152, opened 2025-06-03
SW-7029 Add GPS coordinates to recorded trees

sgrimm:
In monitoring observations, we record the GPS coordinates of each plant the user
observes, but we weren't doing the same for the trees that are recorded in biomass
observations. Add GPS coordinates as an optional attribute for recorded trees.


sgrimm:
* **#3152** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3152?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3152?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -280,6 +282,7 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
                         TREE_GROWTH_FORM_ID,
                         TREE_NUMBER,
                         TRUNK_NUMBER,
+                        recordedTreesGpsCoordinatesField,
```
ice1080, line 285: What's the benefit here of the separate variable as opposed to just doing `GPS_COORDINATES`?

sgrimm: It's a workaround for a bug/limitation in the community edition of jOOQ. Basically jOOQ gets confused if you include a geometry field in a multiset because of the way multisets are implemented under the hood, so you have to explicitly tell it how to render the value in the multiset JSON object. The separate variable is just so we don't have to recalculate that expression every time.

ice1080: Ah that makes sense. Thanks!

ice1080 (approved):

PR #3158, opened 2025-06-06
SW-6808 Implemented CRUD for disclaimers

tommylau523:
This PR implements functions to create, read, accept and delete disclaimers. 

Read and accept current disclaimer are exposed via external facing APIs.

Creating, deleting, and listing disclaimers and user acceptance records will be functions that are exposed through the admin UI, which will be part of the PR that follows.


tommylau523:
* **#3158** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/customer/db/DisclaimerStoreTest.kt
```diff
@@ -0,0 +1,315 @@
...
+          acceptedOn = Instant.ofEpochSecond(3001))
+
+      val disclaimerId2 =
+          insertDisclaimer(
+              content = "Disclaimer 2",
+              effectiveOn = Instant.ofEpochSecond(6000),
+          )
+
+      clock.instant = Instant.ofEpochSecond(2999)
+      assertEquals(null, store.fetchCurrentDisclaimer(), "All disclaimers are in the future")
```
sgrimm, line 208: Could use `assertNull`.

src/main/kotlin/com/terraformation/backend/customer/model/DisclaimerModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.jooq.Record
+
+data class DisclaimerModel(
+    val id: DisclaimerId,
+    val content: String,
+    /** The time when the current user has accepted the disclaimer, if they had */
+    val acceptedOn: Instant? = null,
+    val effectiveOn: Instant,
+    /** List of all users that had accepted the disclaimer and the time of acceptance. */
+    val users: Map<UserId, Instant>? = null,
```
sgrimm: Doing it this way is more concise, but it feels odd to me to have the same model class hold either the current user's acceptance time or all users' times. Especially in the all-users case, you end up with a nullable field that should never actually be allowed to be null.

That said, maybe it's not worth the added complexity here for the sake of a slightly cleaner object model.

src/main/kotlin/com/terraformation/backend/customer/db/DisclaimerStore.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    val currentDisclaimer = fetchCurrentDisclaimer()
+
+    if (currentDisclaimer == null) {
+      throw IllegalStateException("No current disclaimer exists.")
+    }
+
+    if (currentDisclaimer.acceptedOn != null) {
+      throw IllegalStateException("User has already accepted the current disclaimer.")
+    }
```
sgrimm: Maybe we should treat this as a no-op instead. Log an info message, perhaps, but don't treat the request as failed.

src/main/kotlin/com/terraformation/backend/customer/db/DisclaimerStore.kt
```diff
@@ -0,0 +1,123 @@
...
+          .from(this)
+          .fetch { DisclaimerModel.of(it, usersMultiset) }
+    }
+  }
+
+  /**
+   * Fetch the latest effective disclaimer. Omits future disclaimers. Null if no effective
+   * disclaimer exists.
+   */
+  fun fetchCurrentDisclaimer(): DisclaimerModel? {
```
sgrimm: If this returns null, will callers ever do anything other than throw an exception? If not, then maybe this should return `DisclaimerModel` and throw the exception itself.

sgrimm (approved):
At some point perhaps we'll want to make this i18n-friendly (multiple versions of the content, etc.) but this seems fine for now. A few minor comments inline but seems generally fine to me.

src/main/kotlin/com/terraformation/backend/customer/db/DisclaimerStore.kt
tommylau523: This will only happen before the first disclaimer gets uploaded. Main reason is that we do not want the client API call to fail.

src/main/kotlin/com/terraformation/backend/customer/model/DisclaimerModel.kt
tommylau523: Perhaps one way to do so is to remove acceptedOn, and rely only on the Map. So we can always have a map, either one/zero entry, or all users' entries.

tommylau523: I will split it into UserDisclaimerModel vs DisclaimerModel. Should be straightforward enough.

PR #3172, opened 2025-06-10
SW-7037 Recalculate site populations from admin UI

sgrimm:
In cases where nursery batches have been deleted, there can be a discrepancy
between the outplanting withdrawal totals from the withdrawal log and the reported
plant populations that are tracked in `planting_site_populations` and its sibling
tables.

We're figuring out a more robust plan for this going forward, but until that's
in place, we need to be able to clear up the discrepancies in the numbers.

Add an admin UI function to reset the populations to the totals from the
withdrawal log.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/DeliveryStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/DeliveryStore.kt
@@ -30,6 +30,7 @@ import com.terraformation.backend.nursery.db.WithdrawalNotFoundException
 import com.terraformation.backend.tracking.model.DeliveryModel
 import com.terraformation.backend.tracking.model.PlantingModel
 import jakarta.inject.Named
+import java.math.BigDecimal
 import java.time.Instant
 import java.time.InstantSource
 import org.jooq.DSLContext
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3172** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3172?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3172?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/DeliveryStore.kt
ice1080, line 1: I might just be missing the context for these populations tables, but if they are just calculations from other tables, couldn't they be replaced with views instead?

sgrimm: They're an optimization. Over time you'll have hundreds or thousands of deliveries to a planting site and you don't want to have to aggregate across all those rows every time someone visits a plants dashboard.

ice1080: So then I guess my question would be why not a materialized view with a refresh triggered by the events in these files?

sgrimm: They actually started out as views!

sgrimm: See commit b114cd630 for a bit more context.

sgrimm: If I'm not mistaken, materialized views can't be incrementally refreshed, so to keep the numbers current, you'd have to recalculate the totals for all planting sites/zones/subzones every time an outplanting withdrawal happened to any site. It's possible I'm wrong, though.

ice1080: Ya it definitely makes sense to me for moving away from views. I don't understand quite as much why a materialized view couldn't handle it since it would be doing the same queries that we are now, but also I don't care much either way as you end up with the same result and the same amount of db computation either way.

ice1080: 🤔 Actually now that I say that, that's not correct, since your methods only refresh one site/zone/subzone at a time, and a refresh would do the whole MV.

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/DeliveryStore.kt
ice1080: Ah, github didn't update while I was typing. I see your comment now and came to the same conclusion ha.

ice1080: I have seen a case in oracle that did incremental materialized view refreshing, but iirc it was pretty complicated and might have been an oracle-only feature. I didn't interact with it much, and it's possible we were hacking it in some way too.

sgrimm: The other thing to note here is that this new refresh method should be very rarely called. It's hopefully a one-time fixup for each of a handful of affected planting sites. In normal operation, we update the tables much more selectively (e.g., "add 5 to the total for this site+species" rather than "recalculate the totals from scratch for this site.")

PR #3176, opened 2025-06-13
SW-7052 Endpoints to update metric targets

tommylau523:


tommylau523:
* **#3176** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3176?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -361,6 +361,87 @@ class ReportStore(
...
+    if (!metricExists) {
+      throw IllegalStateException(
+          "Project metric $metricId is not associated with project $projectId")
+    }
+
+    val reportIds = targets.keys
+    validateReportsInProject(reportIds, projectId)
+    if (!updateSubmitted) {
+      validateReportsUpdatable(reportIds)
+    }
```
ice1080, line 394: Do we need to validate that reports aren't already approved and/or published even if `updateSubmitted` is true?

tommylau523: The product answer is no. 

Accelerator admins are supposed to be able to update the targets for reports in any statuses. The review report endpoint bypasses these checks as well and allow for the updates. 

The new target will not be part of a published report, until the next publish. So that gives some guardrail on what a funder may see at least.

ice1080: Right, I knew it wouldn't directly affect a published report, just wasn't sure if we needed to block them if a report had ever been published.

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
tommylau523: It is a fair question to ask!! 

Right now we have a possible interaction of report approved -> published -> report set to another status like Needs Update -> no longer publishable.

This raises a slight alarm if I am thinking about it like a state machine, that an "Approved" state isn't a terminal state and can lead to funky behaviors.

PR #3186, opened 2025-06-16
SW-7057 Set quantities to 0 on batch delete

sgrimm:
When the user deletes a batch that has had withdrawals, keep it around but set its
remaining quantities to zero. This will remove it from the list of active batches
in the web app, while preserving its entries in the withdrawal log.

The reasoning is that users are using the delete function just to clear irrelevant
batches out of their inventory, not to forget the fact that the batches ever
existed.

However, for batches that have had no withdrawals, we continue to delete them
outright. This will have no effect on the nursery's withdrawal log, but will
remove batches that were entered by mistake and never used for anything.


sgrimm:
* **#3186** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3200, opened 2025-07-01
SW-7108 Add optional unit field to project specific metrics

ice1080:


src/main/resources/db/migration/0350/V387__ProjectMetricsUnit.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE accelerator.project_metrics
+    ADD COLUMN unit VARCHAR(25);
```
sgrimm: Do we specifically want a database-enforced length limit here, as opposed to enforcing the limit in the application code?

IMO it'd be better to make this `TEXT` and enforce the length limit with a validation rule on the request payload. That way if a client passes in a value that's too long, it'll get back a validation failure error message rather than a generic HTTP 500 from a failed INSERT statement. See, for example, `CreateOrganizationRequestPayload`. One thing to know about using field validations is that you need a `@Valid` annotation on the request handler parameter (again, see the code for that endpoint for an example).

sgrimm (changes requested):

ice1080:
* **#3202** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3200** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3200?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3200?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0350/V387__ProjectMetricsUnit.sql
ice1080: Good thought, updated! I'm still trying to forget oracle world where everything is VARCHAR2 or CLOB, so I always think TEXT columns are CLOBs (which are stored very differently than everything else).

sgrimm (approved):

ice1080:
### Merge activity

* **Jul 2, 3:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3200).
* **Jul 2, 3:59 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3200).

PR #3203, opened 2025-07-02
SW-7113 Fix Seed Fund Reports with SDGs

ice1080:
Seed Fund Reports previously saved and retrieved SDGs by their enum value, but this was changed to their UN number. Update `forJsonValue` to first retrieve by UN number and then by enum value.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/accelerator/model/SustainableDevelopmentGoalTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/accelerator/model/SustainableDevelopmentGoalTest.kt
@@ -24,4 +24,60 @@ class SustainableDevelopmentGoalTest {
     assertEquals(
         SustainableDevelopmentGoal.Partnerships, SustainableDevelopmentGoal.forJsonValue("17"))
   }
+
+  @Test
+  fun `forJsonValue works correctly for names`() {
+    assertEquals(
+        SustainableDevelopmentGoal.NoPoverty, SustainableDevelopmentGoal.forJsonValue("NoPoverty"))
+    assertEquals(
+        SustainableDevelopmentGoal.ZeroHunger,
+        SustainableDevelopmentGoal.forJsonValue("ZeroHunger"))
+    assertEquals(
+        SustainableDevelopmentGoal.GoodHealth,
+        SustainableDevelopmentGoal.forJsonValue("GoodHealth"))
+    assertEquals(
+        SustainableDevelopmentGoal.QualityEducation,
+        SustainableDevelopmentGoal.forJsonValue("QualityEducation"))
+    assertEquals(
+        SustainableDevelopmentGoal.GenderEquality,
+        SustainableDevelopmentGoal.forJsonValue("GenderEquality"))
+    assertEquals(
+        SustainableDevelopmentGoal.CleanWater,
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3203** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3203?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3203?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/accelerator/model/SustainableDevelopmentGoalTest.kt
```diff
@@ -24,4 +24,60 @@ class SustainableDevelopmentGoalTest {
     assertEquals(
         SustainableDevelopmentGoal.Partnerships, SustainableDevelopmentGoal.forJsonValue("17"))
   }
+
+  @Test
+  fun `forJsonValue works correctly for names`() {
+    assertEquals(
+        SustainableDevelopmentGoal.NoPoverty, SustainableDevelopmentGoal.forJsonValue("NoPoverty"))
```
sgrimm, line 31: Not convinced it's useful to cover every single enum value here rather than just testing whether name lookup works at all, but if we want full coverage, maybe something like this rather than listing the values all out explicitly?
```kotlin
SustainableDevelopmentGoal.entries.forEach {
  assertEquals(it, SustainableDevelopmentGoal.forJsonValue(it.name))
}
```

sgrimm (approved):
This change seems like a sign we've failed to specify our payload formats adequately; I would have expected this mismatch to cause a TypeScript error in the client code. I guess maybe this value is coming from a variable rather than an explicit payload field. The change itself seems fine aside from a quibble with the test.

ice1080:
@sgrimm the error is actually a failure on retrieval from the DB, unrelated to FE payloads. We changed from saving the enum value in the DB to the UN number, and same with retrieval. I had missed that this was used in the jsonified body of the annual seed fund reports, and thus we had some original enum values stored.

src/test/kotlin/com/terraformation/backend/accelerator/model/SustainableDevelopmentGoalTest.kt
ice1080: I'll update to just testing a couple! Wasn't sure how often we wanted full coverage

sgrimm:
Ah, okay. In that case, would it be better to do a database migration to convert the existing values?

sgrimm:
Though to be clear: I think this is fine for now. We can do a migration later and remove this at that point.

ice1080:
Sounds good, I'll merge for now. 
The DB migration would have to be a hardcode with the current enum values to numbers, inside the json bodies. I wasn't sure how often we did something like that. (This assumes that we would do the migration with flyway and sql. Alternatively we could do it with the java code instead.)

ice1080:
### Merge activity

* **Jul 2, 5:06 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3203).

PR #3204, opened 2025-07-02
SW-7117 Allow retrieval of multiple published funder project details

ice1080:
Previously the endpoint only allowed for one projectId to be retrieved at a time. Update the endpoint to receive a set of projectIds. Return this in a different property so that the FE doesn't break. 
Eventually the original code returning only one project's details should be removed once the FE is only using the list of project details returned.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/funder/FunderProjectService.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/FunderProjectService.kt
@@ -1,7 +1,6 @@
 package com.terraformation.backend.funder
 
 import com.terraformation.backend.customer.model.requirePermissions
-import com.terraformation.backend.db.ProjectNotFoundException
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.funder.db.PublishedProjectDetailsStore
 import com.terraformation.backend.funder.model.FunderProjectDetailsModel
@@ -11,11 +10,14 @@ import jakarta.inject.Named
 class FunderProjectService(
     private val publishedProjectDetailsStore: PublishedProjectDetailsStore,
 ) {
-  fun fetchByProjectId(projectId: ProjectId): FunderProjectDetailsModel {
+  fun fetchByProjectId(projectId: ProjectId): FunderProjectDetailsModel? {
     requirePermissions { readProjectFunderDetails(projectId) }
 
     return publishedProjectDetailsStore.fetchOneById(projectId)
-        ?: throw ProjectNotFoundException(projectId)
+  }
+
+  fun fetchListByProjectIds(projectIds: Set<ProjectId>): List<FunderProjectDetailsModel> {
+    return projectIds.mapNotNull { fetchByProjectId(it) }
   }
 
   fun publishProjectProfile(model: FunderProjectDetailsModel) {
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FundingProjectsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FundingProjectsController.kt
@@ -41,7 +41,8 @@ class FundingProjectsController(
       if (model == null) {
         throw ProjectNotFoundException(projectIds.toList()[0])
       }
-      return GetFundingProjectResponsePayload(details = FunderProjectDetailsPayload(model))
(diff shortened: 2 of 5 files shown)


ice1080:
* **#3204** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/funder/FunderProjectService.kt
```diff
@@ -11,11 +10,14 @@ import jakarta.inject.Named
...
-  fun fetchByProjectId(projectId: ProjectId): FunderProjectDetailsModel {
+  fun fetchByProjectId(projectId: ProjectId): FunderProjectDetailsModel? {
     requirePermissions { readProjectFunderDetails(projectId) }
 
     return publishedProjectDetailsStore.fetchOneById(projectId)
-        ?: throw ProjectNotFoundException(projectId)
+  }
+
+  fun fetchListByProjectIds(projectIds: Set<ProjectId>): List<FunderProjectDetailsModel> {
+    return projectIds.mapNotNull { fetchByProjectId(it) }
```
sgrimm, line 20: How many project IDs are we going to be seeing here? Worth considering making a multi-project fetch function in the underlying store class to grab everything in one database round-trip? If the list of IDs is going to be low single digits, then that's probably overkill.

src/main/kotlin/com/terraformation/backend/funder/api/FundingProjectsController.kt
```diff
@@ -30,11 +31,22 @@ class FundingProjectsController(
...
+  @Operation(summary = "Get published project details displayable to funders")
+  fun getProjects(@PathVariable projectIds: Set<ProjectId>): GetFundingProjectResponsePayload {
+    // Remove this section once FE is handling the list of projects returned
+    if (projectIds.size == 1) {
+      val model: FunderProjectDetailsModel? =
+          funderProjectService.fetchByProjectId(projectIds.toList()[0])
+      if (model == null) {
+        throw ProjectNotFoundException(projectIds.toList()[0])
+      }
+      return GetFundingProjectResponsePayload(details = FunderProjectDetailsPayload(model))
```
sgrimm: Would it be worth also populating `projects` in this case so the FE can switch to reading from that without requiring another BE change?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/funder/FunderProjectService.kt
ice1080: I think 95% of use cases will be 1, and the other 5% will be 2-3. I also considered that, but came to the conclusion of overkill

src/main/kotlin/com/terraformation/backend/funder/api/FundingProjectsController.kt
ice1080: That's a good thought, will add that. I meant for it to work like that, but didn't notice that the if doesn't do that

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/funder/FunderProjectServiceTest.kt
```diff
@@ -37,22 +40,62 @@ class FunderProjectServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
-    val funderId = insertUser(type = UserType.Funder)
-    switchToUser(funderId)
-    assertThrows<ProjectNotFoundException> { service.fetchByProjectId(projectId) }
+  @Nested
+  inner class FetchByProjectId {
+    @Test
+    fun `throws error if user can't read project funder details`() {
+      val funderId = insertUser(type = UserType.Funder)
+      switchToUser(funderId)
+      assertThrows<ProjectNotFoundException> { service.fetchByProjectId(projectId) }
```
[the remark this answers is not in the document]
ice1080: This is incorrect. It still throws an error from the require permissions check.

PR #3213, opened 2025-07-07
SW-6826 Update history on zone rename

sgrimm:
We want a planting site's current zone names to match the names in the most recent
edit history entry for the zone; otherwise historical views of the planting site
will be inconsistent. Make admin-UI zone name edits apply retroactively, with a
tooltip to clarify the behavior.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
@@ -1038,31 +1038,58 @@ class PlantingSiteStore(
             ?: throw PlantingZoneNotFoundException(plantingZoneId)
     val edited = editFunc(initial)
 
-    with(PLANTING_ZONES) {
-      dslContext
-          .update(PLANTING_ZONES)
-          .set(ERROR_MARGIN, edited.errorMargin)
-          .set(MODIFIED_BY, currentUser().userId)
-          .set(MODIFIED_TIME, clock.instant())
-          .set(NAME, edited.name)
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#3213** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -1038,32 +1038,59 @@ class PlantingSiteStore(
...
+
+        with(PLANTING_ZONE_HISTORIES) {
+          val plantingZoneHistoryId =
+              dslContext
+                  .select(ID)
+                  .from(PLANTING_ZONE_HISTORIES)
+                  .where(PLANTING_ZONE_ID.eq(plantingZoneId))
+                  .orderBy(ID.desc())
+                  .limit(1)
+                  .fetchSingle(ID)
```
ice1080, line 1077: Is there any chance that the most recent planting zone histories row is not the one that we need, and/or not the only one that needs updated? (I'm not real familiar with this table so feel free to just say "no possibility").

sgrimm: It'd take someone uploading a new map at the same time this rename operation was happening, but good point; I can guard against that easily enough.

ice1080 (approved):

PR #3217, opened 2025-07-10
SW-7141 Updated with new pre-screen checks

tommylau523:


tommylau523:
* **#3217** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3217?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3217?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/i18n/Messages_en.properties
```diff
@@ -34,8 +34,8 @@ accessionState.Active=Active
 # Accession active/inactive states (active = seeds are still available, inactive = used up)
 accessionState.Inactive=Inactive
 application.failure.modulesIncomplete=Application modules are incomplete.
-# {0} is a country name; {1} and {2} are numbers
-applicationPreScreen.failure.badSize=Projects in {0} must be between {1} and {2} hectares.
+# {1} is the project type; {1} is a country name; {2} and {3} are numbers
+applicationPreScreen.failure.badSize={0} projects in {1} must be between {2} and {3} hectares.
```
sgrimm: I don't think this is going to be very i18n-friendly. I assume the project type is "Terrestrial" or "Mangrove." In English, you can use both of those words as adjectives with any noun. But in some languages, adjectives change depending on what nouns they're modifying. Spanish is one of them: adjectives have singular and plural forms as well as feminine and masculine forms.

The way to do this such that it'll be grammatically correct is to have a separate string for each project type rather than using a placeholder like this. That way translators can make the project type agree with the word for "projects." For example,

```
applicationPreScreen.failure.badSize.Mangrove=Mangrove projects in {0} must be between {1} and {2} hectares.
```

sgrimm (dismissed):

ice1080 (approved):

tommylau523:
@ice1080 Requesting a new review because this PR has completely changed due to requirements.

src/main/kotlin/com/terraformation/backend/accelerator/db/ApplicationStore.kt
```diff
@@ -705,28 +703,25 @@ class ApplicationStore(
...
+      val minimumHectares =
+          perCountryMinimumTotalHectares[projectCountriesRow.code!!] ?: defaultMinimumHectares
+      val minimumMangroveHectares = perCountryMinimumMagroveHectares[projectCountriesRow.code!!]
+
+      val mangroveLandUseArea =
+          applicationVariableValues.landUseModelHectares[LandUseModelType.Mangroves]?.toDouble()
+              ?: 0.0
+
+      if ((totalLandUseArea < minimumHectares || totalLandUseArea > defaultMaximumHectares) &&
+          (minimumMangroveHectares?.let { mangroveLandUseArea < it } ?: true)) {
```
ice1080: Is this supposed to be `||` instead of `&&`?

tommylau523: This is supposed to be `||`. The idea is that it has to either meet the total minimum, or meet the mangrove minimum. I was thinking of ways to make this look better haha.

tommylau523: Maybe something like

if ((total < minTotal && mangrove < minMangrove) || total > maxTotal)
would this be clearer?

ice1080: I don't think order necessarily needs to change, but right now it's saying this:
> if total land use is too small or large, AND mangrove land use is less than min then throw an error

But I'm saying that's incorrect and the AND should be an OR, like this:
```kt
      if (totalLandUseArea < minimumHectares || totalLandUseArea > defaultMaximumHectares ||
          (minimumMangroveHectares?.let { mangroveLandUseArea < it } ?: true)) {
```

tommylau523: "if total land use is too small or large, AND mangrove land use is less than min then throw an error"

This statement is correct. One case we can think of is, if total land use is below the requirement, but mangrove land use is above the minimum, we want to pass.

ice1080: So the min Mangrove check is only supposed to happen when total land use is too small or large?

tommylau523: We can think of it that way. The requirement specified that one of the following has to pass
1) total > 15k
2) total > 3k for select countries
3) mangrove > 1k for select countries. 

But I think there may be an issue here with bypassing the max.

ice1080: Ya I follow what you're saying with the max. I was talking about the mangrove min, but I didn't notice this requirement in the story:
> An application passes the prescreen if they pass any one of the 3 requirements

So I think we're good, unless you want to adjust the max

ice1080 (commented):
This looks good to me. Not sure if that graphite comment is valid or note

tommylau523:
### Merge activity

* **Jul 17, 9:44 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3217).

PR #3252, opened 2025-07-29
Disable H2 console

sgrimm:
By default, Spring Boot supports an admin console for the embedded H2 database
engine. We don't directly use H2, but it's used by one of our dependencies, and
Spring detects its presence and sets up the console. The console isn't accessible
because there's no access control rule for it in our security configuration and
we have a deny-by-default security policy, so it's doing nothing but eating memory
and wasting time at application start. Turn it off.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -61,6 +61,10 @@ spring:
       jsonColumnType: "JSONB"
       uuidColumnType: "UUID"
 
+  h2:
+    console:
+      enabled: false
+
   jackson:
     default-property-inclusion: non_absent
 


sgrimm:
* **#3252** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3252?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3252?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

ice1080:
How did you notice that it was enabled? For some reason I thought the h2 console was disabled by default

sgrimm:
Ah, I didn't dig enough. Looks like it's a dev-only dependency. I was seeing a log message about it in my dev environment and I assumed it was enabled in production too. I'm actually not sure why we need that as a dev-only dependency in the first place. I added it as part of the Spring Boot 3 upgrade; maybe it was recommended in a migration guide or something. I'll try removing it and see what happens.

sgrimm:
Ah, I see, it's needed for `./gradlew generateOpenApiDocs` which starts up the server without a database, but there are some components that need a database so we give them a dummy in-memory H2 one. So this change is still fine; it'll be a no-op but harmless in staging/prod.

PR #3253, opened 2025-07-29
Allow table names as search prefixes

sgrimm:
Previously, the root prefix in a search API call had to be specified as a sublist
path starting from the `organizations` search table. This could be annoying
when searching against tables several steps removed from organizations.

Update the search API such that you can specify the root prefix either as a
sublist path or as the name of a search table as defined in the `SearchTables`
class. This only applies to the root prefix; the relative prefixes in field names
still have to be paths starting from the root prefix.

We can update clients to use table names instead of paths, after which we can get
rid of path-based prefixes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
@@ -31,10 +31,10 @@ import org.springframework.web.bind.annotation.RestController
 class SearchController(
     private val clock: Clock,
     private val messages: Messages,
-    tables: SearchTables,
-    private val searchService: SearchService
+    private val searchService: SearchService,
+    private val searchTables: SearchTables
 ) {
-  private val organizationsTable = tables.organizations
+  private val organizationsTable = searchTables.organizations
 
   @Operation(summary = "Searches for data matching a supplied set of search criteria.")
   @PostMapping
@@ -153,7 +153,7 @@ class SearchController(
   private fun resolvePrefix(prefix: String?): SearchFieldPrefix {
     val table =
         if (prefix != null) {
-          organizationsTable.resolveTable(prefix)
+          searchTables[prefix] ?: organizationsTable.resolveTable(prefix)
         } else {
           organizationsTable
         }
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3253** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3253?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
```diff
@@ -153,7 +153,7 @@ class SearchController(
   private fun resolvePrefix(prefix: String?): SearchFieldPrefix {
     val table =
         if (prefix != null) {
-          organizationsTable.resolveTable(prefix)
+          searchTables[prefix] ?: organizationsTable.resolveTable(prefix)
         } else {
```
ice1080, line 157: Technically this affects some search tables whose name matches their sublist name in the `OrganizationsTable`, e.g. `batches`, `draftPlantingSites`, `facilities`, `inventories`, etc because of the join logic. Whether that's an issue or not for existing queries, I'm not sure 🤔

sgrimm: This is just for root prefixes, not for prefixes on field names.

sgrimm: But my PR description could make that distinction clearer.

ice1080: I think my comment actually still applies. If there existed a search request like this:
```json
"prefix": "batches",
"fields": ["batchNumber"]
```

then before this would have gotten the batches that belong to the org, but now this would retrieve all batches that match the `BatchesTable.conditionForVisibility`

ice1080: This is only true because the searchTable name for BatchesTable is `batches`, and the sublist for batches on the OrganizationsTable is also named `batches`. I think technically to not break anything you'd have to flip the line to this:
```kotlin
organizationsTable.resolveTableOrNull(prefix) ?: searchTables[prefix]
```

Not necessarily suggesting that because it has other issues since you would still want to throw an error if neither of them found it, just mentioning it as an example.

sgrimm: `resolvePrefix()` doesn't return any context information about how the prefix resolved to a particular search table, though; it just wraps a single `SearchTable` in a `SearchPrefix` that has no sublists. A request of
```
"prefix": "batches",
"fields": ["batchNumber"]
```
with no search filters will return all visible batch numbers with or without this change (just tried it on staging to make sure I wasn't missing something).

ice1080: Yup you're absolutely right, I totally forgot that it's just setting the root table for the prefix, not getting the table relative to the OrganizationsTable

ice1080 (approved):

PR #3255, opened 2025-07-30
Create Accelerator Projects DB View

ice1080:
Create a DB view to consolidate all the definitions of accelerator projects/orgs.

This may later be dropped in favor of data model changes to simplify the definition.

### DIFF
--- a/src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
+++ b/src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
@@ -16,6 +16,5 @@ FROM (SELECT p.id    as project_id,
                          ON p.id = pad.project_id) as accelerator_projects
 WHERE application_id IS NOT NULL
    OR participant_id IS NOT NULL
-   OR cohort_id IS NOT NULL
    OR has_org_accelerator_tag IS TRUE
 ;
--- a/src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
+++ b/src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
@@ -4,22 +4,18 @@ FROM (SELECT p.id    as project_id,
              a.id    as application_id,
              part.id as participant_id,
              part.cohort_id,
-             CASE
-                 WHEN EXISTS (SELECT 1
-                              FROM organization_internal_tags oit
-                                       JOIN internal_tags it ON it.id = oit.internal_tag_id
-                              WHERE oit.organization_id = o.id
-                                AND it.name = 'Accelerator') THEN true
-                 ELSE false
-                 END as has_org_accelerator_tag
+             EXISTS (SELECT 1
+                     FROM organization_internal_tags oit
+                     WHERE oit.organization_id = p.organization_id
+                       AND oit.internal_tag_id = 4)
+                     as has_org_accelerator_tag
       FROM projects p
-               JOIN organizations o ON o.id = p.organization_id
                LEFT JOIN accelerator.applications a ON p.id = a.project_id
                LEFT JOIN accelerator.participants part ON p.participant_id = part.id
                LEFT JOIN accelerator.project_accelerator_details pad
                          ON p.id = pad.project_id) as accelerator_projects
 WHERE application_id IS NOT NULL
    OR participant_id IS NOT NULL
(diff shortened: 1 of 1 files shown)


ice1080:
* **#3255** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3255?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
```diff
@@ -0,0 +1,25 @@
...
+             part.id as participant_id,
+             part.cohort_id,
+             CASE
+                 WHEN EXISTS (SELECT 1
+                              FROM organization_internal_tags oit
+                                       JOIN internal_tags it ON it.id = oit.internal_tag_id
+                              WHERE oit.organization_id = o.id
+                                AND it.name = 'Accelerator') THEN true
+                 ELSE false
+                 END as has_org_accelerator_tag
```
sgrimm: Can this just be `EXISTS (SELECT 1 ...) AS has_org_accelerator_tag`, without the `CASE`? Since `EXISTS (...)` is a boolean expression, I think that should work.

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
```diff
@@ -0,0 +1,25 @@
...
+      FROM projects p
+               JOIN organizations o ON o.id = p.organization_id
```
sgrimm: I don't think you need to join with `organizations` -- you could use `p.organization_id` instead of `o.id` in the `EXISTS` clause above, and I think that's the only place the table is being referenced.

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
```diff
@@ -0,0 +1,25 @@
...
+               LEFT JOIN accelerator.applications a ON p.id = a.project_id
+               LEFT JOIN accelerator.participants part ON p.participant_id = part.id
+               LEFT JOIN accelerator.project_accelerator_details pad
+                         ON p.id = pad.project_id) as accelerator_projects
+WHERE application_id IS NOT NULL
+   OR participant_id IS NOT NULL
+   OR COHORT_ID IS NOT NULL
```
sgrimm: Nit: Lower-case the column name (not that it really matters, but for consistency)

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM (SELECT p.id    as project_id,
+             a.id    as application_id,
+             part.id as participant_id,
+             part.cohort_id,
+             CASE
+                 WHEN EXISTS (SELECT 1
+                              FROM organization_internal_tags oit
+                                       JOIN internal_tags it ON it.id = oit.internal_tag_id
+                              WHERE oit.organization_id = o.id
+                                AND it.name = 'Accelerator') THEN true
```
sgrimm: I think it's okay to hardwire the internal tag ID here and remove the join with `internal_tags`, since the ID is a constant.

sgrimm (changes requested):

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
ice1080: Ah ya I was wondering about that

src/main/resources/db/migration/0350/V391__AcceleratorProjectsView.sql
ice1080: Ah ya good thought

sgrimm (approved):

ice1080:
Oh actually before I merge that, @sgrimm is it possible for a project to have a participant without a cohort and still be an accelerator project? Or should I remove the `participant_id is not null` check?

sgrimm (approved):

PR #3251, opened 2025-07-28
SW-7192 Add sublist criteria support to search api

ice1080:
Update NestedQueryBuilder to support sublist querying in the search api. This is accomplished by using the previously-passed `criteria` to add a search condition to a sublist's multisetQuery.

Also update `SeachController` to add the necessary property to the payload for sublist searching. 

(this should all be discussed with Steve to ensure there are no other potential issues with adding this support)

(if this approach is confirmed to be the way we want to go, better documentation is needed on NestedQueryBuilder since the top large comment says that sublist filtering is not currently supported - lines 266-285)

PR #3249, opened 2025-07-28
SW-7192 Pass criteria map and move searchService methods to NestedQueryBuilder

ice1080:
More preparation for sublist criteria support. Attempting to isolate changes to make it easier for review of downstream PRs.

PR #3248, opened 2025-07-28
SW-7192 Change criteria in search service to be a map of prefix to SearchNode

ice1080:
This is some pre-work for making sublist searching work in the search api, which has been accomplished in this PR stack by allowing `criteria` to be a map of prefix to SearchNode, and then applying a filter condition when adding each select field (if relevant) instead of only at the end of query building. See downstream PRs for more details.

PR #3256, opened 2025-07-31
SW-7201 Make Project Variable Views support list variables and projects without values

ice1080:
Previously the `project_variable_values` view would not return any projects that didn't have any values set, and also didn't work with list variables. Even though the view is being created to assist with the Matrix View feature which won't support list variables, this change will better support the search api moving forward.

Create a new view `project_variables` that is based on the recent view `accelerator_projects` that retrieves the latest variables for each project, and adds `is_list` to its columns. 

Recreate the view `project_variable_values` to include `list_position` and both `url` and `title` for link variables, as well as return the scalar values for variables to simplify the search api.

Update the search api to match the new structure.

Note that table variables and section variables may cause issues with this approach, and can be excluded in the view if preferred.

Requires changes to FE to match new search api structure (https://github.com/terraware/terraware-web/pull/4314).


ice1080:
* **#3257** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3251** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3250** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3250?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3249** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3248** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3248?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3256** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariableValuesTable.kt
```diff
@@ -48,56 +44,12 @@ class ProjectVariableValuesTable(tables: SearchTables) : SearchTable() {
...
-                  DSL.select(VARIABLE_SELECTS.IS_MULTIPLE)
-                      .from(VARIABLE_SELECTS)
-                      .where(
-                          VARIABLE_SELECTS.VARIABLE_ID.eq(PROJECT_VARIABLE_VALUES.VARIABLE_ID)))),
+          integerField("listPosition", PROJECT_VARIABLE_VALUES.LIST_POSITION),
+          textField("textValue", PROJECT_VARIABLE_VALUES.TEXT_VALUE),
+          bigDecimalField("numberValue", PROJECT_VARIABLE_VALUES.NUMBER_VALUE),
+          dateField("dateValue", PROJECT_VARIABLE_VALUES.DATE_VALUE),
+          uriField("linkUrl", PROJECT_VARIABLE_VALUES.LINK_URL),
+          textField("linkTitle", PROJECT_VARIABLE_VALUES.LINK_TITLE),
```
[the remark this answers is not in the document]
ice1080: That field is already in the list

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
```diff
@@ -497,10 +497,14 @@ val EMBEDDABLES =
...
             .withColumns("project_id", "land_use_model_type_id"),
+        EmbeddableDefinitionType()
+            .withName("project_variable_id")
+            .withTables("project_variables")
+            .withColumns("project_id", "stable_id"),
         EmbeddableDefinitionType()
             .withName("project_variable_value_id")
             .withTables("project_variable_values")
-            .withColumns("project_id", "stable_id"),
+            .withColumns("project_id", "variable_value_id"),
```
sgrimm: Is this needed? `variable_value_id` should be unique already.

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariableValuesTable.kt
sgrimm: Can we alphabetize the list of fields for consistency with the other search tables?

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariablesTable.kt
```diff
@@ -0,0 +1,57 @@
...
+
+class ProjectVariablesTable(tables: SearchTables) : SearchTable() {
+  override val primaryKey: TableField<out Record, out Any?>
+    get() = PROJECT_VARIABLES.PROJECT_VARIABLE_ID
+
+  override val sublists: List<SublistField> by lazy {
+    with(tables) {
+      listOf(
+          projectVariableValues.asMultiValueSublist(
+              "variableValues",
```
sgrimm: `values` would also be an okay name for this to keep it shorter.

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariablesTable.kt
```diff
@@ -0,0 +1,57 @@
...
+
+  override val fields: List<SearchField> =
+      listOf(
+          idWrapperField("projectId", PROJECT_VARIABLES.PROJECT_ID) { ProjectId(it) },
+          idWrapperField("variableId", PROJECT_VARIABLES.VARIABLE_ID) { VariableId(it) },
+          stableIdField("stableId", PROJECT_VARIABLES.STABLE_ID),
+          textField("variableName", PROJECT_VARIABLES.NAME),
+          booleanField("isList", PROJECT_VARIABLES.IS_LIST),
+          booleanField("isMultiSelect", PROJECT_VARIABLES.IS_MULTI_SELECT),
+          enumField("variableType", PROJECT_VARIABLES.VARIABLE_TYPE_ID, false),
```
sgrimm: My hunch is that if we start using the search API for interacting with variables/values, we'll eventually want to split these out into a separate `variables` search table and link it to this table as a single-value sublist so we can query against the full list of variables whether or not any projects have values for them.

But that'll require thinking about permissions, which will be a whole other rabbit hole, so this is good for now. If/when we go that route, we can put `AliasField`s here for backward compatibility.

src/main/resources/db/migration/0350/V392__ProjectVarValsAccelerator.sql
```diff
@@ -0,0 +1,52 @@
...
+CREATE OR REPLACE VIEW accelerator.project_variables AS
+WITH latest_variables AS (SELECT stable_id,
+                                 MAX(id) as variable_id
+                          FROM docprod.variables
+                          GROUP BY stable_id),
+     project_variables AS (SELECT project_id,
+                                  lv.variable_id
+                           FROM docprod.variable_values vv
+                                    JOIN latest_variables lv ON lv.variable_id = vv.variable_id
+                           GROUP BY project_id, lv.variable_id)
```
sgrimm: Alternately: use `SELECT DISTINCT` instead of `GROUP BY`. Either way works, but I think the former maybe makes the intent a little clearer.

src/main/resources/db/migration/0350/V392__ProjectVarValsAccelerator.sql
```diff
@@ -0,0 +1,52 @@
+DROP VIEW accelerator.project_variable_values;
+
+CREATE OR REPLACE VIEW accelerator.project_variables AS
```
sgrimm, line 3: Might be worth adding this view to `R__Comments.sql` to document that it includes all accelerator projects, whether or not they have any variables set. In the absence of having been part of discussing this, I would find it surprising to query this and get back rows with null variable IDs.

src/main/resources/db/migration/0350/V392__ProjectVarValsAccelerator.sql
```diff
@@ -0,0 +1,52 @@
...
+       vlv.title as link_title
+FROM docprod.variable_values vv
+         JOIN latest_values lv on (
+    lv.project_id = vv.project_id
+        AND lv.variable_id = vv.variable_id
+        AND lv.list_position = vv.list_position
+        AND lv.variable_value_id = vv.id
+    )
+         LEFT JOIN docprod.variable_link_values vlv ON vlv.variable_value_id = vv.id
+WHERE vv.is_deleted <> true
```
sgrimm: Alternately: `WHERE NOT vv.is_deleted`

src/main/resources/db/migration/0350/V392__ProjectVarValsAccelerator.sql
```diff
@@ -0,0 +1,52 @@
...
+       vv.date_value,
+       vlv.url   as link_url,
+       vlv.title as link_title
+FROM docprod.variable_values vv
+         JOIN latest_values lv on (
+    lv.project_id = vv.project_id
+        AND lv.variable_id = vv.variable_id
+        AND lv.list_position = vv.list_position
+        AND lv.variable_value_id = vv.id
+    )
```
sgrimm: Alternately (may be easier to read, or may not be, your call):
```
JOIN latest_values lv ON (
    (lv.project_id, lv.variable_id, lv.list_position, lv.variable_value_id) =
    (vv.project_id, vv.variable_id, vv.list_position, vv.id)
)
```

src/main/resources/i18n/Messages_en.properties
```diff
@@ -769,20 +769,22 @@ search.projectDeliverables.status=Deliverable status
 search.projectDeliverables.submissionId=Deliverable submission ID if submission exists
 search.projectDeliverables.type=Deliverable type
 search.projectLandUseModelTypes.landUseModelType=Project land use model type
+search.projectVariables.isList=Boolean for List Variables
+search.projectVariables.isMultiSelect=Boolean for Select Variables
```
sgrimm: I try to word the field names in this file as if they're column headers in a CSV export (which is one thing they're actually used for), so maybe something like, "Variable is list" would be more succinct here.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
```diff
@@ -0,0 +1,249 @@
...
+
+  private val clock = TestClock()
+  private val searchService: SearchService by lazy { SearchService(dslContext) }
+  private val searchTables = SearchTables(clock)
+
+  @BeforeEach
+  fun setUp() {
+    insertOrganization()
+    insertOrganizationUser(
+        userId = inserted.userId, organizationId = inserted.organizationId, role = Role.Admin)
```
sgrimm: No need for `organizationId` here (the most recently inserted one is the default).

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
```diff
@@ -0,0 +1,249 @@
...
+    val dateVariableId = insertVariable(stableId = dateStableId, type = VariableType.Date)
+    val oldLinkVariableId = insertVariable(stableId = linkStableId, type = VariableType.Link)
+    val newLinkVariableId =
+        insertVariable(
+            stableId = linkStableId,
+            type = VariableType.Link,
+            replacesVariableId = oldLinkVariableId)
+
+    // Add a connection between the project and each variable
+    insertValue(variableId = newVariableId, projectId = projectId, textValue = "NewVarOldVal")
```
sgrimm: Could leave `projectId` out of these calls since it's the most recently inserted project which is the default.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
```diff
@@ -0,0 +1,249 @@
...
+                mapOf(
+                    "projectId" to "$projectId",
+                    "stableId" to dateStableId,
+                    "variableId" to "$dateVariableId",
+                    "variableType" to "Date",
+                ),
+            ),
+            cursor = null)
+
+    val actual = Locales.GIBBERISH.use { searchService.search(prefix, fields, NoConditionNode()) }
```
sgrimm, line 109: It's harmless, but curious why you're switching locales here.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
```diff
@@ -0,0 +1,249 @@
...
+
+    assertJsonEquals(expected, actual)
+  }
+
+  @Test
+  fun `can retrieve nested variableValues from projects`() {
```
sgrimm: Maybe I'm missing it, but I don't see this test fetching any values.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableValueSearchTest.kt
```diff
@@ -170,6 +245,34 @@ class ProjectVariableValueSearchTest : DatabaseTest(), RunsAsUser {
     assertJsonEquals(expected, actual)
   }
 
+  @Test
+  fun `old variable with value doesn't show up when new version does not have value`() {
```
sgrimm, line 228: Worth having a test case for deleted values?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariablesTable.kt
ice1080: Isn't the same basically true about DB views in general though? That you can always just query the tables directly (ignoring permissions)? To me having a search table that has the latest values per variable for each project doesn't mean that you can't also have a search table(s) for variables and/or values

sgrimm: I was too vague; by "permissions" I meant that we'd need to figure out which users have permission to query the list of variable definitions since a theoretical `variables` table won't have any project or organization IDs to hang permission checks on.

But maybe "all users can see all variable definitions" is the answer; I guess that's how the variable listing endpoints work right now anyway.

ice1080: Indeed, but I was more responding to your comment of:
> if we start using the search API for interacting with variables/values, we'll eventually want to split these out into a separate variables search table and link it to this table as a single-value sublist.

In my opinion there doesn't have to be only one search table to access a field, similar to how views are basically additional ways to access data. But maybe there's other considerations I'm missing.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
ice1080: Ah that was a copy of another test. Looks like we have several other tests that do that, so I must have copied it from that. I'll clean up these tests per the feedback, and then a separate PR should be used to fix the rest of them.

src/test/kotlin/com/terraformation/backend/accelerator/ProjectVariableSearchTest.kt
ice1080: Another artifact of copying. I think I saw 7 other tests doing it so assumed that we were wanting to do that for most searchService tests.

jooq/src/main/kotlin/com/terraformation/backend/jooq/Config.kt
ice1080: Good thought, will remove and update primary key on search table.

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariablesTable.kt
sgrimm: Ah, I see what you mean. It's definitely possible to have fields in more than one place. In this case I'm not sure it buys us a lot since from the client's point of view the only difference is whether you pass a field name of, say, `variableName` vs. `variable_name` (to do a flattened reference to the sublist).

Anyway, definitely not worth worrying about in the context of this PR. It's easy to revise later if we need to.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/table/ProjectVariablesTable.kt
ice1080: Ahh good point. I was thinking that making this change would require FE changes to keep things working, but so long as sublists are done correctly and `AliasField`s are added, I see what you mean now.

ice1080:
Because of the changes here to the search structure of project variables and values, I've got a corresponding FE PR that I'll wait to be approved before merging this PR.

PR #3302, opened 2025-08-15
SW-7256 Add permissions check for addInternalUser

ice1080:

PR #3301, opened 2025-08-15
SW-7256 Rename column from user to internal to match table

ice1080:
Rename column to match table name that it references.
Fix a comment for multiple tf contacts.

PR #3303, opened 2025-08-15
SW-7256 Store/Service updates for project internal users

ice1080:
Update ProjectService and ProjectStore to allow retrieval and insert of project internal users.


ice1080:
* **#3304** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3304?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3303** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3303?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3302** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3301** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3301?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -29,4 +45,18 @@ class ProjectService(
       plantingSiteStore.assignProject(projectId, plantingSiteIds)
     }
   }
+
+  fun addInternalUserRole(
+      projectId: ProjectId,
+      userId: UserId,
```
ice1080, line 41: I couldn't decide if this should be in ProjectStore or ProjectService. Happy to switch it if preferred.

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -29,4 +45,18 @@ class ProjectService(
...
+      role: ProjectInternalRole? = null,
+      roleName: String? = null
+  ) {
+    val user = userStore.fetchOneById(userId)
+    if (user.globalRoles.isEmpty()) {
+      throw IllegalStateException("Only global users can be added as project internal roles.")
```
sgrimm: Kind of a nit and maybe even me being wrong about terminology since the product folks have gone all-in on it, but "global user" isn't really a concept in the code or data model. There are users, and there are global roles (where the term "global" just means the roles aren't scoped to specific organizations) and users can have global roles. If I were writing this, I'd probably make it say something like, "User has no global roles."

But I'm biased because I came up with the term "global role," and "global user" has always kind of rubbed me the wrong way because it sort of implies the underlying user is of a different type. So you can ignore this if you like.

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
```diff
@@ -16,7 +20,19 @@ class ProjectService(
...
+    private val userStore: UserStore,
 ) {
+  companion object {
+    /** These ProjectInternalRoles should also become TF Contacts. */
+    private val TF_CONTACT_PROJECT_ROLES =
+        setOf(ProjectInternalRole.ProjectLead, ProjectInternalRole.RestorationLead)
+
+    fun roleShouldBeTfContact(role: ProjectInternalRole?): Boolean {
+      return role in TF_CONTACT_PROJECT_ROLES
+    }
```
sgrimm: This might be cleaner as an extension method on `ProjectInternalRole`.

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
sgrimm: I think it's fine here. We violate this rule, but in general I think we should avoid store classes calling other store classes just so we don't end up with circular dependencies.

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -151,4 +223,12 @@ class ProjectStore(
...
     }
   }
+
+  private fun fetchOrgIdByProjectId(projectId: ProjectId): OrganizationId {
+    return dslContext
+        .select(PROJECTS.ORGANIZATION_ID)
+        .from(PROJECTS)
+        .where(PROJECTS.ID.eq(projectId))
+        .fetchOne(PROJECTS.ORGANIZATION_ID) ?: throw ProjectNotFoundException(projectId)
+  }
```
sgrimm: `ParentStore` already has a method for this (and, despite my comment about stores calling stores, `ParentStore` is safe to call from here since it only depends on `DSLContext`.)

src/test/kotlin/com/terraformation/backend/customer/ProjectServiceTest.kt
```diff
@@ -135,6 +162,13 @@ class ProjectServiceTest : DatabaseTest(), RunsAsUser {
     insertFacility()
   }
 
+  @Test
+  fun `Role should be tf contact`() {
```
sgrimm: Nit: method names should generally start with lower-case letters.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/customer/ProjectService.kt
ice1080: Ya I feel conflicted here about the implentation overall, because we might one day add a global role that should _not_ be able to be a project internal user...

src/test/kotlin/com/terraformation/backend/customer/ProjectServiceTest.kt
ice1080: Fixed, though as a note we have 19 tests that don't match this paradigm.
Regex to see them - `@Test\n? *fun .?[A-Z]`

src/test/kotlin/com/terraformation/backend/db/ProjectInternalRoleTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+
+internal class ProjectInternalRoleTest {
+
+  @Test
+  fun `role should be tf contact`() {
+    assertTrue { ProjectInternalRole.ProjectLead.shouldBeTfContact() }
+    assertTrue { ProjectInternalRole.RestorationLead.shouldBeTfContact() }
+    assertFalse { ProjectInternalRole.Consultant.shouldBeTfContact() }
```
sgrimm: These assertions (or at least the two `assertTrue` ones) should have messages so we know which one failed without having to look at the line numbers in the stack traces.

sgrimm (approved):

ice1080:
### Merge activity

* **Aug 19, 5:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3303).
* **Aug 19, 6:05 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3303) rebased this pull request as part of a merge.
* **Aug 19, 6:13 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3303).

PR #3307, opened 2025-08-18
SW-7265 Add Admin UI to Delete Org Users

ice1080:
We've never had a way to delete users from organizations that they don't belong to. Now that orgs can have multiple TF Contacts, these can start to pile up in an org because adding a new one no longer replaces the old one.
Add a section to the Admin UI (and corresponding endpoint) to delete users from an organization, only for Super Admins.

Currently this does not allow deleting the user if it's the last user in an org, but this may be added if product prefers.

### DIFF
--- a/src/main/resources/templates/admin/index.html
+++ b/src/main/resources/templates/admin/index.html
@@ -149,6 +149,22 @@
         </form>
     </th:block>
 
+    <th:block th:if="${canRemoveOrganizationUser}">
+        <h2>Remove Organization User</h2>
+
+        <form method="POST" action="/admin/removeOrganizationUser">
+            <label for="removeOrgUserEmail">Email (user must already exist)</label>
+            <input id="removeOrgUserEmail" type="email" name="email" required/>
+            <label for="removeOrgUserOrganizationId">Organization</label>
+            <select id="removeOrgUserOrganizationId" name="organizationId">
+                <option th:each="organization : ${allOrganizations}" th:value="${organization.id}"
+                        th:text="|${organization.name} (${organization.id})|">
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3307** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3307?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3307?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/templates/admin/index.html
```diff
@@ -149,6 +149,22 @@ <h2>Add Organization User</h2>
         </form>
     </th:block>
 
+    <th:block th:if="${canRemoveOrganizationUser}">
+        <h2>Remove Organization User</h2>
+
```
sgrimm, line 154: Maybe add a blurb here to note that you can only remove users who are listed as TF contacts for the organization. If you try to remove a regular user, I believe the permission check will fail.

sgrimm (dismissed):

src/main/resources/templates/admin/index.html
ice1080: Ah thanks for noting this. This was the last addition to the PRD and I didn't look closely enough at the existing permissions to realize that only members of an org can remove other members. This new request on the PRD is actually to allow super admins to be able to remove any member from any org, so I'll update to make that work and re-request review

sgrimm (approved):

PR #3309, opened 2025-08-19
Remove alias on search sql sublists

ice1080:
The addition of the alias to search sublists causes some search queries to be very slow. Remove this until it can be added correctly.

PR #3310, opened 2025-08-19
Remove unnecessary left joins from sublist filtering

ice1080:
When aliases were added to sublists in the search api, they can cause queries to take 2 orders of magnitude longer, which sometimes causes health checks to fail and connections to reset. Aliasing on sublist tables was removed in https://github.com/terraware/terraware-server/pull/3309 to fix this, but this causes sublist filtering to not be possible, thus breaking the Matrix View which is the first (and currently only) usage of sublist filtering. However sublist filtering actually only needed aliasing because of an extra left join on the same table that is being queried. 
Remove this unnecessary left join, only for sublist filtering.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -4,6 +4,7 @@ import com.terraformation.backend.search.field.AliasField
 import com.terraformation.backend.search.field.SearchField
 import com.terraformation.backend.search.table.AccessionsTable
 import com.terraformation.backend.util.MemoizedValue
+import kotlin.random.Random
 import org.jooq.Condition
 import org.jooq.DSLContext
 import org.jooq.Field
@@ -848,7 +849,7 @@ class NestedQueryBuilder(
       val queryBuilder = NestedQueryBuilder(dslContext, prefix.withSublist(sublistName))
       if (criteria != null) {
         queryBuilder.addCondition(
-            filterResults(SearchFieldPrefix(relativeField.searchTable), criteria))
+            filterResults(SearchFieldPrefix(relativeField.searchTable), criteria, true))
       }
       queryBuilder
     }
@@ -1126,10 +1127,16 @@ class NestedQueryBuilder(
     }
   }
 
-  /** Returns a condition that filters search results based on a list of criteria. */
+  /**
+   * Returns a condition that filters search results based on a list of criteria.
+   *
+   * Aliasing the table name can cause slow search results if used on high level subqueries. It
+   * should only be added to subqueries inside multisets, so [includeAlias] is false by default.
+   */
   fun filterResults(
       rootPrefix: SearchFieldPrefix,
       criteria: SearchNode? = null,
+      includeAlias: Boolean = false,
   ): Condition {
     // Filter out results the user doesn't have the ability to see. NestedQueryBuilder will include
     // the visibility check on the root table, but not on parent tables.
@@ -1142,9 +1149,12 @@ class NestedQueryBuilder(
 
     val primaryKey = rootTable.primaryKey
 
+    val random = Random.nextInt()
+    val alias = "${rootTable.fromTable.name}_${random}"
+    val fromTable = if (includeAlias) rootTable.fromTable.`as`(alias) else rootTable.fromTable
+
     val subquery =
-        joinWithSecondaryTables(
-                DSL.select(primaryKey).from(rootTable.fromTable), rootPrefix, rootCriterion)
+        joinWithSecondaryTables(DSL.select(primaryKey).from(fromTable), rootPrefix, rootCriterion)
             .where(conditions)
 
     // Ideally we'd preserve the type of the primary key column returned by the subquery, but that
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -4,7 +4,6 @@ import com.terraformation.backend.search.field.AliasField
 import com.terraformation.backend.search.field.SearchField
 import com.terraformation.backend.search.table.AccessionsTable
 import com.terraformation.backend.util.MemoizedValue
-import kotlin.random.Random
 import org.jooq.Condition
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3310** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
```diff
@@ -1142,9 +1149,12 @@ class NestedQueryBuilder(
 
     val primaryKey = rootTable.primaryKey
 
+    val random = Random.nextInt()
+    val alias = "${rootTable.fromTable.name}_${random}"
```
sgrimm: Probably should have caught this on the earlier version of the PR, but: wouldn't you need to use this alias in place of the table name in the rest of the subquery too, not just in the `FROM` clause? Simple example to demonstrate what I mean:
```sql
SELECT mytable.col1
FROM mytable
WHERE mytable.id IN (
   SELECT mytable.id
   FROM mytable AS some_random_alias -- this PR adds the alias here
   WHERE mytable.col2 = 'search value'
)
```
But this would actually need to be
```sql
SELECT mytable.col1
FROM mytable
WHERE mytable.id IN (
   SELECT some_random_alias.id -- alias used here
   FROM mytable AS some_random_alias
   WHERE some_random_alias.col2 = 'search value' -- and here
)
```
Otherwise the `SELECT` and `WHERE` clauses will reference the values from the instance of the table in the outer query, not the subquery.

Or am I analyzing this wrong?

ice1080: 🤔  No I think you're right, and I think that's actually what caused the expensive queries before. I'll convert this to a draft and keep thinking about it. It may take me a while to get back into this headspace to be able to do it correctly. 
This PR technically fixes the sql error by aliasing tables that are showing up multiple times in a query, and it does somehow work according to all our unit tests (which imo are pretty thorough around the search service et al) but isn't actually building the sql "the right way" which bothers me.

sgrimm: Now I don't feel so puzzled about why I'd previously thought it'd be more complicated to add sublist filtering. It should definitely be possible, though.

Might also be worth figuring out if we can write a test that fails with the alias just in the `FROM` clause. I suspect the behavior difference is masked for older queries by the fact that the table in the outer query is being filtered by the criteria in the inner query, so the aliased table ends up just getting ignored in terms of picking which rows end up in the result. (But in the subquery, all its rows are fetched and joined since there are no filters on the aliased table, hence the slow queries.)

ice1080: Ya this will definitely be more complicated, since `filterResults` is using `rootTable.primaryKey`, `rootCriterion.toCondition`, `conditionForVisibility`, etc and all of these are referencing the original table not the aliased table. I'm not sure if there's a way to "convert" these from their original table to an alias with jooq; will keep looking into that.

Regarding writing a test though, I still don't understand how all of my tests worked at all if what you're saying is true. I have tests that show that the top level rows are _not_ filtered, only the results. I don't understand how that would work if the alias table is not being filtered. I agree that the sql seems like it should be filtering the top level rows, but the results seem to be doing what we actually want. Very curious

ice1080: As I'm looking at the sql that's generated, I think there may actually be a way to do this without needing an alias. I believe that it is still retrieving the correct results, but might be doing this with an unnecessary `left outer join` on the table it's querying on. And because that's the same table, the DB complains that the table is referenced twice.
Looking into how to fix just this.

src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
```diff
@@ -1142,9 +1148,15 @@ class NestedQueryBuilder(
 
     val primaryKey = rootTable.primaryKey
 
+    // the criteria for a sublist filter will already be in conditions instead of the secondary
+    // table joins since the conditions don't need the table joined
+    val secondaryTablesCriteria = if (isSublistFilter) NoConditionNode() else rootCriterion
```
ice1080, line 1153: @sgrimm this _seems_ to fix the problem without the need for an alias. What I haven't figured out yet is if `rootCriterian` might have other sublists referenced that would be used in `joinWithSecondaryTables` that are not based on the same table as the current sublist.

ice1080: I _think_ that with the way that sublist filtering works, `rootCriterion` will always only be needed for the `conditions` and thus shouldn't be in the `secondaryTablesCriteria`, but it's possible I'm not thinking through everything. Tests for this are in `SearchServiceSublistSearchTest` and include `AndNode` and `OrNode` for these

sgrimm: Sounds plausible to me, and I'm not immediately thinking of any gotchas. Let's make sure to test this in the web app with a meaningful amount of accession data on staging, though, to make sure the perf problem doesn't come back.

sgrimm (approved):
PR title could use a rewrite before merging, but the change seems reasonable.

src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
ice1080: I will definitely confirm, but I believe that this no longer touches the accession search sql since it is only for sublist filtering which is only used by the Matrix View currently.

ice1080: Confirmed. The release tag `v20250807.1` (which is before the search service updates were merged) and this branch both generate identical search sql for accessions (including a value in the search input on the FE). We are good to go

PR #3312, opened 2025-08-20
SW-7276 Update Data Model for Survival Rate

ice1080:
Update `observed_` tables to add `survival_rate` columns.

Add table for specifying t0 observation or estimated density for a given plot and species.


ice1080:
* **#3315** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3315?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3312** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3312?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3312?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
```diff
@@ -0,0 +1,33 @@
...
+CALL add_survival_rate_column('observed_site_species_totals');
+
+DROP PROCEDURE add_survival_rate_column;
+
+CREATE TABLE tracking.t0_plot
+(
+    id                         BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    monitoring_plot_id         BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    observation_id             BIGINT REFERENCES tracking.observations ON DELETE CASCADE,
+    species_id                 BIGINT REFERENCES species ON DELETE CASCADE,
```
sgrimm, line 11: Is the idea that there's either manually-entered per-species data (multiple rows) or a single row with an observation ID?

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
```diff
@@ -0,0 +1,33 @@
...
+    estimated_planting_density NUMERIC,
+
+    CONSTRAINT t0_plot_null_check CHECK (
+        (observation_id IS NULL AND estimated_planting_density IS NOT NULL AND
+         species_id IS NOT NULL) OR
+        (observation_id IS NOT NULL AND estimated_planting_density IS NULL AND species_id IS NULL)
+    ),
+
+    CONSTRAINT unique_plot_observation UNIQUE (monitoring_plot_id, observation_id),
```
sgrimm: Each monitoring plot should only pick a single observation as its t0 source, right? If so, then I think this needs to be done differently or we'll allow multiple observation IDs for the same plot.

A partial unique index will do the trick. I don't think you can do this as a unique constraint (the partial-index syntax isn't supported).
```sql
CREATE UNIQUE INDEX ON tracking.t0_plot (monitoring_plot_id)
WHERE observation_id IS NULL;
```

sgrimm (changes requested):

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
ice1080: That's correct!

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
```diff
@@ -0,0 +1,33 @@
...
+    BEGIN
+        EXECUTE 'ALTER TABLE tracking.' || table_name || '
+                ADD COLUMN survival_rate INTEGER';
+    END;
+$$ LANGUAGE plpgsql;
+
+CALL add_survival_rate_column('observed_plot_species_totals');
+CALL add_survival_rate_column('observed_subzone_species_totals');
+CALL add_survival_rate_column('observed_zone_species_totals');
+CALL add_survival_rate_column('observed_site_species_totals');
```
sgrimm: This seems like overkill to replace four one-line `ALTER TABLE` statements.

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
sgrimm: Gah, that should be `IS NOT NULL`, oops.

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
```diff
@@ -0,0 +1,33 @@
...
+
+DROP PROCEDURE add_survival_rate_column;
+
+CREATE TABLE tracking.t0_plot
+(
+    id                         BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    monitoring_plot_id         BIGINT NOT NULL REFERENCES tracking.monitoring_plots ON DELETE CASCADE,
+    observation_id             BIGINT REFERENCES tracking.observations ON DELETE CASCADE,
```
sgrimm, line 10: This column could use an index.

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
ice1080: Ah I see now. I was thinking that the unique constraint on `monitoring_plot_id, species_id` would handle this, but apparently it doesn't apply when `species_id` is null.

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
```diff
@@ -0,0 +1,28 @@
...
+        (observation_id IS NOT NULL AND estimated_planting_density IS NULL AND species_id IS NULL)
+    ),
+
+    CONSTRAINT unique_plot_species UNIQUE (monitoring_plot_id, species_id)
+);
+
+CREATE UNIQUE INDEX ON tracking.t0_plot (monitoring_plot_id)
+WHERE observation_id IS NOT NULL;
+
+CREATE INDEX ON tracking.t0_plot (monitoring_plot_id);
```
sgrimm: You don't need this index since there's already an index from the unique constraint on `(monitoring_plot_id, species_id)` that the DB can use for this. The other two are needed, though.

sgrimm (approved):

src/main/resources/db/migration/0350/V399__PlotSurvivalRate.sql
ice1080: Ahh that makes sense. I was only thinking about species_id which is not handled, but I understand now.

ice1080:
### Merge activity

* **Aug 22, 6:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3312).
* **Aug 22, 6:01 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3312).

PR #3319, opened 2025-08-22
Update Internal Users api to combine calls

ice1080:
As part of FE implementation, it was realized that it would be easier to update internal users with a single endpoint instead of one for adding a single one and removing a single one (especially since updating the role for a user required first calling the endpoint to remove it then calling the endpoint to add it with the new role).
Update the api to accept the full list of users and roles that will be Project Internal Users. Update the service to parse out existing users vs requested users in order to determine which to delete and which to add. For users that need to be updated, add them to both lists.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
@@ -9,14 +9,15 @@ import com.terraformation.backend.customer.event.ProjectInternalUserRemovedEvent
 import com.terraformation.backend.customer.event.ProjectRenamedEvent
 import com.terraformation.backend.customer.model.ExistingProjectModel
 import com.terraformation.backend.customer.model.NewProjectModel
+import com.terraformation.backend.customer.model.ProjectInternalUserModel
 import com.terraformation.backend.customer.model.ProjectModel
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.ProjectNameInUseException
 import com.terraformation.backend.db.ProjectNotFoundException
 import com.terraformation.backend.db.accelerator.ParticipantId
+import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.ProjectId
-import com.terraformation.backend.db.default_schema.ProjectInternalRole
 import com.terraformation.backend.db.default_schema.UserId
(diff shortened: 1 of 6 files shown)


ice1080:
* **#3319** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/db/ProjectStore.kt
```diff
@@ -132,38 +129,45 @@ class ProjectStore(
       with(PROJECT_INTERNAL_USERS) {
         dslContext
             .insertInto(this)
-            .set(PROJECT_ID, projectId)
-            .set(USER_ID, userId)
-            .set(PROJECT_INTERNAL_ROLE_ID, role)
-            .set(ROLE_NAME, roleName)
+            .columns(PROJECT_ID, USER_ID, PROJECT_INTERNAL_ROLE_ID, ROLE_NAME)
```
sgrimm: FYI, you can add the columns as additional arguments to `insertInto` rather than calling `.columns`.

src/main/kotlin/com/terraformation/backend/customer/model/ProjectInternalUserModel.kt
```diff
@@ -0,0 +1,10 @@
+package com.terraformation.backend.customer.model
+
+import com.terraformation.backend.db.default_schema.ProjectInternalRole
+import com.terraformation.backend.db.default_schema.UserId
+
+class ProjectInternalUserModel(
+    val userId: UserId,
+    val role: ProjectInternalRole? = null,
+    val roleName: String? = null,
```
sgrimm, line 9: Maybe add an `init` block to check that exactly one of these is non-null?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/customer/model/ProjectInternalUserModel.kt
ice1080: We have this checked at the DB level as well but I'm always a fan of more being more explicit

ice1080:
### Merge activity

* **Aug 22, 11:35 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3319).

PR #3330, opened 2025-08-25
SW-7295 Add Project Internal Users to Search API

ice1080:
The People table on the user side of Terraware needs to display all project internal users and their internal roles at an organization level. Instead of a separate query for internal users for each project in the organization, add the internal users to the search api.


ice1080:
* **#3330** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3330?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3330?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
```diff
@@ -21,6 +23,10 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
               "organizationMemberships",
               USERS.ID.eq(ORGANIZATION_USERS.USER_ID),
           ),
+          projectInternalUsers.asMultiValueSublist(
+              "internalUsers",
```
sgrimm: I don't have a suggestion I love, but this name seems a little unclear: it doesn't seem intuitive to me that a sublist of `users` called `internalUsers` gets you a list of that user's _projects_.

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
```diff
@@ -45,15 +51,29 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
...
+                    DSL.exists(
+                        DSL.selectOne()
+                            .from(PROJECTS)
+                            .join(PROJECT_INTERNAL_USERS)
+                            .on(PROJECTS.ID.eq(PROJECT_INTERNAL_USERS.PROJECT_ID))
+                            .where(USERS.ID.eq(PROJECT_INTERNAL_USERS.USER_ID))
+                            .and(
+                                PROJECTS.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)
+                            )
+                    ),
```
sgrimm, line 75: These "we only have project, but we actually need org" clauses are making me wonder if we should add an organization ID column to `project_internal_users`.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
ice1080: Oh that's my fault, I meant to rename that. I'll come up with something else.

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
ice1080: We could definitely add it. It's a bit redundant and could technically cause data integrity issues but I know we do the same with at least monitoring plots having both subzone id and site id

ice1080: I'm going to merge this PR as-is since that's a bit of scope creep for this change. Let me know if you think this change is worth doing in another PR. (alternatively could change to a view that is basically just doing a join on projects to get the org id)

ice1080:
### Merge activity

* **Aug 26, 3:37 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3330).

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
sgrimm: I don't think it's needed per se, but let's keep it in mind as a possibility if we keep writing SQL clauses like this one repeatedly. A view is a good thought and is probably the right place to start, but if we were to go with a physical table column, we could avoid the integrity issues by using a composite foreign key, like `FOREIGN KEY (project_id, organization_id) REFERENCES projects (id, organization_id)`. We'd need a unique key/index for those two columns on the projects table.

PR #3332, opened 2025-08-26
SW-7133 Added recorded plants multiset to monitoring plot observation results

tommylau523:
As part of the change, we added GPS coordinates to plants in the test scenarios. The results are compared with the inputs. The inputs are expected to sort by monitoring plot number. The results are guaranteed to follow the order of the inputs.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
@@ -59,6 +59,7 @@ import com.terraformation.backend.tracking.model.ObservationRollupResultsModel
 import com.terraformation.backend.tracking.model.ObservationSpeciesResultsModel
 import com.terraformation.backend.tracking.model.ObservedPlotCoordinatesModel
 import com.terraformation.backend.tracking.model.PlantingSiteDepth
+import com.terraformation.backend.tracking.model.RecordedPlantModel
 import com.terraformation.backend.tracking.model.ReplacementDuration
 import com.terraformation.backend.tracking.model.ReplacementResult
 import io.swagger.v3.oas.annotations.Operation
@@ -705,6 +706,16 @@ data class RecordedPlantPayload(
     val speciesName: String?,
     val status: RecordedPlantStatus,
 ) {
+  constructor(
+      model: RecordedPlantModel
+  ) : this(
+      certainty = model.certainty,
+      gpsCoordinates = model.gpsCoordinates,
+      speciesId = model.speciesId,
+      speciesName = model.speciesName,
+      status = model.status,
+  )
+
   fun toRow(): RecordedPlantsRow {
     return RecordedPlantsRow(
         certaintyId = certainty,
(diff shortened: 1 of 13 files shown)


tommylau523:
* **#3332** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3332?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3332?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -973,6 +977,34 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
         )
       }
 
+  private val recordedPlantsGpsField = RECORDED_PLANTS.GPS_COORDINATES.forMultiset()
+
```
ice1080: Does this need to be a separate variable? Seems like the rest of the columns are used the same amount of times

tommylau523: This is consistent with other GPS coordinates columns. I think it is less verbose than adding `.forMultiset()` to every usage.

ice1080: Ah gotcha, I thought that was only needed when defining it as a separate variable

ice1080 (approved):

tommylau523:
Appreciate the quick review @ice1080. I should have set it to draft but I still have to add unit tests. Will re-request a review once ready.

ice1080:
> Appreciate the quick review @ice1080. I should have set it to draft but I still have to add unit tests. Will re-request a review once ready.

No problem @tommylau523 . As a note you can create PRs as drafts without publishing in Graphite in the UI or cli, instead of publishing them and then converting them back to draft

tommylau523:
> > Appreciate the quick review @ice1080. I should have set it to draft but I still have to add unit tests. Will re-request a review once ready.
> 
> No problem @tommylau523 . As a note you can create PRs as drafts without publishing in Graphite in the UI or cli, instead of publishing them and then converting them back to draft

Noted. I tend to click publish right after without thinking haha. But it is now ready to be reviewed.

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -838,6 +849,7 @@ data class ObservationMonitoringPlotResultsPayload(
     val photos: List<ObservationMonitoringPlotPhotoPayload>,
     @Schema(description = "Number of live plants per hectare.") //
     val plantingDensity: Int,
+    val plants: List<RecordedPlantPayload>?,
```
sgrimm, line 887: Should this perhaps be fetched separately?

There can be hundreds of recorded plants in each monitoring plot, so adding the list here will potentially result in huge "get observation results" response payloads. We'll always download the entire plants lists for every monitoring plot even if the user is only viewing a list of site-level results in the UI.

tommylau523: I can add a "depth" option to give us some flexibility. I think we can default it to subzone since most of the time we don't even need plot level details.

tommylau523: Actually maybe search API is best for this.

tommylau523: I will add it to search in a followup PR. For now I added a depth settings.

PR #3331, opened 2025-08-26
SW-7278 Store t0 plot density for species in plot

ice1080:
Instead of storing either an observation or multiple species' densities, store a species density always. Use a separate table for storing observations use for t0 data. Calculate density when observations are used by using the total live plus total dead from that observation.

Set plot density to numeric because it will likely be specified in plants per hectare and we want to store in plants per plot.

PR #3333, opened 2025-08-26
SW-7279 Calculate Survival Rate for Observed Totals with T0 Density

ice1080:
Calculate survival rate in all `observed_<bla>_species_totals`​ tables using `plot_t0_density`​ table.

Add tests for this.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -2105,6 +2105,7 @@ class ObservationStore(
             plantCountsBySpecies,
             cumulativeDeadFromCurrentObservation,
         )
+        updateSiteSurvivalRate(observationId, plantingSiteId)
       }
     }
   }
@@ -2379,6 +2380,29 @@ class ObservationStore(
           .update(this)
           .set(SURVIVAL_RATE, TOTAL_LIVE.times(100).div(denominator))
           .where(PLANTING_ZONE_ID.eq(plantingZoneId))
+          .and(OBSERVATION_ID.eq(observationId))
+          .and(SPECIES_ID.isNotNull)
+          .execute()
+    }
+  }
+
+  private fun updateSiteSurvivalRate(
+      observationId: ObservationId,
+      plantingSiteId: PlantingSiteId,
+  ) {
+    with(OBSERVED_SITE_SPECIES_TOTALS) {
+      val denominator =
+          DSL.field(
+              DSL.select(DSL.sum(PLOT_T0_DENSITY.PLOT_DENSITY))
+                  .from(PLOT_T0_DENSITY)
+                  .where(PLOT_T0_DENSITY.monitoringPlots.PLANTING_SITE_ID.eq(plantingSiteId))
+                  .and(PLOT_T0_DENSITY.SPECIES_ID.eq(this.SPECIES_ID))
+          )
+
+      dslContext
+          .update(this)
+          .set(SURVIVAL_RATE, TOTAL_LIVE.times(100).div(denominator))
+          .where(PLANTING_SITE_ID.eq(plantingSiteId))
           .and(OBSERVATION_ID.eq(observationId))
           .and(SPECIES_ID.isNotNull)
           .execute()
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -2094,6 +2094,7 @@ class ObservationStore(
               plantCountsBySpecies,
               cumulativeDeadFromCurrentObservation,
           )
+          updateZoneSurvivalRate(observationId, plantingZoneId)
         }
 
         updateSpeciesTotalsTable(
@@ -2343,21 +2344,41 @@ class ObservationStore(
           DSL.field(
               DSL.select(DSL.sum(PLOT_T0_DENSITY.PLOT_DENSITY))
                   .from(PLOT_T0_DENSITY)
-                  .where(
-                      PLOT_T0_DENSITY.MONITORING_PLOT_ID.`in`(
-                          DSL.select(MONITORING_PLOTS.ID)
-                              .from(MONITORING_PLOTS)
-                              .where(MONITORING_PLOTS.PLANTING_SUBZONE_ID.eq(plantingSubzoneId))
-                      )
-                  )
+                  .where(PLOT_T0_DENSITY.monitoringPlots.PLANTING_SUBZONE_ID.eq(plantingSubzoneId))
                   .and(PLOT_T0_DENSITY.SPECIES_ID.eq(this.SPECIES_ID))
-                  .groupBy(this.SPECIES_ID)
           )
 
       dslContext
           .update(this)
           .set(SURVIVAL_RATE, TOTAL_LIVE.times(100).div(denominator))
           .where(PLANTING_SUBZONE_ID.eq(plantingSubzoneId))
+          .and(OBSERVATION_ID.eq(observationId))
+          .and(SPECIES_ID.isNotNull)
+          .execute()
+    }
+  }
+
+  private fun updateZoneSurvivalRate(
+      observationId: ObservationId,
+      plantingZoneId: PlantingZoneId,
+  ) {
+    with(OBSERVED_ZONE_SPECIES_TOTALS) {
+      val denominator =
+          DSL.field(
+              DSL.select(DSL.sum(PLOT_T0_DENSITY.PLOT_DENSITY))
+                  .from(PLOT_T0_DENSITY)
+                  .where(
+                      PLOT_T0_DENSITY.monitoringPlots.plantingSubzones.PLANTING_ZONE_ID.eq(
+                          plantingZoneId
+                      )
+                  )
+                  .and(PLOT_T0_DENSITY.SPECIES_ID.eq(this.SPECIES_ID))
+          )
+
+      dslContext
+          .update(this)
+          .set(SURVIVAL_RATE, TOTAL_LIVE.times(100).div(denominator))
+          .where(PLANTING_ZONE_ID.eq(plantingZoneId))
           .and(OBSERVATION_ID.eq(observationId))
           .and(SPECIES_ID.isNotNull)
           .execute()
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -2082,6 +2082,7 @@ class ObservationStore(
               plantCountsBySpecies,
               cumulativeDeadFromCurrentObservation,
           )
+          updateSubzoneSurvivalRate(observationId, plantingSubzoneId)
         }
 
         if (plantingZoneId != null) {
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3333** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3333?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3333?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3331** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3331?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2201,36 +2209,36 @@ class ObservationStore(
...
+                .set(certaintyField, speciesKey.certainty)
+                .set(speciesIdField, speciesKey.id)
+                .set(speciesNameField, speciesKey.name)
+                .set(totalLiveField, totalLive)
+                .set(totalDeadField, totalDead)
+                .set(totalExistingField, totalExisting)
+                .set(cumulativeDeadField, cumulativeDead)
+                .set(permanentLiveField, permanentLive)
+                .set(mortalityRateField, mortalityRate)
+                .set(survivalRateField, survivalRate)
```
ice1080, line 2241: Had to change to `.set` since jooq doesn't allow a mix of raw kotlin values and DSL values

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -0,0 +1,447 @@
...
+            mapOf(subzoneId to survivalRates1And2),
+            mapOf(zoneId to survivalRates1And2),
+            mapOf(plantingSiteId to survivalRates1And2),
+        ),
+        "Plots 1 and 2 survival rates",
+    )
+  }
+
+  @Test
+  fun `survival rate is calculated for a site using all data`() {
```
ice1080, line 224: This test was created using data from the spreadsheet, hence why there's so much data

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2122,6 +2128,7 @@ class ObservationStore(
       isPermanent: Boolean,
       totals: Map<RecordedSpeciesKey, Map<RecordedPlantStatus, Int>>,
       cumulativeDeadFromCurrentObservation: Boolean = false,
+      survivalRateDensityCondition: Condition,
```
ice1080, line 2131: I couldn't think of a better way to do this without having to pass it in to each. If someone knows of a way for this, I'm all ears.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: FYI, you can wrap the raw values in `DSL.value()` in cases like this. But using `.set` is fine too.

ice1080: Right, I knew that was possible, but it felt cleaner to use `.set` than 11 instances of `DSL.value` (should have clarified in my comment)

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -0,0 +1,445 @@
...
+    val speciesIdField =
+        table.field("species_id", SQLDataType.BIGINT.asConvertedDataType(SpeciesIdConverter()))!!
+    val survivalRateField = table.field("survival_rate", Int::class.java)!!
+
+    return dslContext
+        .select(idField, speciesIdField, survivalRateField)
+        .from(table)
+        .where(speciesIdField.isNotNull)
+        .fetch()
+        .filter { it[idField] != null }
```
sgrimm, line 439: Can this ever be null?

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -0,0 +1,445 @@
...
+  ) {
+    val idFields =
+        listOf(
+            OBSERVED_PLOT_SPECIES_TOTALS.MONITORING_PLOT_ID,
+            OBSERVED_SUBZONE_SPECIES_TOTALS.PLANTING_SUBZONE_ID,
+            OBSERVED_ZONE_SPECIES_TOTALS.PLANTING_ZONE_ID,
+            OBSERVED_SITE_SPECIES_TOTALS.PLANTING_SITE_ID,
+        )
+
+    val levelNames = listOf("Plot", "Subzone", "Zone", "Site")
```
sgrimm: Maybe bikeshedding, but this could be `listOf("Plot" to OBSERVED_PLOT_SPECIES_TOTALS.MONITORING_PLOT_ID, ...)` just to reduce the order dependency (though there's still an order dependency between the field list and `expected`, of course).

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
sgrimm: Would it make sense to use the actual spreadsheet (or a CSV export of it) here? We do that in some of the other rate-calculation tests. The main reason I wrote the tests that way was so that if the sheet gets changed because we decided to change the calculations, we can just re-export it and the test code won't have to change.

But obviously that requires writing code to process the CSVs, which may be more trouble than it's worth in this case.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2283,13 +2291,21 @@ class ObservationStore(
                 .execute()
           }
 
+          val survivalRate =
+              if (isPermanent && speciesKey.id != null) {
+                totalLiveField.plus(totalLive).mul(100).div(survivalRateDenominator)
+              } else {
+                DSL.inline(null, SQLDataType.INTEGER)
```
sgrimm: Could use `DSL.castNull()`.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2283,13 +2291,21 @@ class ObservationStore(
                 .execute()
           }
 
+          val survivalRate =
+              if (isPermanent && speciesKey.id != null) {
+                totalLiveField.plus(totalLive).mul(100).div(survivalRateDenominator)
```
sgrimm, line 2296: This will do integer math, right? Are we okay with that in terms of rounding behavior? (Same for the similar calculation above.)

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2202,35 +2210,35 @@ class ObservationStore(
...
+        val survivalRateDenominator =
+            DSL.field(
+                DSL.select(DSL.sum(PLOT_T0_DENSITY.PLOT_DENSITY))
+                    .from(PLOT_T0_DENSITY)
+                    .where(survivalRateDensityCondition)
+                    .and(PLOT_T0_DENSITY.SPECIES_ID.eq(speciesKey.id))
+            )
+        val survivalRate =
+            if (isPermanent && speciesKey.id != null) {
+              DSL.`val`(totalLive).mul(100).div(survivalRateDenominator)
```
sgrimm: `DSL.value` is a synonym for `DSL.val` but the name is more Kotlin-friendly.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: Can't think of a cleaner way offhand, given that you have to do different joins in each case.

sgrimm (approved):
A few minor comments inline, but this seems reasonable.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
ice1080: Ah I did see that after using `val`. I saw that we use both, but I should have paid better attention to notice that we almost always use `value`. Will update

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
ice1080: I think it must be casting to numeric before dividing since `survivalRateDenominator` is numeric, and then after the division casting back to integer and doing rounding at that point. I saw the correct rounding happening in the tests so was just assuming that was the case

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
ice1080: It's definitely more work, but with how complex this stuff is I rather like that idea. I was super nervous already while copying all the data of accidentally touching one of the cells because even `undo` doesn't go back to previous random values so every value would have changed. I saved several screenshots during it ha. I'll look into this

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
ice1080: Yes, not every species will have a survival rate, as seen on lines 333-334. This is because the `totals` map in `updateSpeciesTotalsTable` only contains species that had recorded plants, so if a species had recorded plants in a previous observation and no recorded plants in the latest, it will not show up in the map which we are looping through to do these inserts. Thus the survival rate will be null. I haven't yet understood what issues this will cause, if any, when calculating survival rates across all species for plots, subzones, zones, and sites.

ice1080: I _think_ it won't cause any issues since I'll have to do a sum anyway, and nulls are treated as 0 in that case.

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
ice1080: The more I think about this, the more I lean towards waiting for now. This scenario already covers quite a bit of different situations, so I'll wait till the second time around to write the parser stuff for it. Might be worth extracting from `ObservationResultsStoreTest` for reuse.

sgrimm (approved):
New changes look good, and the plan to hold off changing the spreadsheet-driven tests makes sense.

PR #3334, opened 2025-08-27
SW-7275 Observations: Export Locations GPX file - include plot type information

constanzauanini:
Added Zone Name, Subzone Name and Plot type to GPX file.
Also simplified the name of the corners (expect SW).


src/main/kotlin/com/terraformation/backend/tracking/model/AssignedPlotDetails.kt
```diff
@@ -18,30 +18,33 @@ data class AssignedPlotDetails(
...
 ) {
   fun gpxWaypoints(messages: Messages): List<GpxWaypoint> {
+    val plotType = if (model.isPermanent) "Permanent" else "Temporary"
+
     return listOf(
         GpxWaypoint(
             boundary.coordinates[SOUTHWEST].y,
             boundary.coordinates[SOUTHWEST].x,
-            messages.monitoringPlotSouthwestCorner(plotNumber),
+            "${messages.monitoringPlotSouthwestCorner(plotNumber)} - Plot type: $plotType - Planting zone name: $plantingZoneName - Planting subzone name: $plantingSubzoneName",
```
sgrimm: This will return English text in the GPX file even if the user is using the site in Spanish or French (but it'll use the translation for "southwest corner").

We'll want to move this text to `Messages_en.properties` so it can be translated. The simplest approach might be to edit the strings for `monitoringPlotSouthwestCorner` and the other corners so they include placeholders for the additional values. Then add parameters to the methods for these strings in the `Messages` class. Those monitoring plot corner strings are only used in GPX export so it's safe to change them.

We'll also need translatable strings for the plot types.

sgrimm (changes requested):

src/main/resources/i18n/Messages_en.properties
```diff
@@ -119,13 +119,15 @@ manifestCsvVariableNameNotUniqueWithinParent=No two children of the same parent
...
+monitoringPlotNortheastCorner={0} NE - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotNorthwestCorner={0} NW
+monitoringPlotNorthwestCorner={0} NW - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotSoutheastCorner={0} SE
+monitoringPlotSoutheastCorner={0} SE - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotSouthwestCorner={0} Southwest Corner
+monitoringPlotSouthwestCorner={0} Southwest Corner - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
```
sgrimm: This should be "SW" instead of "Southwest Corner" to match the changes to the other strings.

sgrimm (approved):

src/main/resources/i18n/Messages_en.properties
constanzauanini: In the jira ticket it says that we want to keep the "long" description just for southwest corner

constanzauanini: Oh, you are right! We only want to keep the word "Corner" for SW

src/main/resources/i18n/Messages_en.properties
```diff
@@ -119,13 +119,15 @@ manifestCsvVariableNameNotUniqueWithinParent=No two children of the same parent
...
+monitoringPlotNortheastCorner={0} NE - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotNorthwestCorner={0} Northwest Corner
+monitoringPlotNorthwestCorner={0} NW - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotSoutheastCorner={0} Southeast Corner
+monitoringPlotSoutheastCorner={0} SE - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
 # {0} is plot name
-monitoringPlotSouthwestCorner={0} Southwest Corner
+monitoringPlotSouthwestCorner={0} SW Corner - Plot type: {1} - Planting Zone Name: {2} - Planting Subzone Name: {3}
```
sgrimm: Just looked at the Jira task and it looks like the SW corner string should have the word "Plot" at the beginning. (But not the other corners.)

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/i18n/Messages.kt
```diff
@@ -207,17 +207,65 @@ class Messages {
 
   fun batchCsvQuantityInvalid() = getMessage("batchCsvQuantityInvalid")
 
-  fun monitoringPlotNortheastCorner(plotNumber: Long) =
-      getMessage("monitoringPlotNortheastCorner", plotNumber)
+  fun monitoringPlotNortheastCorner(
```
tommylau523: The changes you made in this file will only update the name of the waypoint. I think we do want to make them short-handed like Plot 5 NE. But the requirement reads like we want the zone/subzone info to be elsewhere.

tommylau523:
The GPX schema only allow certain element tags. For waypoints, you can find it here: https://www.topografix.com/gpx/1/1/#type_wptType

I think the best place to put the additional data is <desc> to ensure it works on most GPS devices.

sgrimm:
Sorry to keep doing this to you, but I just merged #3335 to add llm-autotranslate; if you can rebase and generate the translations, that'd be great.

tommylau523 (approved):

PR #3338, opened 2025-08-28
SW-7306 Fetch Data grouped by monitoring plots, assign multiple monitoring plots at once

ice1080:
Previously there was discussion about only updating survival rate settings for one monitoring plot at a time, but that was replaced with a single save button for an entire planting site.

Group t0 data fetched for survival rate settings by monitoring plot.
Replace assign-t0-endpoints with single endpoint to assign multiple plots' data in one call, to match designs.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
@@ -1,0 +1,20 @@
+package com.terraformation.backend.tracking
+
+import com.terraformation.backend.tracking.db.T0PlotStore
+import com.terraformation.backend.tracking.model.PlotT0DataModel
+import jakarta.inject.Named
+
+@Named
+class T0PlotService(
+    private val t0PlotStore: T0PlotStore,
+) {
+  fun assignT0PlotsData(plotsList: List<PlotT0DataModel>) {
+    plotsList.forEach {
+      if (it.observationId == null) {
+        t0PlotStore.assignT0PlotSpeciesDensities(it.monitoringPlotId, it.densityData)
+      } else {
+        t0PlotStore.assignT0PlotObservation(it.monitoringPlotId, it.observationId)
+      }
+    }
+  }
+}
--- a/src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
@@ -3,17 +3,21 @@ package com.terraformation.backend.tracking
 import com.terraformation.backend.tracking.db.T0PlotStore
 import com.terraformation.backend.tracking.model.PlotT0DataModel
 import jakarta.inject.Named
+import org.jooq.DSLContext
 
 @Named
 class T0PlotService(
+    private val dslContext: DSLContext,
     private val t0PlotStore: T0PlotStore,
 ) {
   fun assignT0PlotsData(plotsList: List<PlotT0DataModel>) {
-    plotsList.forEach { model ->
-      if (model.observationId == null) {
-        t0PlotStore.assignT0PlotSpeciesDensities(model.monitoringPlotId, model.densityData)
-      } else {
-        t0PlotStore.assignT0PlotObservation(model.monitoringPlotId, model.observationId)
+    dslContext.transaction { _ ->
+      plotsList.forEach { model ->
+        if (model.observationId == null) {
+          t0PlotStore.assignT0PlotSpeciesDensities(model.monitoringPlotId, model.densityData)
+        } else {
+          t0PlotStore.assignT0PlotObservation(model.monitoringPlotId, model.observationId)
+        }
       }
     }
   }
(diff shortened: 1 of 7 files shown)


ice1080:
* **#3338** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3338?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
```diff
@@ -84,40 +98,41 @@ class T0PlotStore(
...
+            throw IllegalArgumentException("Plot density must not be negative")
+          }
+          insertQuery = insertQuery.values(monitoringPlotId, it.speciesId, it.plotDensity)
+        }
+
+        insertQuery.onDuplicateKeyUpdate().set(PLOT_DENSITY, DSL.excluded(PLOT_DENSITY)).execute()
+      }
+
+      densities.forEach {
+        eventPublisher.publishEvent(
```
ice1080: Wasn't sure if this is correct to publish an event for each item in the densities. Can update this to only one event per monitoring plot if better.

src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
```diff
@@ -0,0 +1,20 @@
...
+class T0PlotService(
+    private val t0PlotStore: T0PlotStore,
+) {
+  fun assignT0PlotsData(plotsList: List<PlotT0DataModel>) {
+    plotsList.forEach {
+      if (it.observationId == null) {
+        t0PlotStore.assignT0PlotSpeciesDensities(it.monitoringPlotId, it.densityData)
+      } else {
+        t0PlotStore.assignT0PlotObservation(it.monitoringPlotId, it.observationId)
+      }
```
sgrimm: This spans enough lines that it'd make sense to declare a named variable instead of using `it`.

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
sgrimm: Mostly depends on what's consuming the events, but I'd probably lean toward one event per monitoring plot.

sgrimm: I assume a followup PR will add something that uses these events?

src/test/kotlin/com/terraformation/backend/tracking/db/T0PlotStoreTest.kt
```diff
@@ -265,22 +264,28 @@ internal class T0PlotStoreTest : DatabaseTest(), RunsAsDatabaseUser {
   }
 
   @Nested
-  inner class AssignT0PlotSpeciesDensity {
+  inner class AssignT0PlotSpeciesDensities {
```
sgrimm, line 266: Worth having a test case for updating existing densities where the new value is missing a species that was originally there? Not sure what the desired behavior is in that case. I think the current implementation will leave the old value as-is, which is fine if that's what we want, but a test case would make it clear that it's an intentional choice.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
ice1080: Correct ya. I was trying to publish events as early as possible in the epic so that I don't forget where to add them later on. 
Will update to one event per plot

src/test/kotlin/com/terraformation/backend/tracking/db/T0PlotStoreTest.kt
ice1080: Good thought. I think we'll actually want to assume that the list is everything that should be in the plot and delete existing ones in the monitoring plot.

src/main/kotlin/com/terraformation/backend/tracking/T0PlotService.kt
ice1080: @sgrimm one other thing I forgot to ask about was if we're good with each plot being in its own transaction, or if we usually do something like this in all one transaction?

sgrimm: At a high level, it boils down to how you want it to behave on failure: if you return an error to the client, do you want the operation to be partially completed or not completed at all? The answer is situational. In this case, I'd probably wrap the entire thing in a transaction so the update request either succeeds completely or makes no changes.

ice1080: That's what I've generally done in the past, but wasn't sure if we had any other instance of preferring partial updates so just double checking

sgrimm (approved):

PR #3339, opened 2025-08-29
SW-7306 Fetch withdrawal densities per species per plots

ice1080:
The survival rate settings page needs to display densities calculated from withdrawal data to allow the user to select these values.

Add an endpoint to fetch this data.

All names and locations of classes and the endpoint open to change suggestions.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/db/NurseryWithdrawalStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/db/NurseryWithdrawalStore.kt
@@ -9,8 +9,9 @@ import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONE
 import com.terraformation.backend.nursery.model.NurserySpeciesModel
 import com.terraformation.backend.nursery.model.PlotSpeciesModel
 import jakarta.inject.Named
-import java.math.BigDecimal
 import org.jooq.DSLContext
+import org.jooq.impl.DSL
+import org.jooq.impl.SQLDataType
 
 @Named
 class NurseryWithdrawalStore(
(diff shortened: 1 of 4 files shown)


ice1080:
* **#3339** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3339?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3339?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryWithdrawalStore.kt
```diff
@@ -0,0 +1,52 @@
...
+) {
+  fun fetchSiteSpeciesByPlot(plantingSiteId: PlantingSiteId): List<PlotSpeciesModel> {
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    with(MONITORING_PLOTS) {
+      return dslContext
+          .select(
+              ID,
+              PLANTING_SUBZONE_POPULATIONS.SPECIES_ID,
+              (PLANTING_SUBZONE_POPULATIONS.TOTAL_PLANTS / PLANTING_SUBZONES.AREA_HA).`as`(
```
tommylau523: Another option is to create a field.

```
val densityField = PLANTING_SUBZONE_POPULATIONS.TOTAL_PLANTS / PLANTING_SUBZONES.AREA_HA

dslContext.select(..., densityField)...

...

density = record[densityField]
```

Should avoid some type casting since it should evaluate the type for you that way.

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryWithdrawalStore.kt
ice1080: Looks like it still requires casting at the field level, since the numerator is an int and denominator is numeric. I still like it better though, will go with that. 
```kotlin
val densityField =
    PLANTING_SUBZONE_POPULATIONS.TOTAL_PLANTS.cast(SQLDataType.NUMERIC) /
        PLANTING_SUBZONES.AREA_HA
```
Additionally this pointed out a rounding issue, so fixing that as well

src/main/kotlin/com/terraformation/backend/nursery/db/NurseryWithdrawalStore.kt
```diff
@@ -0,0 +1,57 @@
...
+) {
+  fun fetchSiteSpeciesByPlot(plantingSiteId: PlantingSiteId): List<PlotSpeciesModel> {
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    with(MONITORING_PLOTS) {
+      val densityField =
+          DSL.round(
+              PLANTING_SUBZONE_POPULATIONS.TOTAL_PLANTS.cast(SQLDataType.NUMERIC) /
+                  PLANTING_SUBZONES.AREA_HA
+          )
```
ice1080, line 28: Have to cast `total_plants` to numeric or it will do integer division.

PR #3350, opened 2025-09-04
SW-7333 Add survival rate calculations to ObservationResults models

ice1080:
Add survival rate calculations to ObservationResults models.

Add survival rate to ObservationController payloads.


ice1080:
* **#3353** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3353?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3350** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3350?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/resources/tracking/observation/DisjointSubzones/SiteStats.csv
```diff
@@ -1,3 +1,3 @@
-1,1,1,1,2,2,2,2,3,3,3,3
-Planting Density,Estimated # Plants,# Species,Mortality Rate,Planting Density,Estimated # Plants,# Species,Mortality Rate,Planting Density,Estimated # Plants,# Species,Mortality Rate
-22,33000,1,33%,11,16500,1,50%,33,49500,1,43%
\ No newline at end of file
+1,1,1,1,1,2,2,2,2,2,3,3,3,3,3
+Planting Density,Estimated # Plants,# Species,Mortality Rate,Survival Rate,Planting Density,Estimated # Plants,# Species,Mortality Rate,Survival Rate,Planting Density,Estimated # Plants,# Species,Mortality Rate,Survival Rate
+22,33000,1,33%,67%,11,16500,1,50%,100%,33,49500,1,43%,133%
```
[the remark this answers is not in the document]
ice1080: This is possible because t0 might not have observed all plants, and also natural regeneration could increase survival rate.

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
```diff
@@ -2108,6 +2153,25 @@ class ObservationResultsStoreTest : DatabaseTest(), RunsAsUser {
       plantsRows
           .groupBy { it.monitoringPlotId!! }
           .forEach { (plotId, plants) ->
+            if (observationNum == 0 && plotId in permanentPlotIds) {
+              insertPlotT0Observation(monitoringPlotId = plotId)
+              plants
+                  .filter {
+                    it.certaintyId == RecordedSpeciesCertainty.Known &&
+                        it.speciesId != null &&
+                        it.statusId != RecordedPlantStatus.Existing
```
sgrimm, line 2158: Does this need to exclude dead plants too? (Same with the similar block above.)

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
```diff
@@ -1491,22 +1508,24 @@ class ObservationResultsStoreTest : DatabaseTest(), RunsAsUser {
...
-          val positionInColumnGroup = (index - 1) % 8
-          positionInColumnGroup != 3 &&
-              positionInColumnGroup != 4 &&
+          val positionInColumnGroup = (index - 1) % 10
+          positionInColumnGroup != 4 &&
               positionInColumnGroup != 5 &&
-              positionInColumnGroup != 6
+              positionInColumnGroup != 6 &&
+              positionInColumnGroup != 7 &&
+              positionInColumnGroup != 8
```
sgrimm: Maybe a range check makes more sense here now. (Honestly, maybe it made more sense before, too!)

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
```diff
@@ -308,7 +344,17 @@ data class ObservationPlantingZoneRollupResultsModel(
...
+                  mortalityRate to permanentPlants.toDouble()
+                }
+              }
+              .calculateWeightedStandardDeviation()
+      val survivalRate = species.calculateSurvivalRate()
+      val survivalRateStdDev =
+          completedMonitoringPlots
+              .mapNotNull { plot ->
+                plot.survivalRate?.let { survivalRate ->
+                  val sumDensity = plot.species.mapNotNull { it.t0Density }.sumOf { it }
```
sgrimm, line 356: Can you use `sum()` instead of `sumOf { it }` here?

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
```diff
@@ -510,7 +584,7 @@ fun Collection<Int>.calculateStandardDeviation(): Int? {
  * Calculate standard deviation from a collection of pairs of (data, weight). Weight must be the
  * number of samples, so we can correctly apply the (n-1) Bessel's correction.
  */
-fun Collection<Pair<Int, Int>>.calculateWeightedStandardDeviation(): Int? {
+fun Collection<Pair<Int, Double>>.calculateWeightedStandardDeviation(): Int? {
```
sgrimm, line 587: Why the change to `Double` here? It looks like the callers all have `Int`s and just cast them to `Double` to call this function.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
ice1080: `t0Density` is a BigDecimal, and I couldn't get the method to work with `Numeric` so used a `Double`. The loss of precision shouldn't matter here since these rates are converted to Int anyway. I'm open to other ideas though

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
ice1080: From what I can tell, there is no `sum()` available for `Iterable<BigDecimal>`?

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreTest.kt
ice1080: Nope, `t0Density` is based on `Live` and `Dead` plants in the first observation!

ice1080: When it's assigned with an observation, it's supposed to represent all planted plants from as early as possible, so even though it's supposed to be the survival rate since t0, the density includes dead plants because those were supposedly also planted (likely). It makes for a slightly interesting scenario because the survival rate at the t0 observation could actually be lower than 100%. 
I like to think about it as `t0Density` is the number of plants planted in that plot sometime between actual planting and the first observation.

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
sgrimm: Ah, my bad, I assumed it was `Int`. This is fine, then.

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
sgrimm: You're right. For some reason I forgot that this wasn't an integer. Looks good as is.

PR #3357, opened 2025-09-05
SW-7360 Add Created/Modified fields to Project Internal Users

ice1080:
We want to be able to sort a project’s internal users by modified time but project_internal_users table does not have this field. 

Add created/modified by/time fields. Default these to the timestamp and user they were added with if Project Leads or Restoration Leads. Use the current timestamp and system user for the rest. (All currently in prod are Project Leads).

Add created and modified time to the controller payload for retrieving project internal users.

### DIFF
--- a/src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
+++ b/src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
@@ -1,0 +1,39 @@
+ALTER TABLE project_internal_users
+    ADD COLUMN created_time TIMESTAMP WITH TIME ZONE DEFAULT NOW();
+ALTER TABLE project_internal_users
+    ADD COLUMN modified_time TIMESTAMP WITH TIME ZONE DEFAULT NOW();
+
+ALTER TABLE project_internal_users
+    ADD COLUMN created_by BIGINT REFERENCES users;
+ALTER TABLE project_internal_users
+    ADD COLUMN modified_by BIGINT REFERENCES users;
+
+-- Most internal users have been Project Leads, which includes TF Contacts
+UPDATE project_internal_users piu
+SET (created_by, modified_by, created_time, modified_time) = (
+    SELECT ou.created_by, ou.modified_by, ou.created_time, ou.modified_time
+    FROM organization_users ou
+    WHERE ou.user_id = piu.user_id
+    AND ou.organization_id = (select organization_id from projects where id = piu.project_id)
+    AND ou.role_id = 5 -- TF CONTACT
+)
+WHERE piu.project_internal_role_id in (1, 2) -- only do this for users that became TF Contacts through internal roles
+;
+
+-- If any others remain, use system user
+UPDATE project_internal_users piu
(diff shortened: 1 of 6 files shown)


ice1080:
* **#3357** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
```diff
@@ -0,0 +1,39 @@
...
+    WHERE ou.user_id = piu.user_id
+    AND ou.organization_id = (select organization_id from projects where id = piu.project_id)
+    AND ou.role_id = 5 -- TF CONTACT
+)
+WHERE piu.project_internal_role_id in (1, 2) -- only do this for users that became TF Contacts through internal roles
+;
+
+-- If any others remain, use system user
+UPDATE project_internal_users piu
+SET (created_by, modified_by, created_time, modified_time) = (
```
sgrimm: Since the time columns default to `NOW()`, you could skip setting them here.

sgrimm (approved):

sgrimm:
Ah, looks like I was too quick to approve this; tests need to populate these new columns since they're non-nullable.

src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
ice1080: I actually tried that first and it threw an error when setting to not null. I wonder if you have to set to not null with a default at the same time 🤔

ice1080:
Ah that's right. Will fix

src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
ice1080: Ya it doesn't like that, and I can't set not null and default in the same statement (plus the default already exists). I think the default is only used when inserting after not null has been enforced.
> ERROR: column "created_time" of relation "project_internal_users" contains null values

ice1080: Interesting, this discovered cases where users have `Project Lead` but not TF Contact. Looking into this

ice1080: Ok there we go, that's a little cleaner.

src/main/resources/db/migration/0400/V403__InternalUsersModified.sql
```diff
@@ -0,0 +1,39 @@
...
+    SELECT u.id, u.id, NOW(), NOW()
+    FROM users u
+    WHERE u.user_type_id = 4 -- system user
+)
+WHERE created_by IS NULL;
```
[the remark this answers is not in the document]
ice1080: All environments have this user

PR #3365, opened 2025-09-08
SW-7346 Add activity media file upload endpoint

sgrimm:
Add an endpoint clients can use to upload media files related to activities.
Currently, only photo files (JPEG, PNG) are supported.

The server will extract the photo's captured date and geolocation from EXIF
metadata if available in the file.

This only includes the upload (write) path; reading will be added in a separate
change.


sgrimm:
* **#3365** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3365?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3365?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
```diff
@@ -74,6 +85,26 @@ class ActivitiesController(
...
   }
+
+  @Operation(summary = "Uploads media for an activity.")
+  @PostMapping("/{id}/media", consumes = [MediaType.MULTIPART_FORM_DATA_VALUE])
+  @RequestBodyPhotoFile
+  fun uploadActivityMedia(
+      @PathVariable id: ActivityId,
+      @RequestPart("file") file: MultipartFile,
+  ): UploadActivityMediaResponsePayload {
+    val contentType = file.getPlainContentType(SUPPORTED_PHOTO_TYPES)
```
ice1080, line 96: Will there be a diff endpoint for videos? This seems specific to photos

src/main/kotlin/com/terraformation/backend/accelerator/db/ActivityMediaService.kt
```diff
@@ -0,0 +1,186 @@
...
+            val capturedDate =
+                exifMetadata?.let { extractCapturedDate(it) } ?: getCurrentDate(activityId)
+            val geolocation = exifMetadata?.let { extractGeolocation(it) }
+
+            dslContext
+                .insertInto(ACTIVITY_MEDIA_FILES)
+                .set(ACTIVITY_MEDIA_FILES.FILE_ID, fileId)
+                .set(ACTIVITY_MEDIA_FILES.ACTIVITY_ID, activityId)
+                .set(ACTIVITY_MEDIA_FILES.ACTIVITY_MEDIA_TYPE_ID, ActivityMediaType.Photo)
+                .set(ACTIVITY_MEDIA_FILES.IS_COVER_PHOTO, false)
```
ice1080, line 101: Same comment here, the method is named `storeMedia` but seems like it should be `storePhoto` unless the media type is passed in (or extracted from the file metadata?)

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
sgrimm: No, the plan is to use something else instead of `SUPPORTED_PHOTO_TYPES` here once video support lands. I didn't want to make this accept videos initially since they wouldn't work right.

src/main/kotlin/com/terraformation/backend/accelerator/db/ActivityMediaService.kt
sgrimm: Once videos are supported, this value will get figured out dynamically from the content type. Or at least that's the plan at the moment, though it's possible I'll have to do something else once I figure out how video will actually work.

ice1080 (approved):

PR #3368, opened 2025-09-09
Fix race in InputStreamCopierTest

sgrimm:
One of the tests had a race condition that caused it to fail occasionally.

What the test is verifying is difficult to test 100% deterministically (it's
kind of the halting problem) but we can make it less susceptible to races using
explicit signaling, and make it pass rather than fail if it's running on an
extremely loaded host such that its sleep time isn't long enough to trigger the
condition it's testing for.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/util/InputStreamCopierTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/util/InputStreamCopierTest.kt
@@ -4,12 +4,14 @@ import java.io.ByteArrayInputStream
 import java.io.IOException
 import java.io.InputStream
 import java.net.SocketTimeoutException
+import java.time.Duration
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3368** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
Verified that this doesn't fail intermittently by changing `@Test` to `@RepeatedTest(5000)` locally and running it. The old version failed about 5% of the time; the new one passes consistently. (Obviously that's not a guarantee that there's no intermittent failure, but it's at least _better_ than before.)

ice1080 (approved):

sgrimm:
### Merge activity

* **Sep 9, 5:06 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3368).

PR #3376, opened 2025-09-10
SW-7365 Remove t0 observation if changing to manual species data

ice1080:
If a user decides to change from an observation as the t0 source of truth to manual species values, the endpoint throws an exception. Update the endpoint to delete the existing t0 observation instead of throwing an exception.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
@@ -53,48 +53,52 @@ class T0PlotStore(
   fun assignT0PlotObservation(monitoringPlotId: MonitoringPlotId, observationId: ObservationId) {
     requirePermissions { updateT0(monitoringPlotId) }
 
-    with(PLOT_T0_OBSERVATIONS) {
-      dslContext
-          .insertInto(this)
-          .set(MONITORING_PLOT_ID, monitoringPlotId)
-          .set(OBSERVATION_ID, observationId)
-          .onDuplicateKeyUpdate()
-          .set(OBSERVATION_ID, observationId)
-          .execute()
-    }
+    dslContext.transaction { _ ->
+      with(PLOT_T0_OBSERVATIONS) {
+        dslContext
+            .insertInto(this)
+            .set(MONITORING_PLOT_ID, monitoringPlotId)
+            .set(OBSERVATION_ID, observationId)
+            .onDuplicateKeyUpdate()
+            .set(OBSERVATION_ID, observationId)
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3376** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3376?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
```diff
@@ -104,10 +104,10 @@ class T0PlotStore(
...
     requirePermissions { updateT0(monitoringPlotId) }
 
-    if (plotHasObservationT0(monitoringPlotId)) {
-      throw IllegalStateException(
-          "Cannot assign species density to plot $monitoringPlotId which already has observation assigned"
-      )
+    with(PLOT_T0_OBSERVATIONS) {
+      // delete t0 observations associated with this plot so that retrieval doesn't return incorrect
+      // data
+      dslContext.deleteFrom(this).where(MONITORING_PLOT_ID.eq(monitoringPlotId)).execute()
```
sgrimm: Wrapping this and the other write operations in a transaction would be a good idea.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
ice1080: I considered that but was thinking that these would likely never be called outside of the context of the `T0PlotService` which already wraps them so that all plots are updated in one transaction. Is it common for us to have a transaction at multiple levels to be safe in cases like this? I know that it won't create a new transaction so am happy to add it, just wasn't sure.

sgrimm: My rough rule of thumb, which I probably violate all over the place, is that private methods can rely on callers' transactions but public ones should create their own.

ice1080: Makes sense; updated!

ice1080:
### Merge activity

* **Sep 10, 6:40 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3376).

PR #3375, opened 2025-09-10
Update tests and comments for merging other species

ice1080:
Update a few comments to make survival rate calculation more clear.

Remove unused method and its test file.

Add tests around survival rate in general and mergeOtherSpecies specifically to ensure survival rate is recalculated when `Other` species are identified.

PR #3370, opened 2025-09-09
SW-7284 Add Survival Rate to Funder Reports

ice1080:
Add Survival Rate System Metric.

Set System Metric when gathering values for funder reports.

PR #3380, opened 2025-09-11
Exclude Plots that have no observations from Survival Rate calculations

ice1080:
Exclude Plots that have no observations from Survival Rate calculations.

Add Summary suffix to observation summary test CSVs to clear understanding.

Fix survival rate tests to include check for area survival rates across all species.

Include t0 Densities in survival rate denominator for species that have never been recorded in observations.


ice1080:
* **#3380** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3380?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3380?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3375** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3375?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3370** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/test/resources/tracking/observation/SurvivalRateNoObservations/Subzones.csv
```diff
@@ -0,0 +1,5 @@
+,,
```
sgrimm, line 1: Not a big deal, but it's a shame this needs to be here.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -752,6 +752,12 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
                                     PLANTING_SUBZONE_ID
                                 )
                             )
+                            .and(
+                                PLOT_T0_DENSITY.MONITORING_PLOT_ID.`in`(
+                                    DSL.select(OBSERVATION_PLOTS.MONITORING_PLOT_ID)
+                                        .from(OBSERVATION_PLOTS)
```
sgrimm: Does this need to filter out temporary or ad-hoc plots?

sgrimm (dismissed):

src/test/resources/tracking/observation/SurvivalRateNoObservations/Subzones.csv
ice1080: Ya that was due to the `Finished Planting?` column per observation in some other observation scenarios. Alternatively we could clean this up by adding a boolean to the reader method that includes observation number. It's technically always ignored, it just make it easier to read the spreadsheets. Alternatively could change the headers to something like `Finished Planting - Observation 1`, etc. I'll leave that for a different PR for now since it would affect every observation scenario test

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
ice1080: Good question. Right now, we only allow t0 density to be set on permanent plots. Eventually we'll allow temporary, so I felt like not limiting it made sense because it's already limited to what it needs now, and in the future when we add temp plots' densities then this wouldn't need updated. I'm good either way though, that was just initial thinking.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -1106,13 +1141,33 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
         )
       }
 
+  private fun sumT0Field(plotCondition: Condition) =
+      with(PLOT_T0_DENSITY) {
+        DSL.field(
+            DSL.select(
+                    DSL.case_()
+                        .`when`(DSL.count().gt(0), DSL.sum(PLOT_DENSITY))
+                        .else_(DSL.castNull(SQLDataType.NUMERIC))
```
sgrimm: I'm surprised `DSL.sum(PLOT_DENSITY)` wouldn't work here in place of the `CASE` (as in, evaluate to null if there are no rows).

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
ice1080: Good catch. I must have added this when I had some weird other logic going on, or maybe I incorrectly thought this was the issue. Anyway, changing this still has all tests passing 👍

ice1080:
### Merge activity

* **Sep 12, 11:57 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3380).

PR #3381, opened 2025-09-11
SW-7367 Allow temporary read access to files

sgrimm:
To allow a video streaming provider to convert user-uploaded videos to the correct
formats for streaming, the provider needs to be able to access the original file,
and of course it won't know how to authenticate to Terraware.

Add an endpoint that works similarly to features like S3's pre-signed URLs: we can
generate a URL that includes a token that allows unauthenticated access to a file
for a certain amount of time. For our purposes, there's no need for anything as
complex as signed keys; it's sufficient to just generate a hard-enough-to-guess
token and check it against our database.

The admin UI has a new page that super-admins can use to generate tokens for
testing.

### DIFF
--- a/src/main/resources/db/migration/0400/V405__FileAccessTokens.sql
+++ b/src/main/resources/db/migration/0400/V405__FileAccessTokens.sql
@@ -1,0 +1,10 @@
+CREATE TABLE file_access_tokens (
+    token TEXT PRIMARY KEY,
+    file_id BIGINT NOT NULL REFERENCES files ON DELETE CASCADE,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    expires_time TIMESTAMP WITH TIME ZONE NOT NULL
+);
+
+CREATE INDEX ON file_access_tokens (expires_time);
+CREATE INDEX ON file_access_tokens (file_id);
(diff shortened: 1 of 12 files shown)


sgrimm:
* **#3381** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3381?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0400/V405__FileAccessTokens.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE file_access_tokens (
+    token TEXT PRIMARY KEY,
+    file_id BIGINT NOT NULL REFERENCES files ON DELETE CASCADE,
```
ice1080, line 3: What is a case where the token would be null? (i.e. should this be NOT NULL)

sgrimm: `PRIMARY KEY` implies `NOT NULL`.

ice1080: 🤦 wasn't even paying attention to that. My bad

ice1080 (approved):

PR #3383, opened 2025-09-12
SW-7367 Mux API integration

sgrimm:
Add code to send video files to Mux so they're available for streaming in the web
app. Nothing calls this code yet; subsequent changes will add Terraware API
endpoints and admin UI controls to interact with Mux.


sgrimm:
* **#3398** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3398?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3383** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3383?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3397** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3397?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
```diff
@@ -88,6 +88,9 @@ class SecurityConfig(
         // Allow unauthenticated users to fetch files using temporary access tokens.
         authorize("/api/v1/files/tokens/**", permitAll)
 
+        // Allow webhook requests from external services.
+        authorize("/api/v1/webhooks/**", permitAll)
```
ice1080, line 92: Should this be just `mux` webhooks in case we add other webhook controllers that need diff auth?

src/main/kotlin/com/terraformation/backend/file/mux/MuxPayloads.kt
```diff
@@ -0,0 +1,74 @@
...
+ * Mux API and webhook payloads. These are abbreviated; they only contain the fields we care about.
+ * The actual payloads include a lot of additional information.
+ */
+
+@Suppress("EnumEntryName")
+enum class MuxApiStatus(val assetStatus: MuxAssetStatus) {
+  preparing(MuxAssetStatus.Preparing),
+  ready(MuxAssetStatus.Ready),
+  errored(MuxAssetStatus.Errored),
+}
```
ice1080, line 18: Is the benefit of not using `MuxAssetStatus` just that we can add to this without adding to `MuxAssetStatus`?

src/main/kotlin/com/terraformation/backend/file/mux/MuxService.kt
```diff
@@ -0,0 +1,249 @@
...
+   * be accessible without authentication for this amount of time; it should be long enough to let
+   * Mux retry if there's a transient error.
+   */
+  private val fileAccessTokenExpiration = Duration.ofHours(24)
+
+  /**
+   * URL path prefix of the "download a file without authentication" endpoint. The file access token
+   * generated by [FileService] will be added as a path element beneath this path.
+   */
+  private val fileAccessEndpoint = "/api/v1/files/tokens"
```
ice1080, line 56: I'm assuming there's no way to get this dynamically from `FilesController` right? Not super important, just curious

src/main/kotlin/com/terraformation/backend/file/mux/MuxService.kt
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+
+  /**
+   * Parses a Mux-provided signing key. Mux's private keys are provided as base64-encoded strings
+   * that decode to PEM-formatted RSA private keys.
+   */
+  private fun parseSigningKey(): PrivateKey {
+    val pemString =
+        Base64.getDecoder().decode(config.mux.signingKeyPrivate).toString(Charsets.US_ASCII)
```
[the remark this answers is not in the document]
sgrimm: This isn't `toString()`, it's `toString(Charset)` which is a Kotlin extension.

src/main/kotlin/com/terraformation/backend/file/mux/MuxPayloads.kt
sgrimm: These are the status strings the API can return. I wanted to allow for us to add our own internal statuses (like maybe related to error recovery) without affecting how we parse status values in Mux payloads.

src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
sgrimm: I don't _think_ so; unless those webhooks use Terraware login sessions as their authentication mechanism, we'll still need to bypass the usual "is this request from a logged-in Terraware user?" check and use some other custom auth check like I'm doing with the webhook signature checking here.

src/main/kotlin/com/terraformation/backend/file/mux/MuxService.kt
sgrimm: Not dynamically, as far as I know, short of using reflection.

I considered defining it as a global `const val` but that seemed like it'd be more annoying than helpful since you'd no longer be able to see the URL path directly in the controller annotations. But obviously doing it this way has the downside that it's possible for the paths to get out of sync if we change them.

ice1080: Makes sense. I doubt it'll ever change, but another option could be to add a comment to `FilesController` saying that `MuxService` needs to be updated if the request mapping ever changes. That's just my thoughts though, not any kind of requirement or anything

ice1080 (approved):

sgrimm:
### Merge activity

* **Sep 17, 5:59 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3383).

PR #3391, opened 2025-09-15
Add non-recorded species to survival rate calculations

ice1080:
Add species that have never been recorded in an observation to survival rate calculations (specifically only if they have a t0 density specified). 

Cleanup sumT0Field code that is no longer necessary.

Fix funder report calculation.


ice1080:
* **#3391** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3391?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3391?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -563,28 +568,40 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
...
+                .fullOuterJoin(PLOT_T0_DENSITY)
+                .on(
+                    PLOT_T0_DENSITY.MONITORING_PLOT_ID.eq(MONITORING_PLOT_ID)
+                        .and(PLOT_T0_DENSITY.SPECIES_ID.eq(SPECIES_ID))
+                )
+                .where(
+                    MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID)
+                        .or(PLOT_T0_DENSITY.MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID))
+                )
+                .and(OBSERVATION_ID.eq(OBSERVATIONS.ID).or(OBSERVATION_ID.isNull))
```
sgrimm, line 604: I think this approach is fine, but reading it, another possible approach occurred to me: using `UNION` and `DISTINCT ON` to split this up into two separate subqueries whose results get combined and deduplicated. I think it'd eliminate the need for the `COALESCE`s. But it might end up being more lines of code than this approach.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
ice1080: I'll try that out locally and see see which one reads better 👍

ice1080: (for reference) tested locally and it required a lot of jooq boilerplate. We decided to keep the original coalesce version.

PR #3394, opened 2025-09-16
SW-7313 Used up accessions should allow updates

ice1080:
When an accession was last withdrawn by seed and the latest observed quantity was not seed, and the accession is currently used up, it throws an error during `calculateRemaining` if the estimated seed count wasn't an exact match of the remaining weight.

Fix the error and add tests.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
@@ -251,10 +251,13 @@ data class AccessionModel(
           // always measured in seeds), we can't rely on the seeds-to-weight calculation precisely
           // matching the remaining weight quantity because weights have limited precision. Force
           // the remaining quantity to zero in that case.
+          // Also force the remaining quantity to zero if the accession is currently used up and the
+          // new state is also used up
           if (
-              latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
+              (latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
                   mostRecentWithdrawal?.withdrawn?.units == SeedQuantityUnits.Seeds &&
-                  mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount
+                  (mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount ||
+                      (state == AccessionState.UsedUp && existing.state == AccessionState.UsedUp)))
           ) {
             SeedQuantityModel.of(BigDecimal.ZERO, latestObservedQuantity.units)
           } else {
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3394** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
```diff
@@ -251,10 +251,13 @@ data class AccessionModel(
...
           // the remaining quantity to zero in that case.
+          // Also force the remaining quantity to zero if the accession is currently used up and the
+          // new state is also used up
           if (
-              latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
+              (latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
                   mostRecentWithdrawal?.withdrawn?.units == SeedQuantityUnits.Seeds &&
-                  mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount
+                  (mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount ||
+                      (state == AccessionState.UsedUp && existing.state == AccessionState.UsedUp)))
```
ice1080: (original code)
```kt
|| (state == AccessionState.UsedUp && existing.state == AccessionState.UsedUp)
```
This is not the correct approach to this problem. The real problem is that `mostRecentWithdrawal.withdrawn.quantity.toInt()` is non-zero, but `existing.estimatedSeedCount` is now 0. This should be updated to a different approach, but I'm not sure I fully understand what that approach should be regarding the state.

src/main/kotlin/com/terraformation/backend/seedbank/model/Accession.kt
```diff
@@ -251,10 +251,14 @@ data class AccessionModel(
...
+          // Also force the remaining quantity to zero if the accession is currently used up and the
+          // new state is also used up
           if (
-              latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
+              (latestObservedQuantity.units != SeedQuantityUnits.Seeds &&
                   mostRecentWithdrawal?.withdrawn?.units == SeedQuantityUnits.Seeds &&
-                  mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount
+                  (mostRecentWithdrawal.withdrawn.quantity.toInt() == existing.estimatedSeedCount ||
+                      (state == AccessionState.UsedUp &&
+                          existing.remaining?.quantity == BigDecimal.ZERO)))
```
ice1080: I'm also not sure if this is the right approach. It seems to work for all our tests and my 2 new ones, but I'm not sure if there are others we're not thinking of.

sgrimm: I can't think of a clearly-better approach offhand, so let's give this a whirl. One minor thing: I think you could use `remaining` instead of `existing.remaining` since this code will only be executed if the two are equal.

sgrimm (approved):

ice1080:
### Merge activity

* **Sep 16, 8:55 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3394).

PR #3395, opened 2025-09-16
SW-7385 Add created/modified fields to plot t0 tables

ice1080:
Add created/modified fields to `plot_t0_density` and `plot_t0_observation` tables.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
@@ -79,7 +79,19 @@ class T0PlotStore(
 
       with(PLOT_T0_DENSITY) {
         // ensure no leftover densities for species that are not in this observation
-        dslContext.deleteFrom(this).where(MONITORING_PLOT_ID.eq(monitoringPlotId)).execute()
+        dslContext
+            .deleteFrom(this)
+            .where(MONITORING_PLOT_ID.eq(monitoringPlotId))
+            .and(
+                SPECIES_ID.notIn(
+                    DSL.select(OBSERVED_PLOT_SPECIES_TOTALS.SPECIES_ID)
+                        .from(OBSERVED_PLOT_SPECIES_TOTALS)
+                        .where(OBSERVED_PLOT_SPECIES_TOTALS.OBSERVATION_ID.eq(observationId))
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3395** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3395?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3395?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
```diff
@@ -121,13 +151,32 @@ class T0PlotStore(
         // ensure no leftover densities for species that are not in this request
         dslContext.deleteFrom(this).where(MONITORING_PLOT_ID.eq(monitoringPlotId)).execute()
 
-        var insertQuery = dslContext.insertInto(this, MONITORING_PLOT_ID, SPECIES_ID, PLOT_DENSITY)
+        var insertQuery =
+            dslContext.insertInto(
+                this,
```
ice1080: Note that we're deleting rows in this plot before inserting, which kindof defeats the purpose of the modified fields. I could alternatively only delete ones that aren't in the list of species passed in, or could remove the modified columns. I'm open to either.

sgrimm: To be pedantic: I'd say it's the creation time that becomes meaningless here; the modified time ends up being correct. Outside of this transaction, deleting then inserting would be indistinguishable from updating in place, aside from the created by/time changing on each write.

I'd probably leave created by/time out completely, given that they won't ever be different from modified by/time. But that's just a preference; if you disagree, I'm fine with merging this as is.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
ice1080: The more I think about it, the more I think I'd prefer to make this method actually do an update instead of delete/insert. I recognize we'll likely never use the created columns, but I am personally a fan of traceable data when possible.

sgrimm: Sure, that works too.

sgrimm (approved):

PR #3399, opened 2025-09-17
SW-7277 Publish rate-limited t0 data set event per org per planting site

ice1080:

PR #3402, opened 2025-09-18
SW-7277 Send t0 data changed email on event

ice1080:
Send t0 data changed email on rate limited event.

Only include changed densities in change model.

Sort monitoring plots and species in email.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
@@ -885,34 +885,16 @@ class EmailNotificationService(
   fun on(event: RateLimitedT0DataAssignedEvent) {
     systemUser.run {
       val organization = organizationStore.fetchOneById(event.organizationId)
-      val plantingSiteId =
-          parentStore.getPlantingSiteId(event.monitoringPlots.first().monitoringPlotId)!!
-      val plantingSite = plantingSiteStore.fetchSiteById(plantingSiteId, PlantingSiteDepth.Plot)
-      val speciesIds =
-          event.monitoringPlots.flatMap { it.speciesDensityChanges }.map { it.speciesId }.toSet()
-      val species = speciesStore.fetchSpeciesByIdCollection(speciesIds)
-
-      val plotDetailsMap =
-          plantingSite.plantingZones
-              .flatMap { it.plantingSubzones }
-              .flatMap { it.monitoringPlots }
-              .associate { it.id to it.plotNumber }
-
-      event.monitoringPlots.forEach {
-        it.monitoringPlotNumber = plotDetailsMap[it.monitoringPlotId]!!
-        it.speciesDensityChanges.forEach {
-          it.speciesScientificName = species[it.speciesId]!!.scientificName
-        }
-        it.speciesDensityChanges = it.speciesDensityChanges.sortedBy { it.speciesScientificName }
-      }
+      val plantingSite =
+          plantingSiteStore.fetchSiteById(event.plantingSiteId, PlantingSiteDepth.Site)
 
       val model =
           T0PlotDataSet(
               config,
               organizationName = organization.name,
-              plantingSiteId = plantingSiteId,
+              plantingSiteId = event.plantingSiteId,
               plantingSiteName = plantingSite.name,
-              monitoringPlots = event.monitoringPlots.sortedBy { it.monitoringPlotNumber },
+              monitoringPlots = event.monitoringPlots,
           )
       emailService.sendOrganizationNotification(
           event.organizationId,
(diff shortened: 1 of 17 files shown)


ice1080:
* **#3402** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3399** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3399?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -879,6 +881,48 @@ class EmailNotificationService(
...
+              .flatMap { it.monitoringPlots }
+              .associate { it.id to it.plotNumber }
+
+      event.monitoringPlots.forEach {
+        it.monitoringPlotNumber = plotDetailsMap[it.monitoringPlotId]!!
+        it.speciesDensityChanges.forEach {
+          it.speciesScientificName = species[it.speciesId]!!.scientificName
+        }
+        it.speciesDensityChanges = it.speciesDensityChanges.sortedBy { it.speciesScientificName }
+      }
```
sgrimm: Mutating an event in the middle of an event listener will make the data flow harder to reason about IMO. If there are two listeners for this event class, we'll start having to care about what order they're called in to know whether we'd see these changes or not. Immutable event objects mean the event listeners can be called in any order and the events can be passed to other threads to do work in parallel without having to worry about race conditions.

I think a better approach here would be to define a new model class that has the data the email needs, and make the email model include a list of that class rather than a list of `PlotT0DensityChangedModel`.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
ice1080: That's fine with me, I think what I had been mainly trying to avoid was adding a bunch of data into the event since rate limited events get saved to the db and I didn't see other events with much data in them at all. But if we're good with that then I'll update.

src/main/kotlin/com/terraformation/backend/tracking/db/T0PlotStore.kt
```diff
@@ -255,19 +255,21 @@ class T0PlotStore(
...
+            )
+      }
+    }
 
     existingDensities.keys
-        .filter { it !in changeList }
+        .filter { it !in newDensities }
         .forEach { speciesId ->
           changeList[speciesId] =
               SpeciesDensityChangedModel(
```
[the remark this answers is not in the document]
ice1080: existingDensity is a BigDecimal so it can't be null

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
```diff
@@ -879,6 +881,35 @@ class EmailNotificationService(
     }
   }
 
+  @EventListener
+  fun on(event: RateLimitedT0DataAssignedEvent) {
+    if (event.monitoringPlots.flatMap { plot -> plot.speciesDensityChanges }.isEmpty()) {
+      // changes were reversed before the event was eventually refired
+      return
+    }
```
sgrimm, line 889: Is this a sign we need a way for rate-limited events to delete themselves when they're combined?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/email/EmailNotificationService.kt
ice1080: There are 2 other events that return early in `EmailNotificationService`, though I don't think either are rate limited. I think that is one approach for sure. Perhaps we could change `combine` to return `T?` instead of `T`, and when null is returned after `combine`, we remove the event?

sgrimm: Yeah, that's the same thought I had. Obviously not a topic for this PR, though (it's already big enough!)

ice1080:
### Merge activity

* **Sep 19, 9:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3402).
* **Sep 19, 9:22 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3402) rebased this pull request as part of a merge.
* **Sep 19, 9:32 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3402).

PR #3419, opened 2025-09-22
Upgrade to Java 25

sgrimm:


sgrimm:
* **#3419** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

build.gradle.kts
```diff
@@ -274,7 +274,7 @@ sourceSets.main { java.srcDir("build/generated/kotlin") }
 sourceSets.test { java.srcDir("build/generated-test/kotlin") }
 
 java {
-  toolchain { languageVersion = JavaLanguageVersion.of(24) }
+  toolchain { languageVersion = JavaLanguageVersion.of(25) }
   targetCompatibility = JavaVersion.VERSION_23
```
ice1080, line 278: Will we ever change the `targetCompatibility` to > 23? I tried it once and it broke some stuff so wasn't sure what was required

ice1080 (approved):

build.gradle.kts
sgrimm: The catch with that is that it needs to match the target compatibility of the Kotlin compiler, and right now Kotlin (or at least the Kotlin Gradle plugin) only supports up to 23. We'll bump it when we upgrade Kotlin to a version that supports 24+.

ice1080: Ahh it's kotlin based; that makes sense. Thanks!

PR #3420, opened 2025-09-22
SW-7408 Fix observations payload NPE

ice1080:
Completing observations is broken with this error:
```
{"error":{"message":"Field value cannot be null: plants[0].id"},"status":"error"}
```
which is caused by a new required id field. The field was added for response payloads, but perhaps it wasn't realized this was used in request payloads as well (specifically the `completeObservation` endpoint).
Make this field optional.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
@@ -732,7 +732,7 @@ data class RecordedPlantPayload(
     val certainty: RecordedSpeciesCertainty,
     @Schema(description = "GPS coordinates where plant was observed.") //
     val gpsCoordinates: Point,
-    val id: RecordedPlantId,
+    val id: RecordedPlantId?,
     @Schema(
         description = "Required if certainty is Known. Ignored if certainty is Other or Unknown."
     )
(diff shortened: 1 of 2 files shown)


ice1080:
Will end up merging this manually per the [confluence page](https://terraformation.atlassian.net/wiki/spaces/SW/pages/482017308/Deploying+code#Fix-a-bug-in-production-without-pulling-in-other-changes-from-main) requirements. Once approved, will merge manually and close the PR.

sgrimm (approved):
This is okay as a quick emergency fix, but I think we should define separate payloads for request and response since they're no longer identical.

Otherwise the OpenAPI schema will say this field is optional even in responses, and clients will have to treat it that way if they want to be type-safe.

ice1080:
> This is okay as a quick emergency fix, but I think we should define separate payloads for request and response since they're no longer identical.
> 
> Otherwise the OpenAPI schema will say this field is optional even in responses, and clients will have to treat it that way if they want to be type-safe.

I have wondered about that since we sometimes distinguish and sometimes do not. I'll do that work as a follow up to this once it's released. 
I was actually assuming that this would be a breaking change to the FE because typescript would need to handle the nullability, but in this case we're only doing equality checks with the `id` property for now, so we got off lucky this time.

PR #3427, opened 2025-09-23
SW-7411 Add db column and table for temp plots in survival rates

ice1080:
We need to store whether or not a planting site should include temp plots in its survival rate calculations. 
Add this column to `planting_sites`.
We also need to store plant density at a zone level instead of at a plot level for temp plots. 
Add this table.


ice1080:
* **#3428** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3427** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3427?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0400/V411__TempPlotsT0Density.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE tracking.planting_sites
+    ADD COLUMN survival_rate_includes_temp_plots BOOLEAN;
```
ice1080: Can also add `NOT NULL DEFAULT FALSE` if preferred

sgrimm: Yeah, I think that's preferable. The fewer nulls we have to deal with, the better.

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -565,6 +565,8 @@ COMMENT ON TABLE tracking.recorded_species_certainties IS '(Enum) Levels of cert
 
 COMMENT ON TABLE tracking.tree_growth_forms IS '(Enum) Growth form of each species in a biomass observation.';
 
+COMMENT ON TABLE tracking.zone_t0_temp_density IS 'Density for a zone per species, in plants per hectare. Only applies to temporary plots and only if site_temp_inclusions is set to true for the zone''s planting site.';
```
sgrimm: I guess `site_temp_inclusions` was a previous version of the column name?

src/main/resources/db/migration/0400/V411__TempPlotsT0Density.sql
```diff
@@ -0,0 +1,14 @@
...
+
+CREATE TABLE tracking.zone_t0_temp_density
```
sgrimm: The rest of the schema uses `planting_zone` instead of just `zone` in names. Maybe that was a mistake that just makes things needlessly verbose, but may as well stay consistent with the existing names.

sgrimm (changes requested):

src/main/resources/db/migration/R__Comments.sql
ice1080: I originally had it as a fully separate table before realizing it could just be a column on `planting_sites`. Will update!

src/main/resources/db/migration/0400/V411__TempPlotsT0Density.sql
ice1080: I've seen the alternative before, and less verbose ends up leading to increased confusion and mistakes, so I'm all about keeping things consistent for clarity's sake 👍

src/main/resources/db/migration/0400/V411__TempPlotsT0Density.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE tracking.planting_sites
+    ADD COLUMN survival_rate_includes_temp_plots BOOLEAN NOT NULL DEFAULT FALSE;
+
+CREATE TABLE tracking.planting_zone_t0_temp_density
```
sgrimm: Nit (sorry I didn't mention this earlier): "densities" would be a more consistent term than "density" here since we use plural table names.

sgrimm (approved):

src/main/resources/db/migration/0400/V411__TempPlotsT0Density.sql
ice1080: Ah we must have missed this on the other table (`plot_t0_density`). Will fix here and push a separate PR to fix that one.

PR #3428, opened 2025-09-23
SW-7411 Set boolean for Temp Plots in Survival Rate calculation

ice1080:
Add endpoint and update store for setting boolean on planting site to include temp plots in survival rate calculation.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
@@ -196,6 +196,21 @@ class PlantingSitesController(
     return SimpleSuccessResponsePayload()
   }
 
+  @Operation(
+      summary = "Sets the boolean for including temp plots in site survival rate calculations."
+  )
+  @PutMapping("/{id}/survivalRateIncludesTempPlots")
+  fun updateTempPlotsSurvivalRate(
+      @PathVariable("id") id: PlantingSiteId,
+      @RequestParam("survivalRateIncludesTempPlots") survivalRateIncludesTempPlots: Boolean,
+  ): SimpleSuccessResponsePayload {
+    plantingSiteStore.updatePlantingSite(id, emptyList()) { model: ExistingPlantingSiteModel ->
+      model.copy(survivalRateIncludesTempPlots = survivalRateIncludesTempPlots)
+    }
+
+    return SimpleSuccessResponsePayload()
+  }
+
   @ApiResponse200
   @ApiResponse409(
       description = "The planting site is in use, e.g., there are plantings allocated to the site."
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3428** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3427** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
ice1080, line 1: This endpoint didn't feel like it made sense in the t0 plot controller since it would then need access to the `plantingSiteStore`, but I can adjust if needed.
It also doesn't look like we have any required request params, so I can also change that to a path variable (or convert both the boolean and site id to a payload).

sgrimm: Why does this need to be a separate endpoint? Can we add the flag to `UpdatePlantingSiteRequestPayload` instead?

ice1080: We could, but then the caller would have to send the full payload of the planting site back for just setting this one flag, since several of the properties are required on it. I'm fine with that if that's what's preferred, wasn't sure if that's common practice when updating a single property on a page that isn't displaying a form with the rest of the properties

sgrimm: Doing a full PUT to update a single field is our typical flow, rather than defining a separate endpoint for each field.

From an API design point of view, it's not really a matter of whether or not there's a form shown to anyone, just a matter of keeping the API consistent. Often when I've implemented the APIs for this kind of thing, it's been before the designs were ready, so I had no idea what would or wouldn't be shown in the forms anyway; I just made sure to include all the writable fields in the PUT payload.

Obviously it's not great that the client has to pass along data it doesn't want to touch. Long term, I think the better alternative is to implement PATCH semantics where clients can omit the fields they don't want to modify. That'll be the cleanest way to do this kind of thing and I think we should try to figure it out at some point. It's a little tricky to implement given the muddy distinction between "field has a value of null" vs. "field isn't present" on the server side (also on the client, to some extent, though at least JS has separate `null` and `undefined` concepts). Doable, but hasn't warranted the effort yet.

sgrimm (approved):

PR #3430, opened 2025-09-24
SW-7403 Move thumbnail management to new service

sgrimm:
In preparation for generating thumbnail images for videos, restructure the
internal APIs for fetching and deleting thumbnails.

Previously, the flow for generating a new thumbnail looked like

    caller
    -> FileService.readFile
    -> ThumbnailStore.getThumbnailData
    -> FileStore.readFile (to fetch original file)

Now `FileService` is only concerned with reading and writing files, not with
transforming them to different sizes. The flow now looks like

    caller
    -> ThumbnailService.readFile
    -> ThumbnailStore.getThumbnailData
    -> FileStore.readFile (to fetch original file)

In addition to having a cleaner separation of concerns, this avoids a problem
when video thumbnail support is added. For videos, we'll need to fetch
thumbnails from Mux, which will mean calling `MuxService`. `MuxService` depends
on `FileService`. If we called `MuxService` from `FileService`, there'd be a
circular dependency between the two services, which we try to avoid whenever
possible.  With this new structure, the call to `MuxService` can happen from
`ThumbnailService` instead.

Handling deletion of thumbnails for video files would have introduced a similar
circular dependency. To break that cycle, `FileService` now publishes an event
when it deletes a file, rather than explicitly calling `ThumbnailStore` to
delete thumbnails. The thumbnails are deleted by an event listener. This also
separates the concerns more cleanly.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/ActivityMediaService.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/ActivityMediaService.kt
@@ -25,6 +25,7 @@ import com.terraformation.backend.db.accelerator.tables.references.ACTIVITY_MEDI
 import com.terraformation.backend.db.default_schema.FileId
 import com.terraformation.backend.file.FileService
 import com.terraformation.backend.file.SizedInputStream
+import com.terraformation.backend.file.ThumbnailService
 import com.terraformation.backend.file.event.VideoFileDeletedEvent
 import com.terraformation.backend.file.event.VideoFileUploadedEvent
 import com.terraformation.backend.file.model.NewFileMetadata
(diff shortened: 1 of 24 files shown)


sgrimm:
* **#3430** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3430?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3430?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3431, opened 2025-09-24
SW-7432 Convert Plot T0 Densities to plants per hectare

ice1080:
Convert `plot_t0_densities` to represent plants per hectare instead of plants per plot. 

Update survival rate calculations to assume densities are in plants per hectare.

Fix `T0PlotStore` to not treat observation and manually assigned densities as both the same unit (which was incorrect before), and to instead treat observations as plants per plot (requiring conversion) and manual densities as plants per hectare.

Update tests to match.


ice1080:
* **#3441** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3439** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3439?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3438** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3434** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3434?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3433** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3432** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3432?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3431** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/util/Constants.kt
```diff
@@ -2,3 +2,5 @@ package com.terraformation.backend.util
 
 /** Number of square meters in a hectare. */
 const val SQUARE_METERS_PER_HECTARE: Double = 10000.0
+
+const val HECTARES_IN_PLOT: Double = 0.09
```
sgrimm: Nit: I think "per" rather than "in" would make it slightly clearer than this is a conversion factor that's constant across all plots, as opposed to, say, a jOOQ name for a column whose value could differ from plot to plot.

src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
```diff
@@ -4795,35 +4800,75 @@ class ReportStoreTest : DatabaseTest(), RunsAsDatabaseUser {
             plantingCompletedTime = plantingCompletedDate1.atStartOfDay().toInstant(ZoneOffset.UTC),
         )
     val subzone1plot1 = insertMonitoringPlot(permanentIndex = 1)
-    insertPlotT0Density(speciesId = speciesId, plotDensity = BigDecimal.valueOf(10))
-    insertPlotT0Density(speciesId = otherSpeciesId, plotDensity = BigDecimal.valueOf(11))
+    insertPlotT0Density(
+        speciesId = speciesId,
+        plotDensity =
+            BigDecimal.valueOf(10).divide(hectaresInPlot, divisionScale, RoundingMode.HALF_UP),
```
sgrimm: Extracting this `divide()` call out into a little helper function (perhaps an extension method on `BigDecimal`) that also gets used in the application code would mean the scale and rounding mode are only defined in one place, and the tests would be more concise too.

src/test/kotlin/com/terraformation/backend/tracking/db/T0PlotStoreTest.kt
```diff
@@ -346,10 +351,38 @@ internal class T0PlotStoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      assertEquals(
+          BigDecimal("888.8888888889"),
+          dbDensity,
+          "Should convert density to plants/ha",
+      )
+      assertEquals(
+          10,
+          dbDensity.scale(),
+          "Should store plot density with a scale of 10",
+      )
```
sgrimm: Could this assertion ever fail if the previous one succeeded?

sgrimm (changes requested):

src/test/kotlin/com/terraformation/backend/tracking/db/T0PlotStoreTest.kt
ice1080: I thought that scales weren't compared and that `BigDecimal("12") == BigDecimal("12.0")` but I just tested in a playground and looks like I'm wrong there. So no it could never fail, but since we are explicitly setting the scale in the code, I would still like to keep an explicit check in the test. I'll convert to an `assertAll` though if that would be better?

src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
ice1080: I like this, but am having a hard time coming up with a name. Are you referring to a helper that converts plants per plot to plants per hectare? Or rather just something like a `terrawareDivide` that has our common scale and rounding mode?

ice1080: I assumed the latter and went with `divideHalfUp`. Open to other names, or let me know if you wanted a divide method like `plantsPerPlotToPerHectare` that includes the `HECTARES_PER_PLOT` const usage in it.

sgrimm: I was thinking it'd be a plot-to-hectares conversion so we don't have to specify `HECTARES_PER_PLOT` everywhere.

sgrimm (approved):
This is reasonable, but if you want to tweak the helper function to do the hectare/plot conversion, I think we can make the tests and code more concise.

ice1080:
### Merge activity

* **Sep 25, 9:28 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3431).

PR #3437, opened 2025-09-25
Add assertSetEquals to compare sets in tests

sgrimm:
We use sets to compare unordered collections of items in a lot of our tests.
The default JUnit `assertEquals` will correctly detect when two sets aren't
equal, but the assertion failure message is hard to read, especially when
the two sets have the same items but in a different order.

Add a helper function to compare sets and produce a more human-friendly failure
message if they aren't equal. The items are sorted by their string representations
and are separated by newlines.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/Helpers.kt
+++ b/src/test/kotlin/com/terraformation/backend/Helpers.kt
@@ -65,6 +65,24 @@ fun assertGeometryEquals(expected: Geometry?, actual: Geometry?, message: String
   }
 }
 
+/**
+ * Asserts that two sets are equal. If not, produces an assertion failure message with the items
+ * sorted by their string representations, each item on a separate line, to make it easier to spot
+ * differences.
+ */
+fun <T> assertSetEquals(expected: Set<T>, actual: Set<T>, message: String? = null) {
+  if (expected != actual) {
+    assertEquals(expected.toPrettyString(), actual.toPrettyString(), message)
+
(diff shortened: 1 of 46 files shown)


sgrimm:
* **#3437** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
This isn't motivated by a specific failure I was seeing; it just occurred to me as a nice-to-have when I was updating a test to compare using sets. I did most of this using a search-and-replace on `assertEquals(setOf(...),...)` so I probably missed some assertions that could be converted, but we can update those over time.

src/test/kotlin/com/terraformation/backend/Helpers.kt
```diff
@@ -65,6 +65,24 @@ fun assertGeometryEquals(expected: Geometry?, actual: Geometry?, message: String
...
+
+    // Should never get here unless the string representations of the items are incomplete.
+    assertEquals(
+        expected,
+        actual,
+        "Sets are not equal, but their items' string representations are the same",
+    )
+  }
+}
+
```
sgrimm, line 85: Implementation is here

ice1080 (approved):
Thanks for adding this!

PR #3442, opened 2025-09-26
SW-7346 Add hidden, list position to activity media

sgrimm:
Add a boolean field to indicate whether a media file should be hidden on the map.

Add an integer field to indicate the list position of each file. This can be
updated to rearrange the list.

Since maintaining the list position requires additional database operations, move
the media file deletion logic into `ActivityMediaStore`. `ActivityMediaService`
no longer directly interacts with the database.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
@@ -115,6 +115,7 @@ class ActivitiesController(
   @RequestBodyPhotoFile
   fun uploadActivityMedia(
       @PathVariable activityId: ActivityId,
+      @RequestPart("listPosition", required = false) listPositionStr: String?,
       @RequestPart("file") file: MultipartFile,
   ): UploadActivityMediaResponsePayload {
     val contentType = file.getPlainContentType(supportedMediaTypes)
@@ -125,6 +126,7 @@ class ActivitiesController(
             activityId,
             file.inputStream,
             FileMetadata.of(contentType, filename, file.size),
+            listPositionStr?.toIntOrNull(),
         )
 
     return UploadActivityMediaResponsePayload(fileId)
@@ -198,6 +200,8 @@ data class ActivityMediaFilePayload(
     val fileId: FileId,
     val geolocation: Point?,
     val isCoverPhoto: Boolean,
+    val isHiddenOnMap: Boolean,
+    val listPosition: Int,
     val type: ActivityMediaType,
 ) {
   constructor(
@@ -208,6 +212,8 @@ data class ActivityMediaFilePayload(
       fileId = model.fileId,
       geolocation = model.geolocation,
       isCoverPhoto = model.isCoverPhoto,
+      isHiddenOnMap = model.isHiddenOnMap,
+      listPosition = model.listPosition,
       type = model.type,
   )
 }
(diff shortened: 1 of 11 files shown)


sgrimm:
* **#3442** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ActivityMediaStore.kt
```diff
@@ -28,6 +37,22 @@ class ActivityMediaStore(
...
+    requirePermissions { readActivity(activityId) }
+
+    return fetchByCondition(ACTIVITY_MEDIA_FILES.ACTIVITY_ID.eq(activityId))
+  }
+
+  fun fetchByOrganizationId(organizationId: OrganizationId): List<ActivityMediaModel> {
+    requirePermissions { readOrganization(organizationId) }
+
+    val user = currentUser()
+    return fetchByCondition(
```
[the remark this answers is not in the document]
sgrimm: The event isn't published asynchronously, and the active user when it's published will have permission to delete activities since they have permission to delete the organization.

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
```diff
@@ -125,6 +126,7 @@ class ActivitiesController(
             activityId,
             file.inputStream,
             FileMetadata.of(contentType, filename, file.size),
+            listPositionStr?.toIntOrNull(),
```
ice1080: Will this throw the correct error if `listPosition` is passed as "ABC" instead of a parseable number?

src/main/kotlin/com/terraformation/backend/accelerator/db/ActivityMediaStore.kt
```diff
@@ -78,4 +205,29 @@ class ActivityMediaStore(
...
+          .where(ACTIVITIES.ID.eq(activityId))
+          .forUpdate()
+          .fetchOne() ?: throw ActivityNotFoundException(activityId)
+
+      func()
+    }
+  }
+
+  private fun updateListPositions(activityId: ActivityId, fileIds: List<FileId>) {
+    val positionsMap = fileIds.mapIndexed { index, fileId -> fileId to index + 1 }.toMap()
```
ice1080, line 230: Can we add a quick javadoc or comment that indicates that `fileIds` is the list of `FileId`s already in the desired order? Initially it seemed like this was just setting the `list_position`s to be in the order of `file_id`s numeric ascending

src/test/kotlin/com/terraformation/backend/accelerator/db/ActivityMediaStoreTest.kt
ice1080, line 1: Do we need a test to ensure the list positions stay the same if the requested position is the same as the current? I'm not familiar enough to say this is definitely needed, so if it's not just mark this as resolved.

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
sgrimm: Good point. It ignores invalid numbers right now but it'd be better to reject them.

sgrimm: Also added a `@Schema` annotation to specify that this is a number, though I'm not sure if the code generator in terraware-web will do anything useful with it since this is a multipart request rather than a JSON field.

ice1080 (approved):

PR #3458, opened 2025-10-01
Properly set memory limit for test suite

sgrimm:
Previously, we were setting the memory limit for the Gradle worker process that
runs the test suite by adding it to Gradle's JVM arguments. That no longer works;
instead we need to include it in the configuration for the test-running task.
Otherwise Gradle will use the default limit of 512MB, which isn't enough to
run our tests in parallel without lots of memory pressure that causes frequent
GC pauses.

In a local dev environment, this change cuts test suite execution time in half.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -194,6 +194,7 @@ testing {
           targets {
             all {
               testTask.configure {
+                jvmArgs("-Dfile.encoding=UTF-8", "-Duser.country=US", "-Xmx5120m")
                 systemProperty("java.locale.providers", "SPI,CLDR")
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#3458** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3458?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3458?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3472, opened 2025-10-06
SW-7414 Add temp plots to completePlot survival rate calc

ice1080:
Add temp plots to survival rate calculations. 

Some more refactoring will occur in future PRs (hence the TODO).


ice1080:
* **#3473** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3473?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3472** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/resources/tracking/observation/SurvivalRateSiteTempData/SubzoneRates.csv
ice1080, line 1: As a note, several of the Zone Survival Rates are not yet correct because they are implemented in other methods

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2095,6 +2108,24 @@ class ObservationStore(
     }
   }
 
+  private enum class UpdateSpeciesType {
+    Plot,
+    Subzone,
+    Zone,
+    Site;
+
+    fun <ID : Any> getTempZoneCondition(scopeId: ID): Condition =
```
sgrimm, line 2117: Would it make sense to pull the other table-specific conditions from `updateSpeciesTotalsTable` into this enum too?

Alternately, instead of an enum with a `when`, make this an interface with implementations for plot, subzone, zone, and site; then you wouldn't need to do the `scopeId as Whatever` casts since each implementation would use the correct types for its scope.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: Just reread your TODO and it seems like you were planning to move more things here already. But I think making it an interface+classes would be cleaner than an enum.

ice1080: Yes, working on it now! And I'm actually running into issues with the scopeId generic anyway, so I like the interface+classes idea. Thanks!

ice1080:
### Merge activity

* **Oct 8, 4:29 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3472).

PR #3475, opened 2025-10-08
SW-7497 Added status to Activity

tommylau523:
Activity will need a Do Not Use status, on top of Verified and Not Verified.


tommylau523:
* **#3475** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3475?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3475?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
```diff
@@ -282,6 +285,7 @@ data class UpdateActivityMediaRequestPayload(
 data class UpdateActivityRequestPayload(
     val date: LocalDate,
     val description: String,
+    val status: ActivityStatus,
```
sgrimm, line 296: Wouldn't this need to be copied to the model in `applyTo`?

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/accelerator/api/ActivitiesController.kt
tommylau523: Yes absolutely! Sorry I meant to leave this as draft since it was still WIP, but it is ready now!

sgrimm (approved):

sgrimm:
Seems like this is still under development; should I review it yet?

tommylau523:
> Seems like this is still under development; should I review it yet?

This is ready now. I caught a few edge cases when I was integrating with the FE, so I went and made some changes. Appreciate the reviewing effort and apologies for being hasty with sending the PR out.

src/main/resources/db/migration/0400/V414__ActivityStatus.sql
```diff
@@ -0,0 +1,29 @@
...
+ALTER TABLE accelerator.activities
+    ALTER COLUMN activity_status_id SET NOT NULL;
+
+ALTER TABLE accelerator.activities
+    ADD CONSTRAINT activity_status_verified
+    CHECK (
+        (verified_by IS NULL AND activity_status_id = 1)
+            OR (verified_by IS NOT NULL AND activity_status_id = 2)
+            OR (activity_status_id = 3)
+        );
```
sgrimm: Alternately, maybe something like
```
CHECK (
  CASE activity_status_id
    WHEN 1 THEN verified_by IS NULL
    WHEN 2 THEN verified_by IS NOT NULL
    ELSE TRUE
  END
)
```
which may be easier to follow.

src/test/kotlin/com/terraformation/backend/accelerator/db/ActivityStoreTest.kt
```diff
@@ -522,7 +531,47 @@ class ActivityStoreTest : DatabaseTest(), RunsAsDatabaseUser {
     }
 
     @Test
-    fun `clears verifiedBy and verifiedTime if isVerified goes from true to false`() {
+    fun `sets verifiedBy and verifiedTime if status goes to Verified from Not Verified`() {
```
sgrimm: Nit: this test name and the following one would read a little better as "from X to Y" rather than "to X from Y."

src/main/kotlin/com/terraformation/backend/accelerator/model/ActivityModel.kt
```diff
@@ -24,8 +26,10 @@ data class ExistingActivityModel(
     val projectId: ProjectId,
     val verifiedBy: UserId? = null,
     val verifiedTime: Instant? = null,
-    val isVerified: Boolean = verifiedBy != null,
 ) {
+  val isVerified: Boolean
+    get() = verifiedBy != null
```
sgrimm: The semantics are a little confusing here, though that may reflect the requirements. We have a status of `Verified` and a property `isVerified`, which look like they ought to mean the same thing, but if I'm reading the code right, `isVerified` can be true even when the status isn't `Verified` because we ignore verified by/time if the status is `Do Not Use`.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/model/ActivityModel.kt
tommylau523: Yes, that is because we used this to check whether the verified by/verified time need to be cleared/updated. After some requirements discussions, it makes more sense to treat "Do Not Use" to work similarly as "Not Verified". It is not going to happen routinely but a new verification is required for this state change essentially.

tommylau523:
### Merge activity

* **Oct 9, 4:47 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3475).

PR #3478, opened 2025-10-08
SW-7415 Add temp plots to survival rate recalculations

ice1080:
Add temp plots to survival rate recalculation methods.

Refactor survival rate denominator.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -1563,13 +1563,19 @@ class ObservationStore(
       observationId: ObservationId,
       plantingSiteId: PlantingSiteId,
   ) {
-    data class ZoneInfo(
+    data class SubzoneSpeciesRecord(
+        val certaintyId: RecordedSpeciesCertainty,
+        val speciesId: SpeciesId?,
+        val speciesName: String?,
+        val plantingZoneId: PlantingZoneId,
         val permanentLive: Int,
-        val totalLive: Int,
         val cumulativeDead: Int,
+        val totalLive: Int,
         val survivalRateIncludesTempPlots: Boolean,
     )
-    val liveAndDeadTotals: Map<RecordedSpeciesKey, Map<PlantingZoneId, List<ZoneInfo>>> =
+
+    val liveAndDeadTotals:
+        Map<RecordedSpeciesKey, Map<PlantingZoneId, List<SubzoneSpeciesRecord>>> =
         with(OBSERVED_SUBZONE_SPECIES_TOTALS) {
           val observationIdForSubzoneField =
               DSL.select(OBSERVATIONS.ID)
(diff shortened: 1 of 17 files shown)


ice1080:
* **#3479** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3478** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1599,55 +1605,92 @@ class ObservationStore(
...
+                      recordsForZone.map {
+                          (
+                              _,
+                              _,
+                              _,
+                              _,
+                              permanentLive,
+                              cumulativeDead,
+                              totalLive,
+                              survivalRateIncludesTempPlots) ->
```
sgrimm: With this many fields, maybe it's worth defining another little local class to hold the records from this query rather than using destructuring.

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -476,9 +611,12 @@ class ObservationStoreSurvivalRateCalculationTest : ObservationScenarioTest() {
             actualByIdMap.entries.joinToString("\n") { (id, speciesMap) ->
               "$level $id: {${speciesMap.entries.joinToString(", ") { "SpeciesId: ${it.key} -> SurvivalRate: ${it.value}" }}}"
             }
-        assertEquals(expectedString, actualString, "$message - $level")
+        assertionList += Triple(expectedString, actualString, "$message - $level")
```
sgrimm: Alternately: make `assertionList` a list of functions, so this becomes
```
assertionList += { assertEquals(expectedString, actualString, "$message - $level") }
```
and then the list gets passed directly to `assertAll` below.

sgrimm: Or, maybe better, use `mapIndexed` instead of `forEachIndexed` so the loop returns the list of assertions rather than appending to a mutable list.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1599,55 +1605,92 @@ class ObservationStore(
...
 
           val rowsInserted =
               dslContext
                   .insertInto(OBSERVED_ZONE_SPECIES_TOTALS)
                   .set(CERTAINTY_ID, speciesKey.certainty)
                   .set(CUMULATIVE_DEAD, totalDead)
-                  .set(MORTALITY_RATE, zoneMortalityRate)
+                  .set(
+                      MORTALITY_RATE,
+                      if (totalDead > 0 && totalPermanentLive > 0) zoneMortalityRate else null,
```
sgrimm: Would it be clearer to make `zoneMortalityRate` null if these totals are both 0, so it never has a numeric value that isn't considered valid/usable?

This would become `zoneMortalityRate` instead of the `if` expression, and the usage of it below would change to `zoneMortalityRate?.let { DSL.inline(it) } ?: MORTALITY_RATE`.

sgrimm (approved):
This would have been a bit easier to review if the refactor had been split out into a separate PR from the functionality change.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
ice1080: Do you have a number of fields that you prefer to create a local class? I would have assumed this should have happened a while back since we were already at 6 fields...

sgrimm: Arguably I should have done it from the get-go, but if we want a rubric, maybe "destructuring list doesn't fit on one line" would be a useful one.

ice1080:
> This would have been a bit easier to review if the refactor had been split out into a separate PR from the functionality change.

@sgrimm I'm struggling to agree with this. This PR was adding nearly identical code to `getSurvivalRateTempDenominator` as what was already in `getSurvivalRateDenominator`, so I moved the 15 or so lines into a common method. In the past when I have done refactors like this in separate PRs, I have received requests to clean up the code in the original PR.

sgrimm:
That's fair. I failed to realize that the refactoring was that small.

ice1080:
> That's fair. I failed to realize that the refactoring was that small.

Perhaps the takeaway for me is to not mention something like that in the PR description when it's that small. That was technically the second commit, and graphite automatically includes the commit messages as descriptions initially, so I can just remove something like that in the future.

ice1080:
### Merge activity

* **Oct 8, 10:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3478).
* **Oct 8, 10:18 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3478).

PR #3480, opened 2025-10-08
Assert that t0 data can only be assigned to permanent plots

ice1080:
We discovered that any plot that had ever been permanent in an observation were being included in the Survival Rate settings page. This was causing the FE to assign t0 data to permanent plots which isn't supposed to be allowed. The data was being rolled back when the `T0PlotDataAssignedEvent` occurred and the survival rates were being recalculated. Instead of relying on a different method to assert permanence, add a permanence check to the `assign` methods.


ice1080:
* **#3480** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm (approved):

PR #3481, opened 2025-10-08
Use event-based flow to delete unreferenced files

sgrimm:
Previously, to delete a file from the file store, application code would call
`FileService.deleteFile` and pass it a function to remove the reference to the
file from the appropriate child tables.

That worked well enough for simple cases where there was a 1:1 relationship
between files and references to them. But we're starting to see more complex
1:N relationships, and in those cases, it's not always clear which code should
be responsible for calling `FileService.deleteFile`.

Improve the situation by moving to an event-based scheme. When application code
deletes a reference to a file, instead of doing

    fileService.deleteFile(fileId) {
        deleteRowsThatReferToFile(fileId)
    }

it should now do

    deleteRowsThatReferToFile(fileId)
    eventPublisher.publishEvent(FileReferenceDeletedEvent(fileId))

`FileService` listens for the event and takes care of deleting the file from
the file store if there are no other references to the file.

With this setup, in a 1:N situation, the event can be published as each reference
is deleted, without regard to whether or not there are other references. When the
last one is gone, the file will be deleted.

Since the files in question may be video files that are associated with Mux
assets, and we only want to delete those when the video is no longer
referenced, Mux asset deletion is now handled the same way as deletion of
thumbnail images: it's triggered by the `FileDeletionStartedEvent` that's
published by `FileService` before the file is actually removed from the file
store, rather than by a video-specific event from `ActivityMediaService`. That
way as we add video support to more places in the app, we won't have to
remember to publish the correct deletion event everywhere.

We use a "scan for references" approach rather than maintaining a reference count
because it's simpler to make robust: it's impossible to fail to increment or
decrement the reference count. We also have a relatively low rate of file
deletion, so scanning a bunch of tables for references should be less expensive
overall than keeping reference counts up to date.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/file/FileService.kt
+++ b/src/main/kotlin/com/terraformation/backend/file/FileService.kt
@@ -3,23 +3,16 @@ package com.terraformation.backend.file
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.config.TerrawareServerConfig
 import com.terraformation.backend.daily.DailyTaskTimeArrivedEvent
+import com.terraformation.backend.db.DefaultCatalog
 import com.terraformation.backend.db.FileNotFoundException
 import com.terraformation.backend.db.TokenNotFoundException
-import com.terraformation.backend.db.accelerator.tables.references.ACTIVITY_MEDIA_FILES
-import com.terraformation.backend.db.accelerator.tables.references.REPORT_PHOTOS
-import com.terraformation.backend.db.accelerator.tables.references.SUBMISSION_SNAPSHOTS
 import com.terraformation.backend.db.default_schema.FileId
 import com.terraformation.backend.db.default_schema.tables.daos.FilesDao
 import com.terraformation.backend.db.default_schema.tables.pojos.FilesRow
 import com.terraformation.backend.db.default_schema.tables.references.FILES
 import com.terraformation.backend.db.default_schema.tables.references.FILE_ACCESS_TOKENS
-import com.terraformation.backend.db.default_schema.tables.references.SEED_FUND_REPORT_FILES
-import com.terraformation.backend.db.docprod.tables.references.VARIABLE_IMAGE_VALUES
-import com.terraformation.backend.db.funder.tables.references.PUBLISHED_REPORT_PHOTOS
-import com.terraformation.backend.db.nursery.tables.references.BATCH_PHOTOS
-import com.terraformation.backend.db.nursery.tables.references.WITHDRAWAL_PHOTOS
-import com.terraformation.backend.db.seedbank.tables.references.ACCESSION_PHOTOS
-import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PHOTOS
+import com.terraformation.backend.db.default_schema.tables.references.MUX_ASSETS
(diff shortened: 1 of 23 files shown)


sgrimm:
* **#3481** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3481?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3481?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/file/FileService.kt
```diff
@@ -171,7 +177,32 @@ class FileService(
...
+      // Clean up thumbnails, Mux assets, etc.
+      eventPublisher.publishEvent(FileDeletionStartedEvent(event.fileId, filesRow.contentType!!))
+
+      try {
+        fileStore.delete(storageUrl)
+      } catch (_: NoSuchFileException) {
+        log.warn("File $storageUrl was already deleted from file store")
+      }
+
+      filesDao.deleteById(fileId)
```
ice1080, line 186: Does `FileDeletionStartedEvent` need to be in the same transaction as `deleteById`? Or is it better if it's not?

src/main/kotlin/com/terraformation/backend/file/FileService.kt
```diff
@@ -41,6 +51,26 @@ class FileService(
 ) {
   private val log = perClassLogger()
 
+  /**
+   * Fields that count as references to files. If a file isn't referenced in any of these, it will
+   * be deleted from the file store. This doesn't necessarily include all the foreign-key
+   * relationships with the files table, just ones whose presence should cause the file to be
+   * considered still in use. For example, thumbnails aren't included here.
+   */
```
ice1080, line 52: One option here to ensure we never forget to add to this list would be write a test that checks the list of referencingFields against the list of all fileId FKs, and have exclusions such as the thumbnails. Not sure if it would be worth it though

sgrimm: I guess if I'm going to look that up dynamically for purposes of the test, may as well just look it up dynamically here so the list doesn't need to be manually edited.

src/main/kotlin/com/terraformation/backend/file/FileService.kt
sgrimm: I'm leaving that up to the caller. Unfortunately there's no simple blanket answer because the operation on the file store (S3, etc.) is non-transactional. So you can end up with inconsistent state whether or not there's a transaction involved: the files row still exists but the file is gone from the file store, or vice versa.

sgrimm: We could kind of button it down with something like a two-phase commit setup (e.g., set an "I'm deleting this" flag and commit that, then do the file store operation) but I think that'd be more complicated than it's worth, especially given that callers might have transactions open and you'd then have to work out what the end state should be when the caller's transaction gets rolled back for some reason that has nothing to do with the file deletion.

ice1080 (approved):

PR #3486, opened 2025-10-09
Allow ReadOnly users to access project_internal_users in search requests

ice1080:
The Matrix View now uses the search api to get project leads for projects, but users are currently only visible if they are in the same org as the current user (or if they're a `project_internal_user` and the project is in the same org as the current user. 

Update the `UsersTable` to allow ReadOnly+ users to see `project_internal_users`, even if not in the same org.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
@@ -65,6 +65,16 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
                     ),
                     DSL.exists(
                         DSL.selectOne()
+                            .from(PROJECT_INTERNAL_USERS)
+                            .where(USERS.ID.eq(PROJECT_INTERNAL_USERS.USER_ID))
+                            .and(
+                                currentUser().canReadAllAcceleratorDetails().let {
+                                  if (it) DSL.trueCondition() else DSL.falseCondition()
+                                }
+                            )
+                    ),
+                    DSL.exists(
+                        DSL.selectOne()
                             .from(PROJECTS)
                             .join(PROJECT_INTERNAL_USERS)
                             .on(PROJECTS.ID.eq(PROJECT_INTERNAL_USERS.PROJECT_ID))
--- a/src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
@@ -63,25 +63,26 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
                                 )
                             )
                     ),
-                    if (currentUser().canReadAllAcceleratorDetails())
-                        DSL.exists(
-                            DSL.selectOne()
-                                .from(PROJECT_INTERNAL_USERS)
-                                .where(USERS.ID.eq(PROJECT_INTERNAL_USERS.USER_ID))
-                        )
-                    else
-                        DSL.exists(
-                            DSL.selectOne()
-                                .from(PROJECTS)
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3486** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3486?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3486?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
ice1080, line 1: I kept this as a separate `exists` in the `or` list, but I can combine with the one below it and use a `.or` condition instead if preferred.

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
```diff
@@ -63,6 +63,16 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
...
+                    DSL.exists(
+                        DSL.selectOne()
+                            .from(PROJECT_INTERNAL_USERS)
+                            .where(USERS.ID.eq(PROJECT_INTERNAL_USERS.USER_ID))
+                            .and(
+                                currentUser().canReadAllAcceleratorDetails().let {
+                                  if (it) DSL.trueCondition() else DSL.falseCondition()
+                                }
+                            )
+                    ),
```
sgrimm: Rather than returning a false condition to make this a no-op, maybe make the whole `exists` expression null if the user can't read accelerator details, and make the enclosing list a `listOfNotNull`.

Also, I don't think the `let` is needed here. It could be `if (currentUser().canReadAllAcceleratorDetails())`.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
ice1080: I realized that the block after this one is only necessary if the user _isn't_ ReadOnly+ so instead of doing this:
```kt
listOfNotNull(
  if (currentUser().canReadAllAcceleratorDetails()) {DSL.exists(...)} else null,
  DSL.exists(...
```
I could just do this:
```kt
listOf(
  if (currentUser().canReadAllAcceleratorDetails()) {DSL.exists(...)} else {
    DSL.exists(...
```

src/main/kotlin/com/terraformation/backend/search/table/UsersTable.kt
```diff
@@ -63,16 +63,25 @@ class UsersTable(private val tables: SearchTables) : SearchTable() {
...
-                        DSL.selectOne()
-                            .from(PROJECTS)
-                            .join(PROJECT_INTERNAL_USERS)
-                            .on(PROJECTS.ID.eq(PROJECT_INTERNAL_USERS.PROJECT_ID))
-                            .where(USERS.ID.eq(PROJECT_INTERNAL_USERS.USER_ID))
-                            .and(
-                                PROJECTS.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys)
-                            )
-                    ),
+                    if (currentUser().canReadAllAcceleratorDetails())
```
sgrimm: Ah, this is much nicer, yeah. I'd put these expressions in curly braces so it's easier to see they're children of the if/else.

ice1080:
### Merge activity

* **Oct 10, 4:38 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3486).

PR #3488, opened 2025-10-09
SW-7514 Fix issue where sublist filtering depends on order of search fields

ice1080:
Sublist Filtering was only working if there were no flattened fields in the sublists being filtered (or if a direct property on the sublist was being queried above).
Example:
```json
"prefix": "projects",
"fields": [
  "id",
  "internalUsers.user_firstName",
],
"filters": [{
  "prefix": "internalUsers",
  "search": {
    "operation": "field",
    "field": "role",
    "values": ["Project Lead"]
  }
}]
```
This would only work if the field `internalUsers.role` was added _above_ `internalUsers.user_firstName` because then it would get processed first and the sublist query would get built with the correct criteria. 

The problem was that `fieldPath.prefix` for `internalUsers.user_firstName` is `internalUsers.user_` which causes `criteria[fieldPath.prefix]` to be null when adding the `sublistQuery` object to `sublistQueryBuilders`. 

Change this to `criteria[prefix.relativeSublistPrefix(sublistName)]` which correctly returns `internalUsers`.

Uncomment related sublist-filtering test, and delete duplicate test.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -779,7 +779,8 @@ class NestedQueryBuilder(
     } else if (relativeField.isNested) {
       // Prefix = a.b, fieldName = a.b.c.d => make a sublist "c" to hold field "d"
       val sublistName = getSublistName(relativeField)
-      val sublistQuery = getSublistQuery(relativeField, criteria[fieldPath.prefix])
+      val sublistQuery =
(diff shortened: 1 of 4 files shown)


ice1080:
* **#3488** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm (approved):
Good catch.

PR #3497, opened 2025-10-10
SW-7416 Add temp plots to observation results for plots and subzones

ice1080:
Add temp plots to survival rate calculations in observation results for plots and subzones.

Zones are handled in https://github.com/terraware/terraware-server/pull/3503 and sites are handled in https://github.com/terraware/terraware-server/pull/3504; this was broken up to keep the changes easier to review.


ice1080:
* **#3504** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3503** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3497** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -802,20 +883,25 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
...
                       SURVIVAL_RATE,
                       DSL.`when`(
-                          densityCol.gt(BigDecimal.ZERO),
-                          DSL.coalesce(PERMANENT_LIVE, 0)
-                              .div(densityCol)
-                              .div(DSL.inline(HECTARES_PER_PLOT)),
+                          DSL.coalesce(permDensityCol, BigDecimal.ZERO)
+                              .plus(DSL.coalesce(tempDensityCol, BigDecimal.ZERO))
+                              .gt(BigDecimal.ZERO),
+                          BigDecimal.ZERO,
```
sgrimm, line 889: The result of this `when` will only ever be zero or null, unless I'm reading it wrong; is that the intent here?

ice1080: That is correct. It's inside a coalesce, so it will first get the survival rate if it exists. If it does not, then we want to check if the survival rate should be 0 because the subzone has t0 densities set, or null because it can't be calculated. This was done this way because we're doing a full outer join on the t0 table(s) so that if there are species that were not recorded in observations but that have densities set, then the survival rate should be 0 instead of null.

sgrimm (approved):

ice1080:
### Merge activity

* **Oct 15, 4:03 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3497).
* **Oct 15, 4:04 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3497).

PR #3510, opened 2025-10-15
SW-7417 Add temp plots to survival rate in funder reports

ice1080:


ice1080:
* **#3510** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3510?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3510?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1462,35 +1465,94 @@ class ReportStore(
...
+                      )
+                      .and(OBSERVATION_PLOTS.OBSERVATION_ID.`in`(observationsInReportPeriod))
+                      .orderBy(
+                          OBSERVATION_PLOTS.COMPLETED_TIME.desc(),
+                          OBSERVATION_PLOTS.OBSERVATION_ID.desc(),
+                      )
+                      .limit(1)
+                      .asTable("most_recent")
+              )
+              .where(DSL.field("most_recent.IS_PERMANENT", Boolean::class.java).eq(isPermanent))
```
sgrimm: I saw a similar query in an earlier PR and wondered if it could be made simpler with a window function. Couldn't get that to come out any simpler than this version, but it just occurred to me that maybe this can use something along the lines of
```
DSL.select(IS_PERMANENT.eq(isPermanent))
```
instead of the where/exists to at least get rid of one layer of query nesting. Maybe wrapped in a coalesce to deal with the case of no matching rows.

The query as it exists works fine and isn't hard to understand, but for some reason every time I see this, I get the urge to try to simplify it. Feel free to ignore this comment if you don't think it's worth fussing with.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
ice1080: Good thought. I plan on coming back and cleaning up several of these queries, and trying to make them reusable too. I'll have to do something different than the last PR though, because several of these need to be reused in multiple packages (e.g. accelerator/db and tracking/db)

ice1080:
### Merge activity

* **Oct 16, 6:19 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3510).

PR #3511, opened 2025-10-16
SW-7375 Add user names to report model

ice1080:
We want to be able to display "Terraformation Team" when the current user doesn't have access to view another user but the other user is part of TF. 
Add user names to the report model.


src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
ice1080, line 1: Trying to make this reusable so that more than just the reports payloads can use it

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
```diff
@@ -0,0 +1,32 @@
...
+        email: String,
+        userIsInSameOrg: Boolean,
+    ): SimpleUserModel {
+      return when {
+        (userIsInSameOrg || currentUser().canReadUser(userId)) ->
+            SimpleUserModel(userId, firstName, lastName)
+        (emailIsTf(email)) -> SimpleUserModel(userId, "Terraformation", "Team")
+        else ->
+            throw IllegalArgumentException("User is not in same organization and cannot be viewed")
+      }
```
ice1080, line 36: Could alternatively add two booleans for `userIsVisible` and `userIsTf` to the model so that the FE can display the name if the language is not in english (or could also translate server side)

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1334,6 +1339,47 @@ class ReportStore(
           )
           .convertFrom { results -> results.map { ReportProjectMetricModel.of(it) } }
 
+  private fun userIsInSameOrg() =
+      with(ORGANIZATION_USERS) {
+        DSL.field(
```
ice1080, line 1349: Couldn't initially think of a better way to do this field, but open to ideas as it doesn't seem very clean (e.g. `systemUser` can't enumerate `organizationRoles` so it needs a separate condition

src/main/kotlin/com/terraformation/backend/accelerator/api/ProjectReportsController.kt
```diff
@@ -472,6 +473,14 @@ data class NewAcceleratorReportConfigPayload(
       )
 }
 
+data class SimpleUserPayload(
+    val userId: UserId,
+    val firstName: String? = null,
+    val lastName: String? = null,
```
ice1080: Could also change to just `fullName` instead of separating

sgrimm: That seems good to me especially if we're looking to translate the system user's name.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.terraformation.backend.db.default_schema.UserId
+
+data class SimpleUserModel(
+    val userId: UserId,
+    val firstName: String? = null,
+    val lastName: String? = null,
+) {
+  companion object {
+    private fun emailIsTf(email: String): Boolean =
+        email.endsWith(suffix = "@terraformation.com", ignoreCase = true) ||
```
sgrimm, line 16: This makes me think we need a better way to identify internal users. Like, if we decided to start using `@terraware.io` email addresses or we had contractors logging in with their own email addresses to work on projects, this would break. We check for TF email addresses in a few places in the code already so adding one more isn't the end of the world, and changing it is outside the scope of this PR for sure.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
sgrimm: Server-side translation seems reasonable to me here.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
```diff
@@ -0,0 +1,32 @@
...
+        lastName: String,
+        email: String,
+        userIsInSameOrg: Boolean,
+    ): SimpleUserModel {
+      return when {
+        (userIsInSameOrg || currentUser().canReadUser(userId)) ->
+            SimpleUserModel(userId, firstName, lastName)
+        (emailIsTf(email)) -> SimpleUserModel(userId, "Terraformation", "Team")
+        else ->
+            throw IllegalArgumentException("User is not in same organization and cannot be viewed")
```
sgrimm: I guess we could also make this log an error and show a default name instead of bailing out.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
sgrimm: For the system user, this could always be `DSL.falseCondition()`, I think.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1334,6 +1339,47 @@ class ReportStore(
...
+            DSL.exists(
+                DSL.selectOne()
+                    .from(ORGANIZATION_USERS)
+                    .where(USER_ID.eq(USERS.ID))
+                    .and(
+                        ORGANIZATION_ID.`in`(
+                            if (currentUser().userId == systemUser.userId) emptyList()
```
sgrimm: `currentUser() is SystemUser` would be a more robust test.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
ice1080: I think that makes sense to me as well, especially since Alex just noted that we might need to support "Former User" when a user gets deleted.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
ice1080: I was definitely surprised to see that an email search was how we identified internal users.

ice1080:
* **#3511** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3511?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3511?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1334,6 +1344,53 @@ class ReportStore(
           )
           .convertFrom { results -> results.map { ReportProjectMetricModel.of(it) } }
 
+  private fun userIsInSameOrg() =
+      with(ORGANIZATION_USERS) {
+        DSL.field(
+            if (currentUser() is SystemUser) DSL.falseCondition()
```
sgrimm: Nit: Add curly braces to keep the two branches of the `if` consistently formatted.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1334,6 +1344,53 @@ class ReportStore(
...
+          )
+          .convertFrom { results ->
+            results
+                .map {
+                  SimpleUserModel.create(
+                      it[ID.asNonNullable()],
+                      "${it[FIRST_NAME.asNonNullable()]} ${it[LAST_NAME.asNonNullable()]}",
+                      it[EMAIL.asNonNullable()],
+                      it[sameOrgField],
+                      it[DELETED_TIME.isNotNull.asNonNullable()],
```
sgrimm: Might be simpler to select `DELETED_TIME` and change this to `it[DELETED_TIME] != null`.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1334,6 +1344,53 @@ class ReportStore(
...
+                  )
+                  .from(USERS)
+                  .where(ID.`in`(REPORTS.CREATED_BY, REPORTS.MODIFIED_BY, REPORTS.SUBMITTED_BY))
+          )
+          .convertFrom { results ->
+            results
+                .map {
+                  SimpleUserModel.create(
+                      it[ID.asNonNullable()],
+                      "${it[FIRST_NAME.asNonNullable()]} ${it[LAST_NAME.asNonNullable()]}",
```
sgrimm: I don't think you need the `asNonNullable`s here; they're really only used for type safety but the types are being discarded by string interpolation here anyway.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
```diff
@@ -0,0 +1,39 @@
...
+        userIsDeleted: Boolean,
+        messages: Messages,
+    ): SimpleUserModel {
+      return when {
+        (!userIsDeleted && (userIsInSameOrg || currentUser().canReadUser(userId))) ->
+            SimpleUserModel(userId, fullName)
+        (emailIsTf(email)) -> SimpleUserModel(userId, messages.terraformationTeam())
+        (userIsDeleted) -> SimpleUserModel(userId, messages.formerUser())
+        else -> {
+          log.warn("User {} cannot read name of user {}", currentUser().userId, userId)
```
sgrimm: Should this be an error so we'll get notified about it? I assume this will still be a sign of a bug if it occurs.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
```diff
@@ -0,0 +1,39 @@
...
+        email: String,
+        userIsInSameOrg: Boolean,
+        userIsDeleted: Boolean,
+        messages: Messages,
+    ): SimpleUserModel {
+      return when {
+        (!userIsDeleted && (userIsInSameOrg || currentUser().canReadUser(userId))) ->
+            SimpleUserModel(userId, fullName)
+        (emailIsTf(email)) -> SimpleUserModel(userId, messages.terraformationTeam())
+        (userIsDeleted) -> SimpleUserModel(userId, messages.formerUser())
```
sgrimm: Are the parentheses around the conditions here actually needed?

src/test/kotlin/com/terraformation/backend/customer/model/SimpleUserModelTest.kt
```diff
@@ -0,0 +1,151 @@
...
+      val userId = UserId(1000)
+
+      assertEquals(
+          SimpleUserModel(userId, "Former User"),
+          SimpleUserModel.create(
+              userId = userId,
+              fullName = "Some Name",
+              email = "test@other.com",
+              userIsInSameOrg = true,
+              userIsDeleted = true,
```
sgrimm: I don't think this would ever happen in the actual app unless there's a bug somewhere. When a user is deleted, they get removed from all their orgs.

sgrimm (changes requested):
This looks fine except for the handling of deleted users.

src/main/kotlin/com/terraformation/backend/customer/model/SimpleUserModel.kt
ice1080: Ah, didn't realize they weren't needed in kotlin

src/test/kotlin/com/terraformation/backend/customer/model/SimpleUserModelTest.kt
ice1080: I was more trying to showcase that `isDeleted` takes precedence over readability, but I'll switch to using `canReadUser`

sgrimm (approved):

PR #3530, opened 2025-10-21
SW-7569 Move geolocation, capture time to files table

sgrimm:
We're going to be pulling GPS coordinates from photo and video files in additional
places in the app soon. Currently we only extract them from files that are
attached to activity log entries, and we store them in `activity_media_files`.
We also store GPS locations for observation photos, though those GPS coordinates
are sent separately in the photo upload API request rather than extracted from the
photo files.

Move the geolocation and captured date columns to the files table so they'll be
stored consistently across the app. "Captured date" turns into "Captured local
time" because we may want the time of day rather than just the date; the word
"local" is to clarify that the timestamp is not necessarily in UTC. But in this
initial change, we still only extract the date part of the EXIF timestamp for
activity media files.

For observation photos, the GPS coordinates continue to come from the API request
payload for now, such that there's no client-visible behavior change.

### DIFF
--- a/src/main/resources/db/migration/0400/V419__FilesMetadata.sql
+++ b/src/main/resources/db/migration/0400/V419__FilesMetadata.sql
@@ -1,0 +1,19 @@
+ALTER TABLE files ADD COLUMN captured_local_time TIMESTAMP;
+ALTER TABLE files ADD COLUMN geolocation GEOMETRY(Point);
+
+UPDATE files
+SET geolocation = amf.geolocation,
+    captured_local_time = amf.captured_date
+FROM accelerator.activity_media_files amf
+WHERE amf.file_id = files.id;
+
+UPDATE files
+SET geolocation = op.gps_coordinates
+FROM tracking.observation_photos op
+WHERE op.file_id = files.id;
+
+ALTER TABLE accelerator.activity_media_files DROP COLUMN captured_date;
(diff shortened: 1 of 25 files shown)


sgrimm:
* **#3530** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3530?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3530?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0400/V419__FilesMetadata.sql
ice1080, line 1: Would there be any reason to make `captured_local_time` to be non-null? We could populate it with the `created_time` of the files row, and if we really wanted to, write a script that updates this based on the original files' captured times...

sgrimm: I thought about that, but I'm not sure what the concept of "captured time" would mean for files like Excel spreadsheets or Word documents that get uploaded as deliverable submissions.

sgrimm: I agree the nullability situation is less than ideal here.

ice1080 (approved):

PR #3531, opened 2025-10-21
SW-7570 Add endpoint for checking if all site t0 data has been set

ice1080:
To determine whether all t0 data has been set, the FE currently has to query 3 endpoints and combine all the data, and then the data is never used again on the page.

Add an endpoint to do this determination on the BE since this is used on pages that only need to show an indication for whether or not all t0 data has been set.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
@@ -9,8 +9,11 @@ import com.terraformation.backend.db.tracking.ObservationId
 import com.terraformation.backend.db.tracking.PlantingSiteId
 import com.terraformation.backend.db.tracking.PlantingZoneId
 import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOTS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PLOTS
 import com.terraformation.backend.db.tracking.tables.references.OBSERVED_PLOT_SPECIES_TOTALS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONES
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONE_POPULATIONS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_ZONE_T0_TEMP_DENSITIES
 import com.terraformation.backend.db.tracking.tables.references.PLOT_T0_DENSITIES
 import com.terraformation.backend.db.tracking.tables.references.PLOT_T0_OBSERVATIONS
@@ -28,9 +31,11 @@ import jakarta.inject.Named
 import java.math.BigDecimal
 import java.time.InstantSource
 import kotlin.collections.Map
+import org.jooq.Condition
 import org.jooq.DSLContext
 import org.jooq.Field
 import org.jooq.impl.DSL
+import org.jooq.impl.SQLDataType
 import org.springframework.context.ApplicationEventPublisher
 
 @Named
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3531** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3531?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3531?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -65,6 +70,96 @@ class T0Store(
...
+                )
+              }
+        }
+
+    return siteData.all { datum ->
+      datum.plotHasCompletedObservations &&
+          datum.subzoneWithdrawalDensity != null &&
+          ((datum.isPermanent && datum.plotDensity != null) ||
+              (!datum.isPermanent && datum.zoneDensity != null))
+    }
```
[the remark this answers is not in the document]
ice1080: If a site does not have any data withdrawn to it, then we likely don't want to say that the site doesn't have t0 data set for it. However I have a question out to product to confirm this.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -65,6 +70,96 @@ class T0Store(
     } ?: throw PlantingSiteNotFoundException(plantingSiteId)
   }
 
+  fun fetchAllT0SiteDataSet(plantingSiteId: PlantingSiteId): Boolean {
```
sgrimm, line 73: This seems like a lot of querying to do on every visit to a planting site page for a value that should almost never change over time (I'd expect generally it'll start off false, change to true when someone enters T0 data, then never go false again).

Would it make sense to keep this as a boolean column on `planting_sites` and update it whenever we do a write operation that could cause it to change?

sgrimm: It's possible I'm prematurely optimizing here.

src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
```diff
@@ -166,6 +186,91 @@ internal class T0StoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+    @Test
+    fun `correctly checks all data`() {
+      insertPlantingZoneT0TempDensity(speciesId = speciesId1)
+      insertPlantingZoneT0TempDensity(speciesId = speciesId2)
+      insertPlotT0Density(speciesId = speciesId1, monitoringPlotId = monitoringPlotId)
+      insertPlotT0Density(speciesId = speciesId2, monitoringPlotId = monitoringPlotId)
+      insertPlantingSubzonePopulation(speciesId = speciesId1)
+      insertPlantingSubzonePopulation(speciesId = speciesId2)
+
+      assertEquals(
```
sgrimm: Could use `assertTrue` and `assertFalse` in these tests.

sgrimm (approved):
No objections to this implementation if we do want to compute it on the fly each time.

src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
ice1080: Ah that's much better. For some reason IntelliJ really struggles to find imports for assertions every time. I was looking for that but didn't see it. I should have checked the project instead of relying on IJ

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: This currently is only being requested on the Observations page, but obviously that could change later. I'm curious your thoughts on a materialized view handling this. I know we've talked before about the downside of them not being able to be refreshed incrementally, but perhaps that's ok in this situation since there would only be one row per planting site? 
Alternatively we leave it as is like you're saying. In case helpful, I just did typed this query manually and did an explain plan and total cost for a single planting site ended up being `42.42 - 247.50`

ice1080: And regarding "never go false again", really the only things that would cause it to go false later on would be a change in permanent plots, either with a map change or a plot reassignment change. It also will happen a couple times at the very beginning until there have been observations over all zones/subzones in the site

sgrimm: I guess the best course here is to leave it as is for now and if it becomes an actual performance bottleneck, we can address it.

I think a materialized view would be less than ideal here because you'd end up having to refresh the whole thing (meaning recomputing the result for _all_ planting sites) every time a write happened that might cause the value to change for a single planting site. That'll get more and more expensive as the number of sites goes up over time. It wouldn't save you from having to figure out what writes should trigger a recalculation.

ice1080: A couple follow up questions more for my own understanding.
1. Would you consider that cost to be very expensive per planting site? `42.42 - 247.50` I recognize that it's still a waste since it's being calculated every page visit instead of just when things change, but just curious on the cost. 
2. I recognize that materialized views are most useful when some change(s) causes most/all rows to need to be recalculated, but I'm curious. Even though we'd be recalculating every row when only one is changing, wouldn't that still be cheaper than the current calculation being done on every page visit? (obviously the math would have to be done to determine at what # of planting sites and page visits make it "worth" it, but just asking your opinion)

sgrimm: First off, I wouldn't pay too much attention to the exact numbers of those cost figures. As I understand them, they're an abstraction Postgres's planner uses internally to decide between different possible execution plans. They're estimates based on table statistics, which can be incomplete or wrong. You'd want to run `EXPLAIN ANALYZE` rather than just `EXPLAIN` -- the `ANALYZE` will make it run the query for real and will tell you how long the various steps actually ended up taking.

One thing to be cautious of either way is that the query plan can change depending on the values you plug into the query. For example, if there happen to be table statistics that tell the query planner that 90% of the rows in a table have `planting_site_id = 2`, then it might not bother using an index if you happen to ask for that exact planting site since it's cheaper to scan the whole table. So if you're looking at query plans, you usually want to sample a few different data points, preferably ones that have different distributions of data.

As for whether the recalculation would be more expensive than computing on the fly, I think it's plausible that refreshing a view would be cheaper in terms of total resource consumption. But with on-the-fly calculation, the cost is spread out in little chunks across all the read operations, as opposed to being paid all at once on write operations. To pull numbers out of the air for illustration: it may be better for page loads to take an extra 50 milliseconds than for a "save t0 data" request to take 10 seconds, even if the 50-millisecond cost adds up to minutes or hours of total time across all the page views.

I guess I don't see what advantage a materialized view would bring over a boolean column on a regular table. At the point you know you need to refresh the view, you would know which planting site was triggering the refresh, since it'd always be due to a write to a particular site's data. Instead of refreshing the whole view, you could run this query and update the flag for just that one planting site. Or am I missing something?

ice1080: Not missing anything, I think that covers the points I was thinking of. In my head we would trigger the MV refresh asynchronously so that it doesn't change the request time for "save t0 data". 

The only reason I have preferred MVs over DB columns/tables in the past is essentially the same reasoning for using a view over a table - that there's low to no chance of it getting out of date with the data used to calculate it. In the case of an MV you just have to refresh it at the right times. When using a DB column, we'd have to calculate it when t0 data changes, but also when a planting site map changes, when plot reassignment happens, when withdrawals happen, and possibly others in the future. There's just a higher chance of forgetting one of them.

Now that I'm typing all of this though, I guess that all of those situations are still ones where you'd have to remember to refresh the MV, so there is just as much risk of human error in either situation. 

Ok I'm now convinced that for our app, most of the time a DB table or column is better than an MV. 😂 

I'll still leave things as is for now until we have some more data or a request to speed things up.

ice1080: Thanks for talking through all of this, very helpful to understand the various perspectives and challenges.

PR #3534, opened 2025-10-22
SW-7584 Add searchCount endpoint to search service

ice1080:
In order to support BE pagination, we'll need to have a total row count to know how many pages and rows there are. Rather than adding this to the regular search api (and thus probably over-query this when it's not needed), add a separate endpoint to support this. 

The next PR will limit what is queried.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.search
 
+import com.terraformation.backend.log.debugWithTiming
 import com.terraformation.backend.log.perClassLogger
 import com.terraformation.backend.search.field.SearchField
 import jakarta.inject.Named
@@ -180,5 +181,36 @@ class SearchService(private val dslContext: DSLContext) {
     // SearchField.computeValue() can introduce duplicates that the query's SELECT DISTINCT has no
     // way of filtering out.
     return searchResults.map { it?.get(fieldPathName)?.toString() }.distinct()
+  }
+
+  fun searchCount(
+      rootPrefix: SearchFieldPrefix,
+      fields: Collection<SearchFieldPath>,
+      criteria: Map<SearchFieldPrefix, SearchNode>,
+  ): Long {
+    val exactCriteria = criteria.mapValues { it.value.toExactSearch() }
+    if (exactCriteria != criteria) {
+      val exactResults = searchCount(rootPrefix, fields, exactCriteria)
+      if (exactResults > 0) {
+        return exactResults
+      }
+    }
+
+    return runQueryCount(rootPrefix, fields, criteria)
+  }
+
+  private fun runQueryCount(
+      rootPrefix: SearchFieldPrefix,
+      fields: Collection<SearchFieldPath>,
+      criteria: Map<SearchFieldPrefix, SearchNode>,
+  ): Long {
+    val queryBuilder = buildQuery(rootPrefix, fields, criteria, emptyList())
+    val query = queryBuilder.toSelectQuery()
+
+    val count =
+        log.debugWithTiming("Retrieved count for search query") {
+          dslContext.selectCount().from(query).fetchOne(0, Long::class.java) ?: 0L
+        }
+    return count
   }
 }
--- a/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/SearchService.kt
@@ -185,12 +185,11 @@ class SearchService(private val dslContext: DSLContext) {
 
   fun searchCount(
       rootPrefix: SearchFieldPrefix,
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3535** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3535?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3534** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceCountTest.kt
```diff
@@ -0,0 +1,115 @@
...
+            prefix.resolve("viabilityTest.id"),
+            listOf(testId1.toString(), testId2.toString(), testId3.toString()),
+        )
+
+    insertTestResults(testId1, 1000)
+    insertTestResults(testId2, 2000)
+    insertTestResults(testId3, 3000)
+    insertTestResults(testId4, 5) // excluded from criteria
+
+    assertEquals(6000, searchService.searchCount(prefix, fields, mapOf(prefix to criteria)))
```
sgrimm: In the interest of keeping the tests fast: What does testing with 6000 rows tell us that testing with, say, 60 (or even 6) rows doesn't? Not sure what failure mode this is guarding against.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -181,4 +182,35 @@ class SearchService(private val dslContext: DSLContext) {
...
+        return exactResults
+      }
+    }
+
+    return runQueryCount(rootPrefix, fields, criteria)
+  }
+
+  private fun runQueryCount(
+      rootPrefix: SearchFieldPrefix,
+      fields: Collection<SearchFieldPath>,
```
sgrimm: Do we even need a field list for this?

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceCountTest.kt
ice1080: Good point. I think I wanted to test more than a count of `4`, but this whole test is probably not needed. I did speed check it locally to make sure it ran in <2s, but again the test isn't really telling us anything.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
sgrimm: Commented on this at more length on the other PR.

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
```diff
@@ -181,4 +182,34 @@ class SearchService(private val dslContext: DSLContext) {
     // way of filtering out.
     return searchResults.map { it?.get(fieldPathName)?.toString() }.distinct()
   }
+
+  fun searchCount(
+      rootPrefix: SearchFieldPrefix,
+      fields: Collection<SearchFieldPath>,
```
sgrimm: Could remove the `fields` parameter.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/SearchService.kt
ice1080: Removed in this PR, will close the other PR if we're good with this approach. 
However, I'm not sure if this will work because I am no longer calling `joinFlattenedSublists` for two reasons:
1. It requires `SelectJoinStep<Record>` which doesn't work with `selectCount` which produces `SelectSeekStepN<Record1<Int>>`
1. I wasn't sure if it was needed anymore since we're no longer querying the fields from the flattened sublists.

The potential problem with this is that this could produce different counts because `joinFlattenedSublists` includes all of the `conditionForVisibility` additions. I think I probably need to spend some time trying to come up with a test that breaks what I have now, then write a separate (or update) `joinFlattenedSublists` that works for the count query. Thoughts @sgrimm ?

sgrimm: Agreed, though I wouldn't treat it as super high priority because I believe in practice we don't use flattened sublists in a way that would cause any behavior differences here. But it'd be better to either make it work properly or at least detect a potentially problematic query and throw an error.

ice1080: Ok, I'll work on that in a follow up PR!

ice1080:
### Merge activity

* **Oct 24, 12:42 AM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3534).

PR #3535, opened 2025-10-22
SW-7584 Exclude multisets to simplify search count queries

ice1080:
In order to simplify the queries used for getting search counts, exclude nested fields.

I wasn't sure if other things could be excluded, and/or if this is the best way to exclude multisets. Open to ideas here

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/NestedQueryBuilder.kt
@@ -620,9 +620,10 @@ class NestedQueryBuilder(
   fun addSelectFields(
       fields: Collection<SearchFieldPath>,
       criteria: Map<SearchFieldPrefix, SearchNode>,
+      excludeMultisets: Boolean = false,
   ) {
     assertNotRendered()
-    fields.forEach { addSelectField(it, criteria) }
+    fields.forEach { addSelectField(it, criteria, excludeMultisets) }
   }
 
   /**
@@ -777,6 +778,7 @@ class NestedQueryBuilder(
   private fun addSelectField(
       fieldPath: SearchFieldPath,
       criteria: Map<SearchFieldPrefix, SearchNode>,
+      excludeMultisets: Boolean = false,
   ) {
     val relativeField = fieldPath.relativeTo(prefix)
 
@@ -795,6 +797,9 @@ class NestedQueryBuilder(
         nextSelectFieldPosition += searchField.selectFields.size
       }
     } else if (relativeField.isNested) {
+      if (excludeMultisets) {
+        return
+      }
       // Prefix = a.b, fieldName = a.b.c.d => make a sublist "c" to hold field "d"
       val sublistName = getSublistName(relativeField)
       val sublistQuery =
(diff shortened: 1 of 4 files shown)


ice1080:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3535?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3535** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3535?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3534** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sgrimm (commented):
I commented on the other PR too, but I'm not sure we even need any fields at all for a count query, multiset or otherwise. `SELECT COUNT(*) FROM (SELECT a, b, c FROM x WHERE x.y = z)` should give the same result as `SELECT COUNT(*) FROM (SELECT 1 FROM x WHERE x.y = z)`.

And if we're fussing with the list of fields on the inner query anyway, I guess we could move the `COUNT` there, like `SELECT COUNT(*) FROM x WHERE x.y = z`.

ice1080:
> I commented on the other PR too, but I'm not sure we even need any fields at all for a count query, multiset or otherwise. `SELECT COUNT(*) FROM (SELECT a, b, c FROM x WHERE x.y = z)` should give the same result as `SELECT COUNT(*) FROM (SELECT 1 FROM x WHERE x.y = z)`.
> 
> And if we're fussing with the list of fields on the inner query anyway, I guess we could move the `COUNT` there, like `SELECT COUNT(*) FROM x WHERE x.y = z`.

I tried passing in `emptyList()` to `buildQuery` and it didn't work, though I can't recall exactly why now. I'll go back to that to learn more. It's possible moving the count to the inner query would fix the issue, but let me see what it was first.

sgrimm:
I was imprecise -- when I say "you don't need any fields at all" I mean you don't need any actual `SearchField`s. The `SELECT` clause in the generated SQL still needs at least one argument, but it can be a dummy value rather than a reference to a column (`SELECT 1` being the typical example).

ice1080:
> I was imprecise -- when I say "you don't need any fields at all" I mean you don't need any actual `SearchField`s. The `SELECT` clause in the generated SQL still needs at least one argument, but it can be a dummy value rather than a reference to a column (`SELECT 1` being the typical example).

Right, I assumed that as well. Thankfully I've spent enough time in the search api's code to be aware of how it builds the query. There's definitely a chance I didn't test that correctly though. Just about done with some other work and then will come back to this.

sgrimm:
This isn't needed any more after the change to the parent PR, I believe, but let me know if that's wrong and I should still look at it.

ice1080:
Nah I just hadn't yet posted my comment saying the same thing. You beat me to it

PR #3537, opened 2025-10-23
Render all MJML files in one Node command

sgrimm:
Update the `renderMjml` Gradle task to run a single command to render all the MJML
files rather than running a separate command for each file. The vast majority of
the execution time for that task is from firing up Node and loading all the MJML
modules, so this change makes the task take only slightly longer to render all the
files than it used to take to render each one.

The wrapper script we use to load our plugins is no longer a passthrough to the
MJML command-line interface; it's now specialized for rendering a batch of files.

Since we're rendering MJML using the MJML library's API rather than its 
command-line tool, we need to minify the generated HTML ourselves (minification 
is only supported by MJML's command-line tool). Configure the minifier with the
same options the MJML CLI uses, but add a custom option to ignore Freemarker
template tags like `<#if>`. With that change, we no longer need to wrap
Freemarker tags in `<!-- htmlmin:ignore -->` to avoid HTML parser errors.


sgrimm:
Looking at the test failure. Seems like it's due to a change in word wrapping behavior.

sgrimm:
Trying to get this working led me to discover that we have several MJML files for which the minifier wasn't working at all (even with the old MJML version) because they have `<#if>` and other FTL tags that aren't surrounded with `<!-- htmlmin:ignore -->` comments.

buildSrc/src/main/kotlin/com/terraformation/gradle/RenderMjmlTask.kt
```diff
@@ -62,37 +56,36 @@ constructor(
...
+      renderMjmlFiles(filesToRender)
     }
   }
 
-  /** Runs Yarn to format a single MJML file. */
-  private fun renderMjmlFile(mjmlFile: File, targetFile: File) {
-    Files.createDirectories(targetFile.toPath().parent)
-    val runner = objectFactory.newInstance(YarnExecRunner::class.java)
+  /** Runs Yarn to render one or more MJML files in a single Node process. */
+  private fun renderMjmlFiles(files: List<Pair<File, File>>) {
```
ice1080: Could be helpful to add to the doc comment what the files in the pair represent

ice1080 (approved):

buildSrc/src/main/kotlin/com/terraformation/gradle/RenderMjmlTask.kt
sgrimm: Actually, we don't even need the second file in the pair any more; that was left over from the way this used to work. Thanks.

PR #3538, opened 2025-10-23
SW-7586 Update Planting Site survival rates all-set check

ice1080:
Update survival rate settings all-set check with the following:
- Exclude ad-hoc plots
- Split up perm and temp checks and include check for survival-rates-includes-temp-plots
- Simplify temp data check to distinct per zone
- Refactor for readability

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
@@ -9,7 +9,6 @@ import com.terraformation.backend.db.tracking.ObservationId
 import com.terraformation.backend.db.tracking.PlantingSiteId
 import com.terraformation.backend.db.tracking.PlantingZoneId
 import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOTS
-import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PLOTS
 import com.terraformation.backend.db.tracking.tables.references.OBSERVED_PLOT_SPECIES_TOTALS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONES
@@ -31,11 +30,9 @@ import jakarta.inject.Named
 import java.math.BigDecimal
 import java.time.InstantSource
 import kotlin.collections.Map
-import org.jooq.Condition
 import org.jooq.DSLContext
 import org.jooq.Field
 import org.jooq.impl.DSL
-import org.jooq.impl.SQLDataType
 import org.springframework.context.ApplicationEventPublisher
 
 @Named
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3538** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -540,22 +457,168 @@ class T0Store(
...
-      isPermanent: Field<Boolean?>,
-  ): Condition =
-      DSL.exists(
-          DSL.selectOne()
-              .from(OBSERVATION_PLOTS)
-              .where(
-                  OBSERVATION_PLOTS.MONITORING_PLOT_ID.eq(monitoringPlotIdField)
-                      .and(OBSERVATION_PLOTS.IS_PERMANENT.eq(isPermanent))
-                      .and(OBSERVATION_PLOTS.COMPLETED_TIME.isNotNull)
+  private fun getAllPermPlotSpeciesT0Data(plantingSiteId: PlantingSiteId): Pair<Boolean, Boolean> {
```
sgrimm: I didn't expect a function with this name to return a pair of booleans, but rather a list of all the species T0 data from permanent plots. A comment explaining the return value would help.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -540,22 +457,168 @@ class T0Store(
...
+
+      val permPlotData =
+          dslContext
+              .select(
+                  ID,
+                  plantingSites.SURVIVAL_RATE_INCLUDES_TEMP_PLOTS,
+                  permSpeciesIdField,
+                  PLOT_T0_DENSITIES.PLOT_DENSITY,
+              )
+              .from(this)
```
sgrimm: I had to scroll up to see which table `this` was referring to, and the `plantingSites` implicit join made me think there was a local variable by that name somewhere. I think this would be more readable if each query had its own `with`.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -540,22 +457,168 @@ class T0Store(
...
+      val survivalRateIncludesTempPlots =
+          if (permPlotData.isEmpty()) {
+            dslContext
+                .select(PLANTING_SITES.SURVIVAL_RATE_INCLUDES_TEMP_PLOTS)
+                .from(PLANTING_SITES)
+                .where(PLANTING_SITES.ID.eq(plantingSiteId))
+                .fetchOne(PLANTING_SITES.SURVIVAL_RATE_INCLUDES_TEMP_PLOTS.asNonNullable())!!
+          } else {
+            permPlotData.first().survivalRateIncludesTempPlots
+          }
```
sgrimm: Disregard the following if this change is part one of a series and the things I'm talking about will make more sense in the context of later refactoring steps.

This feels to me like it could be simplified. The query above this is being used to answer two questions: is there any row with a null plot density, and do we need to fetch the survival-rate-includes-temp-plots flag separately?

For the first part of that, it seems like rather than constructing a list of objects, this could be a `SELECT EXISTS(SELECT 1 FROM ... WHERE plot_density IS NULL)`.

And for the second part, we can just fetch the flag from the planting site unconditionally; it's an extra database round-trip but the logic is simpler to follow than this "try to join if we can, but if that doesn't work, fetch it" logic.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -540,22 +457,168 @@ class T0Store(
...
+              .on(PLANTING_SUBZONE_POPULATIONS.PLANTING_SUBZONE_ID.eq(PLANTING_SUBZONES.ID))
+              .where(PLANTING_SITE_ID.eq(plantingSiteId))
+              .and(IS_AD_HOC.eq(false))
+              .and(PERMANENT_INDEX.isNotNull)
+              .fetch { record ->
+                T0SitePermPlotData(
+                    monitoringPlotId = record[ID.asNonNullable()],
+                    survivalRateIncludesTempPlots =
+                        record[plantingSites.SURVIVAL_RATE_INCLUDES_TEMP_PLOTS.asNonNullable()],
+                    speciesId = record[permSpeciesIdField],
```
sgrimm: The monitoring plot ID and species ID don't seem to be used for anything.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -540,22 +457,168 @@ class T0Store(
...
+              .and(PERMANENT_INDEX.isNull)
+              .fetch { record ->
+                T0SiteTempPlotData(
+                    plantingZoneId = record[plantingSubzones.PLANTING_ZONE_ID.asNonNullable()],
+                    speciesId = record[tempSpeciesIdField],
+                    zoneDensity = record[PLANTING_ZONE_T0_TEMP_DENSITIES.ZONE_DENSITY],
+                )
+              }
+
+      return tempPlotData.all { it.zoneDensity != null }
```
sgrimm: Similar to my comment above: the "is there any entry with null density?" logic could be baked into the SQL.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: Ah right, I meant to rename the function. I originally intended to return all the data list but then realized I could return just the pair. I'll add a comment but lmk if returning the full list would be easier to follow.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: Good point. I had been using them for debugging and they were thus required in my head, but we don't actually need them

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: Wrote that comment before seeing the next one, which simplifies the retrieval of `survivalRateIncludesTempPlots` and makes this one unnecessary

sgrimm (approved):
Nice, that's easier to follow.

PR #3567, opened 2025-10-31
SW-7565 Add State enum to ObservationsTable

ice1080:
The state enum of observations was missing from the `ObservationsTable`. Add this.

It seems like we don't have tests for every field in search tables, so no test was added/updated.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/ObservationsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/ObservationsTable.kt
@@ -47,6 +47,7 @@ class ObservationsTable(private val tables: SearchTables) : SearchTable() {
             PlantingSiteHistoryId(it)
           },
           dateField("startDate", OBSERVATIONS.START_DATE),
+          nonLocalizableEnumField("state", OBSERVATIONS.STATE_ID),
           enumField("type", OBSERVATIONS.OBSERVATION_TYPE_ID),
       )
 
--- a/src/main/resources/i18n/Messages_en.properties
+++ b/src/main/resources/i18n/Messages_en.properties
@@ -701,6 +701,7 @@ search.observations.endDate=Observation end date
 search.observations.id=Observation ID
 search.observations.isAdHoc=Observation is ad-hoc
 search.observations.plantingSiteHistoryId=Observation planting site history ID
+search.observations.state=Observation state
 search.observations.startDate=Observation start date
 search.observations.type=Observation type
 search.organizationInternalTags.name=Organization internal tag name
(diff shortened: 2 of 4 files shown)


ice1080:
* **#3567** <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3567?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terraware/terraware-server/3567?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/ObservationsTable.kt
```diff
@@ -47,6 +47,7 @@ class ObservationsTable(private val tables: SearchTables) : SearchTable() {
             PlantingSiteHistoryId(it)
           },
           dateField("startDate", OBSERVATIONS.START_DATE),
+          nonLocalizableEnumField("state", OBSERVATIONS.STATE_ID),
```
sgrimm, line 50: Do we need to make it localizable? Only matters if we're going to show it directly to an end user or include it in a CSV export.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/search/table/ObservationsTable.kt
ice1080: The `ObservationState`​ enum doesn't extend `LocalizableEnum`​ which is why I left it as `nonLocalizableEnum`​. If you would prefer to change that localizable, I can update `Config.kt`​. I was assuming that was the source of truth here

ice1080: @sgrimm looks like you made it non-localizable explicitly in Nov 2023, when adding support for non-localizable enums.

https://github.com/terraware/terraware-server/commit/c3c49c00de3d418ddc5d0373faf1af5dfb6cd0ad

sgrimm: Kind of a chicken-and-egg thing, I guess. I'm sure I made it non-localizable because I wasn't including it in the search API, and the search API is the only place it matters.

sgrimm: If we're not doing CSV export that includes this column, then leaving it non-localizable is fine.

ice1080: Sounds good. For now this is just being added because it needs to be used for filtering out permanent plots that are not in completed or abandoned observations, so I'll wait till it's needed. Thanks!

ice1080:
### Merge activity

* **Oct 31, 10:38 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terraware/terraware-server/3567).

PR #3582, opened 2025-11-05
SW-7607 Reduce boilerplate for polymorphic payloads

sgrimm:
If a payload class includes a property whose type is an interface, the serialized
JSON includes a "discriminator" property to identify which concrete type the
property's value is. The mapping between discriminator values and subclasses is
in annotations that are read by Jackson.

But Springdoc-OpenAPI doesn't automatically reflect the Jackson annotations in
the OpenAPI schema, so we also had to include a second version of the exact same
mappings in the `@Schema` annotation.

We're about to start adding even more polymorphic payloads for different kinds
of events from the event log, so the amount of this boilerplate would have kept
growing over time.

Add logic to auto-populate the schema in cases where the superclass is a sealed
class or interface, and remove the now-redundant discriminator mappings from
the paylaod class schemas.


sgrimm:
* **#3582** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3590, opened 2025-11-07
SW-7684 Add endpoint for species required for t0 data

ice1080:
The previous `getSpeciesWithdrawnToPlantingSite` endpoint only included species that were withdrawn to plots and their existing densities (for pre-population in the FE). We want to also know any species that were recorded in observations but not withdrawn to the plot's subzone.
Create a new endpoint that combines both lists of species. Use the same basic format for the return but make the density optional, such that null densities should be treated as species still required for t0 data, even though they don't have withdrawal densities to pre-populate the FE with.

The existing endpoint `getSpeciesWithdrawnToPlantingSite` will get removed in https://terraformation.atlassian.net/browse/SW-7688.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
@@ -52,7 +52,14 @@ class T0Controller(
   }
 
   @GetMapping("/plantingSite/{plantingSiteId}/species")
-  @Operation(summary = "Lists all the species that have been withdrawn to a planting site.")
+  @Operation(
+      summary =
+          "Lists all the species that have been withdrawn to a planting site or recorded in " +
+              "observations (if not withdrawn).",
+      description =
+          "Species with densities are species that were withdrawn, species with null densities are " +
+              "species that were recorded in observations but not withdrawn to the plot's subzone.",
+  )
   fun getSpeciesWithdrawnToPlantingSite(
       @PathVariable("plantingSiteId") plantingSiteId: PlantingSiteId
   ): GetSitePlotSpeciesResponsePayload {
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
@@ -60,7 +60,7 @@ class T0Controller(
           "Species with densities are species that were withdrawn, species with null densities are " +
               "species that were recorded in observations but not withdrawn to the plot's subzone.",
   )
-  fun getSpeciesWithdrawnToPlantingSite(
+  fun getT0SpeciesForPlantingSite(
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3590** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
```diff
@@ -49,6 +51,23 @@ class T0Controller(
...
+  @GetMapping("/plantingSite/{plantingSiteId}/species")
+  @Operation(
+      summary =
+          "Lists all the species that have been withdrawn to a planting site or recorded in " +
+              "observations (if not withdrawn).",
+      description =
+          "Species with densities are species that were withdrawn, species with null densities are " +
+              "species that were recorded in observations but not withdrawn to the plot's subzone.",
+  )
+  fun getSpeciesWithdrawnToPlantingSite(
```
sgrimm: This method name seems inaccurate (and also, it'll collide with the existing endpoint's method name since OpenAPI endpoint names are a flat namespace; the OpenAPI schema generator will append `_1` to one of them.)

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -77,6 +80,92 @@ class T0Store(
...
+            .join(PLANTING_SUBZONES)
+            .on(PLANTING_SUBZONE_POPULATIONS.PLANTING_SUBZONE_ID.eq(PLANTING_SUBZONES.ID))
+            .where(PLANTING_SUBZONES.PLANTING_SITE_ID.eq(plantingSiteId))
+            .and(densityField.ge(BigDecimal.valueOf(0.05)))
+
+    val observedNotWithdrawnSpecies =
+        DSL.select(
+                MONITORING_PLOTS.ID,
+                OBSERVED_PLOT_SPECIES_TOTALS.SPECIES_ID.`as`("species_id"),
+                DSL.inline(null, BigDecimal::class.java).`as`("density"),
```
sgrimm: Can this use `DSL.castNull()` instead of `DSL.inline()`?

sgrimm (approved):
I think Graphite's comment is correct, but I'm not sure if what it's complaining about is actually the desired behavior; if so, then this looks reasonable to me (with a couple small tweaks).

ice1080:
### Merge activity

* **Nov 7, 10:09 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3590).

PR #3595, opened 2025-11-08
SW-7607 Add event log query endpoint

sgrimm:
`POST /api/v1/events/list` will search the event log for events related to the IDs
specified in the request body, optionally limited to specific kinds of entities.

In this initial change, only observation media events are supported; others will
be added later. Example payload to query the media events for a particular
observation of a particular monitoring plot:

    {
        "organizationId": 1,
        "observationId": 2,
        "monitoringPlotId": 3,
        "subjects": ["ObservationPlotMedia"]
    }

Events are represented using "subjects" and "actions," where the list of actions
is currently limited to create, delete, and field-update. These always use the
same payload format no matter what the event is. The subject's structure varies
depending on the type of entity, but always includes long and short text
descriptions which UI code can use to display line items for events it doesn't
recognize or that don't need custom rendering logic.

In this initial change, the field names in the field-update action aren't
localized; that'll come later. (I say "field names" but there's only one field
so far: the "caption" field on observation photos/videos.)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformer.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformer.kt
@@ -1,0 +1,101 @@
+package com.terraformation.backend.eventlog
+
+import com.terraformation.backend.customer.db.SimpleUserStore
+import com.terraformation.backend.eventlog.api.CreatedActionPayload
+import com.terraformation.backend.eventlog.api.DeletedActionPayload
+import com.terraformation.backend.eventlog.api.EventActionPayload
+import com.terraformation.backend.eventlog.api.EventLogEntryPayload
+import com.terraformation.backend.eventlog.api.EventSubjectName
+import com.terraformation.backend.eventlog.api.EventSubjectPayload
+import com.terraformation.backend.eventlog.api.FieldUpdatedActionPayload
+import com.terraformation.backend.eventlog.api.ObservationPlotMediaSubjectPayload
+import com.terraformation.backend.eventlog.db.EventLogStore
+import com.terraformation.backend.eventlog.model.EventLogEntry
+import com.terraformation.backend.i18n.Messages
+import com.terraformation.backend.log.perClassLogger
+import com.terraformation.backend.tracking.event.ObservationMediaFileEditedEvent
+import com.terraformation.backend.tracking.event.ObservationMediaFilePersistentEvent
+import jakarta.inject.Named
+
+@Named
+class EventLogPayloadTransformer(
+    private val messages: Messages,
+    private val simpleUserStore: SimpleUserStore,
+) {
+  private val log = perClassLogger()
+
+  /**
+   * Returns a list of [EventLogEntryPayload] objects representing a sequence of events. This is not
+   * necessarily a 1:1 mapping; a single event can map to multiple payloads, e.g., if the event
+   * describes an update of multiple fields.
+   *
+   * The logic that produces a payload for a given event can refer to earlier events in the list,
+   * e.g., to pull an entity's name from its "created" event. Therefore, it's important that
+   * [entries] include all the events for each entity whose events are included. This should happen
+   * automatically if the list of events is queried using the event interfaces from
+   * [EventSubjectName].
+   *
+   * The list of entries must be in chronological order; the query methods in [EventLogStore] return
+   * chronological results, so this usually doesn't require any additional work.
+   */
+  fun eventsToPayloads(entries: List<EventLogEntry<PersistentEvent>>): List<EventLogEntryPayload> {
+    val context = EventLogPayloadContext(entries, messages)
+    val users = simpleUserStore.fetchSimpleUsersById(entries.map { it.createdBy }.distinct())
+
+    return entries.flatMap { entry ->
+      val actions = getActions(entry.event)
+      val subject = getSubject(entry.event, context)
+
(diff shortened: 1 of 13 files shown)


sgrimm:
* **#3609** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3609?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3605** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3604** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3604?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3596** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3595** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/eventlog/api/EventActionPayload.kt
```diff
@@ -0,0 +1,28 @@
...
+@JsonTypeName("Created")
+class CreatedActionPayload : EventActionPayload {
+  override fun equals(other: Any?) = other is CreatedActionPayload
+
+  override fun hashCode() = javaClass.hashCode() xor 1
+}
+
+@JsonTypeName("Deleted")
+class DeletedActionPayload : EventActionPayload {
+  override fun equals(other: Any?) = other is DeletedActionPayload
```
ice1080, line 18: I assume "soft deletes" will be considered Field Updated actions? If not then other is DeletedActionPayload might not be correct for `equals`

src/main/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformer.kt
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+  }
+
+  private fun getActions(event: PersistentEvent): List<EventActionPayload> {
+    return when (event) {
+      is ObservationMediaFileEditedEvent ->
+          listOf(
+              FieldUpdatedActionPayload(
+                  "caption", // TODO: i18n
```
ice1080, line 84: Will this TODO be handled in a future PR?

ice1080: I saw these updated in a later PR

src/main/kotlin/com/terraformation/backend/eventlog/api/EventActionPayload.kt
sgrimm: Soft deletes would still be considered Deleted actions.

The actions are meant to represent application-level operations, rather than SQL-level ones. The fact that something was soft-deleted is more of an implementation detail of the "delete" operation than a fundamentally different category of operation.

sgrimm: Maybe another way to think about it: if a user is looking at history in the UI and it includes a soft delete, they'd expect to see "Item X deleted," not, "Item X 'is deleted' flag changed from false to true."

ice1080: That makes sense, I think I was mainly asking about the `equals` method, but perhaps my question under the question is if it's possible to soft delete an entity and then "un-delete" it, for any of our entities. If so, then I think `other is DeletedActionPayload` might not be sufficient for `equals`

ice1080 (approved):
Approving for now since this PR is blocking the stack, though my question is still open about the soft delete being reversible

src/main/kotlin/com/terraformation/backend/eventlog/api/EventActionPayload.kt
sgrimm: But they would still be equal -- a "deleted" payload doesn't have any properties that could differ from one instance to another.

This is a check for value equality, not object identity.

sgrimm: The serialized JSON is always
```json
{
  "type": "Deleted"
}
```

sgrimm: If an entity is undeleted, that'd be reflected in the subject (its `deleted` property wouldn't be set to `true`) but you'd still want to see the deletion action in the event log along with an entry for the subsequent undeletion, which we'd probably either represent as a "field updated" event or we'd have to add a new action type for.

ice1080: Ya that makes sense, I'm following now!

sgrimm:
### Merge activity

* **Nov 17, 7:34 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3595).
* **Nov 17, 7:34 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3595).

PR #3620, opened 2025-11-17
SW-7694 All Set check should check all definitions of plot permanence

ice1080:
When checking if all permanent plots have t0 data set, change the definition to include all of the following (previously was just numbers 1 & 2):

1. currently permanent
1. have a completed or abandoned observation
1. within that observation the plot was also permanent

This helps ensure that if a plot was previously temporary, had an observation, then was changed to permanent via a site geometry change, it will not be included incorrectly in the permanent plot list until it's had a completed observation.

Do the same thing for temporary plots in reverse.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
@@ -375,6 +375,52 @@ internal class T0StoreTest : DatabaseTest(), RunsAsDatabaseUser {
           "All site data set",
       )
     }
+
+    @Test
+    fun `geometry changes plot temporary to permanent`() {
+      insertPlotT0Density(speciesId = speciesId1, monitoringPlotId = monitoringPlotId)
+
+      // plots that are permanent now and have completed observations but were temporary during the
+      // observations are excluded
+      insertMonitoringPlot(plotNumber = 103, permanentIndex = 103)
+      insertObservationPlot(
+          claimedTime = clock.instant(),
+          claimedBy = user.userId,
+          completedTime = clock.instant(),
+          completedBy = user.userId,
+          isPermanent = false,
+      )
+
+      assertTrue(
+          store.fetchAllT0SiteDataSet(plantingSiteId),
+          "Ignore observed plots that weren't permanent at observation",
+      )
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3620** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3620?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3620?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
```diff
@@ -375,6 +375,52 @@ internal class T0StoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+    }
+
+    @Test
+    fun `geometry changes plot permanent to temporary`() {
+      includeTempPlotsInSurvivalRates(plantingSiteId)
+      insertPlotT0Density(speciesId = speciesId1, monitoringPlotId = monitoringPlotId)
+      insertPlantingZoneT0TempDensity(speciesId = speciesId1)
+
+      // plots that are temporary now and have completed observations but were permanent during the
+      // observations are excluded
```
sgrimm, line 407: The temporary-to-permanent case makes sense to me, but I'm not sure I follow why we'd want to exclude plots in this case. Doesn't setting `includeTempPlotsInSurvivalRates` mean that we want to include both permanent and temporary plots, that is, we want to ignore whether or not a plot is permanent?

ice1080: Permanent and temporary plots have different types of checks, so the checks in T0Store are separated by permanence type (temporary plots are checked per zone instead of per plot)

sgrimm (approved):

src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
ice1080: This test is really just proving that out; I recognize what you're saying though. That plot would still be included in the check for permanent plots.

PR #3625, opened 2025-11-17
SW-7650 Fix division by zero error when calculating survival rate

ice1080:
If a t0 density is set to 0 for a species, then some of the survival rate calculations will fail with a `Division by 0` error. Fix this by setting survival rate to 0 in these scenario.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -1653,17 +1653,9 @@ class ObservationStore(
               )
           val survivalRate =
               if (liveAndDeadForZone.first().survivalRateIncludesTempPlots) {
-                DSL.if_(
-                    survivalRateDenominator.eq(BigDecimal.ZERO),
-                    DSL.zero(),
-                    DSL.value(totalLive).mul(100).div(survivalRateDenominator),
-                )
+                getSurvivalRate(DSL.value(totalLive), survivalRateDenominator)
               } else {
-                DSL.if_(
-                    survivalRateDenominator.eq(BigDecimal.ZERO),
-                    DSL.zero(),
-                    DSL.value(totalPermanentLive).mul(100).div(survivalRateDenominator),
-                )
+                getSurvivalRate(DSL.value(totalPermanentLive), survivalRateDenominator)
               }
 
           val rowsInserted =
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3625** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1653,9 +1653,17 @@ class ObservationStore(
               )
           val survivalRate =
               if (liveAndDeadForZone.first().survivalRateIncludesTempPlots) {
-                DSL.value(totalLive).mul(100).div(survivalRateDenominator)
+                DSL.if_(
+                    survivalRateDenominator.eq(BigDecimal.ZERO),
+                    DSL.zero(),
+                    DSL.value(totalLive).mul(100).div(survivalRateDenominator),
+                )
```
sgrimm: Worth adding a helper function for this?

sgrimm: Also, and this is definitely a matter of opinion so feel free to ignore, but I think putting the valid case before the error case is slightly clearer in if/else constructs like this. That is,
```kotlin
DSL.if_(
    survivalRateDenominator.ne(BigDecimal.ZERO),
    DSL.value(totalLive).mul(100).div(survivalRateDenominator),
    DSL.zero(),
)
```

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
ice1080: Interesting. I generally follow Robert Martin's guideline which is not using a negative case first in an if/else, because it can potentially be more confusing, but I feel like this situation is readable either way

ice1080: Oh interesting. This version doesn't work. Took me a second to understand, but it's because of how null propagates. Tests are failing expecting null and instead have 0.

PR #3639, opened 2025-11-19
Add test for localizable update event field names

sgrimm:
"Fields updated" action payloads return localized field names, which means the
field names need to be in the strings file.

For classes that implement `FieldsUpdatedPersistentEvent` and follow the typical
pattern of using `changedFrom` and `changedTo` properties with nested classes
for the field lists, we can sanity check that there are strings for all the
properties. This won't be foolproof but should catch cases where someone follows
the typical pattern but forgets to add strings to `Messages_en.properties`.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/eventlog/PersistentEventTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/eventlog/PersistentEventTest.kt
@@ -1,8 +1,12 @@
 package com.terraformation.backend.eventlog
 
 import com.fasterxml.jackson.annotation.JsonValue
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#3639** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3639?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3639?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

PR #3645, opened 2025-11-21
SW-7695 Support updating biomass details

sgrimm:
Add an `ObservationStore` method to update biomass-specific details for
observation plots, along with a persistent event to record the updates.

Initially, it only supports updating qualitative data (plot description
and soil assessment).

PR #3643, opened 2025-11-21
SW-7695 Add BiomassDetailsCreatedEvent

sgrimm:
Add an event to track the creation of new biomass observation details. Backfill
events for existing biomass observations. This will be used to support tracking
changes to biomass details once an update API is introduced for them.

PR #3646, opened 2025-11-21
Support distinct null and absent payload properties

sgrimm:
To support PATCH-style API semantics, we need to be able to distinguish between
a payload property being absent from the incoming JSON (in which case the existing
value should be left as is) and the property being present with a value of `null`
(in which case the client is asking us to remove the existing value).

Jackson supports using the Java `Optional` class to distinguish those two cases,
but it's a little cumbersome out of the box; add an extension method to simplify
using it for PATCH-style operations.

The distinction between "not required" and "can be null" needs to be reflected
in the OpenAPI schema as well. We could do it by adding `@Schema(nullable = true)`
to payload properties, but that'll add a lot of noise, so instead add logic to
automatically mark all `Optional` payload properties as allowing null values.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
+++ b/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
@@ -257,9 +257,6 @@ fun Geometry.differenceNullable(other: Geometry?): Geometry =
  * example, say an entity has a property `notes` of type `String?`. We'd want to handle a `PATCH`
  * request as follows:
  *
- * In the context of a PATCH endpoint where a property represents a possible change to an existing
- * value, we'd want to handle it as follows, respectively:
- *
  * | JSON               | Desired end result      |
  * |--------------------|-------------------------|
  * | `{}`               | Keep the original notes |
--- a/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
+++ b/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
@@ -13,6 +13,7 @@ import java.time.LocalDate
 import java.time.LocalTime
 import java.time.ZoneId
 import java.time.ZonedDateTime
+import java.util.Optional
 import org.geotools.api.referencing.FactoryException
 import org.geotools.api.referencing.crs.CoordinateReferenceSystem
 import org.geotools.geometry.jts.JTS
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#3647** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3647?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3646** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3645** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3645?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3644** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3644?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3643** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3643?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/util/Extensions.kt
```diff
@@ -248,3 +249,42 @@ fun Geometry.nearlyCoveredBy(other: Geometry, minCoveragePercent: Double = 99.99
...
+
+/**
+ * Applies this `Optional` as a replacement for an existing value.
+ *
+ * This is primarily used with payloads for `PATCH` endpoints that can update nullable values. For
+ * example, say an entity has a property `notes` of type `String?`. We'd want to handle a `PATCH`
+ * request as follows:
+ *
+ * In the context of a PATCH endpoint where a property represents a possible change to an existing
+ * value, we'd want to handle it as follows, respectively:
```
sgrimm: Oops, forgot to remove the initial draft of this text.

sgrimm:
### Merge activity

* **Nov 21, 8:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3646).
* **Nov 21, 8:44 PM UTC**: [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3646) rebased this pull request as part of a merge.
* **Nov 21, 8:53 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3646).

PR #3660, opened 2025-11-24
Removed oneOf annotations for Search Payloads

tommylau523:
This will results in correct OpenAPI doc. This will also allow for correct type generation in the FE without circular references. 

The `openapi-ts` package used in the FE still has a bug that requires manual overwrite. That script will be updated at the same time as this PR to ensure correctness.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/api/SearchPayloads.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/api/SearchPayloads.kt
@@ -85,13 +85,6 @@ data class SearchSortOrderElement(
         "A search criterion. The search will return results that match this criterion. The " +
             "criterion can be composed of other search criteria to form arbitrary Boolean " +
             "search expressions. TYPESCRIPT-OVERRIDE-TYPE-WITH-ANY",
-    oneOf =
-        [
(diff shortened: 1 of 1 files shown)


tommylau523:
* **#3660** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3660?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3660?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm (approved):

sgrimm:
I believe the mobile app uses the search API too.

tommylau523:
> I believe the mobile app uses the search API too.

I am making a change in the FE typegen script to make sure it doesn't change what existing. It is functional with some slight script changes but our FE code uses the overridden types so it will require some work there as well.

sgrimm:
Is this PR still relevant?

tommylau523:
I can close it for now. this change will need to happen when we stop using generate-types in the FE and the mobile app. So it will be a little while.

PR #3661, opened 2025-11-24
SW-7725 Utilize correct t0 when a plot changes from perm to temp or vice versa

ice1080:
When a plot changes from temp to permanent or permanent to temp, the observation results will use the incorrect t0 densities if there was an observation on the plot before the change. 

Additionally, the survival rate can be incorrect on a later observation, as calculated in the `completeObservation` method.

Fix these.

Also fix when a plot is removed from a subzone.


ice1080:
* **#3661** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3661?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3661?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/Queries.kt
```diff
@@ -26,9 +33,98 @@ fun latestObservationForSubzoneField(observationIdField: Field<ObservationId?>)
               // Prefer results from the same observation as the `observationId` if it exists.
               // True is considered greater than false, so sorting by an "=" expression in
               // descending order means the matches come first.
-              OBSERVATION_ID.eq(observationIdField).desc(),
+              OBSERVATIONS.ID.eq(observationIdField).desc(),
```
ice1080, line 42: This is a fix to a bug that showed up when I started adding the `observation_requested_subzones`​ to the tests

src/main/kotlin/com/terraformation/backend/tracking/db/Queries.kt
```diff
@@ -26,9 +33,98 @@ fun latestObservationForSubzoneField(observationIdField: Field<ObservationId?>)
...
               OBSERVATIONS.COMPLETED_TIME.desc(),
           )
           .limit(1)
     }
+
+fun plotIsInObservationResult(
+    monitoringPlotIdField: Field<MonitoringPlotId?>,
+    observationIdField: Field<ObservationId?>,
+    isPermanent: Boolean,
+): Condition {
```
ice1080, line 62: If preferred I can work on simplifying this to a single query (I recognize that it's technically a single query now, just split into variables). Many of the subqueries are only used once. I had them split out as I was writing them because I was trying to grok the query while writing it

sgrimm: I think keeping it split up will make it easier to understand in the future.

sgrimm: But maybe a KDoc comment describing what the goal of the function is would be helpful to future maintainers.

sgrimm (approved):

sgrimm (approved):

ice1080:
### Merge activity

* **Dec 3, 7:23 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3661).

PR #3668, opened 2025-11-25
Move withLockedObservation to separate class

sgrimm:
In preparation for moving some biomass-related update functions to `BiomassStore`,
split the observation locking function out into its own class that can be a
dependency of both `BiomassStore` and `ObservationStore`. Otherwise we would have
to make `BiomassStore` depend on `ObservationStore` which we try to avoid.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
@@ -67,6 +67,7 @@ import com.terraformation.backend.tracking.db.InvalidObservationEndDateException
 import com.terraformation.backend.tracking.db.InvalidObservationStartDateException
 import com.terraformation.backend.tracking.db.ObservationAlreadyStartedException
 import com.terraformation.backend.tracking.db.ObservationHasNoSubzonesException
+import com.terraformation.backend.tracking.db.ObservationLocker
 import com.terraformation.backend.tracking.db.ObservationNotFoundException
 import com.terraformation.backend.tracking.db.ObservationPlotNotFoundException
 import com.terraformation.backend.tracking.db.ObservationRescheduleStateException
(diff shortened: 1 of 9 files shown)


sgrimm:
* **#3669** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3669?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3668** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3667** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3667?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3666** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
```diff
@@ -136,19 +137,21 @@ class ObservationServiceTest : DatabaseTest(), RunsAsDatabaseUser {
   private val eventPublisher = TestEventPublisher()
   private val fileStore = InMemoryFileStore()
   private val muxService: MuxService = mockk()
+  private val observationLocker: ObservationLocker by lazy { ObservationLocker(dslContext) }
```
tommylau523, line 140: Should we move this to `DatabaseTest`? I am just wondering if it may be used more in the future too with writing observation results and other operations.

tommylau523 (approved):

src/test/kotlin/com/terraformation/backend/tracking/ObservationServiceTest.kt
sgrimm: We could, but I think at that point we'd probably want to think about whether we want to do the same thing for other commonly-used application classes like `ParentStore`. The approach I've been following so far is to keep `DatabaseBackedTest` free of any Spring beans except the jOOQ-generated DAOs (and even those might make sense to get rid of since the list of them is getting quite long). Otherwise we end up having a bunch of irrelevant application code included in all the database tests, e.g., `ObservationLocker` would be exposed to the document producer tests.

sgrimm:
### Merge activity

* **Nov 25, 7:22 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3668).
* **Nov 25, 7:33 PM UTC**: [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3668) rebased this pull request as part of a merge.
* **Nov 25, 7:43 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3668).

PR #3704, opened 2025-12-08
SW-7810 Fix survival rates when unrecorded species in later observations

ice1080:
When a species (with a t0 density) is recorded in an observation, and then not recorded in a later observation but other species in the plot are, the survival rate is incorrect (specifically, that species' t0 density is not being included in the survival rate denominator). Fix the full outer join on `plot_t0_densities` to correctly include all relevant rows.


ice1080:
* **#3704** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
ice1080, line 1: Note that this clause is still required in all of these multisets, outside of the `join` clauses:

```
.and(OBSERVATION_ID.eq(OBSERVATIONS.ID).or(OBSERVATION_ID.isNull))
```

sgrimm (approved):
Thanks!

PR #3723, opened 2025-12-12
[WIP] Observation scenario DSL

sgrimm:
Initial proof-of-concept implementation of a DSL for verifying observation
scenarios, along with a working example.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/db/ExampleScenarioTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/db/ExampleScenarioTest.kt
@@ -1,0 +1,167 @@
+package com.terraformation.backend.tracking.db
+
+import com.terraformation.backend.RunsAsDatabaseUser
+import com.terraformation.backend.TestClock
+import com.terraformation.backend.TestEventPublisher
+import com.terraformation.backend.customer.db.ParentStore
+import com.terraformation.backend.customer.model.TerrawareUser
+import com.terraformation.backend.db.DatabaseTest
+import com.terraformation.backend.db.IdentifierGenerator
+import com.terraformation.backend.db.default_schema.Role
+import com.terraformation.backend.gis.CountryDetector
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#3723** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3723?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/db/ExampleScenarioTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        }
+        // Default plot type
+        plot(311) {
+          species(0, live = 6)
+          species(1, live = 7)
+        }
+      }
+
+      expect(observation = 1) {
+        survivalRate(12)
```
ice1080, line 129: Is this inside the planting site, or does it just assume the most recent planting site?
It would be nice to be able to handle multiple planting sites. That's what the `ReportStore` test does, and it's 5000+ lines long because of it 😬

ice1080 (commented):
Disregard if this is already handled, but here's a few situations that you might need to handle:
1. Plot changes from permanent to temporary or vice versa
2. Plot was in substratum and is no longer
3. Plot changes substratum

src/test/kotlin/com/terraformation/backend/tracking/db/ExampleScenarioTest.kt
sgrimm: I think I can can make it support multiple planting sites without too much grief. Didn't think about that because all the observation-results tests operate on one planting site (but with multiple observations).

ice1080: As I think about it more, I'm unsure if this would even work for that test, so I would say don't even worry about it for now; let's wait till we need it, if that ever happens

sgrimm:
Closing this draft to reduce PR clutter while I work on getting it cleaned up and ready for real review.

PR #3729, opened 2025-12-15
SW-7843 Include is-original flag in photo subjects

sgrimm:
The event log subject for observation media files needs to include the flag that
indicates whether or not the file was uploaded as part of the original observation
data; that way the client can differentiate or exclude those files from the
observation history in the UI.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/eventlog/api/EventSubjectPayload.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/api/EventSubjectPayload.kt
@@ -221,6 +221,12 @@ data class ObservationPlotMediaSubjectPayload(
     override val deleted: Boolean?,
     val fileId: FileId,
     override val fullText: String,
+    @Schema(
+        description =
+            "True if this file was uploaded as part of the original submission of observation " +
+                "data; false if it was uploaded later."
+    )
+    val isOriginal: Boolean,
     val mediaKind: MediaKind,
     val monitoringPlotId: MonitoringPlotId,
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#3729** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3729?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3729?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/eventlog/api/EventSubjectPayload.kt
```diff
@@ -221,6 +221,12 @@ data class ObservationPlotMediaSubjectPayload(
     override val deleted: Boolean?,
     val fileId: FileId,
     override val fullText: String,
+    @Schema(
+        description =
+            "True if this file was uploaded as part of the original submission of observation " +
+                "data; false if it was uploaded later."
+    )
+    val isOriginal: Boolean,
```
ice1080, line 229: Will the lack of a default value cause any issues, or do all values already have this flag set?

sgrimm: It's been in the event objects from the start, so there shouldn't be an issue. (This is a response payload, not a request payload.)

ice1080 (approved):

PR #3732, opened 2025-12-15
SW-7856 Update to new logic for project level survival rates in reports

ice1080:
The logic for handling multiple planting sites for a survival rate had not yet been implemented for reports. Implement this using the `ObservationResultsStore` in order to decrease duplication and because the logic for checking multiple planting sites' survival rates is even more complex than what we have for reports currently.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
@@ -62,25 +62,16 @@ import com.terraformation.backend.db.nursery.tables.references.BATCHES
 import com.terraformation.backend.db.nursery.tables.references.BATCH_WITHDRAWALS
 import com.terraformation.backend.db.nursery.tables.references.WITHDRAWAL_SUMMARIES
 import com.terraformation.backend.db.seedbank.tables.references.ACCESSIONS
-import com.terraformation.backend.db.tracking.MonitoringPlotId
 import com.terraformation.backend.db.tracking.PlantingSiteId
 import com.terraformation.backend.db.tracking.RecordedSpeciesCertainty
 import com.terraformation.backend.db.tracking.tables.references.DELIVERIES
-import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOT_HISTORIES
 import com.terraformation.backend.db.tracking.tables.references.OBSERVATIONS
-import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PLOTS
 import com.terraformation.backend.db.tracking.tables.references.OBSERVED_SITE_SPECIES_TOTALS
-import com.terraformation.backend.db.tracking.tables.references.OBSERVED_SUBZONE_SPECIES_TOTALS
 import com.terraformation.backend.db.tracking.tables.references.PLANTINGS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SUBZONES
-import com.terraformation.backend.db.tracking.tables.references.PLANTING_ZONE_T0_TEMP_DENSITIES
-import com.terraformation.backend.db.tracking.tables.references.PLOT_T0_DENSITIES
 import com.terraformation.backend.i18n.Messages
-import com.terraformation.backend.tracking.db.latestObservationForSubzoneField
-import com.terraformation.backend.tracking.db.observationIdForPlot
 import com.terraformation.backend.tracking.model.calculateSurvivalRate
-import com.terraformation.backend.util.HECTARES_PER_PLOT
 import jakarta.inject.Named
 import java.math.BigDecimal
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3732** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3732?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
```diff
@@ -1503,18 +1503,27 @@ class ObservationResultsStore(private val dslContext: DSLContext) {
...
+                                      SPECIES_ID.isNull
+                                          .and(OBSERVED_SITE_SPECIES_TOTALS.SPECIES_ID.isNull)
+                                          .and(CERTAINTY_ID.ne(RecordedSpeciesCertainty.Unknown))
+                                          .and(
+                                              OBSERVED_SITE_SPECIES_TOTALS.CERTAINTY_ID.ne(
+                                                  RecordedSpeciesCertainty.Unknown
+                                              )
+                                          )
+                                  )
+                          )
```
ice1080, line 1526: Would love to know of a cleaner way to do this if possible

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
ice1080, line 1: I couldn't find a way to do this in a `.convertFrom` block, but if that's possible then it would remove the post-processing required, and the changes to `updateReportSystemMetricWithTerrawareData`

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1527,190 +1557,47 @@ class ReportStore(
...
-          .convertFrom { it.toInt() }
+    speciesPairs.forEach { (speciesList, includesTempPlots) ->
+      sumDensity += speciesList.mapNotNull { it.t0Density }.sumOf { it }
+      numKnownLive +=
+          if (includesTempPlots) {
+            speciesList.sumOf { it.latestLive }
+          } else {
+            speciesList.sumOf { it.permanentLive }
+          }
+    }
```
ice1080, line 1599: As a note, I couldn't use the existing `unionSpecies` and `species.calculateSurvivalRate` because the latter has to know if the site includes temp plots in the calculation and since this is a multi-site operation, it depended on each site.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
```diff
@@ -1527,190 +1557,47 @@ class ReportStore(
...
-      )
+  // Calculate survival rate by fetching observations from observationResultsStore
+  private fun calculateSurvivalRateForReport(reportId: ReportId): Int? {
+    val observationIds =
+        dslContext
+            .select(OBSERVATIONS.ID)
+            .from(OBSERVATIONS, REPORTS)
+            .where(REPORTS.ID.eq(reportId))
+            .and(OBSERVATIONS.ID.`in`(observationsInReportPeriod))
+            .fetch { it[OBSERVATIONS.ID]!! }
```
sgrimm, line 1570: I'm not sure I understand this query. Is it doing an existence check on the report ID, and then, if the report exists, doing existence checks on the observation IDs?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStore.kt
sgrimm: I think the `or` is needed for the nulls, but I guess it'd maybe be slightly clearer to check for equality with `RecordedSpeciesCertainty.Other` rather than inequality with `Unknown`. You don't care about `Known` at that point since the species ID wouldn't be null.

sgrimm (changes requested):
Seems reasonable in general, but one of the queries is unclear to me and could use some comments describing what it does.

src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
ice1080: This one is basically using the `observationsInReportPeriod` to get the latest observation id per planting site within the reporting period, which is obviously unclear if you don't already know what's in `observationsInReportPeriod`. Will add comments

sgrimm: Ah, okay, I didn't catch that `observationsInReportPeriod` is a subquery here, not just a `Collection<ObservationId>`. Makes more sense now.

sgrimm (approved):

PR #3733, opened 2025-12-16
Flip metric publishability of survival/mortality rates

ice1080:
We are ready to merge survival rates, so need to make the system metric publishable, and want to make mortality rate un-publishable (which will still keep it in published reports).
Flip the flags in the type codes sql.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
@@ -4177,6 +4177,171 @@ class ReportStoreTest : DatabaseTest(), RunsAsDatabaseUser {
     private val fileId1 by lazy { insertFile() }
     private val fileId2 by lazy { insertFile() }
 
+    @BeforeEach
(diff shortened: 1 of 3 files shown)


ice1080:
* **#3733** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
Test failure here seems legit.

src/test/kotlin/com/terraformation/backend/accelerator/db/ReportStoreTest.kt
ice1080, line 1: This only changed the 50 to 51 and the metric being checked. The rest of the changes are because I moved the `@BeforeEach`to above the tests instead of below them

sgrimm (approved):

PR #3738, opened 2025-12-17
SW-7874 Delete mortality rate system metric and temporary project metrics

ice1080:
Product decided that they wanted to do a clean break from mortality rate, even in older reports (including published). They have already recorded all current mortality rate values in case these values are needed in older reports later on.
Delete the mortality rate system metric.

Additionally, a project in prod had setup a project metric for survival rate temporarily until the real survival rate was merged which is happening in the next release. Product has also already recorded all current survival rate values for this project metric in order to go back into old reports and republish them with an overridden value of what was used with this project metric.
Delete this project metric in prod as well.


ice1080:
* **#3738** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/R__TypeCodes.sql
ice1080, line 1: I wasn't sure if we normally renumber enums like this when a middle one is deleted. Lmk if `Hectares Planted` and `Survival Rate` should be renumbered to 5 and 6 respectively.

tommylau523 (approved):

PR #3750, opened 2025-12-19
SW-7829 Rename request and response payloads for zone->strata

ice1080:
Update payload names to ensure both forward and backwards compatibility with request and response class names (and endpoints).

Updating property names and descriptions will come in a later PR.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/MonitoringResultsPayloads.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/MonitoringResultsPayloads.kt
@@ -239,7 +239,49 @@ data class ObservationMonitoringPlotResultsPayload(
     get() = media
 }
 
+@Schema(description = "Use ObservationSubstratumResultsPayload instead", deprecated = true)
 data class ObservationPlantingSubzoneResultsPayload(
+    val areaHa: BigDecimal,
+    val completedTime: Instant?,
+    val estimatedPlants: Int?,
+    val monitoringPlots: List<ObservationMonitoringPlotResultsPayload>,
+    val mortalityRate: Int?,
+    val mortalityRateStdDev: Int?,
+    val name: String,
+    val plantingDensity: Int,
+    val plantingDensityStdDev: Int?,
+    val plantingSubzoneId: PlantingSubzoneId?,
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3750** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/SubstrataController.kt
```diff
@@ -53,21 +57,54 @@ class PlantingSubzonesController(
...
   }
+
+  @Operation(summary = "Updates the planting-completed state of a substratum.")
+  @PutMapping("/substrata/{id}")
+  fun updateSubstrata(
+      @PathVariable id: PlantingSubzoneId,
+      @RequestBody payload: UpdateSubstratumRequestPayload,
+  ): SimpleSuccessResponsePayload {
+    plantingSiteStore.updatePlantingSubzoneCompleted(id, payload.plantingCompleted)
+
```
ice1080, line 68: The old endpoint should only be needed until the FE uses the new one, because the mobile app doesn't use this. I was unable to get OpenAPI to allow a request payload that works for either name while still keeping the old name in the generated schema in the FE, which was necessary because the FE references it directly using `components['schemas']['UpdatePlantingSubzoneRequestPayload']`.

src/main/kotlin/com/terraformation/backend/tracking/api/MonitoringResultsPayloads.kt
ice1080, line 1: For this file, `sealed interface` would not work for simplifying the api to work with either the old or the new, because the FE references these types and then extends them. Using `sealed interface` adds `Record<string, never>` meaning that they could not be extended and failed the typescript check. Instead of this, I created new types with some new properties. Once Josh is back from vacation, we can test a removal of the old types and ensure that the mobile app still works.

tommylau523 (approved):

PR #3771, opened 2025-12-29
SW-7857 Use web app's internal role name translations

sgrimm:
The Spanish and French translations for project internal role names didn't match
between terraware-web and this repo, so when the frontend tried to use a role as a
search filter, it didn't work because the search string wasn't recognized as a
valid enum value. Update the translations to match the ones in terraware-web.

### DIFF
--- a/src/main/resources/i18n/Enums_es.properties
+++ b/src/main/resources/i18n/Enums_es.properties
@@ -446,13 +446,13 @@ public.ProjectInternalRole.ClimateImpactLead=Responsable del impacto climático
 # 910885ab
 public.ProjectInternalRole.Consultant=Consultor
 # 0095925f
-public.ProjectInternalRole.GISLead=Responsable del GIS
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#3771** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
Do we need to update the file for auto translate to know what the correct translations of all of those should be, for use in other strings?

sgrimm:
Autotranslate will leave these values alone unless we change the English strings, in which case we'd need new translations anyway.

PR #3774, opened 2025-12-29
SW-7829 Rename zone->strata in api response payloads

ice1080:
Rename properties in response payloads and add `get`ters for simple properties, and new response classes for more complex renames. Add deprecations around old property names.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/DeliveriesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/DeliveriesController.kt
@@ -51,6 +51,7 @@ class DeliveriesController(
   }
 }
 
+// response payload
 data class PlantingPayload(
     val id: PlantingId,
     @Schema(description = "If type is \"Reassignment To\", the reassignment notes, if any.")
@@ -61,7 +62,7 @@ data class PlantingPayload(
                 "will be negative."
     )
     val numPlants: Int,
-    val plantingSubzoneId: SubstratumId?,
+    val substratumId: SubstratumId?,
     val speciesId: SpeciesId,
     val type: PlantingType,
 ) {
@@ -71,12 +72,17 @@ data class PlantingPayload(
       id = model.id,
       notes = model.notes,
       numPlants = model.numPlants,
-      plantingSubzoneId = model.substratumId,
+      substratumId = model.substratumId,
       speciesId = model.speciesId,
       type = model.type,
   )
+
+  @Deprecated("Use substratumId instead.")
+  val plantingSubzoneId: SubstratumId?
+    get() = substratumId
 }
 
+// response payload
 data class DeliveryPayload(
     val id: DeliveryId,
     val plantings: List<PlantingPayload>,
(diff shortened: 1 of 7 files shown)


ice1080:
* **#3774** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/DeliveriesController.kt
```diff
@@ -51,6 +51,7 @@ class DeliveriesController(
   }
 }
 
+// response payload
```
sgrimm, line 54: Rather than adding comments like this everywhere, should we maybe rename these classes to make it clear which ones are for requests vs. responses? Or perhaps to avoid cluttering up the names, we can define marker interfaces or annotations to indicate what direction a payload is for.

I don't love the "add a comment" approach because it's easy to get wrong without anything catching the mistake. With an interface or annotation we could add a test to make sure we don't miscategorize any classes.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/api/DeliveriesController.kt
ice1080: Many of the comments will go away since they are on classes that are being replaced by the new classes. Unfortunately we can't rename the classes because most are referenced directly in the FE code. I know you mentioned changing those to reference the payload of an endpoint, but afaik that only works for classes that are the top level payload of an endpoint, and something like `PlantingPayload` doesn't work because it is in a list inside a response payload (and there are some that are 3 or 4 levels deep instead of just 1). 

I'm open to ideas for a better approach here, and can add an annotation if that is preferred (would probably do that in a separate PR). There are definitely cases where some classes are used in both requests and responses (including in my own T0Controller). I think I'd like to move away from that being possible with something like the tests you're talking about. Let me know thoughts.

sgrimm: My preference is an annotation or marker interface, but just to be thorough about the rename option, the FE code could refer to child object classes by walking the type structures. For example, if a payload class had a `blahBlah` child object, you could do something like
```
export type TopLevelPayload =
  paths['/api/v1/whatever']['post']['requestBody']['content']['application/json'];

export type BlahBlah = TopLevelPayload['blahBlah'];
```
That said, if these comments are just temporary, then I'm fine with them for now. We shouldn't block this PR on a discussion of how to mark our payload classes.

ice1080: As a note, looks like we have about 280 instances of `type <bla> = components['schemas']['SomeType']` and about 210 instances of using `paths`, so switching all over to `paths` might take quite a while. I'll work on converting the ones that directly reference a type with 'zone' in the name for now

ice1080: Actually it looks like none of the types defined in `src/types` have been defined using `paths` since these are supposed to be path-agnostic types :thinking_face: Hmm still thinking how I want to handle this in the FE. Anyway, will merge this one for now.

PR #3778, opened 2025-12-30
SW-7829 Rename zone->strata in tracking api request payloads

ice1080:
Rename properties in request payloads by providing the ability to specify either the old or the new. Do this for files in the `tracking` folder to keep the PR small.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
@@ -716,7 +716,29 @@ data class UpdatedPlantingSeasonPayload(
   fun toModel() = UpdatedPlantingSeasonModel(endDate = endDate, id = id, startDate = startDate)
 }
 
+@Schema(description = "Use NewSubstratumPayload", deprecated = true)
 data class NewPlantingSubzonePayload(
+    @Schema(oneOf = [MultiPolygon::class, Polygon::class]) //
+    val boundary: Geometry,
+    val name: String,
+) {
+  fun validate() {
+    if (boundary !is MultiPolygon && boundary !is Polygon) {
+      throw IllegalArgumentException("Substratum boundaries must be Polygon or MultiPolygon")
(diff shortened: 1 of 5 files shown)


ice1080:
* **#3778** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3778?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3778?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
```diff
@@ -785,19 +835,27 @@ data class PlantingSiteValidationProblemPayload(
...
     )
-    val plantingSubzone: String?,
+    val substratum: String?,
     val problemType: PlantingSiteValidationFailureType,
 ) {
   constructor(
       model: PlantingSiteValidationFailure
   ) : this(model.conflictsWith, model.stratumName, model.substratumName, model.type)
+
+  @Deprecated("Use stratum instead")
```
sgrimm, line 852: Should this use `@Schema` to indicate the deprecation instead? Not sure if the OpenAPI generator will pick up `@Deprecated`.

sgrimm (approved):
Seems like a reasonable approach.

src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
ice1080: OpenAPI does actually pick this up too! I've been using this on the getters to enforce the deprecation at a kotlin level too, so that we don't accidentally use these properties in the BE code as well. Here's the generated-schema output for this property:

```typescript
            /**
             * @deprecated
             * @description Use stratum instead
             */
            plantingZone?: string;
```

PR #3781, opened 2025-12-31
Test disjoint substratum observations with map edit

sgrimm:
Add a test case to verify (and document) how survival rates are calculated when
there are geometry changes that affect substrata that aren't observed in later
observations.

This is also a trial run of specifying the expected values in a more
self-documenting way; it includes a new null-safe wrapper for `Map` that allows
us to do null-safe array-style gets without having to litter the code with `!!`
modifiers.

The test uncovered a bug in the test DSL's site editing logic; it wasn't
maintaining consistent stable IDs when substrata moved between strata.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
@@ -16,6 +16,7 @@ import com.terraformation.backend.mockUser
 import com.terraformation.backend.point
 import com.terraformation.backend.tracking.db.ObservationScenarioTest
 import com.terraformation.backend.tracking.model.ObservationResultsModel
+import com.terraformation.backend.util.nullSafeMapOf
 import com.terraformation.backend.util.toPlantsPerHectare
 import io.mockk.every
 import java.math.BigDecimal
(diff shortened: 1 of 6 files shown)


sgrimm:
* **#3781** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3781?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3781?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/tracking/scenario/ExpectedObservationResult.kt
```diff
@@ -32,12 +34,19 @@ abstract class ExpectedObservationResult<
...
     survivalRate = expected
     survivalRateSet = true
     assertions.add { assertEquals(expected, actualResult.survivalRate, "$name survival rate") }
   }
 
+  /** Returns the ratio between two integers as a percentage value with half-up rounding. */
+  fun percent(numerator: Int, denominator: Int): Int {
+    return BigDecimal(numerator * 100)
+        .divide(BigDecimal(denominator), 0, RoundingMode.HALF_UP)
+        .toInt()
```
ice1080: I can't remember off hand for BigDecimal, is it possible to change `Int` to `Number` in the args, in order to support doubles? It's not real important, just was curious in case we ever wanted a decimal density

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -1272,6 +1273,118 @@ class ObservationStoreSurvivalRateCalculationTest : ObservationScenarioTest() {
...
+          substratum(3) {
+            survivalRate(percent(obsLive[1][3], densities[3]))
+            plot(3) { survivalRate(percent(obsLive[1][3], densities[3])) }
+          }
+        }
+      }
+
+      observation(2) { plot(1) { species(0, live = obsLive[2][1]) } }
+
+      expectResults(observation = 2) {
```
ice1080, line 1342: Should this also check `stratum(2)`?

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
```diff
@@ -1272,6 +1273,118 @@ class ObservationStoreSurvivalRateCalculationTest : ObservationScenarioTest() {
...
+        }
+        stratum(2) {
+          substratum(2) { plot(2) } // Moved from stratum 1
+          substratum(3) { plot(3) }
+        }
+      }
+
+      observation(3) { plot(3) { species(0, live = obsLive[3][3]) } }
+
+      expectResults(observation = 3) {
```
ice1080, line 1371: Same question for `stratum(1)`

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreSurvivalRateCalculationTest.kt
sgrimm: There shouldn't be any results for stratum 2 in this observation, I think, because it wasn't observed. We roll its results from observation 1 up into the site-level rate, but we don't include it in the results. But maybe it makes sense to add an assertion that there aren't any results for it, to make that explicit.

ice1080: Ah, I couldn't remember if this used `observationResultsStore.fetchOneById` or `fetchSummariesForPlantingSite` under the hood, which would change whether or not stratum 2 is in the results (I believe)

sgrimm: No worries; adding an explicit "shouldn't have a result for this [sub]stratum" assertion makes the test both more robust and more useful as documentation.

src/test/kotlin/com/terraformation/backend/tracking/scenario/ExpectedObservationResult.kt
sgrimm: Yeah, may as well make these `Number`, I guess. Good call.

ice1080 (approved):

PR #3803, opened 2026-01-06
SW-7891 Get plot observations with species densities

ice1080:
We want to display the species in each observation when selecting observations in addition to when viewing survival rate settings.
Add an endpoint to return the list of observations and their species densities for each plot. This will replace the data currently returned by the search api.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
@@ -11,7 +11,9 @@ import com.terraformation.backend.db.tracking.PlantingSiteId
 import com.terraformation.backend.db.tracking.StratumId
 import com.terraformation.backend.tracking.T0Service
 import com.terraformation.backend.tracking.db.T0Store
+import com.terraformation.backend.tracking.model.ObservationSpeciesDensityModel
 import com.terraformation.backend.tracking.model.OptionalSpeciesDensityModel
+import com.terraformation.backend.tracking.model.PlotObservationSpeciesDensityModel
 import com.terraformation.backend.tracking.model.PlotSpeciesModel
 import com.terraformation.backend.tracking.model.PlotT0DataModel
 import com.terraformation.backend.tracking.model.SiteT0DataModel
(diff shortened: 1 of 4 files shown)


ice1080:
* **#3803** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3803?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/T0Controller.kt
```diff
@@ -52,7 +56,21 @@ class T0Controller(
...
 
-  @GetMapping("/site/{plantingSiteId}/species")
+  @Operation(
+      summary =
+          "Lists all permanent plots with completed observations, and the observations' recorded species for a planting site"
+  )
+  @GetMapping("/site/{plantingSiteId}/observations")
+  fun getT0PlotObservationsForPlantingSite(
+      @PathVariable plantingSiteId: PlantingSiteId
+  ): GetT0SitePlotObservationsResponsePayload {
```
sgrimm: Naming nit: This isn't returning T0 data. Maybe include the word "density" in these names rather than "t0"?

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -83,6 +85,87 @@ class T0Store(
...
+          .join(OBSERVATIONS)
+          .on(OBSERVATIONS.ID.eq(OBSERVATION_ID))
+          .join(MONITORING_PLOTS)
+          .on(MONITORING_PLOTS.ID.eq(MONITORING_PLOT_ID))
+          .where(OBSERVATIONS.PLANTING_SITE_ID.eq(plantingSiteId))
+          .and(completedObservationsCondition)
+          .and(OBSERVED_PLOT_SPECIES_TOTALS.CERTAINTY_ID.eq(RecordedSpeciesCertainty.Known))
+          .and(MONITORING_PLOTS.SUBSTRATUM_ID.isNotNull)
+          // currently permanent:
+          .and(MONITORING_PLOTS.PERMANENT_INDEX.isNotNull)
```
sgrimm, line 119: Why would we not want to return data for plots that were permanent during the observation we're looking at, even if they no longer are?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: The survival rate settings only allow the user to set t0 for currently permanent plots, so I am excluding others here. Right now this exclusion is happening in the FE because we use the data for currently temporary plots for the temporary tab.

PR #3805, opened 2026-01-07
SW-7207 Add validation to search api filter prefixes

ice1080:
The `filters` field in the search api payload allows for filtering sublists to only contain values, instead of only filtering the top level results. Currently, non-existent filter prefixes do nothing and don't filter the sublist. Add validation to throw an error for non-existent filter prefixes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
@@ -76,6 +76,8 @@ class SearchController(
     val rootPrefix = resolvePrefix(payload.prefix)
     val count = if (payload.count > 0) payload.count else Int.MAX_VALUE
 
+    validateFilterPrefixes(payload, rootPrefix)
+
     return SearchResponsePayload(
         searchService.search(
             rootPrefix,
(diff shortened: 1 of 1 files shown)


ice1080:
* **#3805** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3805?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3805?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/api/SearchController.kt
```diff
@@ -76,6 +76,8 @@ class SearchController(
     val rootPrefix = resolvePrefix(payload.prefix)
     val count = if (payload.count > 0) payload.count else Int.MAX_VALUE
 
+    validateFilterPrefixes(payload, rootPrefix)
+
```
ice1080, line 80: Could alternatively move this into the `toSearchCriteria` so that we're not going through the list of filters twice

sgrimm: Nah, I think this is fine; the validation this is doing is pretty cheap.

sgrimm (approved):

PR #3847, opened 2026-01-26
SW-7997 Fix NPE in prod around null survival rates

ice1080:
Commits [f6364eda](https://github.com/terraware/terraware-server/commit/f6364eda) and [1c8bfaf1](https://github.com/terraware/terraware-server/commit/1c8bfaf1) allowed for report system metrics to have a null `system_value` which was needed to accurately show survival rates in reports. It updated the check to use the `system_time` for null checking and if it needed to be calculated.
This was then tested with reports that didn't have a survival rate system metric, but was not tested for reports that had this system metric but had a null value and a non-null `system_time`.

Specifically, the issue is from the case statements for the various system metric types. When using a case statement, even if only one of the cases is nullable, jOOQ doesn’t understand this and requires that all of them be able to handle null. In this situation, the cases for non-survival rate system metrics required that their `convertFrom` handle null values, because survival rate could be null.

was:
`.convertFrom { it.toInt() }`
needed:
`.convertFrom { it?.toInt() }`

This only showed up when a  survival rate system metric had a null `system_value` and a non-null `system_time`.

Fix the `convertFrom`s in `ReportStore`.
Add tests for survival rate with a non-null `system_time` and null `system_value`.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ReportStore.kt
@@ -1579,7 +1579,7 @@ class ReportStore(
                     .where(PROJECT_ID.eq(REPORTS.PROJECT_ID))
                     .and(COLLECTED_DATE.between(REPORTS.START_DATE, REPORTS.END_DATE))
             )
-            .convertFrom { it.toInt() }
+            .convertFrom { it?.toInt() }
       }
 
   private val withdrawnSeedlingsField =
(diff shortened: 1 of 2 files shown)


ice1080:
* **#3847** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3847?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3847?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #3849, opened 2026-01-29
Add admin UI page for manual splat generation

sgrimm:
To make it easier for people to test different combinations of settings
when generating 3D Gaussian splatting models from videos, add an admin UI
page that exposes all the processing script's parameters and allows the
same video to be processed multiple times.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
+++ b/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
@@ -146,6 +146,16 @@ class SplatService(
                 .execute()
           }
 
+      if (rowsInserted == 0 && force) {
+        with(SPLATS) {
+          dslContext
+              .update(SPLATS)
+              .set(ASSET_STATUS_ID, AssetStatus.Preparing)
+              .where(FILE_ID.eq(fileId))
+              .execute()
+        }
+      }
+
       if (rowsInserted == 1 || force) {
         val requestMessage =
             SplatterRequestMessage(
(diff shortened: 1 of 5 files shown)


sgrimm:
* **#3849** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3849?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3849?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -137,19 +146,22 @@ class SplatService(
                 .execute()
           }
 
-      if (rowsInserted == 1) {
+      if (rowsInserted == 1 || force) {
```
ice1080, line 159: If `force` is true, would we want to change the onConflict handling to set the asset_status_id back to `Preparing`?

ice1080 (commented):
The codex and grpahite comments also seem valid

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
sgrimm: Did it as a separate update statement because it looked ugly to make the on-conflict clause conditional on the force flag.

ice1080 (approved):

PR #3857, opened 2026-01-30
SW-8013 Support retrieval of splat annotations

ice1080:
Add table to hold Splat Annotations.

Add endpoint to retrieve annotations for a given observationId and fileId.


ice1080:
* **#3858** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3858?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3857** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3857?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3857?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/db/DatabaseBackedTest.kt
```diff
@@ -3045,6 +3050,78 @@ abstract class DatabaseBackedTest {
...
+          .set(TEXT, text)
+          .set(LABEL, label)
+          .set(
+              POSITION,
+              DSL.function(
+                  "st_pointz",
+                  GeometryBinding.dataType,
+                  DSL.`val`(position.x),
+                  DSL.`val`(position.y),
+                  DSL.`val`(position.z),
```
ice1080: The GeometryBinding class forces all geometry to use an `st_transform` function with a hardcoded `SRID.LONG_LAT`, which is incorrect for this case and causes inserts to fail because `LONG_LAT` requires only x/y. I'm guessing there's a better way to handle this than what I have here.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -198,6 +202,49 @@ class SplatService(
...
+      observationId: ObservationId,
+      fileId: FileId,
+  ): List<SplatAnnotationModel> {
+    ensureObservationFile(observationId, fileId)
+    ensureSplat(fileId)
+
+    val position3D =
+        DSL.function("ST_Force3DZ", GeometryBinding.dataType, SPLAT_ANNOTATIONS.POSITION)
+    val cameraPosition3D =
+        DSL.function("ST_Force3DZ", GeometryBinding.dataType, SPLAT_ANNOTATIONS.CAMERA_POSITION)
```
ice1080: See below comment about `GeometryBinding` class. In this case it forces retrieval of geometries to X/Y points without Z. I'm very open to a better way to handle this.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -198,6 +199,41 @@ class SplatService(
     }
   }
 
+  fun listSplatAnnotations(
+      observationId: ObservationId,
+      fileId: FileId,
+  ): List<SplatAnnotationModel> {
+    ensureObservationFile(observationId, fileId)
```
sgrimm: The approach I've been taking in structuring these service methods is to split them into an entity-type-specific entrypoint (which would look like the method you've declared here) and an implementation that doesn't care what kind of entity the splat's file is associated with. My assumption is that we'll eventually want to support splats for things other than observations.

So concretely, my suggestion here is to move everything under this comment into a private helper function that just takes a file ID.

May as well move the public function up higher in the file so the public API comes first, though admittedly that's not something we're consistent about in general.

src/main/resources/db/migration/0400/V443__SplatAnnotations.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE splat_annotations (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    file_id BIGINT REFERENCES files,
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    title TEXT NOT NULL,
+    text TEXT,
```
sgrimm: jOOQ will properly quote it, but we should probably avoid using a PostgreSQL keyword as a column name (`text`).

src/main/resources/db/migration/0400/V443__SplatAnnotations.sql
```diff
@@ -0,0 +1,22 @@
...
+    label TEXT,
```
sgrimm, line 10: What's the difference between "text" and "label"? Maybe add a blurb for one or both of them to `R__Comments.sql` to clarify?

src/main/resources/db/migration/R__Comments.sql
```diff
@@ -218,6 +218,7 @@ COMMENT ON TABLE species_problem_types IS '(Enum) Specific types of problems tha
 COMMENT ON TABLE species_problems IS 'Problems found in species data. Rows are deleted from this table when the problem is marked as ignored by the user or the user accepts the suggested fix.';
 
 COMMENT ON TABLE splats IS 'Information about 3D Gaussian splatting models generated from video files.';
+COMMENT ON TABLE splat_annotations IS 'Annotations that should be displayed inside splat models.';
```
sgrimm, line 224: Nit: Separate tables with blank lines.

sgrimm (changes requested):

src/main/resources/db/migration/R__Comments.sql
ice1080: Will do. We had several other instances of this (a few of them mine), so fixing that as well so that we're consistent

src/main/kotlin/com/terraformation/backend/splat/api/SplatsController.kt
```diff
@@ -157,7 +159,7 @@ data class SplatAnnotationPayload(
     val fileId: FileId,
     val label: String?,
     val position: CoordinatePayload,
-    val text: String?,
+    val bodyText: String?,
```
sgrimm, line 162: Alphabetize, maybe.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/splat/api/SplatsController.kt
ice1080: Went ahead and did that in the next PR

src/main/resources/db/migration/0400/V443__SplatAnnotations.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE splat_annotations (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    file_id BIGINT REFERENCES files,
```
sgrimm: This should be NOT NULL, right?

ice1080: Ah good catch. I was referencing V442, and not thinking when I removed it as the primary key.

ice1080:
### Merge activity

* **Feb 2, 6:51 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3857).

PR #3868, opened 2026-02-03
SW-8015 Support mobile app background video uploads

sgrimm:
There is a limitation on background file uploads in the mobile app: they are
always submitted with a content type of `application/octet-stream` even though
their request bodies are multipart form data. This prevents the default request
handling logic from parsing the request body.

Add an alternate version of the observation media upload endpoint that accepts
these requests.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/Extensions.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/Extensions.kt
@@ -3,6 +3,7 @@ package com.terraformation.backend.api
 import com.terraformation.backend.file.SizedInputStream
 import jakarta.ws.rs.NotSupportedException
 import java.io.ByteArrayOutputStream
+import org.apache.commons.fileupload2.core.FileItemInput
 import org.apache.tika.mime.MimeTypes
 import org.geotools.api.feature.simple.SimpleFeature
 import org.geotools.geojson.feature.FeatureJSON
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3868** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
```diff
@@ -593,6 +599,85 @@ class ObservationsController(
...
+                fakeContentType
+              } else {
+                super.getHeader(name)
+              }
+        }
+
+    val fileUpload = JakartaServletFileUpload()
+    val items = fileUpload.getItemIterator(wrappedRequest)
+    var payload: UploadPlotMediaRequestPayload? = null
+    var response: UploadPlotPhotoResponsePayload? = null
```
ice1080: We may want to change the response name from `Photo` to `Media` to match the request at some point

src/main/kotlin/com/terraformation/backend/api/Extensions.kt
ice1080, line 1: Might be worth adding tests for these at some point

ice1080 (approved):

ice1080:
Codex comment also seems valid

sgrimm:
Graphite's comment seems valid to me too -- guess we've never actually exercised that code path. And like you said, we should have caught it in tests.

src/main/kotlin/com/terraformation/backend/api/Extensions.kt
sgrimm: Done!

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationsController.kt
sgrimm: Renamed this payload class.

ice1080 (approved):

PR #3879, opened 2026-02-09
SW-8035 Add ApplicationsTable as single sublist to Projects

ice1080:
In order to support the new Quick Filter for the matrix view of `Application Projects`, add an `Applications` search table and include it as a single value sublist on `Projects`.


ice1080:
* **#3879** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3879?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3879?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/search/table/ApplicationsTable.kt
```diff
@@ -0,0 +1,44 @@
...
+    return if (currentUser().canReadAllAcceleratorDetails()) {
+      DSL.trueCondition()
+    } else {
+      DSL.exists(
+          DSL.selectOne()
+              .from(PROJECTS)
+              .join(APPLICATIONS)
+              .on(APPLICATIONS.PROJECT_ID.eq(PROJECTS.ID))
+              .where(PROJECTS.ORGANIZATION_ID.`in`(currentUser().organizationRoles.keys))
+      )
```
sgrimm: I believe this will let users query the entire applications table (for all orgs/projects) if they're in an org that has an application. Not a massive problem since they'll only be able to look at IDs and statuses, but still not ideal.

It'd be better to instead make this table inherit visibility from the projects table.

src/main/resources/i18n/Messages_en.properties
```diff
@@ -564,6 +564,8 @@ search.accessions.withdrawalNotes=Notes (withdrawal)
 search.accessions.withdrawalPurpose=Purpose
 search.accessions.withdrawalQuantity=Quantity of seeds withdrawn
 search.accessions.withdrawalUnits=Units of measurement of quantity withdrawn
+search.applications.id=ID
```
sgrimm: This should include the entity name (so "Application ID") since it's used as a column header in server-side CSV exports. Just having "ID" as the column header will be ambiguous.

sgrimm (changes requested):

src/main/resources/i18n/Messages_en.properties
ice1080: Would you like me to change `search.accessions.id` as well? That's what I had used as a reference for this. Also `search.batches.id` is `ID (seedling batch)`, would that be better as `Seedling Batch ID` (or `Batch ID`)?

ice1080: I changed `accessions.id` but not batches!

src/main/kotlin/com/terraformation/backend/search/table/ApplicationsTable.kt
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          idWrapperField("id", APPLICATIONS.ID) { ApplicationId(it) },
+          enumField("status", APPLICATIONS.APPLICATION_STATUS_ID),
+      )
+
+  override val inheritsVisibilityFrom: SearchTable
+    get() = tables.projects
```
sgrimm, line 30: You'll also need to implement `joinForVisibility()`.

src/main/resources/i18n/Messages_fr.properties
```diff
@@ -951,6 +951,10 @@ search.accessions.withdrawalPurpose=Objectif
 search.accessions.withdrawalQuantity=Quantité de graines prélevées
 # 6cf2a280
 search.accessions.withdrawalUnits=Unités de mesure de la quantité prélevée
+# 0ba65e03
+search.applications.id=ID
```
sgrimm: Looks like the es/fr strings didn't get updated.

sgrimm (changes requested):

src/main/resources/i18n/Messages_fr.properties
ice1080: Good catch

sgrimm (approved):

ice1080:
### Merge activity

* **Feb 9, 10:27 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/3879).

PR #3884, opened 2026-02-10
Add position columns to splats table

ice1080:
We are wanting to add a per-splat origin position and camera starting position. To handle this, add `origin_position_...` and `camera_position_...` columns to the `splats` table to hold this. Only provide an endpoint to retrieve this info, since editing origins and camera positions opens up quite a few design questions that can be figured out later. These will be manually inserted for now.

Also include the list of annotations in this endpoint, and deprecate the `/annotations` endpoint, since the annotations list is also general information that would be retrieved/needed at the same time.


ice1080:
* **#3884** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3884?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3884?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0400/V445__SplatInformation.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE splat_information (
```
sgrimm: Curious why this is a separate table rather than additional columns on the main splats table.

ice1080: Ah good thought, that makes more sense than a separate table. I was thinking of how the response entity for a file can't include the rest of the data, but that doesn't mean that the table itself can't contain the info (I recognize that `splats` is only storing the `file_id` anyway, just had a block in my head for some reason).

sgrimm (approved):

PR #3893, opened 2026-02-11
SW-7931 SW-7945 Add phase to project accelerator details endpoints

ice1080:
Replace `cohortPhase` with `phase` in GET project accelerator details endpoints. Add `cohortPhase` as deprecated getter for `phase` for backwards compatibility.

Add `phase` to PUT endpoint to update cohort phase for now. Add requirement that document properties are non-null if phase is non-null (per [PRD](https://terraformation.atlassian.net/wiki/spaces/PM/pages/2038464515/PRD+Removing+Cohorts+from+Terraware#Project-Data-(Updates))).

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
@@ -8,6 +8,8 @@ import com.terraformation.backend.accelerator.model.TRACKED_ACCUMULATED_METRICS
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.ProjectNotFoundException
+import com.terraformation.backend.db.accelerator.CohortId
+import com.terraformation.backend.db.accelerator.CohortPhase
 import com.terraformation.backend.db.accelerator.SystemMetric
 import com.terraformation.backend.db.accelerator.tables.references.COHORTS
 import com.terraformation.backend.db.accelerator.tables.references.PROJECT_ACCELERATOR_DETAILS
@@ -181,31 +183,7 @@ class ProjectAcceleratorDetailsStore(
       }
 
       if (existing.phase != updated.phase) {
-        val now = clock.instant()
-        if (existing.cohortId != null) {
-          dslContext
-              .update(COHORTS)
-              .set(COHORTS.PHASE_ID, updated.phase)
-              .set(COHORTS.MODIFIED_BY, currentUser().userId)
-              .set(COHORTS.MODIFIED_TIME, now)
-              .where(COHORTS.ID.eq(existing.cohortId))
-              .execute()
-        } else {
-          with(COHORTS) {
-            dslContext
-                .insertInto(COHORTS)
-                .set(
-                    NAME,
-                    DSL.select(PROJECTS.NAME).from(PROJECTS).where(PROJECTS.ID.eq(projectId)),
-                )
-                .set(PHASE_ID, updated.phase)
-                .set(CREATED_BY, currentUser().userId)
-                .set(CREATED_TIME, now)
-                .set(MODIFIED_BY, currentUser().userId)
-                .set(MODIFIED_TIME, now)
-                .execute()
-          }
-        }
+        updateProjectPhase(projectId, existing.cohortId, updated.phase)
       }
 
       if (existing.landUseModelTypes != updated.landUseModelTypes) {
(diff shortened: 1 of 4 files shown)


ice1080:
* **#3893** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3893?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3893?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
```diff
@@ -171,6 +175,14 @@ class ProjectAcceleratorDetailsStore(
             .execute()
       }
 
+      if (existing.phase != updated.phase) {
+        dslContext
```
ice1080: The downside here is that it requires that the FE get deployed at the same time as the BE, otherwise the BE will set the phase to null because the request won't be updated yet to include phases. Might need some intermediate BE and FE PRs that just includes the phase in the request.

sgrimm: I think our deploys are close enough together to not worry about this. An accelerator admin would have to be updating a project in the middle of a deploy to hit this.

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
```diff
@@ -171,6 +180,14 @@ class ProjectAcceleratorDetailsStore(
...
+            .update(COHORTS)
+            .set(COHORTS.PHASE_ID, updated.phase)
+            .where(COHORTS.ID.eq(existing.cohortId))
+            .execute()
```
sgrimm: What if the project isn't in a cohort, e.g., because it's being assigned a phase for the first time? Do we want to create a cohort on the fly so as not to break the "has phase == is in cohort" invariant?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
ice1080: Good question, I'll check with product on this. I have tests that ensure no issues when no cohort, but we might want an intermediate setup until we finish getting rid of cohorts

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
ice1080: Makes sense, I was partially worried about staging as well, but maybe I'll wait to merge this till the FE PR is ready too

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
ice1080: Ah actually, there's a separate problem with this, because the retrieval for it is going to use the cohorts table, so it's not actually being saved _unless_ we create a new cohort on the fly.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
ice1080: (just for records' sakes) product confirmed that we want to create temporary cohorts on the fly with names = project.name

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
```diff
@@ -266,4 +244,58 @@ class ProjectAcceleratorDetailsStore(
...
+        dslContext
+            .update(PROJECTS)
+            .setNull(PROJECTS.COHORT_ID)
+            .where(PROJECTS.ID.eq(projectId))
+            .execute()
+
+        val projectsInCohort =
+            dslContext.fetchCount(PROJECTS, PROJECTS.COHORT_ID.eq(existingCohortId))
+        if (projectsInCohort == 0) {
+          dslContext.deleteFrom(COHORTS).where(COHORTS.ID.eq(existingCohortId)).execute()
```
sgrimm, line 293: We should make sure deleting a cohort won't fail if the cohort has data (modules, I guess primarily).

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/accelerator/db/ProjectAcceleratorDetailsStore.kt
ice1080: Are you thinking I should delete the rows from cohort_modules before this? Or only delete the cohort if there are no modules?

ice1080: If we do delete the cohort_modules rows, this also opens up the question: should we keep the modules themselves if they no longer belong to any cohorts? We could also handle that when we do the later migration to project modules...

sgrimm: Deleting the cohort modules should be enough, I think. The underlying modules are supposed to be reusable across cohorts.

PR #3895, opened 2026-02-13
SW-7936 Add projectModules sublist as placeholder for FE work

ice1080:
In order to change over the Modules list `Cohorts` column to `Projects`, we first need a `projectModules` property on the `ModulesTable`. Until we have an actual `project_modules` table, make this point to `cohort_modules` so that we can update the FE to use this sublist.


ice1080:
* **#3895** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080:
I'm not real sure if this is the right approach. I'm trying to pre-work what I can for the FE but if it's better to close this and wait till we have an actual `project_modules` table, that's also fine with me.

sgrimm (approved):
This seems reasonable as long as we're not including both sublists in the same query (it'd probably produce SQL with a duplicate join).

PR #3896, opened 2026-02-13
SW-8055 Remove unused accelerator organization searches

sgrimm:
The web app no longer needs to search for organizations with the Accelerator
internal tag, so remove that ability from the accelerator organizations
endpoint and change the implementation to only return organizations that have
projects with phases or applications.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/api/AcceleratorOrganizationsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/api/AcceleratorOrganizationsController.kt
@@ -15,13 +15,11 @@ import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
 import io.swagger.v3.oas.annotations.Operation
-import io.swagger.v3.oas.annotations.Parameter
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestMapping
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3897** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3896** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3896?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3896?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/accelerator/api/AcceleratorOrganizationsController.kt
```diff
@@ -35,32 +33,9 @@ class AcceleratorOrganizationsController(
...
-      @Parameter(
-          description = "Whether to also include projects that have been assigned to participants."
-      )
-      @RequestParam
-      includeParticipants: Boolean?,
-      @Parameter(
-          description = "Whether to load all organizations with a project with an application."
-      )
-      @RequestParam
-      hasProjectApplication: Boolean?,
```
ice1080, line 52: The FE is still passing these in, that won't cause any issues with the requests right? I wonder if we have pages that might change behavior with this change

sgrimm: They'll just be ignored, so that should be harmless. The FE, based on my read of the code when I was working on it yesterday, is calling this endpoint to get a list of organization ID -> name mappings that it only ever uses to look up org names for accelerator projects. That should continue to work correctly with this change.

PR #3927, opened 2026-02-19
SW-8059 List accelerator projects based on phase

sgrimm:
Change `AcceleratorProjectService` to query projects with non-null phases, rather
than projects in cohorts. Stop exposing cohort IDs in the API endpoint that lists
all accelerator projects.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/accelerator/AcceleratorProjectService.kt
+++ b/src/main/kotlin/com/terraformation/backend/accelerator/AcceleratorProjectService.kt
@@ -3,7 +3,6 @@ package com.terraformation.backend.accelerator
 import com.terraformation.backend.accelerator.model.AcceleratorProjectModel
 import com.terraformation.backend.customer.model.requirePermissions
 import com.terraformation.backend.db.AcceleratorProjectNotFoundException
-import com.terraformation.backend.db.accelerator.tables.references.COHORTS
 import com.terraformation.backend.db.accelerator.tables.references.PROJECT_VOTE_DECISIONS
 import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.default_schema.ProjectId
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#3927** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/3927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
> Stop exposing cohort IDs in the API endpoint that lists all accelerator projects.

Technically `ModuleEventProject` was also used in the endpoint that created events as well, which is where the issue was with the FE.

sgrimm:
> Technically ModuleEventProject was also used in the endpoint that created events as well, which is where the issue was with the FE.

I don't think that's the case, unless I'm missing something. `EventsController.createEvent` takes a payload of type `CreateModuleEventRequestPayload`, which doesn't have cohorts. Is that the endpoint you're referring to?

ice1080:
> I don't think that's the case, unless I'm missing something. EventsController.createEvent takes a payload of type CreateModuleEventRequestPayload, which doesn't have cohorts. Is that the endpoint you're referring to?

Ah you're correct. Looks like this is all more messiness in the FE. It creates an event and then also updates the event projects, which sends a list of projectIds. But the code is setup to require a list of `ModuleEventProject`s be passed to the parent, and it's not till way later that the rest of the data is thrown out and only the project ids are kept.

ice1080:
So because the code was relying on `ModuleEventPayload`, it was thrown off. Sorry to make you do this PR when it wasn't actually needed, though it does definitely simplify the changes needed in the FE

PR #4042, opened 2026-02-27
Create initial claude SKILL.md for creating db migrations

ice1080:
Create an initial `SKILL.md` for claude to be able to create db migration files and modify the relevant files. This has most of the content necessary to create new tables, but should be added to as needed to support other db migrations.

Note that claude code wrote the initial version of this file. I modified to more accurately represent what we might want from this, and to correct some errors it had.


ice1080:
* **#4042** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4042?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4042?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4042).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

.claude/skills/db-migration/SKILL.md
```diff
@@ -0,0 +1,328 @@
+---
+name: db-migration
+description: Creates a new database migration file with jOOQ code generation config, schema documentation, type codes, and localization. Use when making changes to the database schema.
```
sgrimm: I've heard that skills are picked up more reliably if the description is more "when to use this" rather than "what this does," since the LLM uses the description to decide whether or not a skill is relevant.

.claude/skills/db-migration/SKILL.md
```diff
@@ -0,0 +1,328 @@
...
+---
+
+# Creating a database migration
+
+## Instructions
+
+Create a separate `TaskCreate` for each applicable step below, then work through them in order,
```
sgrimm, line 10: What does "TaskCreate" refer to here?

.claude/skills/db-migration/SKILL.md
```diff
@@ -0,0 +1,328 @@
...
+
+Migrations are grouped into subdirectories in groups of 50 (e.g., `0400/` contains V400–V449,
+`0450/` contains V450–V499). Create a new subdirectory when crossing a 50-migration boundary.
+
+## Step 2: Create the Numbered Migration File
+
+Create `src/main/resources/db/migration/NNNN/VNNN__Description.sql` where `NNNN` is the
+directory (multiple of 50 rounded down) and `NNN` is the migration number.
+
+Make sure and add a new line at the end of the file.
```
sgrimm: Nit: "new line" should be "newline" since it refers to a character

.claude/skills/db-migration/SKILL.md
```diff
@@ -0,0 +1,328 @@
...
+
+Run `./gradlew testClasses` immediately after creating the migration to catch SQL errors early.
```
sgrimm: Could run `./gradlew generateJooqClasses` instead (takes less time to run if successful).

sgrimm (approved):
Sure, can't hurt to give this a spin. I am skeptical that it'll be faster/easier to describe a specific table structure to Claude than to describe it by writing the SQL, but only one way to find out.

.claude/skills/db-migration/SKILL.md
ice1080: To give more context on this PR (I should have put this in the description originally), I actually had claude code create the original version of this skill, giving it detailed explanations of what to do. It originally used the same "Copy this checklist" with "Task Progress" format. I then used it to create #4041 , just to see how it'd do. It missed a few steps, and when I asked why it skipped them, it told me that it only created one `TaskCreate` for the whole task, instead of creating one for each in the list. I told it that it should treat each of these like tasks that it _needs_ to do every time and just mark completed if not applicable, so it modified the Instructions to add the `TaskCreate` specification. I can't seem to find much description on this from some initial searching, outside of this which isn't part of anthropic:  
https://github.com/Piebald-AI/claude-code-system-prompts/blob/main/system-prompts/tool-description-taskcreate.md.   
I haven't used the new description on another migration yet (was just about to) to see if it does any better with this.

ice1080:
> Sure, can't hurt to give this a spin. I am skeptical that it'll be faster/easier to describe a specific table structure to Claude than to describe it by writing the SQL, but only one way to find out.

​I oftentimes map out my table structure in my notes before writing anything, to make sure I'm not missing anything. While it probably takes about the same amount of time as writing the SQL manually, I've found that I'm faster at telling it what to do and doing something else while it runs in the background, and then review at the end of things (obviously depends on the size of the task/migration). If no one else ends up using this, that's totally fine and I'll at least have learned more about writing skills.

.claude/skills/db-migration/SKILL.md
ice1080: I'll make it a bit more clear, though that is what the last sentence is (which matches the `adding-persistent-event` skill) (and actually our `adding-localizable-strings` skill just describes what it dose; we should probably update that)

sgrimm (approved):

PR #4063, opened 2026-03-04
SW-5441 Add destination batch project ID search field

sgrimm:
In the withdrawal history in the UI, for nursery transfers where the destination
batch is set to a different project than the originating batch, we want to show
both project names rather than just the originating batch's project name.

Ideally this would be purely a client-side change: add the search field for the
destination batch's project name to the query. But that doesn't work because it
would cause the query to reference the `batches` search table twice with different
join criteria, which the search API doesn't currently support.

Instead, solve it by adding a use-case-specific field to the `batchWithdrawals`
search table that fetches the destination batch's project name using a subquery.
The frontend code can then add the new field to its query without causing a
double join.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/search/table/BatchWithdrawalsTable.kt
+++ b/src/main/kotlin/com/terraformation/backend/search/table/BatchWithdrawalsTable.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.search.table
 
+import com.terraformation.backend.db.default_schema.tables.references.PROJECTS
 import com.terraformation.backend.db.nursery.tables.references.BATCHES
 import com.terraformation.backend.db.nursery.tables.references.BATCH_WITHDRAWALS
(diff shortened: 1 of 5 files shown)


sgrimm:
* **#4063** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4063?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4063?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4063).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ice1080 (approved):

PR #4080, opened 2026-03-06
SW-8146 Fix indicators showing up multiple times in reports

ice1080:
Indicator targets were joined without filtering by project and year, and we're instead filtered by those in the where clauses. Move these to the join itself (this was already fixed for published reports during the PR, but was not tested thoroughly there either).


sgrimm (approved):
That'd do it!

PR #4083, opened 2026-03-06
SW-8093 Add current year progress to published report model and api

ice1080:
Retrieve the current year's progress list for published reports. For the current quarter, use the published value. For other quarters, use unpublished values. Exclude future quarters.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
@@ -135,7 +135,7 @@ data class PublishedReportIndicatorPayload(
     val classId: IndicatorClass?,
     @Schema(
         description =
-            "If the indicator is cumulative, the list of actual values for all quarters in the" +
+            "If the indicator is cumulative, the list of actual values for all quarters in the " +
                 "report's year. Note that only the report's quarter will be a published value, the " +
                 "rest will be current values whether or not they are the same as their published " +
                 "counterparts."
--- a/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
@@ -21,6 +21,7 @@ import com.terraformation.backend.db.default_schema.ProjectId
 import com.terraformation.backend.db.default_schema.UserId
 import com.terraformation.backend.funder.PublishedReportService
 import com.terraformation.backend.funder.db.PublishedReportStore
+import com.terraformation.backend.funder.model.PublishedCumulativeIndicatorProgressModel
 import com.terraformation.backend.funder.model.PublishedReportIndicatorModel
 import com.terraformation.backend.funder.model.PublishedReportModel
 import io.swagger.v3.oas.annotations.Operation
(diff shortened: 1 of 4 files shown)


ice1080:
* **#4085** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4085?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4084** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4084?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4083** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4083?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4083).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/test/kotlin/com/terraformation/backend/funder/db/PublishedReportStoreTest.kt
```diff
@@ -620,6 +629,250 @@ class PublishedReportStoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      )
+      assertEquals(
+          listOf(
+              PublishedCumulativeIndicatorProgressModel(quarter = ReportQuarter.Q1, value = 50),
+              PublishedCumulativeIndicatorProgressModel(quarter = ReportQuarter.Q2, value = 60),
+          ),
+          indicator.currentYearProgress,
+          "currentYearProgress is populated regardless of indicator class; consumers decide whether to use it",
+      )
+    }
```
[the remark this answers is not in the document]
ice1080: This is correct, but something I also missed in ReportStore. I'll fix it in a separate PR for both

ice1080:
Turns out I messed this one up, as the progress should return the published value for the current quarter and unpublished values for previous quarters (I implemented as all unpublished). Changing back to draft to fix.

src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
```diff
@@ -120,6 +133,14 @@ data class PublishedReportIndicatorPayload(
     val baseline: BigDecimal?,
     val category: IndicatorCategory,
     val classId: IndicatorClass?,
+    @Schema(
+        description =
+            "If the indicator is cumulative, the list of actual values for all quarters in the" +
```
sgrimm: This string fragment is missing a trailing space.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/funder/api/FunderReportsController.kt
ice1080: Thanks, IntelliJ in conjunction with spotless sometimes causes this to happen when commenting, saving (which runs spotless in my case) and uncommenting. Will fix up

ice1080:
Re-requesting review since the changes were pretty large

sgrimm (approved):
This is growing into a pretty sizable bit of SQL querying. Hopefully it'll stay manageable.

ice1080:
I feel like that often happens when we use multisets instead of separate queries. This one isn't as complex as observations, so should be easier to move to separate (simpler) queries if we decide. Though in that case you end up with with more complex logic to handle situations of querying multiple reports. 

The other comment I'll mention is that `PublishedReportStore` uses generics to reduce code which worked well initially but might be getting more complex than its benefit. `ReportStore` has 3 separate methods for the indicator types, but then you have a lot of nearly identical code duplication that differs in jOOQ tables and columns.

If there's a way to tell jOOQ that 2 or 3 tables implement the same interface, that could simplify things, though I haven't seen an example of that yet.

PR #4161, opened 2026-03-18
SW-8202 Build terraware-server on Buildkite

sgrimm:
Port the GitHub Actions workflows to Buildkite. This is a fairly straight
conversion and doesn't attempt to do things like parallelize build steps
for performance; in fact, it actually does _less_ parallel execution than
the GHA workflows because the scripts and docs builds are treated as steps
in the main pipeline rather than as separate workflows.
    
Some GitHub Actions plugins have Buildkite equivalents, but some don't;
the latter are turned into shell scripts here.


sgrimm:
* **#4161** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4161).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

sgrimm:
This is mostly machine-converted from the GHA workflows so far; I'll be updating it as I test it out, so no need to review yet.

sgrimm:
When this is merged, we'll need to update the repo settings to expect a build result from Buildkite rather than from GHA, but I'm leaving it alone for now since changing it will affect the GHA builds.

bhamiltoncx:
I love this!

When you say the GHA pipelines were machine-converted, does that include the `bash` scripts? Or were those written specifically for Buildkite?

IMHO, the amount of footguns in `bash` syntax is worth considering avoiding it if we don't have to use it as a scripting language. (I say that as someone who's been writing `bash` scripts since the 1990s..)

Python and TypeScript are both pretty good alternatives in this day and age.

sgrimm:
> When you say the GHA pipelines were machine-converted, does that include the bash scripts? Or were those written specifically for Buildkite?

They're a combination. Claude Code did a first cut of the conversion and it created several of the bash scripts, mostly the ones that replace GHA plugins. But it didn't move the individual build commands into scripts; it left them in the pipeline (a pipeline step can have a list of commands). I moved everything into scripts because scripts give us the ability to define control flow and error handling.

As for bash vs. alternatives, I agree bash isn't ideal as a programming language generally. But a lot of the scripts are just "run this command, then run that command, abort if anything fails" rather than more sophisticated error recovery and such, so I don't think most of the classic bash footguns are going to be big risks for us. And running sequences of commands is much simpler in bash than in something like Python, where every

    var=$(command --arg --arg)

turns into something like

    result = subprocess.run(["command", "--arg", "--arg"], capture_output=True, text=True, check=True)
    var = result.stdout.strip()

Obviously you could reduce that clutter with helper functions to some extent.

I'm not 100% opposed to using something other than bash. I didn't do it here because it would have added even more incidental complexity and made the pipeline harder for people to read through.

bhamiltoncx:
For sure, I don't feel strongly. But it often turns out folks copy-paste what's already there, so it just depends on what outcome we want.

sgrimm:
One other thing to be aware of when reviewing this: for obvious reasons, I didn't fully test the steps that only run on release tags or when the branch is `main`. I tested _some_ of that code, but it's possible the first time we do a release with this setup, it'll fail and I'll have to fix something.

ice1080:
> When this is merged, we'll need to update the repo settings to expect a build result from Buildkite rather than from GHA, but I'm leaving it alone for now since changing it will affect the GHA builds.

As a note it's possible to have both for a time if we'd prefer that, and we can make GHA be gating and Buildkite not, and vice versa. My last company had PRs with Jenkins checks and GHA checks and a mix of both were required

.buildkite/scripts/build-and-test.sh
```diff
@@ -0,0 +1,50 @@
...
+    rm -rf "$GRADLE_USER_HOME/caches/build-cache-1"
+fi
+
+./gradlew --build-cache downloadDependencies yarn
+
+echo "--- :gradle: Generate jOOQ classes"
+./gradlew --build-cache generateJooqClasses
+
+echo "--- :gradle: Check code style"
+./gradlew --build-cache spotlessCheck
```
ice1080: Looks like we could also put `org.gradle.caching=true` in gradle.properties so we don't have to pass `--build-cache` to every gradle command, if preferred

.buildkite/scripts/run-external-tests.sh
```diff
@@ -0,0 +1,17 @@
...
+# Some secrets need to be exposed with different names than the values they're stored under
+# in the Secrets Manager secret.
+
+export TERRAWARE_ATLASSIAN_ACCOUNT="$TEST_ATLASSIAN_ACCOUNT"
+export TERRAWARE_ATLASSIAN_APIHOST="$TEST_ATLASSIAN_HOST"
+export TERRAWARE_ATLASSIAN_APITOKEN="$TEST_ATLASSIAN_APITOKEN"
+export TERRAWARE_ATLASSIAN_SERVICEDESKKEY="$TEST_ATLASSIAN_SERVICE_DESK_KEY"
+export TERRAWARE_MAPBOX_APITOKEN="$TEST_MAPBOX_APITOKEN"
+
+./gradlew --build-cache -x generateJooqClasses test --tests='*ExternalTest'
```
ice1080: Nice addition of `-x generateJooqClasses` to save time. Do we want that in the `build-and-test` line for running the unit tests too, or is it cached because it's in the same script?

sgrimm:
Sure, if we want both we can use both, but in this PR I'm deleting the GHA workflows so the build never happens (that's why it still shows up as a pending check in the PR status).

.buildkite/scripts/notify-pr.sh
```diff
@@ -0,0 +1,19 @@
...
+    exit 0
+fi
+
+echo "Posting comment on PR #${MERGED_PR_NUMBER}"
+
+curl -s -X POST \
+    -H "Authorization: token ${GITHUB_TOKEN}" \
+    -H "Accept: application/vnd.github.v3+json" \
+    "https://api.github.com/repos/terraware/terraware-server/issues/${MERGED_PR_NUMBER}/comments" \
+    -d '{"body": "Staging deployment complete."}'
```
ice1080, line 19: I wonder if we'll eventually want to create our own plugin for things like this, where GHA had one before and we might reuse it across a few repos. Idk how easy that is though (or if you could define it for the whole org?)

.buildkite/scripts/deploy-docs.sh
ice1080, line 1: Do we not want to use the `gitlog.sh` in here anymore? (I recognize it doesn't seem very useful, just checking)

ice1080 (approved):
A few questions but looks good overall

.buildkite/scripts/build-and-test.sh
sgrimm: I was a little hesitant to do that just because I didn't want to affect builds in local dev environments, but you're right, that would be better than passing the flag everywhere.

.buildkite/scripts/deploy-docs.sh
sgrimm: Lines 26-31 are that script, but with an extra fetch which is needed because we're doing a shallow clone of the repo.

.buildkite/scripts/notify-pr.sh
sgrimm: Creating our own plugins is definitely on the table. My thought was to wait and see what it looks like once people have ported a few more of our builds over; at that point we'll have a good idea of where the repetition is annoying. Some GHA plugins end up being just a couple of shell commands, so we'd probably just keep those inline.

.buildkite/scripts/run-external-tests.sh
sgrimm: It's cached in that case because it's using the same Gradle daemon. In theory it _should_ be cached here too since we're restoring the build cache, but it looks like there's a problem with the jOOQ Gradle plugin we're using, where it doesn't specify its inputs in a way that plays nice with the build cache.

.buildkite/scripts/build-and-test.sh
ice1080: Ya I was partially wondering if it'd be worth doing something like this somewhere in the build scripts, but obviously that's a little messy
```sh
echo "org.gradle.caching=true" >> gradle.properties
```

PR #4215, opened 2026-04-01
Buildkite: Use ECR credential helper to avoid sending Docker credentials in plaintext

bhamiltoncx:
This PR migrates off sending ECR credentials in the command line and onto the ECR credentials helper plugin.

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -37,3 +37,6 @@ export GRADLE_USER_HOME
 export IS_CD
 export MERGED_PR_NUMBER
 export TIER
+
+# Install the ECR credential helper so the ecr plugin can configure it.
+.buildkite/scripts/install-deps.sh --ecr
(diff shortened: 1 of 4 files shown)


sgrimm:
FYI, this repo is public, so you might want to avoid linking to private repos in PR descriptions or comments.

bhamiltoncx:
> FYI, this repo is public, so you might want to avoid linking to private repos in PR descriptions or comments.

Ahh, I keep forgetting that. Thanks, cleaned them up.

sgrimm:
Oh, wait a minute, we don't push terraware-server to ECR, so I think this won't help anyway. The `docker login` in terraware-server is logging into Docker Hub, not ECR. Sorry I spaced on that when you were asking me if it made sense to port this over.

bhamiltoncx:
Oh! Yeah, no need if we don't use ECR here.

OK, the equivalent would be [a credential helper](https://docs.docker.com/reference/cli/docker/login/#credential-helpers), but fixing that is probably fairly low-pri.

PR #4218, opened 2026-04-02
SW-8233 In Settings, show plots that have not yet been observed

constanzauanini:


src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -85,6 +86,45 @@ class T0Store(
...
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    val observed =
+        DSL.exists(
+            DSL.selectOne()
+                .from(OBSERVATION_PLOTS)
+                .join(OBSERVATIONS)
+                .on(OBSERVATIONS.ID.eq(OBSERVATION_PLOTS.OBSERVATION_ID))
+                .where(OBSERVATION_PLOTS.MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID))
+                .and(completedObservationsCondition)
```
constanzauanini, line 102: Not sure if this completed observation condition is needed here?

ice1080: `completedObservationsCondition` ensures that the `observation_plots` row for the plot isn't for a future or incomplete observation. The plot won't be `observed` unless it is in a completed observation. 

However I was just looking through the code and data model, and I think it's possible for a plot to have an `observation_plots` row in a completed observation, but still have never been observed. I think we need to add another check for `OBSERVATION_PLOTS.STATUS_ID.eq(ObservationPlotStatus.Completed)`.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -85,6 +86,45 @@ class T0Store(
         (!survivalRateIncludesTempPlots(plantingSiteId) || isAllTempT0DataSet(plantingSiteId))
   }
 
+  fun fetchMonitoringPlotsT0Status(
+      plantingSiteId: PlantingSiteId
+  ): List<MonitoringPlotT0StatusModel> {
```
ice1080, line 92: We'll need unit tests for this method as well before it can be merged

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -85,6 +86,45 @@ class T0Store(
...
+                .where(OBSERVATION_PLOTS.MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID))
+                .and(completedObservationsCondition)
+        )
+
+    val t0set =
+        DSL.exists(
+            DSL.selectOne()
+                .from(PLOT_T0_DENSITIES)
+                .where(PLOT_T0_DENSITIES.MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID))
+        )
```
ice1080, line 121: I think for `t0set` to be true, it can't just have one row in `plot_t0_densities`; I believe it will need to have a row for every species that has been withdrawn to that plot's substratum

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -85,6 +86,45 @@ class T0Store(
...
+  ): List<MonitoringPlotT0StatusModel> {
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    val observed =
+        DSL.exists(
+            DSL.selectOne()
+                .from(OBSERVATION_PLOTS)
+                .join(OBSERVATIONS)
+                .on(OBSERVATIONS.ID.eq(OBSERVATION_PLOTS.OBSERVATION_ID))
+                .where(OBSERVATION_PLOTS.MONITORING_PLOT_ID.eq(MONITORING_PLOTS.ID))
```
ice1080, line 101: I'm trying to figure out if we also need `.and(OBSERVATION_PLOTS.IS_PERMANENT.eq(true))` to ensure that the plot wasn't observed when it was temporary. That might not matter though 🤔  I'll get back to you on this

ice1080: Question asked in Jira

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
constanzauanini: If a plot doesn't have withdrawn species, would that be set = true or set = false?

src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
```diff
@@ -1594,6 +1595,154 @@ internal class T0StoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+
+      val results = store.fetchMonitoringPlotsT0Status(plantingSiteId)
+      val resultIds = results.map { it.monitoringPlotId }
+
+      assertTrue(resultIds.contains(monitoringPlotId), "permanent plot should be in results")
+      assertFalse(
+          resultIds.contains(tempPlotId),
+          "temp plot (no permanentIndex) should be excluded",
+      )
+      assertFalse(resultIds.contains(adHocPlotId), "ad-hoc plot should be excluded")
```
ice1080: I've had a hard time getting claude to correctly create tests the right way the first time. Instead of multiple asserts on the same object, we usually create an expected object and do an `assertEquals` on the `store.fetch...`. You should be able to tell the agent to use `assertEquals` with an expected object and it will fix these to match (there are others like this in the same file)

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: Good question. I think that it then needs to have t0 set for all species that have been observed (even if it's set to 0). We might want to instead refactor the `permanentPlotSpecies` that is inside `isAllPermT0DataSet` to a separate function, because it already finds all plots with observations and their species that have been withdrawn, unions species that have been recorded, and unions species that have densities added. Perhaps we'd be able to pass in a boolean arg to optionally skip the observations check because I don't think we want it for this use case

ice1080: To answer your question more directly, if a plot doesn't have withdrawn species but t0 is set for all species observed, then set = true, otherwise set = false.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
ice1080: Ok sounds like we don't need that line, so marking this comment as resolved

ice1080 (approved):

PR #4232, opened 2026-04-08
SW-8233 Add substratumObserved to plots in allSet endpoint

constanzauanini:


src/test/kotlin/com/terraformation/backend/tracking/db/T0StoreTest.kt
```diff
@@ -1747,10 +1794,81 @@ internal class T0StoreTest : DatabaseTest(), RunsAsDatabaseUser {
...
+          substratumId = substratumId,
+      )
+
+      assertEquals(
+          true,
+          store
+              .fetchMonitoringPlotsT0Status(plantingSiteId)
+              .single { it.monitoringPlotId == monitoringPlotId }
+              .substratumObserved,
+      )
```
ice1080, line 1831: These need updated again to expected models instead of just checking the flag

ice1080 (approved):

PR #4299, opened 2026-04-20
SW-8276 Added simplified sites and site histories table and upserts on creation and updates

tommylau523:
This PR added the new tables to store simplified versions of planting sites/strata/substrata, as well as implemented the trigger to generate them on create/edit.  
  
The trigger is placed after planting site history creation at any point.


tommylau523:
* **#4302** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4301** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4301?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4299** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4299?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4299?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4299).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/main/kotlin/com/terraformation/backend/util/GeometrySimplifier.kt
```diff
@@ -11,6 +12,7 @@ import org.locationtech.jts.simplify.TopologyPreservingSimplifier
  * [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm)
  * method.
  */
+@Named
```
ice1080: I don't think I agree that this should be an injected class. It's methods are fully static. It seems like we should update this to use the companion object so that we can call it with something like `GeometrySimplifier.simplify` instead of injecting it and calling it with `geometrySimplifier.simplify`

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
```diff
@@ -639,11 +652,76 @@ class PlantingSiteStore(
...
+          }
+        }
+      }
+    }
+  }
+
+  fun upsertSimplifiedPlantingSiteHistory(
+      plantingSiteId: PlantingSiteId,
+      plantingSiteHistoryId: PlantingSiteHistoryId,
+  ) {
```
ice1080, line 693: Is it possible to ever update the boundary of a `planting_site_history` row, or will this one always be an insert? (same comment for other history tables)

ice1080 (commented):
Next time this could be split up into the db work and the app work to reduce the PR size. I had to scroll through most of it to find the db-specific files/changes before I could start looking at the app changes. 
I'm only partway through reviewing but wanted to leave some feedback before getting back to this later

tommylau523:
> Next time this could be split up into the db work and the app work to reduce the PR size. I had to scroll through most of it to find the db-specific files/changes before I could start looking at the app changes.
> I'm only partway through reviewing but wanted to leave some feedback before getting back to this later

​Will do. Sorry about that!

src/main/kotlin/com/terraformation/backend/util/GeometrySimplifier.kt
tommylau523: This is mainly useful for mocking and testing. Although I could move it be a function paramter instead. Then I can add a separate interface and inject a mocked version of it for testing.

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteStore.kt
tommylau523: In theory, a history table should only be inserted. I think there may be a scenario where we update/tune the simplification process and therefore need to backfill histories. This gives us that flexibility.

ice1080 (approved):
The rest looks good to me. I still think we might be better off without injecting a static class, but that's a preference thing so I'll leave it up to you

src/main/kotlin/com/terraformation/backend/util/GeometrySimplifier.kt
tommylau523: @ice1080 I was able to setup mocking a static method! It now doesn't require an injection.

ice1080 (approved):

PR #4310, opened 2026-04-24
Update defaults for feature extraction/matcher

alexTerraformation:
Switch to lightglue/aliked by default at 5 fps extracted

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/admin/AdminSplatsController.kt
+++ b/src/main/kotlin/com/terraformation/backend/admin/AdminSplatsController.kt
@@ -33,6 +33,12 @@ class AdminSplatsController(
     if (!model.containsAttribute("stepArgs")) {
       model.addAttribute("stepArgs", emptyMap<String, String>())
     }
+    if (!model.containsAttribute("featureMatcherSubcommand")) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/terraformation/backend/admin/AdminSplatsController.kt
ice1080, line 1: Don't we want to do this at the splatter level so that all jobs receive these defaults?

ice1080 (approved):
Approving, but I think we want this to be defaults in splatter instead of admin

src/main/kotlin/com/terraformation/backend/admin/AdminSplatsController.kt
alexTerraformation: The changes in this file are just cosmetic to pre-fill the form. Would you prefer me to remove them?

ice1080: Ah gotcha, I thought you had said yesterday that we'd be changing the defaults for all jobs. Go ahead and merge whenever

alexTerraformation: That's fair, I can make the change in splatter instead.

PR #4318, opened 2026-04-27
Ensure files for splats are also observation media files

tommylau523:
This PR guarantees that an observation file used to generate splats is also an organization media file. This is to support organization media files APIs and features for splats are available.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
+++ b/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
@@ -36,6 +36,7 @@ import java.time.InstantSource
 import kotlin.io.path.Path
 import kotlin.io.path.pathString
 import org.jooq.DSLContext
+import org.jooq.impl.DSL
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.context.event.EventListener
@@ -142,6 +143,9 @@ class SplatService(
     val organizationId =
         parentStore.getOrganizationId(observationId)
             ?: throw ObservationNotFoundException(observationId)
+
+    // Ensure that an observation media file used for splats is also an organization media file
+    insertOrganizationMediaFileFromObservationFile(observationId, fileId)
 
     generateSplat(organizationId, fileId, force, params, runBirdnet)
   }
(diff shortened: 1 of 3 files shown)


tommylau523:
* **#4318** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4318?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4318?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4318).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -715,6 +719,41 @@ class SplatService(
...
+    val organizationId =
+        parentStore.getOrganizationId(observationId)
+            ?: throw ObservationNotFoundException(observationId)
+
+    requirePermissions {
+      readObservation(observationId)
+      createOrganizationMedia(organizationId)
+    }
+
+    ensureObservationFile(observationId, fileId)
```
ice1080, line 735: I think we've already done this check on line 141

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -715,6 +719,41 @@ class SplatService(
...
+
+    with(ORGANIZATION_MEDIA_FILES) {
```
ice1080, line 737: I think my only concern with inserting this is if we use this for anything else in the future; it could cause confusion if observation media are in a list of org media

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -715,6 +719,41 @@ class SplatService(
     }
   }
 
+  private fun insertOrganizationMediaFileFromObservationFile(
+      observationId: ObservationId,
+      fileId: FileId,
+  ) {
+    val organizationId =
+        parentStore.getOrganizationId(observationId)
+            ?: throw ObservationNotFoundException(observationId)
```
ice1080, line 728: I think we could require organizationId in the args since it's already been retrieved on line 143

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -715,6 +719,41 @@ class SplatService(
...
+    requirePermissions {
+      readObservation(observationId)
+      createOrganizationMedia(organizationId)
+    }
+
+    ensureObservationFile(observationId, fileId)
+
+    with(ORGANIZATION_MEDIA_FILES) {
+      dslContext
+          .insertInto(ORGANIZATION_MEDIA_FILES, ORGANIZATION_ID, FILE_ID, CAPTION)
```
ice1080, line 739: What happens if the caption gets updated on one of the rows? There's nothing keeping these in sync and it might not be obvious which one to use when querying one or the other

tommylau523: I think this points to us needing more thinking on this. Maybe caption should live in the files table. I am not really sure for the time being.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
tommylau523: This is definitely a valid concern. Perhaps it is best to have two sets of functionalities for now, one for observation splats, and another for organization splats. I can expose the observation ones in a new controller instead.

ice1080: Don't we have that with `OrganizationSplatsController` and `SplatsController` (the latter being the one for observations)?

PR #4404, opened 2026-05-18
Upgrade ktor to v3.5.0

sgrimm:
This requires upgrading the Kotlin coroutines library to a newer version than the
one that's specified by Spring Boot.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -136,6 +136,8 @@ dependencies {
   implementation("org.geotools:gt-xml:$geoToolsVersion")
   implementation("org.geotools.xsd:gt-xsd-core:$geoToolsVersion")
   implementation("org.geotools.xsd:gt-xsd-kml:$geoToolsVersion")
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.11.0")
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.11.0")
   implementation("org.jobrunr:jobrunr-spring-boot-3-starter:8.6.0")
   implementation("org.jooq:jooq:$jooqVersion")
   implementation("org.locationtech.jts:jts-core:$jtsVersion")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -56,6 +56,9 @@ group = "com.terraformation"
 
 version = computeGitVersion("0.1")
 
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#4404** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/terraware?ref=merge-queue-instructions-in-stack-comment&prId=4404).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

build.gradle.kts
```diff
@@ -136,6 +136,8 @@ dependencies {
   implementation("org.geotools:gt-xml:$geoToolsVersion")
   implementation("org.geotools.xsd:gt-xsd-core:$geoToolsVersion")
   implementation("org.geotools.xsd:gt-xsd-kml:$geoToolsVersion")
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.11.0")
+  implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.11.0")
```
ice1080: This might be more of a personal preference, but sometimes maintenance of dependencies gets harder when there's no distinction between dependencies actually required and dependencies only added for version resolution, especially as the years add up. I know that in `package.json` , `npm` has a separate section for `dependencies` and `resolutions`. Does gradle have something like this? If not, do you think it'd be worth adding a comment that these 2 are only for version resolution?

sgrimm: Good call. There is indeed a way to be more explicit about this.

sgrimm: This led me on a little detour. Gradle has a built-in way to do transitive dependency version overrides (the `constraints` block) but Spring Boot declares the dependency in a way that takes precedence over that override mechanism, so we need to use Spring Boot's configuration feature instead.

sgrimm: This isn't the first time I've added an explicit dependency just to do a version override. At some point it's probably worth going through the dependencies and figuring out which ones are override-only and marking them as such.

ice1080 (approved):

PR #4463, opened 2026-05-27
SW-8399 Run survival rate recalculation asynchronously after species totals edits

tommylau523:
updateMonitoringSpecies now publishes MonitoringSpeciesTotalsEditedEvent
without inline survival-rate recomputation. A new @EventListener in
ObservationStore enqueues runRecalculateSurvivalRates for the affected
plot via JobRunr. Totals propagation from OBSERVED_*_SPECIES_TOTALS into
OBSERVATION_*_RESULTS stays synchronous via updateObservationResults so
the async recompute reads fresh numerators; the same split is applied to
the later-observations loop.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #4458, opened 2026-05-26
SW-8399 Run T0 survival rate recalculations asynchronously via JobRunr

tommylau523:
The @EventListener handlers for T0PlotDataAssignedEvent and
T0StratumDataAssignedEvent previously ran recalculateSurvivalRates
synchronously inside the T0 assignment transaction, making the user
wait. They now enqueue a JobRunr job so the request returns immediately.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #4464, opened 2026-05-27
SW-8399 Track survival rate recalculation status per planting site

tommylau523:
T0 writes upsert a row in `tracking.planting_site_survival_rate_recalculations`
with `last_modified_time` inside the existing transaction. Async recalc jobs
snapshot that timestamp, run the recalc, and only set last_recalculated_time
if no newer write arrived in the meantime, so rapid-fire writes can't leave
the site falsely marked as clean. The site-level T0 read response exposes a
new survivalRateRecalculationInProgress boolean derived from the two
timestamps.

Co-Authored-By: Claude Opus 4.7 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)


tommylau523:
* **#4464** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4464?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4464?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4463** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4463?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4525](https://github.com/terraware/terraware-server/pull/4525) <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4525?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#4458** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/model/ObservationResultsModel.kt
```diff
@@ -242,6 +242,11 @@ data class ObservationResultsModel(
     val state: ObservationState,
     val strata: List<ObservationStratumResultsModel>,
     override val survivalRate: Int?,
+    /**
+     * True if a survival rate recalculation is pending or in progress for this observation's
+     * planting site. Defaults to `false`; only the V2 results store populates it.
+     */
```
ice1080, line 248: Not sure this comment is needed, and it will fall out of date as soon as we remove the v1 results store

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -54,6 +55,60 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
     return fetchByCondition(OBSERVATIONS.ID.eq(observationId), depth, 1).first()
   }
 
+  /**
+   * Returns true if a survival rate recalculation is pending or in progress for the given planting
+   * site, based on the timestamps tracked in [PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS].
+   */
```
ice1080, line 61: Not sure the comments are necessary on these methods, and they usually fall out of date pretty quick

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
ice1080, line 1: These changes don't seem to be unit tested anywhere

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -54,6 +55,60 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
...
+                          PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS.LAST_RECALCULATED_TIME.isNull
+                              .and(
+                                  PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS.LAST_T0_MODIFIED_TIME
+                                      .isNotNull
+                                      .or(
+                                          PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS
+                                              .LAST_OBSERVATION_MODIFIED_TIME
+                                              .isNotNull
+                                      )
+                              ),
```
ice1080, line 99: Does this also need to check if the `LAST_T0_MODIFIED_TIME` is null and the other 2 are not null, as well as if `LAST_OBSERVATION_MODIFIED_TIME` is null and the other 2 are not null?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -54,6 +55,60 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
...
+   * Returns true if a survival rate recalculation is pending or in progress for the given planting
+   * site, based on the timestamps tracked in [PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS].
+   */
+  fun isSurvivalRateRecalculationInProgress(plantingSiteId: PlantingSiteId): Boolean {
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    return dslContext
+        .select(survivalRateCalculationInProgressField(DSL.value(plantingSiteId)))
+        .fetchOne()
+        ?.value1() == true
```
sgrimm, line 68: Could use `fetchExists` here.

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -462,6 +465,8 @@ class T0Store(
             .set(MODIFIED_TIME, now)
             .execute()
 
+        markRecalculationRequestedForPlot(monitoringPlotId, now)
+
         eventPublisher.publishEvent(
             T0PlotDataAssignedEvent(
                 monitoringPlotId = monitoringPlotId,
```
ice1080, line 472: I think there's technically a chance that this introduces a race condition, since the transaction might take a long time to commit but the async recalculation has already been requested. I'm not sure the best way to handle this

src/main/kotlin/com/terraformation/backend/tracking/db/T0Store.kt
```diff
@@ -701,6 +710,37 @@ class T0Store(
         .execute()
   }
 
+  private fun markRecalculationRequestedForPlot(monitoringPlotId: MonitoringPlotId, now: Instant) {
+    val plantingSiteId =
+        dslContext
+            .select(MONITORING_PLOTS.PLANTING_SITE_ID)
+            .from(MONITORING_PLOTS)
+            .where(MONITORING_PLOTS.ID.eq(monitoringPlotId))
+            .fetchOne(MONITORING_PLOTS.PLANTING_SITE_ID.asNonNullable()) ?: return
```
ice1080, line 719: Could probably replace this and the stratum version with `parentStore`

src/test/kotlin/com/terraformation/backend/tracking/db/observationStore/ObservationStoreRecalculationStatusTest.kt
```diff
@@ -0,0 +1,137 @@
...
+    dslContext
+        .insertInto(PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS)
+        .set(PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS.PLANTING_SITE_ID, plantingSiteId)
+        .set(PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS.LAST_T0_MODIFIED_TIME, t0ModifiedTime)
+        .set(
+            PLANTING_SITE_SURVIVAL_RATE_RECALCULATIONS.LAST_OBSERVATION_MODIFIED_TIME,
+            observationModifiedTime,
+        )
+        .execute()
+  }
```
ice1080, line 116: Worth moving to `DatabaseBackedTest`?

sgrimm (changes requested):
With this setup, rapid writes would still potentially run the recalculation multiple times in parallel, right? It just wouldn't save the results?

This seems more complicated than it needs to be. Here's a simpler approach that would avoid parallel calculations:

1. Add an "additional recalculation requested" column to the recalculation status table, boolean, default false.
2. When a recalculation event comes in, insert a new row into the table with the new column set to false. On conflict, set the new column to true.
3. If you updated the new column to true, return from the event handler; the calculation will happen below.
4. Run the recalculation and save the results.
5. If the new column has been set to true while the calculation was running, set it to false and go to step 4.
6. Delete the recalculation status row.

Then the "is a recalculation active?" check is just a simple `SELECT EXISTS` on the recalculation table, and you'll never run the same calculation twice in parallel.

PR #4517, opened 2026-06-01
Support activity media files without captured times

sgrimm:
Currently we require activity media files to include captured times, which we
extract from EXIF metadata at upload time. But you can have valid images or videos
that lack that EXIF field; rather than bombing out with NPE, we should just return
a null captured date in the API responses.


sgrimm:
* **#4517** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):

sgrimm:
I confirmed that the API change here doesn't cause any type-checking errors in terraware-web.

PR #4523, opened 2026-06-02
SW-8427 Persistent events for scheduled planting dates

sgrimm:
Add a set of persistent event classes for operations on scheduled planting dates.


sgrimm:
* **#4524** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4524?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4523** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/eventlog/api/EventSubjectPayload.kt
```diff
@@ -375,6 +379,63 @@ data class PlantingSeasonSubjectPayload(
...
+      // list the "changed from" date so it shows up like "Scheduled date 2026-01-01 date changed to
+      // 2026-01-05." But for other events, we want whichever date was the active one before that
+      // event.
+      val date =
+          when (event) {
+            is PlantingSeasonScheduledDateCreatedEvent -> event.date
+
+            is PlantingSeasonScheduledDateUpdatedEvent if event.changedFrom.date != null -> {
+              event.changedFrom.date
+            }
```
ice1080: What's the benefit of using `event.changedFrom.date` instead of `event.date` for update events that don't change the date? The comments don't really explain the why

sgrimm: First answer is that there is no `event.date` for update events; it's only on creation events.

But the idea is that if an update event doesn't change the date , `changedFrom.date` will be null, so we need to keep looking back for an event that did change the date, or for the creation event if there were no date updates.

I guess it's worth noting that this is currently impossible since date is the only field and if it doesn't change, we won't publish the update event at all. I wrote this defensively so the logic won't break if we add more updatable fields later.

Or am I misunderstanding the question?

ice1080: No that answers the question. We might want to mention that in the comments, that only the created event contains the `date` field

ice1080 (approved):

sgrimm:
### Merge activity

* **Jun 2, 11:11 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4523).
* **Jun 2, 11:11 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4523).

PR #4525, opened 2026-06-02
SW-8399 Implemented a survival rate calculation status and rerun tracker

tommylau523:
A new row will be inserted to this table if a planting site has a survival rate recalculation run.
If more calculations are triggered (either via species edits or T0 updates), a flag will be set
to true on this row. Another calculation on all observations of the planting site will be triggered.

Co-authored by Claude to implement test cases on concurrency.


tommylau523:
* **#4525** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4525?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4525?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4463** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4463?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4464](https://github.com/terraware/terraware-server/pull/4464) <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4464?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#4458** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ice1080 (commented):
`runRecalculateSurvivalRates` for plot and stratum seem to start with calculating at only the plot level, and then if it's re-requested while still running, it reruns at the entire site level instead of just at the plot level again. If a plot calculation takes 30s and a request comes in again for the same plot, that means it will change to recalculating for the whole site which could take much longer than the plot, and seems like it would be unnecessary since only the plot needed recalculated

tommylau523:
> `runRecalculateSurvivalRates` for plot and stratum seem to start with calculating at only the plot level, and then if it's re-requested while still running, it reruns at the entire site level instead of just at the plot level again. If a plot calculation takes 30s and a request comes in again for the same plot, that means it will change to recalculating for the whole site which could take much longer than the plot, and seems like it would be unnecessary since only the plot needed recalculated

​A recalculation request may come in again due to edits in other plots/strata. The recalculation but have a different scope as a result. Our options here are either to implement a calculation queue and resolve them 1 by 1, or handle all potential edits at once.

src/main/resources/db/migration/0450/V496__SurvivalRateCalculations.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE tracking.planting_site_survival_rate_calculations
+(
+    planting_site_id BIGINT PRIMARY KEY REFERENCES tracking.planting_sites ON DELETE CASCADE,
+    additional_calculation_requested BOOLEAN NOT NULL DEFAULT FALSE
+);
```
ice1080, line 5: Why do we need a separate table instead of using the job scheduler to look at jobs that are processing, and scheduling a new one if one is currently in progress (and then setting up concurrency rules per planting site)?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -1902,6 +1903,50 @@ class ObservationStore(
     recalculateSurvivalRateResults(ObservationResultsSite(stratumId))
   }
 
+  fun recalculateSurvivalRates(plantingSiteId: PlantingSiteId) {
```
sgrimm, line 1906: A lot of of this method seems like it's the same as the `StratumId` version; could they be combined?

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2142,36 +2187,120 @@ class ObservationStore(
...
+      if (!calculationInProgress) {
+        insertSurvivalRateCalculation(plantingSiteId)
+        jobScheduler.enqueue<ObservationStore> {
+          // Do the survival rate recalculation asynchronously to avoid delays on a T0 settings
+          // change.
+          runRecalculateSurvivalRates(plantingSiteId, event.monitoringPlotId)
+        }
+      } else {
+        setSurvivalRateAdditionalCalculationRequested(plantingSiteId, true)
+      }
```
sgrimm: This queuing logic is repeated several times in the PR, with only the body of the queued job differing. Could we extract it into a helper function that takes a lambda argument? (Might have to make it an inline function so JobRunr can properly extract the variables.)

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -3073,6 +3202,59 @@ class ObservationStore(
...
+        )
+      }
+
+  private fun survivalRateAdditionalCalculationRequested(plantingSiteId: PlantingSiteId) =
+      with(PLANTING_SITE_SURVIVAL_RATE_CALCULATIONS) {
+        dslContext
+            .select(ADDITIONAL_CALCULATION_REQUESTED)
+            .from(this)
+            .where(PLANTING_SITE_ID.eq(plantingSiteId))
+            .fetchOne { it[ADDITIONAL_CALCULATION_REQUESTED] } ?: false
```
sgrimm: Could use `.fetchOne(ADDITIONAL_CALCULATION_REQUESTED)` instead of the lambda.

sgrimm (approved):
Logic seems correct. There are some places where we could reduce repetition; up to you whether to address them or not.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
tommylau523: Likely we can refactor the ResultsScope implementations themselves and that would simplify a few of these functions. Will likely improve efficiency as well since it will be a one-shot update instead of an N+1 one. But I will implement that in a followup PR.

tommylau523:
### Merge activity

* **Jun 4, 10:06 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4525).
* **Jun 4, 10:29 PM UTC**: [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4525) couldn't merge this PR because it had merge conflicts.
* **Jun 4, 10:51 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4525).

PR #4527, opened 2026-06-03
SW-8427 Publish events for scheduled date species

sgrimm:
When the list of species quantities for a scheduled planting date is updated,
publish events so the changes will appear in the change history for the planting
date.

This requires reworking some of the update logic; we need to distinguish when
species are created or deleted or updated so we can publish the correct events.
The "delete everything and insert the new list" logic is replaced by a diff-based
implementation.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonScheduledDatesStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonScheduledDatesStore.kt
@@ -3,19 +3,30 @@ package com.terraformation.backend.plantingmanagement.db
 import com.terraformation.backend.auth.currentUser
 import com.terraformation.backend.customer.db.ParentStore
 import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.asNonNullable
 import com.terraformation.backend.db.tracking.PlantingSeasonId
 import com.terraformation.backend.db.tracking.PlantingSeasonStatus
 import com.terraformation.backend.db.tracking.ScheduledPlantingDateId
+import com.terraformation.backend.db.tracking.SubstratumHistoryId
+import com.terraformation.backend.db.tracking.SubstratumId
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SEASONS
 import com.terraformation.backend.db.tracking.tables.references.SCHEDULED_PLANTING_DATES
 import com.terraformation.backend.db.tracking.tables.references.SCHEDULED_PLANTING_DATE_SPECIES
+import com.terraformation.backend.db.tracking.tables.references.STRATA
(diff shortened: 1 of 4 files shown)


sgrimm:
* **#4528** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4528?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4527** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4527?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4527?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4526** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sgrimm:
Sorry about the length of this one; splitting it up by operation would have meant we'd have a revision where only some of the events got published, which could have led to inconsistent event log entries.

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonScheduledDatesStore.kt
```diff
@@ -133,15 +166,19 @@ class PlantingSeasonScheduledDatesStore(
...
-    val oldModel = fetch(model.plantingSeasonId, scheduledDateId)
-    val plantingSiteId =
-        parentStore.getPlantingSiteId(model.plantingSeasonId)
-            ?: throw PlantingSeasonNotFoundException(model.plantingSeasonId)
-    val organizationId =
-        parentStore.getOrganizationId(plantingSiteId)
-            ?: throw PlantingSeasonNotFoundException(model.plantingSeasonId)
+    if (model.species.size != model.species.distinctBy { it.substratumId to it.speciesId }.size) {
+      throw IllegalArgumentException("Species listed multiple times for substratum")
+    }
```
ice1080: Would this make more sense in the model's initializer?

ice1080 (approved):
I'm somewhat surprised product didn't want any kind of combining for the species events, since creating a scheduled date with species for each substratum will cause there to potentially be 10s to 100s of change events at the same time.

sgrimm:
> I'm somewhat surprised product didn't want any kind of combining for the species events, since creating a scheduled date with species for each substratum will cause there to potentially be 10s to 100s of change events at the same time.

Seems likely to be something that gets requested during QA. I think it's best handled on the client side if we need to do it. In the event log I've tried to avoid events having lists of child objects since that makes it harder to set indexes up sensibly.

sgrimm:
### Merge activity

* **Jun 3, 8:59 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4527).

PR #4529, opened 2026-06-03
SW-8438 Data model for scheduled planting date requests

ice1080:
A Planting Date request represents a snapshot of a scheduled planting date with all quantities for the species x substrata combinations. Create tables to store this information, and add a column to `nursery.withdrawals` to eventually support fulfillment of these requests.

The status of these requests will get updated by applicaion events once withdrawals are used to fulfill requests.


ice1080:
* **#4536** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4536?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4532** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4532?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4531** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4531?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4530** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4530?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4529** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4529?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TABLE tracking.planting_date_request_statuses (
+    id INTEGER PRIMARY KEY,
+    name TEXT NOT NULL UNIQUE
+);
+
+CREATE TABLE tracking.planting_date_requests (
+    id BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
+    scheduled_planting_date_id BIGINT NOT NULL REFERENCES tracking.scheduled_planting_dates ON DELETE CASCADE,
+    date DATE NOT NULL,
```
sgrimm, line 8: This is copied from the scheduled planting date, I see in a later PR. Does this get updated if the date is changed?

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
```diff
@@ -0,0 +1,37 @@
...
+    created_by BIGINT NOT NULL REFERENCES users,
+    created_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    modified_by BIGINT NOT NULL REFERENCES users,
+    modified_time TIMESTAMP WITH TIME ZONE NOT NULL,
+    notes TEXT,
+    status_id INTEGER NOT NULL REFERENCES tracking.planting_date_request_statuses,
+
+    UNIQUE (scheduled_planting_date_id)
```
sgrimm: I think it'd be cleaner to use the scheduled planting date ID as the primary key rather than defining a new ID type; we never use the new ID for anything as far as I can tell.

sgrimm (changes requested):

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
ice1080: The `planting_date_request` is a snapshot of a `scheduled_planting_date`, but only when the user chooses to create a request when saving the `scheduled_planting_date`. If a user creates a `planting_date_request`, then later updates the date of the `scheduled_planting_date` and saves without creating a request, the two will intentionally be mismatched. The user can later choose to create a request and the date in the `planting_date_request` row will get updated at that time.

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
ice1080: The id is used in the sub species table, but if we're fine with that pointing to `scheduled_planting_dates.id` as well, I can update it

sgrimm: You'd want it to point to the `scheduled_planting_date_id` column on the requests table so deletes would cascade properly.

ice1080: Ah got it

ice1080: Are we ok with the nursery withdrawals having the column for the request that is pointing to `scheduled_planting_date_id`?

sgrimm: Yes, that seems fine.

One way to look at this is that because they both have unique constraints and are non-nullable, the two IDs are completely equivalent for any row in the requests table. Each of the newly-added IDs maps to exactly one scheduled date ID and vice versa. So switching to using one ID instead of the other is a semantic no-op.

ice1080: Ya I understood that from the time I wrote the migration. I think it mostly just felt weird to have a column on `nursery.withdrawals` that points to a table with a primary key referencing a different table. And as I type that, it's not even that concept that's throwing me off, but rather coming up with a name for the column. I'm leaning towards `scheduled_planting_date_request_id` because it will be of type `scheduled_planting_dates.id` but can only ever point to the `planting_date_requests` table

ice1080: All 4 PRs now updated with this change

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
```diff
@@ -0,0 +1,34 @@
...
+    quantity INTEGER NOT NULL CHECK (quantity >= 0),
+
+    PRIMARY KEY (scheduled_planting_date_id, substratum_id, species_id)
+);
+
+CREATE INDEX ON tracking.planting_date_request_species (substratum_id);
+CREATE INDEX ON tracking.planting_date_request_species (species_id);
+
+ALTER TABLE nursery.withdrawals
+    ADD COLUMN scheduled_planting_date_request_id BIGINT REFERENCES tracking.planting_date_requests (scheduled_planting_date_id) ON DELETE CASCADE;
```
sgrimm: I don't think we want to delete nursery withdrawals when scheduled planting dates are deleted; the inventory totals wouldn't add up correctly. `ON DELETE SET NULL` might be better here (or no `ON DELETE` at all and handle it in application code, if we want to undo the withdrawals).

sgrimm (changes requested):

src/main/resources/db/migration/0450/V495__ScheduledPlantingDateRequests.sql
ice1080: I'm not sure what I was thinking here. Obviously we wouldn't deletes of either item to affect the other. I think `SET NULL` makes sense.

sgrimm (approved):

ice1080:
### Merge activity

* **Jun 4, 4:51 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4529).
* **Jun 4, 4:51 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4529).

PR #4534, opened 2026-06-03
SW-8446 Use WGS84 spheroid for area calculation

sgrimm:
The function to calculate the area in hectares for a geometry projects the
geometry onto the UTM coordinate system, which uses a 1-meter scale. But that
coordinate system is designed to preserve the angles of the vertices of a
shape, not the total area of the shape.

Switch to the area calculator from the GeographicLib library, which takes the
curvature of the earth into account in its calculations. The library only supports
area calculations on simple polygons, so this requires some added logic to
break complex geometries up into their component polygons.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -127,6 +127,7 @@ dependencies {
   implementation("jakarta.inject:jakarta.inject-api:2.0.1")
   implementation("jakarta.ws.rs:jakarta.ws.rs-api:4.0.0")
   implementation("net.coobird:thumbnailator:0.4.21")
+  implementation("net.sf.geographiclib:GeographicLib-Java:2.1")
   implementation("org.apache.tika:tika-core:3.3.1")
   implementation("org.bouncycastle:bcpkix-jdk18on:1.84")
   implementation("org.commonmark:commonmark:0.28.0")
--- a/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
+++ b/src/main/kotlin/com/terraformation/backend/util/Extensions.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.util
 
+import com.terraformation.backend.db.SRID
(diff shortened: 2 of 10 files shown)


sgrimm:
* **#4534** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
Do we need to run a migration to recalculate all areas? And if so we probably need to update quite a bit of observation data too... (or maybe that only matters for area weighted survival rates; not sure what else it affects)

sgrimm:
Converting this to draft for now while awaiting GIS team feedback on the coordinate transformation question raised by Codex.

sgrimm:
To answer the "Do we need to recalculate all the areas?" question: yes, we do, but there's already a button for that in the admin UI, so no new migration code is needed. We should also recalculate all the area-weighted survival rates, which I think it's fine to do by rerunning the V487 migration manually.

ice1080:
> We should also recalculate all the area-weighted survival rates, which I think it's fine to do by rerunning the V487 migration manually.

There's also a recalc survival rates button in admin, though it's currently for an individual site and observation id. Wouldn't be hard to add the ability to rerun for all sites/observations, though if the migration is easier that's fine too

sgrimm:
After discussing more with the GIS people, we realized that their original request to use the Equal Earth projection wasn't actually the right approach, and instead we needed to use an alternate geometry library that does the correct math on the WGS84 spheroid.

Just by looking at the unit tests, it's clear this new library is giving us better results: lots of whole-number values for shapes that should have them.

ice1080 (approved):

PR #4537, opened 2026-06-04
Move duplicated function to utils file

ice1080:
Many different stores need to check if a planting season is already closed. Cleanup this duplicated function by moving it to a utils file.


ice1080:
* **#4537** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4537?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4537?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/plantingmanagement/util/SeasonUtils.kt
```diff
@@ -0,0 +1,23 @@
+package com.terraformation.backend.plantingmanagement.util
+
+import com.terraformation.backend.db.tracking.PlantingSeasonId
+import com.terraformation.backend.db.tracking.PlantingSeasonStatus
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_SEASONS
+import com.terraformation.backend.plantingmanagement.db.PlantingSeasonClosedException
+import com.terraformation.backend.plantingmanagement.db.PlantingSeasonNotFoundException
+import org.jooq.DSLContext
+
+fun validateSeasonNotClosed(dslContext: DSLContext, plantingSeasonId: PlantingSeasonId) {
```
ice1080: This doesn't seem to be a common pattern, to pass in the `DSLContext`, so wasn't sure if this is the right way to do this or not

sgrimm: This is a potential use case for Kotlin 2.4 context parameters. For now, the way you're doing it is fine.

The approach we've used more often, FWIW, is to put this kind of thing in a `@Named` helper class which gets autowired with the `DSLContext`. See, for example, `ObservationLocker`. Advantage is the helper class can be stubbed out when that's useful for tests, but I'm not sure that matters much in this case.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/plantingmanagement/util/SeasonUtils.kt
ice1080: Ah that makes sense. I try to avoid stores referencing other stores, but this obviously not the same as that so I'll move it to something similar to `ObservationLocker`

sgrimm (approved):

ice1080:
### Merge activity

* **Jun 4, 9:59 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4537).

PR #4544, opened 2026-06-05
SW-8410 API to invite accelerator admins

sgrimm:
We want to be able to add new employees to Terraware as accelerator admins rather
than requiring them to first create an account and then adding global roles after
the fact. Add an API endpoint to invite a new user with an initial list of global
roles. Since, unlike the regular invitation flow, the user won't belong to any
organizations initially, define a new email template that doesn't refer to
any organizations.


sgrimm:
* **#4544** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/customer/UserService.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    if (!email.endsWith("@terraformation.com", ignoreCase = true)) {
+      throw IllegalArgumentException("Only @terraformation.com emails may be invited as admins")
+    }
+
+    return dslContext.transactionResult { _ ->
+      val isNewUser = userStore.fetchByEmail(email) == null
+      val user = userStore.fetchOrCreateByEmail(email)
+
+      userStore.updateGlobalRoles(setOf(user.userId), globalRoles)
```
ice1080, line 41: Is there any concern about replacing roles if the user exists? Should it instead add all that don't already exist?

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/customer/UserService.kt
sgrimm: I thought about that, and in the abstract it makes sense, but the way we're setting global roles for accelerator admins is to give them just one role at a time. If someone is `Read Only` and we set them to `TFExpert` in the console's user editing UI right now, the `Read Only` role gets replaced by `TFExpert` rather than the user having both. I'm not sure if having multiple global roles on a user would cause any funny side effects in the console wherever it tries to display a user's role.

ice1080: Ah good point

PR #4548, opened 2026-06-05
SW-7146 Create local store for dropbox and google drive writers

ice1080:
We already have playwright tests for photo uploads that write to the local file store, and have talked about writing tests for document uploads which write to google drive and dropbox currently. Add a local file store option for these so that playwright tests can configure these env variables (in docker-compose.yml) and actually test that the files are uploaded correctly.

This is a very low priority PR, and claude wrote most of it. 


ice1080:
* **#4548** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4548?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4548?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/file/DropboxWriter.kt
ice1080, line 1: Original contents of this file moved to `RemoteDropboxWriter`

src/main/kotlin/com/terraformation/backend/file/GoogleDriveWriter.kt
ice1080, line 1: Original contents of this file moved to `RemoteGoogleDriveWriter`

src/main/kotlin/com/terraformation/backend/file/LocalDropboxWriter.kt
```diff
@@ -0,0 +1,79 @@
...
+    var finalName = name
+    var attempt = 1
+    while (pathToUrl.containsKey("$folderPath/$finalName")) {
+      finalName = suffixedName(name, attempt++)
+    }
+
+    val fullPath = "$folderPath/$finalName"
+    val storageUrl = fileStore.getUrl(Path("/dropbox/${UUID.randomUUID()}"))
+    fileStore.write(storageUrl, inputStream)
+    pathToUrl[fullPath] = storageUrl
```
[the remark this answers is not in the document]
ice1080: This isn't super relevant to the LocalDropboxWriter, since that one should only ever be used when running the server locally

src/main/kotlin/com/terraformation/backend/file/LocalDropboxWriter.kt
```diff
@@ -0,0 +1,79 @@
...
+
+  override fun uploadFile(folderPath: String, name: String, inputStream: InputStream): String {
+    var finalName = name
+    var attempt = 1
+    while (pathToUrl.containsKey("$folderPath/$finalName")) {
+      finalName = suffixedName(name, attempt++)
+    }
+
+    val fullPath = "$folderPath/$finalName"
+    val storageUrl = fileStore.getUrl(Path("/dropbox/${UUID.randomUUID()}"))
```
sgrimm: Would it make sense to preserve the file extension just to make it easier to inspect the files locally?

src/main/kotlin/com/terraformation/backend/file/LocalDropboxWriter.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+
+/**
+ * Stores documents that would normally go to Dropbox in the local [FileStore] instead. Selected
+ * when `terraware.dropbox.use-local-store` is true.
+ *
+ * Dropbox addresses files by path (`/folder/name`). This class stores the bytes under opaque UUID
+ * keys in the [FileStore] and keeps an in-memory map from Dropbox path to storage URL, so that
+ * arbitrary filenames (including ones with spaces) are handled safely. The map is not persisted,
+ * which is acceptable for tests and dev usage.
```
sgrimm: Maybe add a caveat that this won't work if you're running more than one server instance.

sgrimm: For CI runs or local Playwright runs where you're wiping the state each time, this is fine. But for local dev, I think it'd be nice if the stored files survived a server restart, since the actual files will still be sitting there on disk when you stop the server.

I think you could probably make this work by hashing the final name instead of using a random UUID, and checking for existence of the hashed name in the file store to detect collisions rather than using an in-memory hashmap.

src/main/kotlin/com/terraformation/backend/file/LocalGoogleDriveWriter.kt
```diff
@@ -0,0 +1,190 @@
...
+class LocalGoogleDriveWriter(private val fileStore: FileStore) : GoogleDriveWriter {
+  private val log = perClassLogger()
+
+  private val entries = ConcurrentHashMap<String, Entry>()
+
+  override fun findOrCreateFolders(
+      driveId: String,
+      parentFolderId: String,
+      names: List<String>,
+  ): String = names.fold(parentFolderId) { parent, name -> findOrCreateFolder(parent, name) }
```
sgrimm: Do we really need to fully emulate Google Drive's folder semantics? Would it work to, say, calculate a hash of the folder path and drive ID, and include that in the storage URL? Then I think you wouldn't need the `Entry` class, and the file data would persist across server restarts.

src/test/kotlin/com/terraformation/backend/accelerator/SubmissionServiceLocalStoreTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import com.terraformation.backend.file.LocalGoogleDriveWriter
+import com.terraformation.backend.mockUser
+import io.mockk.every
+import jakarta.ws.rs.core.MediaType
+import org.junit.jupiter.api.Assertions.assertArrayEquals
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+class SubmissionServiceLocalStoreTest : DatabaseTest(), RunsAsUser {
```
sgrimm: What is this testing? That `SubmissionService` uses the `DropboxWriter` and `GoogleDriveWriter` that are passed to its constructor, rather than building its own? If so, that's actually already tested by the existing `SubmissionServiceTest` (which would fail if the service didn't use the supplied MockK mocks).

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/file/LocalDropboxWriter.kt
ice1080: Oh you mean in separate filesystems such as containers, makes sense. I read that earlier today and thought you were talking about on a local filesystem and was confused why

sgrimm: It won't work locally either if you launch two terraware-server instances on different listen ports pointing to the same local directory, though admittedly that's not something we do more than once in a blue moon.

ice1080: Ah gotcha, ya I hadn't heard of anyone needing to do that in the last year and a half. Didn't realize it was something that was needed. Either way I'm updating and will include the note

sgrimm: I do it occasionally when I'm testing cluster-aware code like a new major JobRunr release. Other than that, yeah, not a regular part of our workflow.

src/main/kotlin/com/terraformation/backend/file/LocalGoogleDriveWriter.kt
ice1080: I think we can simplify the `Entry` class some for sure. But because of the way google drive works, we still have to store _some_ of the info, whether in a single class like `Entry` or in separate id to info maps. A simple example is the content type. When uploading a file the client passes `contentType`, then `getFileContentType` has to correctly return the content type based on what was passed in

src/main/kotlin/com/terraformation/backend/file/LocalGoogleDriveWriter.kt
```diff
@@ -0,0 +1,191 @@
...
+ * restart. Folder semantics are also simplified: folders have no backing storage and only their
+ * immediate file children are tracked.
+ */
+@ConditionalOnProperty("terraware.google-drive.use-local-store", havingValue = "true")
+@Named
+@Priority(20) // RemoteGoogleDriveWriter takes precedence when both beans are present.
+class LocalGoogleDriveWriter(private val fileStore: FileStore) : GoogleDriveWriter {
+  private val log = perClassLogger()
+
+  private val entries = ConcurrentHashMap<String, Entry>()
```
sgrimm, line 49: I guess if we care about making this more robust when the server is restarted, we could store the added metadata in a second file alongside the main one. But for most local dev purposes, as long as you can retrieve file contents across restarts, it will probably be fine.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/file/LocalGoogleDriveWriter.kt
ice1080: Alternatively could have a metadata file in the top level of the `/google-drive` folder that contains the metadata for all the files in it

ice1080: Finally implemented

PR #4565, opened 2026-06-09
SW-8420 Add method to query since an id to EventLogStore

ice1080:
Planting Season Notifications will use the event log, and wil store a `last_dismissed_event_log_id` per `planting_season_id`. Add a method to be able to fetch a list of `EventLogEntry`s with a separate `sinceEventLogId` per `LongIdWrapper` passed in. 

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
@@ -100,6 +100,40 @@ class EventLogStore(
     return fetchByCondition(condition, requestedConcreteClasses)
   }
 
+  /**
+   * Fetches events matching any of the given IDs, where each ID has its own lower bound on the
+   * event log ID. An event is returned if it matches one of the [sinceEventLogIds] keys and was
+   * logged after that key's associated event log ID. This is the union of a separate "since" query
+   * per ID, which lets callers track a different watermark for each ID (for example, the last
+   * notification each planting season dismissed).
+   */
+  fun <T : PersistentEvent> fetchByIdsSince(
+      sinceEventLogIds: Map<out LongIdWrapper<*>, EventLogId>,
+      requestedClasses: Collection<KClass<out T>>? = null,
+  ): List<EventLogEntry<T>> {
+    require(sinceEventLogIds.isNotEmpty()) { "No IDs specified" }
+
+    val requestedConcreteClasses = requestedClasses?.let { getConcreteClasses(it) }
+    val classConditions = requestedConcreteClasses?.let { getLikeConditions(it) } ?: emptyList()
+
+    val idConditions = sinceEventLogIds.map { (id, sinceEventLogId) ->
+      DSL.and(
+          payloadField(id.eventLogPropertyName, SQLDataType.BIGINT).eq(id.value),
+          EVENT_LOG.ID.gt(sinceEventLogId),
+      )
+    }
+
+    val condition =
+        DSL.and(
+            listOfNotNull(
+                DSL.or(idConditions),
+                if (classConditions.isNotEmpty()) DSL.or(classConditions) else null,
+            )
+        )
+
+    return fetchByCondition(condition, requestedConcreteClasses)
+  }
+
   fun <T : PersistentEvent> fetchLastByIds(
       ids: Collection<LongIdWrapper<*>>,
       requestedClasses: Collection<KClass<out T>>? = null,
--- a/src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
@@ -105,10 +105,11 @@ class EventLogStore(
    * event log ID. An event is returned if it matches one of the [sinceEventLogIds] keys and was
    * logged after that key's associated event log ID. This is the union of a separate "since" query
    * per ID, which lets callers track a different watermark for each ID (for example, the last
-   * notification each planting season dismissed).
+   * notification each planting season dismissed). A null value means there is no lower bound for
+   * that ID, so all of its events are returned.
    */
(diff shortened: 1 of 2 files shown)


ice1080:
* **#4565** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
```diff
@@ -100,6 +100,40 @@ class EventLogStore(
     return fetchByCondition(condition, requestedConcreteClasses)
   }
 
+  /**
+   * Fetches events matching any of the given IDs, where each ID has its own lower bound on the
+   * event log ID. An event is returned if it matches one of the [sinceEventLogIds] keys and was
```
sgrimm, line 105: Would the client pass 0 as the event log ID the first time it was querying a newly-created entity ID? Maybe it'd make sense for the event log ID to be nullable in the map (in which case the `.gt` doesn't appear for the entity ID in question).

src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
```diff
@@ -100,6 +100,40 @@ class EventLogStore(
...
+      )
+    }
+
+    val condition =
+        DSL.and(
+            listOfNotNull(
+                DSL.or(idConditions),
+                if (classConditions.isNotEmpty()) DSL.or(classConditions) else null,
+            )
+        )
```
sgrimm, line 135: We'll want to look at execution plans here to make sure the database isn't doing anything dumb. I've seen PostgreSQL's optimizer get confused by nested AND/OR operations and it ends up doing a full scan when it could have used indexes instead.

If that happens here, restructuring this into a chain of queries connected with `UNION` might be an alternative (each of the queries would only look up one ID).

But it's also possible the database will do the right thing with this query as is, in which case no need to make it more complicated.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
ice1080: Good thought. I had them separated at first to allow for this, but forgot to make it nullable after combining

src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
ice1080: Here's a quick example output from an explain plan, which confirms that each `or` is using the index correctly!

```
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|QUERY PLAN                                                                                                                                                                                          |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|Sort  (cost=16.88..16.88 rows=1 width=88) (actual time=0.058..0.059 rows=3 loops=1)                                                                                                                 |
|  Output: created_by, created_time, event_class, id, payload                                                                                                                                        |
|  Sort Key: event_log.created_time, event_log.id                                                                                                                                                    |
|  Sort Method: quicksort  Memory: 25kB                                                                                                                                                              |
|  Buffers: shared hit=12                                                                                                                                                                            |
|  ->  Bitmap Heap Scan on public.event_log  (cost=8.32..16.87 rows=1 width=88) (actual time=0.022..0.024 rows=3 loops=1)                                                                            |
|        Output: created_by, created_time, event_class, id, payload                                                                                                                                  |
|        Recheck Cond: ((((event_log.payload ->> 'organizationId'::text))::bigint = 1) OR (((event_log.payload ->> 'organizationId'::text))::bigint = 2))                                            |
|        Filter: (((((event_log.payload ->> 'organizationId'::text))::bigint = 1) AND (event_log.id > 1)) OR ((((event_log.payload ->> 'organizationId'::text))::bigint = 2) AND (event_log.id > 2)))|
|        Rows Removed by Filter: 2                                                                                                                                                                   |
|        Heap Blocks: exact=1                                                                                                                                                                        |
|        Buffers: shared hit=6                                                                                                                                                                       |
|        ->  BitmapOr  (cost=8.32..8.32 rows=4 width=0) (actual time=0.007..0.008 rows=0 loops=1)                                                                                                    |
|              Buffers: shared hit=5                                                                                                                                                                 |
|              ->  Bitmap Index Scan on event_log_organizationid_idx  (cost=0.00..4.16 rows=2 width=0) (actual time=0.006..0.006 rows=3 loops=1)                                                     |
|                    Index Cond: (((event_log.payload ->> 'organizationId'::text))::bigint = 1)                                                                                                      |
|                    Buffers: shared hit=4                                                                                                                                                           |
|              ->  Bitmap Index Scan on event_log_organizationid_idx  (cost=0.00..4.16 rows=2 width=0) (actual time=0.001..0.001 rows=2 loops=1)                                                     |
|                    Index Cond: (((event_log.payload ->> 'organizationId'::text))::bigint = 2)                                                                                                      |
|                    Buffers: shared hit=1                                                                                                                                                           |
|Planning:                                                                                                                                                                                           |
|  Buffers: shared hit=131 read=15                                                                                                                                                                   |
|Planning Time: 0.749 ms                                                                                                                                                                             |
|Execution Time: 0.086 ms                                                                                                                                                                            |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

```

src/main/kotlin/com/terraformation/backend/eventlog/db/EventLogStore.kt
ice1080: Updated

sgrimm (approved):

PR #4566, opened 2026-06-09
Improve survival rate calculation performance

sgrimm:
We were repeating some of our complex subqueries multiple times in the survival
rate calculation code, leading to massive query sizes and slow execution.
Restructure the use of the subqueries so they're only needed one time.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -1991,11 +1991,8 @@ class ObservationStore(
             observationIdField,
         )
     val survivalRateDenominator =
-        DSL.coalesce(
-            survivalRatePermanentDenominator.plus(survivalRateTempDenominator),
-            survivalRatePermanentDenominator,
-            survivalRateTempDenominator,
-        )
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#4566** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4566?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sgrimm:
The "run tests" step of the build of this PR finished in 1:59, compared to 2:37 for the most recent build on main. Obviously it's just one sample.

ice1080 (approved):
I understood about 75% of it, but I at least feel confident in our tests

sgrimm:
Would more detailed comments be useful? If so, any parts in particular you think could use more explanation?

ice1080:
> Would more detailed comments be useful? If so, any parts in particular you think could use more explanation?

I don't think anything specifically was hard to understand, in terms of the new code. I think I more didn't quite understand how it reduced the sql generation by an order of magnitude, but also I haven't been as involved in the pre-computed work that you and Tommy have done

PR #4572, opened 2026-06-10
SW-8420 Fetch and combine events for planting season target notifications

ice1080:
Update `PlantingSeasonNotificationService` to also retrieve PlantingSeasonTarget events. Update Planting Season persistent events with a common interface - `PlantingSeasonRelatedPersistentEvent` (I don't love this name, open to ideas). Use a similar combining logic as `PlantingSeasonPersistentEvent`s, but keep it per species and substratum. The FE (or possibly the control) will eventually take this list and retrieve the set of unique species, discarding the quantities and substrata.

A new model type will be defined in the next PR, along the lines of this:

```kotlin
data class PlantingSeasonNotificationModel(
  val plantingSeasonId: PlantingSeasonId,
  val lastEventLogId: EventLogId,
  val eventsList: List<PlantingSeasonRelatedPersistentEvent>
)
```

Additionally, a later PR will rename some of these methods to separate planting management notifications from inventory management.

Co-authored-by: Claude Opus 4.8 [noreply@anthropic.com](mailto:noreply@anthropic.com)

PR #4571, opened 2026-06-10
SW-8420 Fetch and combine events for planting season notifications

ice1080:
Create a `PlantingSeasonNotificationService` to query the `event_log` by org or planting season and group the results by planting season. Include a method to combine update events such that only the first From and the last To for each (value type) are included in a single event, excluding a Closed event.

Co-authored-by: Claude Opus 4.8 [noreply@anthropic.com](mailto:noreply@anthropic.com)

PR #4570, opened 2026-06-10
SW-8420 Add table for planting season notifications

ice1080:
Add table for planting season notifications to track the `last_dismissed_event_log_id` per planting season, since alerts will be grouped by planting season.

PR #4582, opened 2026-06-11
SW-8420 Add model and endpoint for fetching planting season notifications

ice1080:
Update the service to return a model instead of a list of events per season. Add a new enum to distinguish the various notification type (some are as-of-yet unused). Add an endpoint that fetches the notifications based on the category requested (`PlantingSeasonPlanning` will be completed later).

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
+++ b/src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
@@ -67,23 +67,23 @@ data class ExistingPlantingSeasonScheduledDateModel(
 )
 
 enum class PlantingSeasonNotificationType {
-  PLANTING_SEASON_CLOSED,
-  PLANTING_SEASON_PAST_END_DATE,
-  SPECIES_TARGETS_ADDED,
-  SPECIES_TARGETS_UPDATED,
-  ALLOCATION_QUANTITIES_UPDATED,
-  SEASON_WITHDRAWAL_RECORDED,
+  PlantingSeasonClosed,
+  PlantingSeasonPastEndDate,
+  SpeciesTargetsAdded,
+  SpeciesTargetsUpdated,
+  AllocationQuantitiesUpdated,
+  SeasonWithdrawalRecorded,
 }
 
-data class PlantingSeasonNotificationAlertModel(
+data class PlantingSeasonNotificationModel(
     val type: PlantingSeasonNotificationType,
     val speciesScientificNames: Set<String>? = null,
 )
 
-data class PlantingSeasonNotificationModel(
+data class PlantingSeasonNotificationGroupModel(
     val plantingSeasonId: PlantingSeasonId,
     val plantingSeasonName: String,
     val plantingSiteName: String,
     val lastEventLogId: EventLogId,
-    val events: List<PlantingSeasonNotificationAlertModel>,
+    val notifications: List<PlantingSeasonNotificationModel>,
 )
--- a/src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
+++ b/src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
@@ -1,5 +1,6 @@
 package com.terraformation.backend.plantingmanagement
 
+import com.terraformation.backend.db.default_schema.EventLogId
 import com.terraformation.backend.db.default_schema.SpeciesId
 import com.terraformation.backend.db.tracking.PlantingSeasonId
 import com.terraformation.backend.db.tracking.PlantingSeasonStatus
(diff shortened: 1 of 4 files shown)


ice1080:
* **#4582** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4582?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4582?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4572** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4572?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4571** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4571?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4570** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/plantingmanagement/api/PlantingSeasonNotificationsController.kt
```diff
@@ -0,0 +1,85 @@
...
+  InventoryPlanning,
+  PlantingSeasonPlanning,
+}
+
+data class GetPlantingSeasonNotificationsResponsePayload(
+    val notifications: List<PlantingSeasonNotificationPayload>
+) : SuccessResponsePayload
+
+data class PlantingSeasonNotificationAlertPayload(
+    val type: com.terraformation.backend.plantingmanagement.PlantingSeasonNotificationType,
```
sgrimm: Could import this type.

src/main/kotlin/com/terraformation/backend/plantingmanagement/api/PlantingSeasonNotificationsController.kt
```diff
@@ -0,0 +1,85 @@
...
+enum class PlantingSeasonNotificationCategory {
+  InventoryPlanning,
+  PlantingSeasonPlanning,
+}
+
+data class GetPlantingSeasonNotificationsResponsePayload(
+    val notifications: List<PlantingSeasonNotificationPayload>
+) : SuccessResponsePayload
+
+data class PlantingSeasonNotificationAlertPayload(
```
sgrimm: Why does this use the word "alert" instead of "event"?

src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
```diff
@@ -57,3 +58,25 @@ data class ExistingPlantingSeasonScheduledDateModel(
...
     val species: List<PlantingSeasonScheduledDateSpecies> = emptyList(),
 )
+
+enum class PlantingSeasonNotificationType {
+  PLANTING_SEASON_CLOSED,
+  PLANTING_SEASON_PAST_END_DATE,
+  SPECIES_TARGETS_ADDED,
+  SPECIES_TARGETS_UPDATED,
+  ALLOCATION_QUANTITIES_UPDATED,
+  SEASON_WITHDRAWAL_RECORDED,
```
sgrimm: Any reason these are all-caps snake case rather than camel case like most of our enums?

src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
```diff
@@ -57,3 +58,25 @@ data class ExistingPlantingSeasonScheduledDateModel(
...
+    val type: PlantingSeasonNotificationType,
+    val speciesScientificNames: Set<String>? = null,
+)
+
+data class PlantingSeasonNotificationModel(
+    val plantingSeasonId: PlantingSeasonId,
+    val plantingSeasonName: String,
+    val plantingSiteName: String,
+    val lastEventLogId: EventLogId,
+    val events: List<PlantingSeasonNotificationAlertModel>,
```
sgrimm: Similar to my comment on the payload class: the property is "events" but the list entries are "alerts."

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/plantingmanagement/api/PlantingSeasonNotificationsController.kt
ice1080: I actually couldn't decide on good names for these. One of these could be from a single event or from a combination of events, so I didn't want to use that. Notification was already used for the parent payload. I'm very much open to ideas for both of these because I don't like the ambiguity

src/main/kotlin/com/terraformation/backend/plantingmanagement/Models.kt
ice1080: I should have looked at more examples. IntelliJ's search for `enum class` pulled up `IdentifierType` and `LockType` as the first two, so I assumed that was the standard. Will update

src/main/kotlin/com/terraformation/backend/plantingmanagement/api/PlantingSeasonNotificationsController.kt
ice1080: Perhaps this one should be `PlantingSeasonNotificationPayload` and the parent should be `PlantingSeasonNotificationGroupPayload`

ice1080: Updated

ice1080:
### Merge activity

* **Jun 11, 11:02 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4582).

PR #4590, opened 2026-06-11
Speed up SearchServiceNestedFieldsTest

sgrimm:
The "all fields are valid sort keys" test was referencing fields using nested
sublists that, for some tables, could be close to a dozen nesting levels deep.
The generated SQL worked, but the nesting slowed the queries down, and this
test covers hundreds of search fields.

Change the test so that if a field isn't an alias for another field, the
search query uses the field's table as its prefix. The generated SQL becomes
much simpler and runs more quickly.

Alias fields continue to use the old nested-sublists style, but there aren't
many of them.

On my laptop, execution time for the test method goes from 12.5 to 3.3
seconds.

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceNestedFieldsTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceNestedFieldsTest.kt
@@ -1211,14 +1211,12 @@ internal class SearchServiceNestedFieldsTest : SearchServiceTest() {
           val prefix = SearchFieldPrefix(fullyQualifiedField.searchTable)
           val field = prefix.resolve(fieldName.substringAfterLast('.'))
           val sortFields = listOf(SearchSortField(field))
-          assertDoesNotThrow("Sort by $fieldName") {
-            searchService.search(
-                prefix,
-                listOf(field),
-                mapOf(rootPrefix to NoConditionNode()),
-                sortFields,
-            )
-          }
+          searchService.search(
+              prefix,
+              listOf(field),
+              mapOf(rootPrefix to NoConditionNode()),
+              sortFields,
+          )
         }
       }
     }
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#4590** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceNestedFieldsTest.kt
```diff
@@ -1186,23 +1187,39 @@ internal class SearchServiceNestedFieldsTest : SearchServiceTest() {
...
+              sortFields,
+          )
+        } else {
+          // Query and sort by the field directly on its search table rather than in the form of
+          // nested sublists from the root prefix; sublists cause the search code to generate
+          // SQL with lots of nested subqueries, which slows the test down.
+          val prefix = SearchFieldPrefix(fullyQualifiedField.searchTable)
+          val field = prefix.resolve(fieldName.substringAfterLast('.'))
+          val sortFields = listOf(SearchSortField(field))
+          assertDoesNotThrow("Sort by $fieldName") {
```
ice1080: Is this `assertDoesNotThrow` necessary anymore now that one encompasses it on line 1195?

sgrimm: Ugh, no, removing that was the whole point of wrapping it at the higher level and then I forgot to actually remove the thing!

ice1080 (approved):
I didn't review the codex comment, so double check that

src/test/kotlin/com/terraformation/backend/seedbank/search/SearchServiceNestedFieldsTest.kt
```diff
@@ -1186,23 +1187,39 @@ internal class SearchServiceNestedFieldsTest : SearchServiceTest() {
...
             searchService.search(
                 prefix,
-                searchFields,
-                mapOf(prefix to NoConditionNode()),
+                listOf(field),
```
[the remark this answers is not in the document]
sgrimm: Sorting on nested keys is tested in other test cases. The point of this test is to make sure we don't have any fields (or really: any field types) that can't be sorted on at all.

PR #4607, opened 2026-06-15
SW-8434 Add method to recalculate species totals

sgrimm:
Add an ObservationStore method to recalculate all the observed species totals
values from the recorded plants data. It's not used yet, but will be one of the
steps in the "merge observations" operation.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #4608, opened 2026-06-15
SW-8434 Merge observations in admin UI

sgrimm:
Add an admin UI control to merge one observation into another one.

Merging moves all the results from the source observation into the target one.
If the target observation already had results for a given plot, that plot's data
is replaced by the data from the source observation (that is, the merge uses
"overwrite on conflict" semantics).

Aggregate data such as observation results and species totals are recalculated
from the raw recorded-plants data.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
@@ -40,6 +40,7 @@ import com.terraformation.backend.tracking.db.InvalidObservationStartDateExcepti
 import com.terraformation.backend.tracking.db.ObservationAlreadyStartedException
 import com.terraformation.backend.tracking.db.ObservationHasNoSubstrataException
 import com.terraformation.backend.tracking.db.ObservationLocker
+import com.terraformation.backend.tracking.db.ObservationMergeNotAllowedException
 import com.terraformation.backend.tracking.db.ObservationNotFoundException
 import com.terraformation.backend.tracking.db.ObservationPlotNotFoundException
 import com.terraformation.backend.tracking.db.ObservationRescheduleStateException
@@ -820,6 +821,59 @@ class ObservationService(
       observationStore.deleteObservation(observationId)
 
       observationStore.recalculateSurvivalRates(observation.plantingSiteId)
+    }
+  }
+
+  fun mergeObservations(sourceObservationId: ObservationId, targetObservationId: ObservationId) {
+    requirePermissions {
+      manageObservation(sourceObservationId)
+      manageObservation(targetObservationId)
+    }
+
+    if (sourceObservationId == targetObservationId) {
+      throw ObservationMergeNotAllowedException("Cannot merge an observation into itself")
+    }
+
+    val source = observationStore.fetchObservationById(sourceObservationId)
+    val target = observationStore.fetchObservationById(targetObservationId)
+
+    if (source.plantingSiteId != target.plantingSiteId) {
+      throw ObservationMergeNotAllowedException("Observations belong to different planting sites")
+    }
+    if (source.isAdHoc || target.isAdHoc) {
+      throw ObservationMergeNotAllowedException("Cannot merge ad-hoc observations")
+    }
+    if (
+        source.observationType != ObservationType.Monitoring ||
+            target.observationType != ObservationType.Monitoring
+    ) {
+      throw ObservationMergeNotAllowedException("Can only merge monitoring observations")
+    }
+    if (source.plantingSiteHistoryId == null || target.plantingSiteHistoryId == null) {
+      throw ObservationMergeNotAllowedException(
+          "Cannot merge observations that have not been started"
+      )
+    }
+
+    systemUser.run {
+      // Delete the target's media for any plot that an incoming completed source plot overwrites.
+      deleteMediaWhere(
+          OBSERVATION_MEDIA_FILES.OBSERVATION_ID.eq(targetObservationId)
+              .and(
+                  OBSERVATION_MEDIA_FILES.MONITORING_PLOT_ID.`in`(
+                      DSL.select(OBSERVATION_PLOTS.MONITORING_PLOT_ID)
+                          .from(OBSERVATION_PLOTS)
+                          .where(OBSERVATION_PLOTS.OBSERVATION_ID.eq(sourceObservationId))
+                          .and(OBSERVATION_PLOTS.STATUS_ID.eq(ObservationPlotStatus.Completed))
+                  )
+              )
+      )
+
+      dslContext.transaction { _ ->
+        observationStore.mergeObservationData(sourceObservationId, targetObservationId)
+        observationStore.recalculateObservationTotals(targetObservationId)
+        observationStore.recalculateSurvivalRates(source.plantingSiteId)
+      }
     }
   }
 
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -2376,7 +2376,7 @@ class ObservationStore(
    * Moves the raw per-plot data of completed plots from the source observation to the target
    * observation, repoints/drops t0 baselines, and deletes the source observation. Does NOT delete
    * media (the caller handles colliding target media) and does NOT rebuild derived aggregates (call
-   * [recalculateObservationTotals] afterward).
+   * [recalculateObservationTotals] and [recalculateSurvivalRates] afterward).
    */
   fun mergeObservationData(
       sourceObservationId: ObservationId,
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
@@ -41,6 +41,7 @@ import com.terraformation.backend.db.tracking.tables.records.RecordedPlantsRecor
 import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOTS
 import com.terraformation.backend.db.tracking.tables.references.MONITORING_PLOT_HISTORIES
 import com.terraformation.backend.db.tracking.tables.references.OBSERVATIONS
+import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_MEDIA_FILES
 import com.terraformation.backend.db.tracking.tables.references.OBSERVATION_PLOTS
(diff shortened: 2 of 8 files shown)


sgrimm:
* **#4608** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4608?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4608?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4607** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

sgrimm:
Sorry for the large size of this one; I didn't think it made much sense to split the store and service methods into separate PRs since they're so tightly coupled, and that was the only substantial split point I could see.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2367,6 +2368,154 @@ class ObservationStore(
...
+              .set(OBSERVED_PLOT_COORDINATES.OBSERVATION_ID, targetObservationId)
+              .where(OBSERVED_PLOT_COORDINATES.OBSERVATION_ID.eq(sourceObservationId))
+              .and(OBSERVED_PLOT_COORDINATES.MONITORING_PLOT_ID.`in`(completedInSource))
+              .execute()
+          dslContext
+              .update(OBSERVATION_MEDIA_FILES)
+              .set(OBSERVATION_MEDIA_FILES.OBSERVATION_ID, targetObservationId)
+              .where(OBSERVATION_MEDIA_FILES.OBSERVATION_ID.eq(sourceObservationId))
+              .and(OBSERVATION_MEDIA_FILES.MONITORING_PLOT_ID.`in`(completedInSource))
+              .execute()
```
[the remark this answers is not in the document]
sgrimm: Incomplete plots don't have media (the mobile app uploads media after completing the plot).

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -823,6 +824,59 @@ class ObservationService(
...
+    }
+    if (
+        source.observationType != ObservationType.Monitoring ||
+            target.observationType != ObservationType.Monitoring
+    ) {
+      throw ObservationMergeNotAllowedException("Can only merge monitoring observations")
+    }
+    if (source.plantingSiteHistoryId == null || target.plantingSiteHistoryId == null) {
+      throw ObservationMergeNotAllowedException(
+          "Cannot merge observations that have not been started"
```
ice1080, line 854: Is this error message correct? Does an observation's `plantingSiteHistoryId` stay null until the observation has been started?

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2367,6 +2368,154 @@ class ObservationStore(
...
 
+  /**
+   * Merges the observation data from a source observation to a target.
+   *
+   * You probably want `ObservationService.mergeObservations` instead of this.
+   *
+   * Moves the raw per-plot data of completed plots from the source observation to the target
+   * observation, repoints/drops t0 baselines, and deletes the source observation. Does NOT delete
+   * media (the caller handles colliding target media) and does NOT rebuild derived aggregates (call
+   * [recalculateObservationTotals] afterward).
```
ice1080: If you're being thorough, you'd want to also mention calling `recalculateSurvivalRates` in the comment

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2367,6 +2368,154 @@ class ObservationStore(
...
   }
 
+  /**
+   * Merges the observation data from a source observation to a target.
+   *
+   * You probably want `ObservationService.mergeObservations` instead of this.
+   *
+   * Moves the raw per-plot data of completed plots from the source observation to the target
+   * observation, repoints/drops t0 baselines, and deletes the source observation. Does NOT delete
+   * media (the caller handles colliding target media) and does NOT rebuild derived aggregates (call
```
tommylau523, line 2378: Should we throw another exception, if a file is not yet deleted? I am just thinking about cases that this will leave orphaned file artifacts, if we call this function instead of the service one.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
```diff
@@ -2367,6 +2368,154 @@ class ObservationStore(
...
+              .set(OBSERVATION_MEDIA_FILES.OBSERVATION_ID, targetObservationId)
+              .where(OBSERVATION_MEDIA_FILES.OBSERVATION_ID.eq(sourceObservationId))
+              .and(OBSERVATION_MEDIA_FILES.MONITORING_PLOT_ID.`in`(completedInSource))
+              .execute()
+          dslContext
+              .update(PLOT_T0_OBSERVATIONS)
+              .set(PLOT_T0_OBSERVATIONS.OBSERVATION_ID, targetObservationId)
+              .where(PLOT_T0_OBSERVATIONS.OBSERVATION_ID.eq(sourceObservationId))
+              .and(PLOT_T0_OBSERVATIONS.MONITORING_PLOT_ID.`in`(completedInSource))
+              .execute()
```
ice1080, line 2511: I believe that this will also need to insert into `plot_t0_densities` based on the source observation

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
sgrimm: Yes, because we don't know what map version we're using until the observation actually starts. It's possible to create an observation, edit the map, and then start the observation. We use the map at the time of observation start, not the historical map that was active when the observation was first created.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: This will actually already happen: `observation_media_files` doesn't have ON DELETE CASCADE on its observation ID foreign key, so the observation deletion will fail.

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationStore.kt
sgrimm: That table doesn't have an observation ID, so the source observation's densities will already be there, unless I'm misunderstanding how it works.

ice1080: When an observation is used for t0, the `T0Service` assigns the densities via the observed totals for the observations in addition to inserting into `plot_t0_observations`. See `assignT0PlotObservation`. We talked about this when initially implementing and decided that was the best approach since we thought nothing else would be touching those tables. 
Your method could utilize `T0Service`, `T0Store`, or directly insert to `plot_t0_densities`

ice1080: Note that if you move to using the `T0Store`, it already publishes `T0PlotDataAssignedEvent` and will call recalculate automatically, so you'd either be double-calculating due to the caller of your method, or you'd have to optionally disable that publish

sgrimm: I think the recalculation would be a no-op because we're replacing the target observation's observed totals (if any) with the source's, and the existing densities are already derived from the source's numbers.

I suspect I'm just not seeing the failure mode here. An example of the way I'm thinking about this would be something like

* Plot in observation 1: 22 live of species 1, 44 live of species 2
* Observation 1 chosen for t0; densities calculated as ~244 and ~488 respectively.
* Merge observation 1 to observation 2
* Plot in observation 2 after merge: 22 live of species 1, 44 live of species 2

Wouldn't a recalculation of the densities based on observation 2 end up with 244 and 488 for species 1 and 2 again, meaning nothing would have changed?

ice1080: I'm thinking of the opposite merge. Let's say Observation 2 had half the live values, but t0 had selected observation 1. When merging, the plot should now have half the densities

sgrimm: Just to make sure I'm understanding the scenario you're describing:

* Observation 1: 22 live
* Observation 2: 11 live
* Observation 1 chosen for t0
* Merge observation 2 to observation 1

In that case, in the current implementation, the t0 densities would be deleted (by the `t0DropPlots` delete earlier in the code) and you'd have to set them again manually.

That was an intentional design choice on my part, though maybe it was the wrong one. My reasoning was that the merge is blowing away the data the user said to use for t0, and I didn't think it would necessarily be a given that they'd want to use the numbers from a different observation instead. I admit the concept of "a different observation" is muddy here given the merge operation.

My mental model is that when you pick an observation as the source of t0 data, what you're really saying is something like, "Use the numbers we observed on June 3." If the source observation is the June 3 one and you use it for t0, it makes sense to retain the densities post-merge, because they're still based on the data you gathered on June 3. But if the target was the June 3 observation you're using as the t0 source, replacing its densities with the source's numbers would mean the system is disregarding your implied "June 3's numbers were correct" assertion.

Does any of that make sense?

ice1080: Ah I now understand the intentionality behind this decision, and why it was made. I had been thinking of "merge" more as "combine" so that either direction it went would set `plot_t0_densities`, but obviously the description mentions that it's more of an "overwrite on conflict" semantic. No changes needed, I read back through the changes with that in mind and understand things now. Thanks Steve!

ice1080 (approved):

sgrimm:
### Merge activity

* **Jun 15, 8:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4608).
* **Jun 15, 8:45 PM UTC**: [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4608) rebased this pull request as part of a merge.
* **Jun 15, 9:04 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4608).

PR #4612, opened 2026-06-16
SW-8420 Add endpoint for single season's notifications

ice1080:
Update the notifications endpoint to allow retrieving notifications for only one planting season.

Update the category enum class to include the list of notification types that should be retrieved, since a third category is coming later.

Make `PlantingSeasonAllocatedSpeciesPersistentEvent`also be a `PlantingSeasonRelatedPersistentEvent` and add it to the list of `PlantingSeasonPlanning` types.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
+++ b/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
@@ -14,6 +14,7 @@ import com.terraformation.backend.eventlog.model.EventLogEntry
 import com.terraformation.backend.plantingmanagement.PlantingSeasonNotificationGroupModel
 import com.terraformation.backend.plantingmanagement.PlantingSeasonNotificationModel
 import com.terraformation.backend.plantingmanagement.PlantingSeasonNotificationType
+import com.terraformation.backend.plantingmanagement.event.PlantingSeasonAllocatedSpeciesPersistentEvent
 import com.terraformation.backend.plantingmanagement.event.PlantingSeasonPersistentEvent
 import com.terraformation.backend.plantingmanagement.event.PlantingSeasonRelatedPersistentEvent
 import com.terraformation.backend.plantingmanagement.event.PlantingSeasonSpeciesTargetCreatedEvent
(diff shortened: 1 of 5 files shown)


ice1080:
* **#4616** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4615** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4614** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4614?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4612** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
```diff
@@ -30,24 +31,51 @@ class PlantingSeasonNotificationsService(
...
+
+    val requestedTypes = types.toSet()
+    val eventClasses = eventClassesToFetch(requestedTypes)
+    if (eventClasses.isEmpty()) {
+      return emptyList()
+    }
+
+    val condition =
+        if (plantingSeasonId != null) PLANTING_SEASONS.ID.eq(plantingSeasonId)
+        else PLANTING_SEASONS.plantingSites.ORGANIZATION_ID.eq(organizationId)
```
sgrimm: Minor nit: I think this would be easier to read with curly braces.

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
```diff
@@ -30,24 +31,51 @@ class PlantingSeasonNotificationsService(
...
+      plantingSeasonId: PlantingSeasonId?,
+      types: Collection<PlantingSeasonNotificationType>,
   ): List<PlantingSeasonNotificationGroupModel> {
-    val seasonInfoById =
-        fetchSeasonInfoByCondition(
-            PLANTING_SEASONS.plantingSites.ORGANIZATION_ID.eq(organizationId)
-        )
+    require(organizationId != null || plantingSeasonId != null) {
+      "Either organizationId or plantingSeasonId must be specified."
+    }
```
sgrimm: Another option (up to you) is to define two overloads of `getNotifications`, one for org ID and one for planting season ID, and choose which one to call in the controller. Then the type system gives you this validation check for free. Obviously they can both call into a common implementation method that takes a `Condition` parameter.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
ice1080: Just out of curiosity, do you ever think without curly braces is easier to read? Is it if it's all contained in one line, or even then curly?

sgrimm: Yeah, when it all fits on one line is pretty much the only time I think without curly braces is better. Usually only if it's an expression rather than a statement (i.e., I'll use curly braces for "if X do A, else do B" kinds of statements even if they're short).

sgrimm: Obviously that kind of thing is a matter of preference, though.

ice1080: Makes sense. Most of the time I write this stuff, I write it with a ternary operator like groovy and ts/js, and then remember that kotlin doesn't have that and change it to this, thus forgetting the curly braces

ice1080: I find it somewhat interesting that kotlin implemented the elvis operator, but not ternary

sgrimm: I get the lack of a ternary operator and I think it's the right call. C and Java and JS need the ternary operator because their `if` is a statement, but in Kotlin it's an expression so all the ternary operator would be saving you is a few characters per usage site, with the cost of increasing the size of the language syntax.

But the Elvis operator is still useful because if you emulate it with `if`, you need to repeat whatever expression you're testing for nullness.

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
ice1080: I go back and forth on this type of thing, since the controller would still need to throw an error so it's just moving the logic to the controller. E.g.

```kotlin
if (plantingSeasonId != null) {
  service.getNotifications(plantingSeasonId, ...
} else if (organizationId != null) {
  service.getNotifications(organizationId, ...
} else {
  throw ...
```

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonNotificationsService.kt
ice1080: Ah that makes sense

sgrimm (approved):

PR #4614, opened 2026-06-16
SW-8420 Add planting season withdrawal event

ice1080:
Define a new persistent event to track when a withdrawal has been created and associated with a planting season.

For now this is a `PlantingSeasonRelatedPersistentEvent`, but we could alternatively change it to a regular persistent event (e.g. `WithdrawalCreatedPersistentEvent` that has a `plantingSeasonId`, and use type casting if needed in the notification service.

Co-authored-by: Claude Opus 4.8 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)


ice1080:
* **#4616** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4615** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4614** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4612** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/resources/i18n/Messages_en.properties
```diff
@@ -152,6 +152,8 @@ eventSubject.PlantingSeasonScheduledDateSpecies.short=Species
 eventSubject.PlantingSeasonSpeciesTarget.field.quantity=quantity
 eventSubject.PlantingSeasonSpeciesTarget.full=Species {0} ({1} / {2})
 eventSubject.PlantingSeasonSpeciesTarget.short=Species
+eventSubject.PlantingSeasonWithdrawal.full=Withdrawal recorded
```
sgrimm: The full name is to distinguish between subjects of the same type when you're viewing a list that spans multiple subjects. I guess we don't expose withdrawal IDs to end users, but maybe this could have the withdrawal date or something like that. Which would of course require that the event class carry whatever extra information we use here.

Or, if we don't want to be able to distinguish multiple subjects of this type, this could just be "Withdrawal."

src/main/kotlin/com/terraformation/backend/eventlog/api/EventSubjectPayload.kt
```diff
@@ -682,6 +684,30 @@ data class PlantingSeasonSpeciesTargetSubjectPayload(
   }
 }
 
+@JsonTypeName("PlantingSeasonWithdrawal")
+data class PlantingSeasonWithdrawalSubjectPayload(
+    override val fullText: String,
+    val plantingSeasonId: PlantingSeasonId,
+    val plantingSiteId: PlantingSiteId,
+    override val shortText: String,
+    val withdrawalId: WithdrawalId,
```
sgrimm, line 697: It might make sense to also include the nursery's facility ID here, just so we'll already have it in place if we want to query this from the nursery side in the future.

sgrimm (changes requested):
I think having this be a planting-season-related event is fine. We can add other interfaces later on if we want it to also get grouped with, say, nursery-related events. A couple minor bits of feedback but it generally looks good.

sgrimm (approved):

ice1080:
### Merge activity

* **Jun 17, 5:51 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4614).

PR #4617, opened 2026-06-17
Speed up country detector

sgrimm:
When the country detector is asked to get the list of countries for a complex
MultiPolygon, it can be expensive to compute the intersections with all the
country borders. In addition, many country borders are themselves complex
MultiPolygons.

We only care about whether the total area exceeds the minimum coverage
threshold, so we can reduce the expense by considering the component polygons
one at a time and stopping when the total area hits the threshold.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
@@ -108,10 +108,10 @@ class CountryDetector {
                   .flatMap { geometryPolygon ->
                     if (countryBorder is MultiPolygon) {
                       (0..<countryBorder.numGeometries).asSequence().map { it ->
-                        geometryPolygon to countryBorder.getGeometryN(it)
+                        countryBorder.getGeometryN(it) to geometryPolygon
                       }
                     } else {
-                      sequenceOf(geometryPolygon to countryBorder)
+                      sequenceOf(countryBorder to geometryPolygon)
                     }
                   }
                   .runningFold(0.0) { subtotal, (countryPolygon, geometryPolygon) ->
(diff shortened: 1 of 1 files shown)


sgrimm:
* **#4617** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
```diff
@@ -89,14 +90,57 @@ class CountryDetector {
...
+                }
+              }
+
+          if (area >= minCoverageArea) {
+            countryCode
+          } else {
+            null
+          }
+        }
+        .toSet()
```
ice1080, line 128: I haven't thought too deep about this, but is it possible for country borders to have the same pair more than once? I'm guessing so which is what the `.toSet()` is for. If that's true, is there a possibility that the same country overlap will be evaluated more than once, and then the duplicate removed later?

ice1080 (approved):
One potential improvement, though I might be thinking about it wrong

src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
sgrimm: There shouldn't be duplicates since country borders are from a map that's keyed by country code. The `toSet()` is to make the semantics clear: the return value is unordered and values are unique.

sgrimm:
I realized I could get rid of some of the nested "if" conditions by flattening the iteration out a bit using sequences.

ice1080 (approved):

sgrimm:
### Merge activity

* **Jun 17, 4:33 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4617).

PR #4629, opened 2026-06-19
SW-8537 Import WCVP species data

sgrimm:
Add an importer, accessible via the admin UI, that loads the list of species
and distributions from the official distribution of the World Checklist of
Vascular Plants data set.

Nothing uses this data yet.


sgrimm:
* **#4630** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4630?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4629** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4629?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4629?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/species/WcvpImporter.kt
```diff
@@ -0,0 +1,149 @@
...
+      val locationId = row["locationid"] ?: return null
+      val taxonId = row["coreid"]?.let { WcvpTaxonId(it) } ?: return null
+
+      return if (tdwgPattern.matches(locationId)) {
+        WcvpDistributionsRecord(
+            taxonId = taxonId,
+            occurrenceStatus = row["occurrencestatus"],
+            threatStatus = row["threatstatus"],
+            establishmentMeans = row["establishmentmeans"],
+            level3Code = locationId.substringAfter(':'),
```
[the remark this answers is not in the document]
sgrimm: ChatGPT, let me introduce you to the concept of "stacked diffs."

src/main/kotlin/com/terraformation/backend/species/WcvpImporter.kt
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  private fun importDistributions(inputStream: InputStream, taxonIds: Set<WcvpTaxonId>) {
+    log.info("Importing WCVP distribution data")
+
+    var count = 0
+
+    return DistributionParser(inputStream)
+        .sequence()
+        .filter { it.taxonId in taxonIds }
```
ice1080: Do we want to log the unused taxonIds?

src/main/kotlin/com/terraformation/backend/species/WcvpImporter.kt
```diff
@@ -0,0 +1,149 @@
...
+
+      return WcvpTaxaRecord(
+          acceptedNameUsageId = row["acceptednameusageid"]?.let { WcvpTaxonId(it) },
+          family = row["family"],
+          genus = row["genus"],
+          infraspecificEpithet = row["infraspecificepithet"],
+          nomenclaturalStatus = row["nomenclaturalstatus"],
+          originalNameUsageId = row["originalnameusageid"]?.let { WcvpTaxonId(it) },
+          parentNameUsageId = row["parentnameusageid"]?.let { WcvpTaxonId(it) },
+          scientificName = row["scientfiicname"], // Column name is misspelled in the CSV file
```
ice1080, line 159: Wow

src/test/kotlin/com/terraformation/backend/species/WcvpImporterTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.util.zip.ZipFile
+import kotlin.io.path.Path
+import kotlin.io.path.exists
+import org.junit.Assume.assumeNotNull
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+/**
+ * Tests the WCVP importer. Importing the data is fairly resource-intensive, so this test is skipped
+ * unless you specifically ask for it by setting the `TEST_WCVP_IMPORTER` environment variable.
```
ice1080: Do we want to set this up to run once a week or something?

src/test/kotlin/com/terraformation/backend/species/WcvpImporterTest.kt
```diff
@@ -0,0 +1,56 @@
...
+      perClassLogger().info("Downloading WCVP zipfile")
+
+      WcvpImporter.defaultZipFileUrl.toURL().openStream().use { zipFileStream ->
+        Files.copy(zipFileStream, path, StandardCopyOption.REPLACE_EXISTING)
+      }
+    }
+
+    ZipFile(path.toFile()).use { zipFile -> importer.import(zipFile) }
+
+    eventPublisher.assertEventPublished(WcvpImportedEvent())
```
ice1080, line 58: Should we check that some rows were inserted to both tables?

src/test/kotlin/com/terraformation/backend/species/WcvpImporterTest.kt
ice1080: Alternatively we could take their zip file and truncate it to just a few rows and then test with that

sgrimm: So we can detect if they decide to change file formats, you mean?

I think that's going to be such a low-probability event (given that the existing format has been in use for years) that it's fine to just discover it when we go to update the dataset and the import fails. We'll refresh this data infrequently and, I assume, it will never be an urgent time-sensitive operation where the delay from discovering the format has changed and fixing the code would be unacceptable.

ice1080: That makes sense to me, though I guess I would question the benefit of this test at all if we're not going to run it regularly

ice1080 (approved):

src/test/kotlin/com/terraformation/backend/species/WcvpImporterTest.kt
sgrimm: It's for local dev, really. I was running it dozens of times as I worked on this code, and if at some point we go update the code, whoever does it would end up having to basically write the same test locally to check their work. I can update the comment to make that intent clearer.

ice1080: Ah that makes sense. Perhaps another option would be to add a comment to the importer itself mentioning that it can be tested by setting the variable. I recognize that's a bit unorthodox, but it does help the next person working on it to know about the test instead of just assuming `./gradlew test` will include it

ice1080 (approved):

sgrimm:
### Merge activity

* **Jun 22, 5:24 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4629).
* **Jun 22, 5:24 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4629).

PR #4642, opened 2026-06-22
Increase default heap size limit to 5GB

sgrimm:
Previously we weren't specifying a default heap size when the server was launched
in a container using the Docker image, so the JVM used its default of 25% of the
container's memory limit, a bit under 2GB on our 8GB staging servers. Change the
startup process to pass in a heap size argument (5GB by default) which can be
overridden using an environment variable if needed.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,6 +9,7 @@ WORKDIR /app
 # application code. This way, pulling a new version of the application when the
 # dependencies haven't changed will be able to reuse the existing image layer from
 # the previous version.
+COPY --chown=$USER_ID --chmod=755 start.sh ./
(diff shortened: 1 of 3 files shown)


sgrimm:
* **#4642** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4642?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4642?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

sgrimm:
I'll merge #4641 before this one just to double-check that it reduces memory usage as expected.

PR #4649, opened 2026-06-23
SW-8574 Add endpoint for public stats

ice1080:
Create an endpoint to expose public stats about Terraware, including org count, country count, and sum of area under restoration (ha). Exclude orgs tagged with `Internal` org `Testing`.

Expose this endpoint to non-logged-in clients.

Add a new `PublicEndpoint` annotation class for this endpoint.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/api/Annotations.kt
+++ b/src/main/kotlin/com/terraformation/backend/api/Annotations.kt
@@ -61,6 +61,11 @@ annotation class FunderEndpoint
 annotation class InternalEndpoint
 
 @Retention(AnnotationRetention.RUNTIME)
+@Tag(name = "Public")
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+annotation class PublicEndpoint
+
+@Retention(AnnotationRetention.RUNTIME)
 @Target(AnnotationTarget.ANNOTATION_CLASS, AnnotationTarget.FUNCTION)
 @ApiResponse(
     content =
--- a/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
+++ b/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt
@@ -71,6 +71,9 @@ class SecurityConfig(
         // Allow unauthenticated users to check their app versions for compatibility.
         authorize("/api/v1/versions", permitAll)
 
+        // Allow unauthenticated users to fetch aggregate platform statistics.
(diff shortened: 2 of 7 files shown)


ice1080:
* **#4651** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4651?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4649** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4649?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4649?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

sgrimm (commented):
I'm about to sign off for the day, so I'll look at this in more detail tomorrow, but from a quick glance, I am slightly uneasy with the idea that every drive-by web scraper hit will end up triggering a bunch of full table scans and aggregate calculations on the Terraware database. I wonder if we could cache this in memory such that we only grab new stats once an hour (or once a day) per server or somesuch. Or is the requirement that it update in real time?

ice1080:
> I wonder if we could cache this in memory such that we only grab new stats once an hour (or once a day) per server or somesuch. Or is the requirement that it update in real time?

Definitely a valid concern here; ​I like the idea of caching. The request was "dynamic (or at minimum updated once a quarter)", so I think we could even have it cache indefinitely and add an admin endpoint to invalidate the cache if needed. That way new deploys would obviously invalidate it, and we can invalidate more often if needed.

sgrimm:
> add an admin endpoint to invalidate the cache if needed

If we use an in-memory cache, this will have to be cluster-aware.

sgrimm (approved):
The one thing that is maybe not 100% ideal with this cache setup is that the numbers will potentially be slightly inconsistent between servers if they fetch the data at different times. I don't think that actually matters for the target use case, though, and this caching setup has the benefit of being simple and easy to understand.

ice1080:
> The one thing that is maybe not 100% ideal with this cache setup is that the numbers will potentially be slightly inconsistent between servers if they fetch the data at different times. I don't think that actually matters for the target use case, though, and this caching setup has the benefit of being simple and easy to understand.

I'll leave this as-is for now. I think this could be a use-case for a simple table that gets updated, or a materialized view since it always refreshes all the data at the same time and there would only be a small number of rows in it. We can make that improvement in the future if we run into issues.

PR #4658, opened 2026-06-24
SW-7489 Add collectedTime to accession create/update and response payloads

ice1080:
Update the accessions controller to return the new `collectedTime` field in addition to the existing `collectedDate` . For Create/Update requests, accept both.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
+++ b/src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
@@ -117,17 +117,8 @@ class AccessionsV2Controller(
   }
 }
 
-/**
- * Resolves the deprecated date-only `collectedDate` and the new `collectedTime` to a single
- * timestamp. Once web and mobile have fully migrated to `collectedTime`, `collectedDate` can be
- * removed from the request payloads and this helper deleted.
- */
-private fun resolveCollectedTime(collectedDate: LocalDate?, collectedTime: Instant?): Instant? {
-  if (collectedDate != null && collectedTime != null) {
-    throw IllegalArgumentException("Only one of collectedDate and collectedTime may be specified.")
-  }
-  return collectedTime ?: collectedDate?.atStartOfDay(ZoneOffset.UTC)?.toInstant()
-}
+private fun resolveCollectedTime(collectedDate: LocalDate?, collectedTime: Instant?): Instant? =
+    collectedTime ?: collectedDate?.atStartOfDay(ZoneOffset.UTC)?.toInstant()
 
 /**
  * Supported accession states. This is a subset of the values in [AccessionState], minus obsolete
@@ -181,12 +172,8 @@ data class AccessionPayloadV2(
     )
     val active: AccessionActive,
     val bagNumbers: Set<String>?,
-    @Schema(
-        description = "Use collectedTime instead.",
-        deprecated = true,
-    )
     val collectedDate: LocalDate?,
-    @Schema(description = "Date and time the seeds were collected, in UTC.")
+    @Schema(description = "Date and time the seeds were collected.") //
     val collectedTime: Instant?,
     val collectionSiteCity: String? = null,
     val collectionSiteCoordinates: Set<Geolocation>?,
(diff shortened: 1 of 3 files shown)


ice1080:
* **#4659** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4659?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4658** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4658?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4667** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4667?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/seedbank/api/AccessionsV2Controller.kt
```diff
@@ -324,7 +343,18 @@ data class AccessionPayloadV2(
...
 data class CreateAccessionRequestPayloadV2(
     val bagNumbers: Set<String>? = null,
+    @Schema(
+        description = "Use collectedTime instead.",
+        deprecated = true,
+    )
     val collectedDate: LocalDate? = null,
+    @Schema(
+        description =
+            "Date and time the seeds were collected, in UTC. Optional for now to stay " +
```
sgrimm: The "in UTC" part is a little confusing to me. An `Instant` isn't in UTC; it's time-zone independent.

Is this saying that the client needs to override the local time zone on this timestamp with UTC before rendering it in the payload, as opposed to either including the original timezone suffix or _converting_ to UTC?

If so, it might be better to make this a `ZonedDateTime` so you can reject non-UTC timestamps. With `Instant`, if a client passes in, say, `2026-06-24T20:21:22-08:00` it will be interpreted as the same point in time that is also known as `2026-06-25T04:21:22Z` in UTC, which I guess is probably not what we want given that the previous PR uses UTC rather than the project/org time zone for its conversions between dates and timestamps.

sgrimm: Also, if you make it a `ZonedDateTime`, the "override the original time zone" logic can live on the server side and clients can use their local time zones (assuming I'm right about overriding being the intent here).

ice1080: We currently have no usages of `ZonedDateTime` in a request payload, but we do have usages of `Instant` in a request payload (`CompletePlotObservationRequestPayload.observedTime`, as well as `CompleteAdHocObservationRequestPayload` and `UpdateObservationRequestPayload.BiomassUpdateOperationPayload` among a few others) which is why I chose Instant. Is it just the description that's confusing, or do we want to change this to `ZonedDateTime` and have others follow suit?

sgrimm: The other `Instant`s in the API don't have the same "time calculations are hardwired as UTC" aspect, so this is a different beast IMO.

sgrimm: Maybe a clarifying question: If a client passes in `2020-01-01T23:00:00-08:00`, what date value should we get when we turn that into a date in one of the SQL queries in the earlier PR? It'll either be 2020-01-01 or 2020-01-02 depending on how the time zones are handled.

ice1080: Perhaps the comment is throwing us off the same page. I'm assuming we store `tide_time` in UTC, but display it to the user in either their time zone or their org's time zone? This feature is supposed to work the same way

sgrimm: At the risk of being annoyingly pedantic: we don't store it in UTC, we store it as a time-zone-independent value (confusingly, PostgreSQL's `TIMESTAMP WITH TIME ZONE` type is actually an absolute timestamp with no information about time zones).

But you're right that when we read one of those values, we render it in the planting site or org's time zone.

For editing and display, there's no problem with doing the same thing here. The thing I'm pushing on is that we do more than just display it: we convert it to a UTC date (in the earlier PR) and use that UTC date value in date-range SQL queries. So, to use my example, if you have a report with an end date of 2020-01-01, you'd end up excluding this accession because its collection time, when it's converted to UTC, is on 2020-01-02.

sgrimm: I should add that I'm assuming we _wouldn't_ want to exclude the accession from the report in that scenario. But maybe we don't care if accessions get moved to prior or following reports if they're right at the date boundary, or maybe we even specifically want it to work that way.

sgrimm (approved):

ice1080:
### Merge activity

* **Jun 25, 4:53 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4658).
* **Jun 25, 4:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4658) rebased this pull request as part of a merge.
* **Jun 25, 6:09 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4658).

PR #4670, opened 2026-06-25
Drop redundant ecoregion IDs

sgrimm:
We were storing three ID columns for ecoregions: a synthetic database-generated
ID, a unique numeric identifier from the dataset (which we were storing as a
string), and an object identifier which likely comes from ArcGIS.

Drop the synthetic ID and the ArcGIS object ID; they're not adding any value
since we already have a canonical unique numeric ID for each ecoregion.


sgrimm:
* **#4672** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4672?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4671** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4671?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4670** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4670?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4670?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0500/V518__EcoregionIds.sql
```diff
@@ -0,0 +1,37 @@
...
+           ebc.botanical_country_id
+    FROM ecoregions e
+    JOIN ecoregion_botanical_countries ebc ON e.id = ebc.ecoregion_id;
+
+TRUNCATE TABLE ecoregion_countries;
+TRUNCATE TABLE ecoregion_botanical_countries;
+
+-- Move the existing IDs out of the way since they might collide with the new ones.
+-- Both the existing and the new IDs are positive integers.
+UPDATE ecoregions SET id = -id;
```
ice1080, line 21: Won't this leave these negative-id ecoregions in place indefinitely, instead of updating them later on when the importer is used?

sgrimm: No, because they're overwritten with the externally-supplied IDs in the line below this. This is just to prevent a situation where you have synthetic IDs 1, 2, 3, which map to external IDs 3, 4, 5. Without this line, the update would fail with a duplicate key error when it went to update the first entry's ID from 1 to 3, because ID 3 would already exist. With this extra update, the transition goes [1,2,3] -> [-1,-2,-3] and then [-1,-2,-3] -> [3,4,5] which doesn't have any collisions.

ice1080: Ah right, I wasn't thinking about the next line clearly

ice1080 (approved):

sgrimm:
### Merge activity

* **Jun 25, 9:17 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4670).
* **Jun 25, 9:17 PM UTC**: @sgrimm merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4670).

PR #4673, opened 2026-06-25
SW-7489 Use collectedTime to derive collectedDate

ice1080:
Update `AccessionStore` to derive `collectedDate` if `collectedTime` is present. Prefer the timezone in this order: user, org, UTC.

Also derive `collectedTime` from `collectedDate` if `collectedTime` is not present. This was partially implemented before, but was not using the corrct timezone priority.

Update the `AccessionImporter` to set the`collectedTime` using the correct timezone.

Waiting to merge till the [FE PRs](https://github.com/terraware/terraware-web/pull/5951) are all confirmed. 

Co-authored-by: Claude Opus 4.8 \<noreply@anthropic.com>

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreDatabaseTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreDatabaseTest.kt
@@ -3,6 +3,7 @@ package com.terraformation.backend.seedbank.db.accessionStore
 import com.terraformation.backend.db.AccessionSpeciesHasDeliveriesException
 import com.terraformation.backend.db.ProjectInDifferentOrganizationException
 import com.terraformation.backend.db.default_schema.FacilityType
+import com.terraformation.backend.db.default_schema.Role
 import com.terraformation.backend.db.nursery.tables.pojos.BatchesRow
 import com.terraformation.backend.db.seedbank.CollectionSource
 import com.terraformation.backend.db.seedbank.DataSource
@@ -28,7 +29,9 @@ import com.terraformation.backend.seedbank.model.AccessionModel
 import com.terraformation.backend.seedbank.model.Geolocation
 import com.terraformation.backend.seedbank.seeds
 import java.math.BigDecimal
+import java.time.Instant
 import java.time.LocalDate
+import java.time.ZoneId
 import java.time.ZoneOffset
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.declaredMemberProperties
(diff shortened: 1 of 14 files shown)


ice1080:
* **#4673** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreDatabaseTest.kt
```diff
@@ -335,4 +338,110 @@ internal class AccessionStoreDatabaseTest : AccessionStoreTest() {
...
+    val collectedTime = Instant.parse("2024-01-16T03:00:00Z")
+
+    val initial = store.create(accessionModel())
+    store.update(initial.copy(collectedTime = collectedTime))
+
+    assertEquals(
+        LocalDate.of(2024, 1, 16),
+        accessionsDao.fetchOneById(initial.id!!)?.collectedDate,
+    )
+  }
```
sgrimm: I get the desire for completeness, but it feels to me like these, and the ones for the time-to-date conversion below, are testing the exact same `AccessionHelper` logic that is tested in `AccessionStoreCreateTest`. Do we need to comprehensively test every permutation in both places? I think it might be adequate to just test that it uses a time zone other than UTC if there's one configured.

Obviously that would mean we wouldn't catch a bug where someone replaced the `AccessionHelper` call with "user's time zone or UTC if none" logic, so it's a tradeoff, but that doesn't seem like a super likely failure mode to me.

I'm mostly concerned about the size (and execution time) of the test suite increasing without much added bug-detection or intended-behavior-documentation power to show for it.

sgrimm (approved):
Nit: "partially" is misspelled in the PR description.

src/test/kotlin/com/terraformation/backend/seedbank/db/accessionStore/AccessionStoreDatabaseTest.kt
ice1080: Perhaps a better way would be a full test of AccessionHelper's methods in its own test file, and then just one non-utc test like you mentioned

ice1080:
Didn't mean to change to draft. Updated tests now

PR #4684, opened 2026-06-26
SW-8541 Add country code to non-accelerator project API

sgrimm:
Projects can have country codes already, but we only allow accelerator
admins to read and write projects' country codes, and then only if the
projects are in the accelerator. Remove that restriction by adding
country code to the non-accelerator API endpoints.

Track country code changes with a new persistent event.

If the project is in the accelerator, the new event triggers an update
of the country code variable, with the same effects on file naming and
so forth that currently happen when an accelerator admin edits a
project's country.


sgrimm:
* **#4684** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4684?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4684?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #4688, opened 2026-06-29
Remove redundant PathVariable parameters

sgrimm:
Spring automatically uses the parameter name as the path variable name, so there's
no need to specify it explicitly. This was flagged as a warning by IntelliJ.


sgrimm:
* **#4688** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4688?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4688?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ice1080 (approved):
Just want to confirm that this doesn't affect openapi's output at all, so that the FE isn't affected by this

sgrimm:
> Just want to confirm that this doesn't affect openapi's output at all, so that the FE isn't affected by this

The "diff OpenAPI docs against staging" step in the CI build shows no differences, so we should be good.

PR #4691, opened 2026-06-29
Allow Created Events to have initial values

ice1080:
We already include fields that can be updated in `FieldsUpdatedPersistentEvent`, allowing the `action` of the payload to contain the `changedFrom` and `changedTo` values. Product has started to request that some event log events be displayed in the FE with their initial values. Create a new `CreatedFieldPayload` to return this and `FieldsCreatedPersistentEvent` to allow for events to specify these fields. If a Created event has none of these, it will behave the same as before.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/eventlog/FieldsCreatedPersistentEvent.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/FieldsCreatedPersistentEvent.kt
@@ -1,0 +1,79 @@
+package com.terraformation.backend.eventlog
+
+import com.terraformation.backend.i18n.Messages
+
+/**
+ * Common interface for events that represent an entity being created with specific initial field
+ * values. This is used to render [PersistentEvent]s into API payloads for event log queries.
+ * Creation events that implement this interface will produce a
+ * [CreatedActionPayload][com.terraformation.backend.eventlog.api.CreatedActionPayload] that
+ * includes the initial field values; those that don't will produce a plain
+ * [CreatedActionPayload][com.terraformation.backend.eventlog.api.CreatedActionPayload] with no
+ * field data.
+ *
+ * The typical pattern is to define a nested class (usually called `Values`) with a nullable
+ * property for each field whose initial value should be surfaced. Implementing [listInitialFields]
+ * is then a matter of scanning those properties and returning an [InitialField] for each non-null
+ * one. Use [createInitialField] as a helper to skip null values cleanly inside `listOfNotNull`.
+ *
+ * Fields that don't need to appear in the event log can simply be omitted from `Values` or left
+ * null, and they won't show up in the payload.
+ *
+ * For example:
+ * ```kotlin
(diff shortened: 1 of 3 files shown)


ice1080:
* **#4692** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4692?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4691** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4691?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4691?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/eventlog/FieldsCreatedPersistentEvent.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import com.terraformation.backend.i18n.Messages
+
+/**
+ * Common interface for events that represent an entity being created with specific initial field
+ * values. This is used to render [PersistentEvent]s into API payloads for event log queries.
+ * Creation events that implement this interface will produce a
+ * [CreatedActionPayload][com.terraformation.backend.eventlog.api.CreatedActionPayload] that
+ * includes the initial field values; those that don't will produce a plain
+ * [CreatedActionPayload][com.terraformation.backend.eventlog.api.CreatedActionPayload] with no
```
sgrimm: The  double `[]` blocks here don't seem right to me.

src/main/kotlin/com/terraformation/backend/eventlog/FieldsCreatedPersistentEvent.kt
```diff
@@ -0,0 +1,79 @@
...
+       */
+      val value: List<String>?,
+  )
+
+  /**
+   * Returns an [InitialField] if [value] is non-null, null otherwise. This would typically be
+   * called in `listOfNotNull` in the [listInitialFields] implementation.
+   */
+  fun createInitialField(fieldName: String, value: List<String>?): InitialField? {
+    return if (value != null) InitialField(fieldName, value) else null
```
sgrimm: Or `value?.let { InitialField(fieldName, it) }` (but this is fine too).

sgrimm (approved):
This is a good approach. Nice.

src/main/kotlin/com/terraformation/backend/eventlog/FieldsCreatedPersistentEvent.kt
ice1080: Ya Claude did this. I fixed it manually and then had it do something else and it came back to this and added the double `[]` blocks back in. I think it doesn't like imports being used only in KDocs, but I've definitely seen that elsewhere. Anyway, fixed.

PR #4696, opened 2026-06-30
SW-8430 Add "Explanation" monitoring plot media type

sgrimm:
We want users to be able to upload additional plot corner photos with captions
explaining why they're placing a corner more than a certain distance from the
system-generated location. Add a new media type for these photos, `Explanation`,
which is sorted along with the other plot corner photos in the activity log
and, like quadrat photos, requires a position (corner) to be set.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
@@ -332,6 +332,12 @@ class ObservationService(
     val initialRow = fetchMediaFilesRow(observationId, monitoringPlotId, fileId)
     val updatedRow = updateFunc(initialRow)
 
+    if (
+        initialRow.typeId == ObservationMediaType.Explanation && updatedRow.caption.isNullOrBlank()
+    ) {
+      throw IllegalArgumentException("Cannot remove caption from Explanation photo")
+    }
+
     if (initialRow != updatedRow) {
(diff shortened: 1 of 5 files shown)


sgrimm:
* **#4696** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4696?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4696?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -260,10 +260,18 @@ class ObservationService(
     if (metadata.geolocation == null && isOriginal) {
       throw IllegalArgumentException("Geolocation is required for original observation photos")
     }
+
+    if (type == ObservationMediaType.Explanation) {
+      if (caption.isNullOrBlank()) {
+        throw IllegalArgumentException("Caption is required for explanation media file")
+      }
```
sgrimm, line 267: Awaiting confirmation from Product that we want this validation check, but I'm 80% sure we do based on the surrounding requirements.

sgrimm: Confirmed, we want this check.

ice1080 (approved):

PR #4703, opened 2026-07-01
Add fileBatchId to upload file endpoints

ice1080:
Add optional `fileBatchId` param to upload file endpoints for Organization Media and Observation Media. Throws `FileBatchNotFoundException` if batch doesn't exist.

If batch id is specified and batch is of type "Splat", don't autogenerate a splat when an organization video is uploaded. The next PR includes the ability to finish a batch upload.

Co-authored-by: Claude Opus 4.8 [noreply@anthropic.com](mailto:noreply@anthropic.com)

PR #4702, opened 2026-07-01
Add endpoint to create a file batch

ice1080:
Add endpoint to create a new file batch by type.

Co-authored-by: Claude Opus 4.8 [noreply@anthropic.com](mailto:noreply@anthropic.com)

PR #4701, opened 2026-07-01
Create file_batches table to group files

ice1080:
The mobile app is working on experimenting with the ability to record sensor data while recording a virtual walkthough video. We want to be able to store this data as a json file linked to the video. We also need a way to know that they are being uploaded together.

Create a `file_batches` table to group file uploads together. Include an asset status and a batch type, with available options only "Splat" for now.

Co-authored-by: Claude Sonnet 4.6 [noreply@anthropic.com](mailto:noreply@anthropic.com)

PR #4704, opened 2026-07-01
Add endpoint to finish a file batch upload

ice1080:
We need a way to finish a batch upload and tell the system that it is ready for processing. Add a new endpoint and event for this.

A later PR will consume this event.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
@@ -13,6 +13,7 @@ import com.terraformation.backend.db.DeviceNotFoundException
 import com.terraformation.backend.db.EntityNotFoundException
 import com.terraformation.backend.db.EventNotFoundException
 import com.terraformation.backend.db.FacilityNotFoundException
+import com.terraformation.backend.db.FileBatchNotFoundException
 import com.terraformation.backend.db.NotificationNotFoundException
 import com.terraformation.backend.db.OrganizationNotFoundException
 import com.terraformation.backend.db.ProjectNotFoundException
@@ -614,7 +615,7 @@ class PermissionRequirements(private val user: TerrawareUser) {
   fun finishUploadingFileBatch(fileBatchId: FileBatchId) {
     user.recordPermissionChecks {
       if (!user.canFinishUploadingFileBatch(fileBatchId)) {
-        throw AccessDeniedException("No permission to finish uploading file batch $fileBatchId")
+        throw FileBatchNotFoundException(fileBatchId)
       }
     }
   }
--- a/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
+++ b/src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
@@ -37,6 +37,7 @@ import com.terraformation.backend.db.default_schema.AutomationId
 import com.terraformation.backend.db.default_schema.DeviceId
 import com.terraformation.backend.db.default_schema.DeviceManagerId
 import com.terraformation.backend.db.default_schema.FacilityId
+import com.terraformation.backend.db.default_schema.FileBatchId
 import com.terraformation.backend.db.default_schema.GlobalRole
 import com.terraformation.backend.db.default_schema.NotificationId
 import com.terraformation.backend.db.default_schema.OrganizationId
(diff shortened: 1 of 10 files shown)


ice1080:
* **#4706** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4706?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4705** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4705?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4704** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4703** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4702** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4701** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/file/FileService.kt
```diff
@@ -222,6 +223,27 @@ class FileService(
     }
   }
 
+  /**
+   * Marks a file batch as finished uploading. This sets the batch's asset status to
+   * [AssetStatus.Ready] and publishes a [FileBatchFinishedUploadingEvent].
+   */
+  fun finishUploadingFileBatch(fileBatchId: FileBatchId) {
+    val rowsUpdated =
```
ice1080, line 234: Should this be blocked for anyone but the same user that created the batch? And the system user? Or leave it open?

sgrimm: Yeah, requiring the same user (or system user) probably makes sense here to prevent people from messing with other people's batches.

sgrimm: Maybe allow super-admin users too, so we can more easily deal with any "Customer X uploaded a video but the splat was never generated" support requests.

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
```diff
@@ -610,6 +611,14 @@ class PermissionRequirements(private val user: TerrawareUser) {
     }
   }
 
+  fun finishUploadingFileBatch(fileBatchId: FileBatchId) {
+    user.recordPermissionChecks {
+      if (!user.canFinishUploadingFileBatch(fileBatchId)) {
+        throw AccessDeniedException("No permission to finish uploading file batch $fileBatchId")
```
sgrimm: Should this throw `FileBatchNotFoundException` instead?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/customer/model/PermissionRequirements.kt
ice1080: We seem to have both patterns in this file. 83 instances of `NotFound`, and 173 instances of `AccessDenied`. I'll switch to NotFound, but just mentioning

sgrimm: How it's supposed to work is that if you have read access but not write access, you get access denied. If you don't have read access, you get not found, because you shouldn't be able to know that the resource exists at all if you can't see it. In this case, there's no separate read permission for file batches, so either you can finish the batch or you can't see it at all.

Wouldn't be shocking to learn that we've gotten it wrong in a few places, but that's why there are a bunch of both.

ice1080: Ah, that's very helpful to know! I also now see this documented in the kdoc, so that's my fault for not checking there first to understand the behavior. I was following other write operations, but you're right that this is a slightly special case and should be treated more like a read op than a write op

ice1080:
### Merge activity

* **Jul 8, 6:13 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4704).

PR #4706, opened 2026-07-01
Generate splats with data files when batch finishes

ice1080:
Add an optional `data_file_id` to the splats table to be able to link to the new json file that can be uploaded with videos.

When a Splat file batch finishes uploding, generate the splat that was stopped from auto-generating when the video was uploaded. Include the dataFieId from the file batch when inserting the splat row.


ice1080:
* **#4706** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4706?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4706?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4705** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4705?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4704** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4703** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4702** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4701** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/main/resources/db/migration/0500/V531__SplatDataFileId.sql
ice1080, line 1: This can also be folded into the migration from a few PRs previous if preferred.

ice1080:
I'm operating on the assumption that splatter will not fail with the new field in the request message (`dataFile`) and will just not consume it. If that's incorrect, I can update splatter first and make this PR dependent on that one being deployed.

src/main/resources/db/migration/0500/V531__SplatDataFileId.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE splats ADD COLUMN data_file_id BIGINT REFERENCES files ON DELETE SET NULL;
```
sgrimm, line 1: "Data file" seems kind of broad to me, though I get that we don't want the API to be too prescriptive. I'm just thinking of a future scenario where we decide we want _two_ additional files: which one of them would be the "data file?"

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -537,6 +549,64 @@ class SplatService(
...
+    val batchFiles =
+        dslContext
+            .select(FILES.ID, FILES.CONTENT_TYPE)
+            .from(FILES)
+            .where(FILES.FILE_BATCH_ID.eq(event.fileBatchId))
+            .fetch()
+
+    val videoFileId =
+        batchFiles.firstOrNull { it[FILES.CONTENT_TYPE]?.startsWith("video/") == true }?.value1()
+            ?: return
```
sgrimm: Would we want to log an error here? This would indicate a client bug, I assume: there should never be a batch of type `Splat` with no video file.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -537,6 +549,64 @@ class SplatService(
...
+            .where(ORGANIZATION_MEDIA_FILES.FILE_ID.eq(videoFileId))
+            .fetchOne(ORGANIZATION_MEDIA_FILES.ORGANIZATION_ID) ?: return
+
+    if (batchFiles.size != 2) {
+      log.warn(
+          "File batch ${event.fileBatchId} contains an organization video but does not have " +
+              "exactly two files; not generating splat"
+      )
+      return
+    }
```
sgrimm: This will make the supplementary file mandatory, and if the client doesn't want to supply one, it has to skip creating a file batch. Which maybe is fine, but I wonder if it's better to accept video-file-only batches so the client can use the same code path (create batch, upload, finish batch) whether or not it has the second file.

sgrimm (changes requested):
The implementation seems fine generally, but requesting changes for the naming question I raise below. If you disagree with my comment, I'll approve this.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
ice1080: Good thought

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
ice1080: That's fine with me, I'll change this to no more than 2 since we currently have no purpose for more than 2 files. If we want to change that in the future we can come back to this

src/main/resources/db/migration/0500/V531__SplatDataFileId.sql
ice1080: I've been thinking about that, though another challenge then is distinguishing them on the splatter side (unless we only allow one json file). Obviously the files could be read to try to determine what's in them, that just seemed slightly hacky. 

My thought is that this data file will have general metadata about the info in it (e.g. sensors, sample rate, etc), as well as the sensor data itself, and that if we're adding other data then we can just add it to the json. 

I'm mainly wanting to avoid a name such as `json_file_id` or `sensor_file_id` since we don't know for sure yet if this will end up being json, and we also don't want to limit this file to only sensor data. I'm open to other names though, and will try to think of others

ice1080: Another option, if we wanted more than one additional data file in the future, we could create a `splat_data_files` join table to allow more than one, and just migrate existing `data_file_id`s to that

sgrimm (approved):
Tabling the "data file" question until we have a second supplementary file seems reasonable to me.

ice1080:
### Merge activity

* **Jul 8, 6:42 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4706).

PR #4715, opened 2026-07-01
Enable Graphite CI Optimization

alexTerraformation:
Before any other steps, check with Graphite if CI should run

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -64,6 +64,18 @@ agents:
   queue: "x86-8core"
 
 steps:
+  - label: ":graphite: CI optimizer"
+    key: "graphite-ci-optimizer"
+    soft_fail: true
+    plugins:
+      - *shallow-clone
+      - *load-secrets
+      - withgraphite/graphite-ci#main:
+          graphite_token: "${GRAPHITE_CI_OPTIMIZATION_TOKEN}"
+
+  - wait: ~
+    continue_on_failure: true
+
   - label: ":gradle: Build and test"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -64,6 +64,18 @@ agents:
...
+    key: "graphite-ci-optimizer"
+    soft_fail: true
+    plugins:
+      - *shallow-clone
+      - *load-secrets
+      - withgraphite/graphite-ci#main:
+          graphite_token: "${GRAPHITE_CI_OPTIMIZATION_TOKEN}"
+
+  - wait: ~
+    continue_on_failure: true
```
sgrimm, line 77: Won't `continue_on_failure: true` cause CI to always run even if the CI optimizer step says not to?

Looking at the Graphite docs https://graphite.com/docs/stacking-and-ci#option-2-graphite-job-runs-first I don't see that option.

alexTerraformation: I don't think so. Both a "run" and a "skip" from Graphite exit with 0. The continue_on_failure is only meant to catch failures to reach Graphite's service. https://github.com/withgraphite/graphite-ci-buildkite-plugin/blob/main/hooks/pre-checkout#L55

But you're right, Graphite's docs don't do that so I'm fine with removing it if you'd prefer.

sgrimm: Ah, okay, that makes sense. This is fine as is, then, I think.

sgrimm (approved):

PR #4719, opened 2026-07-02
Don't use Graphite plugin with AWS secrets

alexTerraformation:
Graphite's plugin requires tokens to be available at pipeline load time, but AWS secrets are loaded at runtime. Switch to bash script (same as other pipeline steps) to avoid this issue.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -70,8 +70,7 @@ steps:
     plugins:
       - *shallow-clone
       - *load-secrets
-      - withgraphite/graphite-ci#main:
-          graphite_token: "${GRAPHITE_CI_OPTIMIZATION_TOKEN}"
+    command: .buildkite/scripts/graphite-optimizer.sh
 
   - wait: ~
     continue_on_failure: true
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -70,8 +70,7 @@ steps:
     plugins:
       - *shallow-clone
       - *load-secrets
-      - withgraphite/graphite-ci#main:
-          graphite_token: "${GRAPHITE_CI_OPTIMIZATION_TOKEN}"
+    command: .buildkite/scripts/graphite-optimizer.sh
```
sgrimm, line 73: Shouldn't this be replacing the step that's currently checked into main? I think this will be a merge conflict since there are already lines added here.

alexTerraformation: Ah that's my bad. Rebased this change on top of main

.buildkite/scripts/graphite-optimizer.sh
```diff
@@ -0,0 +1,70 @@
...
+  exit 0
+fi
+
+body_file=$(mktemp)
+trap 'rm -f "$body_file"' EXIT
+
+request_json=$(jq -n \
+  --arg token "$GRAPHITE_CI_OPTIMIZATION_TOKEN" \
+  --arg repository "${BUILDKITE_REPO:-}" \
+  --arg pr "${BUILDKITE_PULL_REQUEST:-}" \
```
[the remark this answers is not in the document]
sgrimm: This looks like it mirrors what the original plugin does.

sgrimm (approved):

PR #4721, opened 2026-07-02
Stop repeatedly running plot-placement tests

sgrimm:
When the code for placing monitoring plots on a planting site map was new, we
needed to make sure it was robust, and since it involves randomness, we did that
by running some of the test cases multiple times (up to 25).

That code has been stable for long enough that the value we get from repetition
isn't worth the extra test execution time. Remove the repetition, but keep the
tests around so we can detect if a future change completely breaks plot placement.


sgrimm:
* **#4721** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4721?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #4727, opened 2026-07-06
Improve CountryDetector large geometry handling

sgrimm:
Previously, if we ran CountryDetector on a very large geometry such as one of
the bigger ecoregions, it was theoretically possible for it to leave out a
covered country if the country was too small to account for a minimum
percentage of the geometry's area (3% by default). Change the minimum coverage
percentage to use the country's area rather than the geometry's area in cases
where the country is smaller than the geometry. For sub-country-sized geometries,
it continues to work the same as before.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
+++ b/src/main/kotlin/com/terraformation/backend/gis/CountryDetector.kt
@@ -78,6 +78,10 @@ class CountryDetector {
     result
   }
 
+  val countryAreas: Map<String, Double> by lazy {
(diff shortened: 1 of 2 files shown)


sgrimm:
* **#4727** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4727?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4727?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523 (approved):

PR #4750, opened 2026-07-10
SW-8649 Add description to includeZones parameter to clarify

ice1080:
Add a description to the `includeZones` parameter because at first glance it can seem like it's an old term that needs renamed

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/api/PlantingSitesController.kt
@@ -41,7 +41,6 @@ import com.terraformation.backend.tracking.model.SubstratumHistoryModel
 import com.terraformation.backend.tracking.model.SubstratumModel
 import com.terraformation.backend.util.toMultiPolygon
 import io.swagger.v3.oas.annotations.Operation
-import io.swagger.v3.oas.annotations.Parameter
 import io.swagger.v3.oas.annotations.media.ArraySchema
 import io.swagger.v3.oas.annotations.media.Schema
 import jakarta.ws.rs.BadRequestException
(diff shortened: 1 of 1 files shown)


ice1080:
* **#4750** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4750?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4750?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tommylau523:
I thought we kept this only for backwards compatibility in the first place. includeZones to include plantingZones variables to handle API field name migrations. I don't think we ever set it to true in new use cases.

ice1080:
> I thought we kept this only for backwards compatibility in the first place. includeZones to include plantingZones variables to handle API field name migrations. I don't think we ever set it to true in new use cases.

​This parameter was just added in March, but I couldn't figure out why it was named `includeZones` instead of `includeStrata`, because it happened after the rename :thinking_face:

ice1080:
No you're totally right. I didn't look at the BE close enough

tommylau523:
> > I thought we kept this only for backwards compatibility in the first place. includeZones to include plantingZones variables to handle API field name migrations. I don't think we ever set it to true in new use cases.
> 
> ​This parameter was just added in March, but I couldn't figure out why it was named `includeZones` instead of `includeStrata`, because it happened after the rename :thinking_face:

Hahaha I think it was specifically created for the rename! Mostly for mobile apps which may still have versions that rely on plantingZones.

tommylau523 (approved):

ice1080:
### Merge activity

* **Jul 10, 11:36 PM UTC**: @ice1080 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4750).

PR #4753, opened 2026-07-14
SW-8654 Fix delete seasons with withdrawals for request

ice1080:
The check constraint on `nursery.withdrawals` - `withdrawals_planting_date_request_season` - checks that `scheduled_planting_date_request_id IS NULL OR planting_season_id IS NOT NULL`, which is throwing an error when deleting a planting season that has a withdrawal tied to a scheduled planting date request because the check constraint is violated by the season delete before the cascade deletes the request. Delete the planting date requests (which use `on delete set null`) before deleting the season to fix this issue.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonStore.kt
+++ b/src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonStore.kt
@@ -9,6 +9,7 @@ import com.terraformation.backend.db.default_schema.OrganizationId
 import com.terraformation.backend.db.tracking.PlantingSeasonId
 import com.terraformation.backend.db.tracking.PlantingSeasonStatus
 import com.terraformation.backend.db.tracking.PlantingSiteId
+import com.terraformation.backend.db.tracking.tables.references.PLANTING_DATE_REQUESTS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SEASONS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SEASON_SPECIES_TARGETS
 import com.terraformation.backend.db.tracking.tables.references.PLANTING_SITES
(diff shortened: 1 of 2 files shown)


ice1080:
* **#4753** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonStore.kt
```diff
@@ -257,6 +258,12 @@ class PlantingSeasonStore(
     val organizationId =
         parentStore.getOrganizationId(id) ?: throw PlantingSeasonNotFoundException(id)
 
+    // Delete requests related to the season before the season to meet check constraint requirements
+    dslContext
+        .deleteFrom(PLANTING_DATE_REQUESTS)
```
sgrimm, line 263: Is this a sign the foreign key should be `ON DELETE CASCADE` instead of `ON DELETE SET NULL`?

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/plantingmanagement/db/PlantingSeasonStore.kt
ice1080: The PK for `scheduled_planting_dates` and `planting_date_requests` is already `ON DELETE CASCADE`, but we don't want withdrawals that are tied to requests to get deleted when a season is deleted, so the withdrawals have an FK on `planting_date_requests` that is `ON DELETE SET NULL`. 

From this error, it seems like postgres is deleting the season, then validating the constraints (where the failure is happening), then cascading the delete to the other tables with PK on season.

sgrimm: Ah, got it. Sounds like that's arguably a PostgreSQL bug, then (it's not taking the constraints into account correctly when it figures out what order to do things during a delete).

ice1080: If AI is to be believed, I asked it about this and it said that these are basically postgres action triggers on the season delete, which don't guarantee any ordering (at least not the type of ordering we're wanting). So the `CASCADE` delete may or may not run before the `SET NULL` on `withdrawals`, and this type of thing can happen when not careful. Alternatives included making our own delete triggers, though I'm not sure adding that type of DB trigger complexity is worth it. I'm open to it though

ice1080: According to this link regarding trigger creation:  
\> If multiple triggers of the same kind are defined for the same event, they will be fired in alphabetical order by name.

https://www.postgresql.org/docs/current/sql-createtrigger.html  
  
So alternatively we could create our own triggers instead of the default `ON DELETE` triggers, and name them in a way that runs in the correct order, but that seems a bit fragile

sgrimm: The approach here seems better than that kind of fiddling.

PR #4771, opened 2026-07-20
SW-8692 Read media for splat annotations

ice1080:
Update the GET splat annotations endpoint to include file info for attached media.

Co-Authored-By: Claude Opus 4.8 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

---


src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -683,15 +686,47 @@ class SplatService(
...
+              LABEL,
+              POSITION,
+              CAMERA_POSITION,
+              splatAnnotationMediaMultiset,
+          )
           .from(SPLAT_ANNOTATIONS)
           .where(FILE_ID.eq(fileId))
-          .fetch { record -> SplatAnnotationModel.of(record) }
+          .fetch { record ->
+            SplatAnnotationModel.of(record).copy(media = record[splatAnnotationMediaMultiset])
```
sgrimm: Could skip the extra copy here by passing the multiset field as an optional parameter to `SplatAnnotationModel.of()`.

src/test/kotlin/com/terraformation/backend/splat/api/SplatAnnotationPayloadTest.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.terraformation.backend.db.default_schema.SplatAnnotationId
+import com.terraformation.backend.splat.CoordinateModel
+import com.terraformation.backend.splat.ExistingSplatAnnotationModel
+import com.terraformation.backend.splat.SplatAnnotationMediaModel
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class SplatAnnotationPayloadTest {
+  @Test
+  fun `maps annotation media list into the payload preserving order`() {
```
sgrimm: It's fast so it's pretty harmless, but this seems like it's not a super useful test. We don't unit test model-to-payload conversions anywhere else in the code and there's no complex logic in this case that'd be at risk of breaking.

src/test/kotlin/com/terraformation/backend/splat/SplatServiceTest.kt
```diff
@@ -2010,4 +2010,90 @@ class SplatServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
+          position = 0,
+      )
+
+      val info = service.getObservationSplatInfo(observationId, fileId)
+
+      assertEquals(
+          listOf(SplatAnnotationMediaModel(photoFileId, "image/jpeg", 0)),
+          info.annotations.single().media,
+      )
+    }
```
sgrimm: This test is redundant IMO (what it's testing is fully covered by the test below it.)

sgrimm: Actually, so is the "returns empty media list" test above this one.

sgrimm (approved):

PR #4783, opened 2026-07-21
SW-8694 Add org/project species checked times

sgrimm:
To allow clients to show the last time a species check was run on a project (or on
the organization if it has no projects), add a "species checked" timestamp to both
the project and organization data and return it as part of the organization and
project data, both in the dedicated APIs and the search API.

Nothing populates the value yet.

---


sgrimm:
Converting this to a draft until we've settled some requirements questions; it's possible this timestamp needs to go on the `project_species` table instead.

sgrimm:
Turns out we don't need to track this; we will show the timestamp of the data source instead.

PR #4803, opened 2026-07-27
Convert CSV line endings to CRLF in build process

sgrimm:
Rather than relying on `.gitattributes` to ensure that CSV files use
CRLF line endings, write CRLF line endings as part of the build process.
This gives us consistent line endings in all our source files, and also
plays nicer with jj, which doesn't yet support `.gitattributes`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-src/main/resources/csv/*.csv eol=crlf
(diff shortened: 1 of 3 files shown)


sgrimm:
@codex review

tommylau523 (approved):

PR #4804, opened 2026-07-27
Move numbered DB migrations to subdirectory

sgrimm:
To reduce clutter in the db/migration directory, move all the historical
numbered migrations to a subdirectory. We'll add new migrations using
the new [year]/[month] folder structure defined in commit 0626f22b.


ice1080 (approved):
Approving without seeing CI due to the node issue

sgrimm:
The Node install is working now, but CI is failing on the Spotless format check of SQL files. Which is odd since this change is straight renames of the existing files, no content modifications. Will investigate.

sgrimm:
CI failure was because Spotless is configured to do format checks on SQL files that have changed compared to `main`, and this PR causes all the files to appear to have been changed. Commented out the SQL Spotless check temporarily and will reenable it after this gets merged.

PR #4810, opened 2026-07-29
SW-8694 Add endpoint to accept pending nativities

sgrimm:


src/main/kotlin/com/terraformation/backend/species/db/ProjectSpeciesStore.kt
```diff
@@ -295,6 +297,40 @@ class ProjectSpeciesStore(
...
+  ) {
+    requirePermissions { createSpecies(organizationId) }
+
+    projectIds?.forEach { projectId ->
+      requirePermissions { readProject(projectId) }
+
+      if (parentStore.getOrganizationId(projectId) != organizationId) {
+        throw ProjectInDifferentOrganizationException()
+      }
+    }
```
ice1080, line 312: Minor downside of doing 2N queries for N project ids. Alternative would be to get all org ids for all project ids in one query, not sure if worth it or not

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/species/db/ProjectSpeciesStore.kt
sgrimm: N will usually be low single digits, so I'm not too worried about this one.

PR #4812, opened 2026-07-29
SW-8713 Renamed reports search table to seed fund reports

tommylau523:


ice1080 (commented):
Won't this cause the FE to break?

tommylau523:
> Won't this cause the FE to break?

​Our SeedFundReport does not use search at all.

sgrimm (approved):

PR #4813, opened 2026-07-29
SW-8713 Add accelerator reports to search

tommylau523:
Add a "reports" search table for accelerator reports, reachable both as a
root prefix and as a "reports" sublist of projects.

Report indicators will come later.

---


src/main/kotlin/com/terraformation/backend/search/table/SearchTables.kt
```diff
@@ -89,6 +89,7 @@ class SearchTables(clock: Clock) {
   val projectVariables = ProjectVariablesTable(this)
   val projects = ProjectsTable(this)
   val recordedTrees = RecordedTreesTable(this)
+  val reports = ReportsTable(this)
```
ice1080: Would it be worth naming these `projectReports` or `acceleratorReports` so we don't run into a similar thing again of having to distinguish report types?

ice1080 (commented):
The other changes look good, but the graphite comment seems valid

ice1080 (approved):

PR #4816, opened 2026-07-29
SW-8713 Add indicators to search

tommylau523:
Add search tables for the three kinds of indicators themselves, holding
the details that describe an indicator: name, reference ID, unit,
description, classification, and decimal places. Reports' indicator
values reference them through an "indicator" sublist, and each indicator
exposes its report values through a "reportIndicators" sublist.

Common and auto-calculated indicators are the same for all projects and
aren't restricted. Project indicators are restricted to the users who
can read the project's reports.

---


src/test/kotlin/com/terraformation/backend/accelerator/AcceleratorIndicatorSearchTest.kt
```diff
@@ -0,0 +1,476 @@
...
+    SearchFieldPrefix(searchTables.autoCalculatedIndicators)
+  }
+
+  private lateinit var organizationId: OrganizationId
+  private lateinit var projectId: ProjectId
+
+  @BeforeEach
+  fun setUp() {
+    organizationId = insertOrganization()
+    insertOrganizationUser(userId = inserted.userId, role = Role.Admin)
```
sgrimm, line 52: Harmless to specify it, but you could leave the user ID out here since `inserted.userId` is the default.

sgrimm (approved):
This seems reasonable. General comment, though I don't think it's merge-blocking: Claude, as it often does, generated what I think is an excessive number of test cases. I think several of the tests could be combined, or even eliminated entirely.

PR #4817, opened 2026-07-30
Fix PlantingSeasonStoreTest failures

sgrimm:
The test was relying on planting seasons being returned in a particular
order, but the underlying database query didn't specify an order.


ice1080 (approved):
Thanks for fixing, I need to be more vigilant looking for these when writing my tests/stores

PR #4820, opened 2026-07-31
SW-8718 Use cheaper completion subquery for subareas

sgrimm:
If the user requests site-, stratum-, or substratum-level results,
we previously still pulled the plot-level data from the database
because it was needed in order to tell whether or not the plots
in the larger area had completed their observations.

Replace the plot-level query with a much more streamlined check
for plot completion in cases where the client doesn't request
plot-level results.


sgrimm:
In local testing on a complex planting site with a bunch of observations, this cut execution time of the "fetch stratum results" query by more than half. Obviously the perf difference will depend on the shape of the data.

tommylau523 (approved):

PR #4854, opened 2026-08-07
Move planting site locking function to separate class

sgrimm:
To allow multiple store classes to lock a planting site while they work,
move the lock function to its own service.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteLocker.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteLocker.kt
@@ -1,0 +1,36 @@
+package com.terraformation.backend.tracking.db
+
+import com.terraformation.backend.customer.model.requirePermissions
+import com.terraformation.backend.db.tracking.PlantingSiteId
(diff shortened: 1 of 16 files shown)


src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteLocker.kt
ice1080, line 1: Do we have any other lockers? If so we could make the class name more generic since the method name makes it obvious what is being locked

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/tracking/db/PlantingSiteLocker.kt
sgrimm: We have `ObservationLocker`. I guess it'd make sense to combine them into an `EntityLocker` class.

ice1080: Doesn't have to be done in this PR, just was mentioning it

PR #4855, opened 2026-08-07
SW-8743 Add planting site species targets tables

sgrimm:
We want users to be able to set per-species plant count targets on
planting sites, and also to be able to mark which species should be
planted in which strata. Add database tables to hold that information
and a store class to work with the tables.


src/main/resources/db/migration/2026/08/V20260806.0943__PlantingSiteSpeciesTargets.sql
```diff
@@ -0,0 +1,29 @@
+CREATE TABLE tracking.planting_site_species_targets (
+    planting_site_id BIGINT NOT NULL REFERENCES tracking.planting_sites ON DELETE CASCADE,
+    species_id BIGINT NOT NULL REFERENCES species ON DELETE CASCADE,
+    target_plants INTEGER,
```
ice1080: Could this number ever be large enough that it warrants `NUMERIC` instead of `INTEGER`?

ice1080 (approved):

src/main/resources/db/migration/2026/08/V20260806.0943__PlantingSiteSpeciesTargets.sql
sgrimm: It seems implausible to me that someone would want to plant more than 2 billion plants in a stratum, but will verify that assumption with product; good flag.

ice1080: This is over the whole site, not just in a single stratum, correct?

sgrimm: Yeah, it's site-level. I'm changing this to BIGINT just to be on the safe side.

PR #4857, opened 2026-08-07
SW-8743 Add planting site species targets to search API

sgrimm:
Expose planting site and stratum species targets in the search API.


src/main/kotlin/com/terraformation/backend/search/table/PlantingSiteSpeciesTargetsTable.kt
```diff
@@ -0,0 +1,56 @@
...
+                  PLANTING_SITE_SPECIES_TARGETS.SPECIES_ID.eq(STRATUM_SPECIES_TARGETS.SPECIES_ID),
+              ),
+          ),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          integerField("targetPlants", PLANTING_SITE_SPECIES_TARGETS.TARGET_PLANTS),
```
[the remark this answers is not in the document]
sgrimm: Ha, caught me. I just ran TrackingSearchTest locally, not the whole suite.

src/main/kotlin/com/terraformation/backend/search/table/StratumSpeciesTargetsTable.kt
```diff
@@ -0,0 +1,64 @@
...
+          ),
+      )
+    }
+  }
+
+  override val fields: List<SearchField> =
+      listOf(
+          idWrapperField("speciesId", STRATUM_SPECIES_TARGETS.SPECIES_ID) { SpeciesId(it) },
+          idWrapperField("stratumId", STRATUM_SPECIES_TARGETS.STRATUM_ID) { StratumId(it) },
+      )
```
ice1080, line 57: Seems slightly odd that this table exposes the `speciesId` and `stratumId`, but `PlantingSiteSpeciesTargetsTable` doesn't expose `speciesId` or `plantingSiteId`. Both have these as single value sublists so both can be retrieved that way, but just noting the inconsistency.

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/search/table/StratumSpeciesTargetsTable.kt
sgrimm: Added a comment explaining why these are here.

ice1080: Couldn't the same be said about `speciesId` in `PlantingSiteSpeciesTargetsTable`?

sgrimm: Sure, that's fair. I was thinking there's an actual value there (the target plant count) but if it's null, you'd run into the same issue with the search API dropping the empty child object.

sgrimm: Added species ID to the site-level table.

ice1080 (approved):

PR #4859, opened 2026-08-07
Combine PlantingSiteLocker and ObservationLocker

sgrimm:


src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
```diff
@@ -40,7 +41,6 @@ import com.terraformation.backend.tracking.db.InvalidObservationEndDateException
 import com.terraformation.backend.tracking.db.InvalidObservationStartDateException
 import com.terraformation.backend.tracking.db.ObservationAlreadyStartedException
 import com.terraformation.backend.tracking.db.ObservationHasNoSubstrataException
-import com.terraformation.backend.tracking.db.ObservationLocker
```
tommylau523, line 43: Are we not deleting the existing lockers?

tommylau523 (approved):

src/main/kotlin/com/terraformation/backend/tracking/ObservationService.kt
sgrimm: Huh, that's weird, they should have been deleted but the deletions seem to have been dropped somewhere along the way. Thanks for catching it.

PR #4862, opened 2026-08-10
Fix incorrect sublist in InternalTagsTable

sgrimm:
The sublist's join condition was correct, but it was using the wrong
SearchTable.


ice1080 (approved):
Might be worth a unit test at some point, but obviously that changes this one line PR to many more

sgrimm:
I did consider writing a test for this but I think we need to either decide we're okay with the search API tests not fully covering every field and every sublist of every table (the current situation) in which case not adding a test here doesn't make things any worse than they already were, or we need to write a _lot_ more test code to get full coverage.

PR #4867, opened 2026-08-13
TWS-1214 Allow local accession collected times

sgrimm:
The `collectedTime` field of the create-accession payload is an `Instant` which
requires that the client include a time zone in the value, but there's an issue
with some mobile app versions that is causing them to omit the time zone. Add a
custom deserializer that defaults the time zone to UTC if not supplied.


ice1080 (approved):
Do we want any tests on the deserializer?

sgrimm:
Added a test for the payload, rather than a generic one for the new deserializer, so we can prove that the deserializer is actually being used.

PR #4868, opened 2026-08-14
SW-7604 Add WithdrawalCreatedEventV2 with creation-time field values

tommylau523:
V2 implements WithdrawalEvent as a FieldsCreatedPersistentEvent and adds   
three fields. V1 becomes an UpgradableEvent whose toNextVersion fills them   
in, so existing rows are upgraded and rewritten on read rather than by a migration.

Co-Authored-By: Claude Opus 5 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)


tommylau523:
* **#4868** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4868?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#4869](https://github.com/terraware/terraware-server/pull/4869) <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4869?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#4870](https://github.com/terraware/terraware-server/pull/4870) <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4870?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/seedbank/event/WithdrawalPersistentEvent.kt
```diff
@@ -23,21 +27,73 @@ sealed interface WithdrawalPersistentEvent : PersistentEvent {
...
     override val accessionId: AccessionId,
     override val facilityId: FacilityId,
     override val organizationId: OrganizationId,
-) : EntityCreatedPersistentEvent, WithdrawalPersistentEvent
+) : FieldsCreatedPersistentEvent, WithdrawalPersistentEvent {
+  override fun listInitialFields(messages: Messages) =
+      listOfNotNull(
+          createInitialField("date", date.toString()),
+          createInitialField("withdrawnQuantity", withdrawnQuantity?.toString()),
+          createInitialField("purpose", purpose?.jsonValue),
```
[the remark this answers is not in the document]
tommylau523: FE should localize instead.

src/main/kotlin/com/terraformation/backend/seedbank/event/WithdrawalPersistentEvent.kt
```diff
@@ -23,21 +27,73 @@ sealed interface WithdrawalPersistentEvent : PersistentEvent {
...
     val purpose: WithdrawalPurpose? = null,
     val date: LocalDate,
     val withdrawnQuantity: SeedQuantityModel? = null,
     val batchId: BatchId? = null,
     val notes: String? = null,
     val staffResponsible: String? = null,
+    val withdrawalId: WithdrawalId,
+    val accessionId: AccessionId,
+    val facilityId: FacilityId,
+    val organizationId: OrganizationId,
```
ice1080: I don't think an existing persistent event can add new non-nullable values

tommylau523: The diffs made it look confusing but no new fields added here. It did change into an UpgradableEvent which I wasns’t entirely sure if that is correct.

ice1080: You're right, that's my bad. I see the rest of the values now. Changing it to an `UpgradableEvent` is correct, but I think it actually needs to still extend `WithdrawalPersistentEvent`, meaning that those values should still be overridden (withdrawalId, accessionId, facilityId, organizationId)

tommylau523: I took a look in the existing code, both existing usages of UpgrdableEvent, ProjectRenamedEventV1 and OrganizationRenamedEventV1 do not extend their respective versions of PersistentEvent. But the test cases do use that pattern. 

Seems like extending the PersistentEvent interface will not affect how we store events, but does add some clarity. I have made the change!

src/main/kotlin/com/terraformation/backend/seedbank/event/WithdrawalPersistentEvent.kt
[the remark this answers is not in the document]
sgrimm: Codex is right. The contract for the event log API says the server does localization, and it localizes other enum values already. See, for example, `ViabilityTestUpdatedEventV1`.

The idea is that the client doesn't need to have special-case rendering code for specific fields on specific event classes; it can just display whatever string values the server sends back.

src/main/kotlin/com/terraformation/backend/seedbank/event/WithdrawalPersistentEvent.kt
```diff
@@ -35,9 +38,62 @@ data class WithdrawalCreatedEventV1(
...
+    val withdrawnByUserId: UserId? = null,
+    override val withdrawalId: WithdrawalId,
+    override val accessionId: AccessionId,
+    override val facilityId: FacilityId,
+    override val organizationId: OrganizationId,
+) : FieldsCreatedPersistentEvent, WithdrawalPersistentEvent {
+  override fun listInitialFields(messages: Messages) =
+      listOfNotNull(
+          createInitialField("date", date.toString()),
+          createInitialField("withdrawnQuantity", withdrawnQuantity?.toString()),
```
sgrimm: May as well alphabetize these since they already don't match the order of the constructor arguments.

src/main/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtils.kt
```diff
@@ -47,4 +52,67 @@ class EventUpgradeUtils(
...
+
+    val laterEdits =
+        eventLogStore
+            .fetchByIdsSince(
+                mapOf(withdrawalId to createdEventLogId),
+                listOf(WithdrawalUpdatedEvent::class),
+            )
+            .map { it.event }
+
+    return WithdrawalValuesMissingFromV1(
```
sgrimm: (Not a merge-blocking comment) This makes me wonder if, for cases like this where multiple pieces of data need to be fetched, we should instead move the V1-to-V2 conversion logic entirely into this class, such that instead of `getWithdrawalValuesMissingFromV1()` you'd have something like
```
fun upgrade(original: WithdrawalCreatedEventV1): WithdrawalCreatedEventV2 {
  // fetch extra data from DB, return populated V2 event
}
```
and then the upgrade function in the V1 class is just a call to that function.

Not 100% sure that's the right direction long term, but I worry that we'll end up having to define more and more of these little bundle-of-data classes over time which will get cluttered.

sgrimm (changes requested):

src/main/kotlin/com/terraformation/backend/seedbank/event/WithdrawalPersistentEvent.kt
ice1080: Thanks for checking the code and updating! @sgrimm this is surprising to me that other `UpgradableEvent`s don't extend their versions of PersistentEvent. Is that intentional?

sgrimm: Yes, the idea is that because the event-category-specific `PersistentEvent` subinterfaces are sealed, you can use them to guarantee comprehensiveness in `when` expressions and such.

Adding those interfaces to obsolete event classes would mean you could no longer rely on the typealiases and do `when(event) { is MyEntityChangedEvent -> foo }` but would instead have to explicitly list the V1, V2, etc. variants. You'd have to edit all the comprehensiveness-checking use sites everywhere whenever you added a new version, despite the fact that in reality, none of those use sites will ever need to handle an outdated object because the event log code will upgrade it to the latest version before returning it to callers.

ice1080: Got it, thanks for explaining! @tommylau523 I leave it up to you for if you want to revert with the other changes or not.

src/main/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtils.kt
```diff
@@ -47,4 +53,72 @@ class EventUpgradeUtils(
             ?.name
         ?: throw ProjectNotFoundException(projectId)
   }
+
+  inner class WithdrawalValuesMissingFromV1 {
```
sgrimm: What does this inner class buy us?

src/main/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtils.kt
```diff
@@ -47,4 +53,72 @@ class EventUpgradeUtils(
...
+                  WITHDRAWALS.DESTINATION,
+                  WITHDRAWALS.VIABILITY_TEST_ID,
+                  WITHDRAWALS.WITHDRAWN_BY,
+              )
+              .from(WITHDRAWALS)
+              .where(WITHDRAWALS.ID.eq(original.withdrawalId))
+              .fetchOne()
+
+      // A withdrawal can't be edited before it exists, so every update event for it was published
+      // after the creation event being upgraded.
```
sgrimm: This comment seems redundant to me.

src/main/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtils.kt
```diff
@@ -47,4 +53,72 @@ class EventUpgradeUtils(
...
+                  { it.changedFrom.destination },
+                  { it.changedTo.destination },
+              ),
+          facilityId = original.facilityId,
+          notes = original.notes,
+          organizationId = original.organizationId,
+          purpose = original.purpose,
+          staffResponsible = original.staffResponsible,
+          // A withdrawal's viability test can't be changed after it's created, so the current row
+          // always has the creation-time value.
```
sgrimm: I don't understand what this comment is getting at. It's describing a line of code that has nothing to do with creation times.

src/main/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtils.kt
```diff
@@ -47,4 +53,72 @@ class EventUpgradeUtils(
...
+      laterEdits: List<T>,
+      currentValue: V?,
+      getChangedFrom: (T) -> V?,
+      getChangedTo: (T) -> V?,
+  ): V? {
+    val firstEditOfField = laterEdits.firstOrNull {
+      getChangedFrom(it) != null || getChangedTo(it) != null
+    }
+
+    return if (firstEditOfField != null) getChangedFrom(firstEditOfField) else currentValue
```
sgrimm: Alternately: `firstEditOfField?.let { getChangedFrom(it) } ?: currentValue`

src/test/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtilsTest.kt
```diff
@@ -169,6 +184,188 @@ class EventUpgradeUtilsTest : DatabaseTest(), RunsAsDatabaseUser {
...
+    }
+
+    @Test
+    fun `uses the current withdrawal row when no later edit touched the fields`() {
+      val viabilityTestId = insertViabilityTest()
+      val withdrawalId =
+          insertSeedbankWithdrawal(
+              batchId = null,
+              destination = "Lab",
+              // The withdrawals_test_id_requires_purpose constraint ties the two together.
```
sgrimm: This comment seems redundant to me.

src/test/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtilsTest.kt
```diff
@@ -169,6 +184,188 @@ class EventUpgradeUtilsTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      )
+
+      val upgraded = upgrade(withdrawalId)
+
+      assertEquals("Lab", upgraded.destination, "destination")
+      assertEquals(user.userId, upgraded.withdrawnByUserId, "withdrawnByUserId")
+    }
+
+    // A null changedFrom can mean either "this edit left the field alone" or "this edit gave the
+    // field its first value". Only the changedTo side tells them apart.
```
sgrimm: This is how the fields-updated events all work; doesn't seem useful to re-explain it here.

src/test/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtilsTest.kt
```diff
@@ -169,6 +184,188 @@ class EventUpgradeUtilsTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      )
+
+      assertNull(
+          upgrade(withdrawalId).destination,
+          "Destination was empty at creation time even though the row has one now",
+      )
+    }
+
+    @Test
+    fun `returns nulls if the withdrawal has been deleted`() {
```
sgrimm: This test name references "nulls" but the assertion doesn't verify that anything is null (at least not explicitly).

src/test/kotlin/com/terraformation/backend/eventlog/db/EventUpgradeUtilsTest.kt
```diff
@@ -169,6 +184,188 @@ class EventUpgradeUtilsTest : DatabaseTest(), RunsAsDatabaseUser {
...
+        eventUpgradeUtils.withdrawalValuesMissingFromV1.upgrade(createdEventV1(withdrawalId))
+
+    private fun createdEventV1(withdrawalId: WithdrawalId) =
+        WithdrawalCreatedEventV1(
+            purpose = WithdrawalPurpose.Other,
+            date = withdrawalDate,
+            withdrawalId = withdrawalId,
+            accessionId = accessionId,
+            facilityId = facilityId,
+            organizationId = inserted.organizationId,
```
sgrimm: May as well alphabetize these and the arguments in the V2 function.

src/test/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformerTest.kt
```diff
@@ -566,6 +592,39 @@ class EventLogPayloadTransformerTest : DatabaseTest(), RunsAsDatabaseUser {
     )
   }
 
+  // In English the display names happen to match the enums' database names, so this is the only
+  // coverage that distinguishes a localized value from a raw one.
+  @Test
+  fun `localizes enum values of Created actions`() {
```
sgrimm: This test feels kind of low-value to me. It's testing a real behavior, but it's a behavior of all events that have enum fields. The only bug this would catch would be if we removed the localization code from this specific event class somehow.

If we're going to test for this behavior, IMO it'd be better to do it using a test that dynamically constructs all the enum-carrying event classes and checks all of them. That way we'd catch a newly-added event that wasn't localizing its enums, which is the most likely failure mode here IMO.

sgrimm (changes requested):

sgrimm (approved):

tommylau523:
### Merge activity

* **Aug 28, 10:04 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4868).
* **Aug 28, 10:04 PM UTC**: @tommylau523 merged this pull request with [Graphite](https://app.graphite.com/github/pr/terraware/terraware-server/4868).

PR #4869, opened 2026-08-14
SW-7604 Surface quantity edit notes in the event log

tommylau523:
AccessionQuantityUpdatedEventV1 persists the notes the user entered with a
manual quantity edit, but they were never projected into the API payload, so
clients couldn't show them.

Co-Authored-By: Claude Opus 5 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformer.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/EventLogPayloadTransformer.kt
@@ -177,7 +177,7 @@ class EventLogPayloadTransformer(
       // Events that can self-describe which fields were updated can be transformed generically.
       is FieldsUpdatedPersistentEvent ->
           event.listUpdatedFields(messages).map {
-            FieldUpdatedActionPayload(it.fieldName, it.changedFrom, it.changedTo)
+            FieldUpdatedActionPayload(it.fieldName, it.changedFrom, it.changedTo, it.notes)
           }
 
       // Create and delete actions don't need any event-type-specific arguments because the details
--- a/src/main/kotlin/com/terraformation/backend/eventlog/FieldsUpdatedPersistentEvent.kt
+++ b/src/main/kotlin/com/terraformation/backend/eventlog/FieldsUpdatedPersistentEvent.kt
@@ -72,6 +72,13 @@ interface FieldsUpdatedPersistentEvent : EntityUpdatedPersistentEvent {
        * field is, e.g., a set of enum values. This value should be localized if applicable.
        */
       val changedTo: List<String>?,
+      /**
+       * User-entered notes about why the field changed, if the operation that changed it accepts
+       * notes. These describe the change rather than being a value of the field, so they travel
+       * alongside the values instead of being a field of their own; that keeps them attached to the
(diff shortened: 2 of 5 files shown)


tommylau523:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4869?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4869** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4869?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4869?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4868** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4868?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#4870](https://github.com/terraware/terraware-server/pull/4870) <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4870?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

sgrimm:
FYI, I'm looking at this one. Something about the structure here doesn't sit right with me on an initial skim, but I need to think it through some more before I can either give concrete feedback/suggestions or decide that it's fine as is.

sgrimm:
This event log API change affects _all_ field-updated actions even though it's currently only relevant for one event type. If we think of "notes" as a generic property that various events might have, maybe we'd want it on `EventActionPayload` rather than just on `FieldUpdatedActionPayload` because there's no reason a creation or deletion event couldn't also have notes.

Are "notes" the only additional metadata we'd want to attach to change events, or will we be gradually adding more and more metadata fields to these payloads? Is "notes" the right level of abstraction for the API, or would it be better to represent this in the API as a more generic "human-readable additional details" field that could potentially include text other than human-supplied notes?

Another consideration if we treat this as a generic feature of the event log is that the notes are at the event level, but the API and `FieldsUpdatedPersistentEvent.UpdatedField` expose them at a per-field level, meaning that if we have an event with notes that includes changes to multiple fields, we'd need to either duplicate the notes in all the individual per-field actions or have some kind of convention that the notes only appear on the first action for a particular event. On the API side, this one is tough to address without a larger rework of the event log API since there's currently no concept of a group of related actions that share metadata. But in the internal representation, we could maybe move `notes` up to the `FieldsUpdatedPersistentEvent` interface rather than putting it in the `UpdatedField` data class. That'd at least more closely mirror the underlying structure.

An alternative approach that would sidestep all those questions would be to model this as a change to a field called `notes` the same way we model changes to other fields. That is, there'd be a field-updated action for the quantity change, and then a second field-updated action that says the `notes` field changed from null to whatever string the user supplied. The downside of that approach is that unless we added special-case handling in the FE code, notes would show up as a separate line item in the history.

sgrimm (changes requested):
Marking this as "request changes" based on my feedback from yesterday, but it's really more of a "request discussion."

PR #4875, opened 2026-08-20
SW-8773 Include accession IDs in batch history API

sgrimm:
Add a new batch history entry payload class for transfers of additional
seeds from accessions.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesHistoryController.kt
+++ b/src/main/kotlin/com/terraformation/backend/nursery/api/BatchesHistoryController.kt
@@ -30,7 +30,9 @@ import com.terraformation.backend.db.nursery.tables.pojos.BatchDetailsHistorySub
 import com.terraformation.backend.db.nursery.tables.pojos.BatchQuantityHistoryRow
 import com.terraformation.backend.db.nursery.tables.pojos.BatchWithdrawalsRow
 import com.terraformation.backend.db.nursery.tables.pojos.WithdrawalsRow
+import com.terraformation.backend.db.seedbank.AccessionId
 import com.terraformation.backend.log.perClassLogger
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/terraformation/backend/nursery/api/BatchesHistoryController.kt
```diff
@@ -330,6 +345,35 @@ data class BatchHistoryStatusChangedPayload(
...
+    allOf = [BatchHistoryPayloadCommonProps::class],
+    description = "A transfer from a seed bank that added germinating seedlings to this batch.",
+)
+data class BatchHistoryAddedFromAccessionPayload(
+    val accessionId: AccessionId,
+    val accessionNumber: String,
+    override val createdBy: UserId,
+    override val createdTime: Instant,
+    val germinatingQuantity: Int,
+    override val version: Int,
```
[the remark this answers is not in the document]
sgrimm: Spec says we only allow adding to the germinating quantity, but this is a valid point; will add code to the previous PR in the stack to reject attempts to add to other quantities.

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
ice1080, line 1: Should this be unit tested?

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: Good call. Added a test class for this, and it uncovered a minor bug, so definitely worthwhile.

ice1080 (approved):

PR #4879, opened 2026-08-21
Update data model for splat additional files

ice1080:
We had previously added support for a `data_file_id` for splats, but this assumed there would be only one file. 
Now that we know there will be multiple files, replace this with `splat_additional_files` and an enum for their types. 
Remove the code that was previously reading/writing the `data_file_id` column (and related file batch code).

As a note, `data_file_id` was never used so no data migration was needed.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---


src/main/resources/db/migration/2026/08/V20260821.0801__SplatAdditionalFiles.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE splat_additional_files
+(
+  file_id       BIGINT  PRIMARY KEY REFERENCES files ON DELETE CASCADE,
+  splat_file_id BIGINT  NOT NULL REFERENCES splats ON DELETE CASCADE,
+  type_id       INTEGER NOT NULL REFERENCES splat_additional_file_types,
+
+  UNIQUE (splat_file_id, type_id)
+);
+
+ALTER TABLE splats DROP COLUMN data_file_id;
```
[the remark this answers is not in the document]
ice1080: Will update the description. This field was never used

src/main/resources/db/migration/2026/08/V20260821.0801__SplatAdditionalFiles.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE splat_additional_file_types
+(
+  id   INTEGER PRIMARY KEY,
+  name TEXT NOT NULL
+);
+
+CREATE TABLE splat_additional_files
+(
+  file_id       BIGINT  PRIMARY KEY REFERENCES files ON DELETE CASCADE,
+  splat_file_id BIGINT  NOT NULL REFERENCES splats ON DELETE CASCADE,
```
sgrimm: I don't think `ON DELETE CASCADE` is what we want here. If a splat file is deleted, we need to delete its additional files from the file store, not just their database rows. See `FileService.on(FileReferenceDeletedEvent)`.

src/main/resources/db/migration/R__TypeCodes.sql
```diff
@@ -673,6 +673,13 @@ VALUES (1, 'Name Misspelled'),
        (3, 'Name Is Synonym')
 ON CONFLICT (id) DO UPDATE SET name = excluded.name;
 
+INSERT INTO splat_additional_file_types (id, name)
+VALUES (1, 'frames'),
+       (2, 'imu'),
+       (3, 'session-meta'),
+       (4, 'audio')
```
sgrimm: Might be better to use camel-case identifiers here like `SessionMeta` so the names here match the names of the entries in the generated enum.

sgrimm (changes requested):

src/main/resources/db/migration/2026/08/V20260821.0801__SplatAdditionalFiles.sql
ice1080: Doesn't the second `ON DELETE CASCADE` handle that? The `splat_file_id` is the same file_id as the video file, which is the primary key for the `splats` row

src/main/resources/db/migration/R__TypeCodes.sql
ice1080: The next PR in the stack determines the type based on the file name, and the file name currently comes from the mobile app. I can ask Josh to update the mobile app if we'd prefer that, though need to also update splatter to accept that different enum value as well. Preference?

src/main/resources/db/migration/2026/08/V20260821.0801__SplatAdditionalFiles.sql
sgrimm: Right, I'm saying if we let these rows delete themselves via cascade, we'll effectively end up with orphaned files in S3. There will still be `files` entries for them, but we'll have no way to navigate to them in the data model because they won't be linked to anything else.

ice1080: Yes, you're totally right. I was reading your text of `file store` and thinking database. I was trying to read things too fast since I'm close to time off. Updating now

src/main/resources/db/migration/R__TypeCodes.sql
sgrimm: Having read the other PR now, I'm a little uncertain about what purpose the enum is serving at all. We're basically just using it as a filter on the list of filenames in the batch.

With the enum, if we add a new file type, we need to update the mobile app, terraware-server, and splatter to know about it. If instead terraware-server just passed filenames along to splatter without filtering them, we'd only need to update the mobile app and splatter for a new file type.

Totally possible I'm missing some aspect of this.

src/main/resources/db/migration/2026/09/V20260902.1015__SplatAdditionalFiles.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE splat_additional_files
+(
+    file_id       BIGINT PRIMARY KEY REFERENCES files ON DELETE CASCADE,
+    splat_file_id BIGINT NOT NULL REFERENCES splats,
```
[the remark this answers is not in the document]
ice1080: This is handled in a later PR in the stack

src/main/resources/db/migration/2026/09/V20260902.1015__SplatAdditionalFiles.sql
```diff
@@ -0,0 +1,11 @@
...
+    type          TEXT   NOT NULL,
+
+    UNIQUE (splat_file_id, type)
+);
+
+ALTER TABLE splats
+    DROP COLUMN data_file_id;
```
sgrimm, line 11: We don't care about migrating the existing data files since they're all just test uploads, I'm assuming.

src/test/kotlin/com/terraformation/backend/splat/SplatServiceTest.kt
```diff
@@ -1934,28 +1934,26 @@ class SplatServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      insertOrganizationMediaFile(fileId = videoFileId)
 
       val messageSlot = slot<SplatterRequestMessage>()
       every { sqsTemplate.send(any<String>(), capture(messageSlot)) } returns mockk(relaxed = true)
 
       service.on(FileBatchFinishedUploadingEvent(fileBatchId))
 
-      val splat = dslContext.fetchSingle(SPLATS, SPLATS.FILE_ID.eq(videoFileId))
-      assertEquals(jsonFileId, splat.dataFileId, "Data file ID")
+      dslContext.fetchSingle(SPLATS, SPLATS.FILE_ID.eq(videoFileId))
```
sgrimm: Might be slightly nicer to change this to `fetchOne` and wrap it in an `assertNotNull` so you get an assertion failure rather than a jOOQ exception if the row doesn't exist.

sgrimm (approved):

src/main/resources/db/migration/2026/09/V20260902.1015__SplatAdditionalFiles.sql
ice1080: Correct, that's in the PR description

PR #4880, opened 2026-08-21
Send splat additional files to splatter

ice1080:
When a file batch finishes uploading for a splat video, include the rest of the batch files as part of the payload. Use the file names as their types.
If a filename/type doesn't exist yet, warn and continue with the rest to allow new files to be ready before splatter is ready to consume them.

---

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
+++ b/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
@@ -11,11 +11,13 @@ import com.terraformation.backend.db.default_schema.AssetStatus
 import com.terraformation.backend.db.default_schema.FileBatchType
 import com.terraformation.backend.db.default_schema.FileId
 import com.terraformation.backend.db.default_schema.OrganizationId
+import com.terraformation.backend.db.default_schema.SplatAdditionalFileType
 import com.terraformation.backend.db.default_schema.tables.references.BIRDNET_RESULTS
 import com.terraformation.backend.db.default_schema.tables.references.FILES
 import com.terraformation.backend.db.default_schema.tables.references.FILE_BATCHES
 import com.terraformation.backend.db.default_schema.tables.references.ORGANIZATION_MEDIA_FILES
 import com.terraformation.backend.db.default_schema.tables.references.SPLATS
+import com.terraformation.backend.db.default_schema.tables.references.SPLAT_ADDITIONAL_FILES
 import com.terraformation.backend.db.default_schema.tables.references.SPLAT_ANNOTATIONS
 import com.terraformation.backend.db.default_schema.tables.references.SPLAT_ANNOTATION_MEDIA
 import com.terraformation.backend.db.tracking.MonitoringPlotId
@@ -159,10 +161,11 @@ class SplatService(
       force: Boolean = false,
       params: SplatGenerationParams = SplatGenerationParams(),
       runBirdnet: Boolean = true,
+      additionalFiles: Map<FileId, SplatAdditionalFileType> = emptyMap(),
   ) {
     ensureOrganizationMediaFile(organizationId, fileId)
 
-    generateSplat(organizationId, fileId, force, params, runBirdnet)
+    generateSplat(organizationId, fileId, force, params, runBirdnet, additionalFiles)
   }
 
   fun readOrganizationSplat(organizationId: OrganizationId, fileId: FileId): SizedInputStream {
@@ -407,6 +410,7 @@ class SplatService(
       force: Boolean = false,
       params: SplatGenerationParams,
       runBirdnet: Boolean = false,
+      additionalFiles: Map<FileId, SplatAdditionalFileType> = emptyMap(),
   ) {
     val videoUrl =
         dslContext.fetchValue(FILES.STORAGE_URL, FILES.ID.eq(fileId))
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -482,9 +486,25 @@ class SplatService(
...
+            additionalFiles.forEach { (additionalFileId, type) ->
+              dslContext
+                  .insertInto(SPLAT_ADDITIONAL_FILES)
+                  .set(FILE_ID, additionalFileId)
+                  .set(SPLAT_FILE_ID, fileId)
+                  .set(TYPE_ID, type)
+                  .execute()
+            }
+          }
+        }
```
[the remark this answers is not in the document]
ice1080: Until we need different behavior, the current behavior intentionally only deletes files that are being replaced, so this is intentional.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -482,9 +486,25 @@ class SplatService(
       }
 
       if (rowsInserted == 1 || force) {
+        if (additionalFiles.isNotEmpty()) {
+          with(SPLAT_ADDITIONAL_FILES) {
+            dslContext.deleteFrom(SPLAT_ADDITIONAL_FILES).where(SPLAT_FILE_ID.eq(fileId)).execute()
```
sgrimm, line 491: What happens to these existing files in S3 and the files table?

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -516,6 +536,35 @@ class SplatService(
     }
   }
 
+  private fun splatAdditionalFileType(fileName: String): SplatAdditionalFileType? {
+    val baseName =
+        fileName
+            .substringAfterLast('/')
+            .substringAfterLast('\\')
```
sgrimm: Huh, does the mobile app use backslash-delimited paths on one of the platforms? I thought that was a Windows-ism. Harmless to include, of course.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -516,6 +536,35 @@ class SplatService(
...
+            .substringBeforeLast('.')
+            .lowercase()
+    return SplatAdditionalFileType.entries.firstOrNull { it.jsonValue == baseName }
```
sgrimm: Is this a sign we need to add a `forJsonValueOrNull` function to our generated enum code?

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
ice1080: Basically just there in case someone uploaded a file from windows that included a folder path. Mobile uses forward slash-delimited paths

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
ice1080: Might not hurt, though per the discussion in slack, I'll be moving away from these types anyway

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
ice1080: Basically these questions haven't been answered yet. 
- Should we allow additional files to be updated?
- Do we want to keep old files around if so? 
- Does keeping them around mean the DB or the file store?

I think I'll update this to disallow additional file updates for a splat to simplify. We can always revisit later if needed

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -482,9 +485,26 @@ class SplatService(
...
+          with(SPLAT_ADDITIONAL_FILES) {
+            // TODO delete files too
+            dslContext.deleteFrom(SPLAT_ADDITIONAL_FILES).where(SPLAT_FILE_ID.eq(fileId)).execute()
+
+            additionalFiles.forEach { (additionalFileId, type) ->
+              dslContext
+                  .insertInto(SPLAT_ADDITIONAL_FILES)
+                  .set(FILE_ID, additionalFileId)
+                  .set(SPLAT_FILE_ID, fileId)
+                  .set(TYPE, type)
```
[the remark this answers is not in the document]
ice1080: This is handled in a later PR in the stack

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -159,10 +160,11 @@ class SplatService(
       force: Boolean = false,
       params: SplatGenerationParams = SplatGenerationParams(),
       runBirdnet: Boolean = true,
+      additionalFiles: Map<FileId, String> = emptyMap(),
```
sgrimm: When I first looked at this, I was assuming this was a map of file IDs to S3 keys, but apparently it's a map of file IDs to file types. Might make sense to define a typealias just to make it clearer what the String represents.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -516,6 +536,34 @@ class SplatService(
     }
   }
 
+  private fun splatAdditionalFileType(fileContentType: String): String =
+      fileContentType
+          .substringBefore(';')
+          .trim()
+          .substringAfterLast('+', missingDelimiterValue = "")
+          .replace(Regex("([A-Z]+)([A-Z][a-z])"), "$1-$2")
+          .replace(Regex("([a-z0-9])([A-Z])"), "$1-$2")
```
sgrimm, line 553: I can follow what these regexes do mechanically, but until I sat and stared at them for a while, it wasn't obvious why we'd want these transformations. I _think_ this is transforming camel-case strings to hyphen-delimited. A comment to that effect would make it easier to read.

src/test/kotlin/com/terraformation/backend/splat/SplatServiceTest.kt
```diff
@@ -1934,27 +2025,315 @@ class SplatServiceTest : DatabaseTest(), RunsAsDatabaseUser {
...
+      )
+      assertEquals(
+          listOf(SplatterRequestFileLocation("bucket", "additional.json", expectedType)),
+          messageSlot.captured.additionalFiles,
+          "Request additional files",
+      )
+    }
+
+    @Test
+    fun `derives additional file types from content types rather than file names`() {
```
sgrimm: This feels like a "Claude verifying that the code no longer works the way it did in an earlier iteration of the PR" test to me. If that's really worth verifying, this case could be covered equally well by just tweaking one of the filenames in one of the other test cases.

sgrimm (approved):

PR #4898, opened 2026-09-02
Delete splat additional files from the file store when they are replaced

ice1080:
Publish FileReferenceDeletedEvent for each splat_additional_files row
that is removed, so the underlying file is deleted from the file store
and the files table once nothing else refers to it.

---

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
+++ b/src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
@@ -556,7 +556,8 @@ class SplatService(
 
   /**
    * Removes a splat's additional files from the database and, unless they're still needed, from the
-   * file store.
+   * file store. Additional files are uploaded as organization media, so their organization media
+   * rows are deleted too; otherwise they would keep the files alive in the file store.
    *
    * @param retainedFileIds Files that are about to be inserted as additional files of the same
    *   splat again. Their rows are still deleted, but the files themselves are left alone.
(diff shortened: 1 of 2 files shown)


ice1080:
I'm not super fond of the solution here to address codex's comment about the db transaction. I decided to err on the side of caution at the expense of cleanliness, but can flip the other way if preferred.

src/main/kotlin/com/terraformation/backend/splat/SplatService.kt
```diff
@@ -545,6 +558,41 @@ class SplatService(
...
+    val deletedFileIds =
+        with(SPLAT_ADDITIONAL_FILES) {
+          dslContext
+              .deleteFrom(SPLAT_ADDITIONAL_FILES)
+              .where(SPLAT_FILE_ID.eq(splatFileId))
+              .returning(FILE_ID)
+              .fetch(FILE_ID.asNonNullable())
+        }
+
+    val unreferencedFileIds = deletedFileIds.filterNot { it in retainedFileIds }
```
sgrimm: Could this instead be `.and(FILE_ID.notIn(retainedFileIds))` on the query?

ice1080: I had to refactor and rely on the assumption that types will never swap to the same as some other existing one, but that seems unlikely.

sgrimm (approved):

PR #4901, opened 2026-09-03
Use file extension as fallback in storage paths

ice1080:
When a storage path gets created, if the mime type isn't found in the default list, use the file's original extension (optionally) to determine the storage path's extension.
If this doesn't exist, continue to fall back to .bin.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

### DIFF
--- a/src/test/kotlin/com/terraformation/backend/file/PathGeneratorTest.kt
+++ b/src/test/kotlin/com/terraformation/backend/file/PathGeneratorTest.kt
@@ -1,7 +1,7 @@
 package com.terraformation.backend.file
 
 import java.time.Instant
-import kotlin.io.path.extension
+import kotlin.io.path.Path
 import kotlin.random.Random
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.params.ParameterizedTest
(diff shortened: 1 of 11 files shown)


tommylau523 (approved):

src/test/kotlin/com/terraformation/backend/file/PathGeneratorTest.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.random.Random
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.CsvSource
+
+class PathGeneratorTest {
+  private val timestamp = Instant.parse("2024-03-15T12:34:56Z")
+  private val pathGenerator = PathGenerator(Random(0))
+
+  @CsvSource(
```
tommylau523, line 17: I like this usage of parameterized tests!

src/test/kotlin/com/terraformation/backend/file/PathGeneratorTest.kt
```diff
@@ -0,0 +1,35 @@
...
+  )
+  @ParameterizedTest(name = "{0}")
+  fun `uses extension of content type, falling back to extension of filename`(
+      name: String,
+      contentType: String,
+      filename: String?,
+      expectedExtension: String,
+  ) {
+    val path = pathGenerator.generatePath(timestamp, "category", contentType, filename)
+
```
tommylau523, line 36: Can we maybe add an assertion of expected filename as well? I am thinking of cases where there maybe be multiple dots in the filename as well. For example foo.bar.json

ice1080: I can add a separate test for that if preferred, but it comes from `generateBaseName` and is based on the timestamp passed in plus the `random` object passed in to the class, so these path filenames are never related to the original filenames

tommylau523: I see. You can just ignore the filename bit then. The case on having multiple dots in the filename may still be worth testing on.

PR #4915, opened 2026-09-09
SW-8817 Allow overriding defaults in temporary plot selection

sgrimm:
In preparation for adding support for starting observations in strata
that don't have room for the configured number of plots, allow callers
to override the default temporary plot count and the default list of
permanent plots.

Nothing uses this new functionality yet, but it'll be used in subsequent
changes.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
@@ -69,8 +69,7 @@ data class StratumModel<
   }
 
   /**
-   * Chooses a set of plots to act as temporary monitoring plots. The number of plots is determined
-   * by [numTemporaryPlots].
+   * Chooses a set of plots to act as temporary monitoring plots.
    *
    * This follows some rules:
    * - Plots that are already selected as permanent plots aren't eligible.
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
```diff
@@ -179,6 +182,28 @@ data class StratumModel<
...
+   *   than or equal to [numPermanentPlots] as permanent. Ignored when [permanentPlotIds] is set.
+   */
+  private fun isPermanent(
+      plot: MonitoringPlotModel,
+      permanentPlotIds: Set<MonitoringPlotId>?,
+      includeAllIndexes: Boolean = false,
+  ): Boolean {
+    return when {
+      permanentPlotIds != null -> plot.id in permanentPlotIds
+      plot.permanentIndex != null -> includeAllIndexes || plot.permanentIndex <= numPermanentPlots
```
ice1080, line 202: Is it possible for some of these indices to be larger than numPermanentPlots due to a pending edit? E.g. [1, 2, 4] and `numPermanentPlots` is 3

sgrimm: Yes, that's totally possible, but the behavior here should be correct in that case: we wouldn't want to treat permanent index 4 as a permanent plot in a new observation.

ice1080: Gotcha. That doesn't make sense to me, but I'm not deep enough in the context to understand these details. So long as it's been thought through :+1:

sgrimm: You can think of the permanent index as the ordering of a list of candidate plots. The list can have more candidates than we actually end up choosing. We don't wipe the permanent indexes when `numPermanentPlots` is lowered, because we want to reuse those plots in the same order if the value is subsequently raised again.

ice1080: I have understood that in the past (though forgotten more recently), regarding not resetting permanent indexes. I think I was wondering about a map edit that causes permanent index 3 to no longer be in the substratum, and if `numPermanentPlots` should be compared against the size of the permanent plots list instead of the permanent indexes, but I may be conflating the previous PR stack which was about map edits with this PR stack which is about observations

ice1080 (approved):

PR #4916, opened 2026-09-09
SW-8817 Place temporary plots in substrata with space

sgrimm:
Previously, if a stratum had a substratum that was too small for whatever
number of temporary plots it should contain according to the "spread
temporary plots evenly across substrata" rule, we treated it as a
failure and refused to start the observation.

Change the allocation code so that it instead tries to place the excess
plots in whichever other requested substrata have room for them. If no
requested substrata have room, don't treat it as a failure, but allow
the observation to start with a reduced plot count.

### DIFF
--- a/src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
+++ b/src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
@@ -75,9 +75,13 @@ data class StratumModel<
    * - Plots that are already selected as permanent plots aren't eligible.
    * - Plots that span substratum boundaries aren't eligible.
    * - Plots in substrata that were not requested for the observation aren't eligible.
-   * - Plots must be spread across substrata as evenly as possible: the number of temporary plots
-   *   can't vary by more than 1 between substrata. This even spreading doesn't take eligibility
-   *   into account.
+   * - Plots must be spread across substrata as evenly as possible: the quota assigned to each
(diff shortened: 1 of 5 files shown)


sgrimm:
Converting this to draft while I try to figure out the geometry-related test failures. They don't reproduce in my dev environment and have so far resisted my attempts to use fixed-scale precision models.

src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
```diff
@@ -122,47 +125,80 @@ data class StratumModel<
...
+              count = numPlots,
+              exclusion = searchExclusion,
+              gridOrigin = gridOrigin,
+              searchBoundary = substratum.boundary,
+              permanentPlotIds = permanentPlotIds,
+          )
+
+      squares.forEach { square ->
+        val triangle = middleTriangle(square)
+        searchExclusion = (searchExclusion?.union(triangle) ?: triangle).toMultiPolygon()
```
ice1080, line 172: Might could be simplified using `fold` instead of `forEach`, but I leave that style preference up to you

ice1080 (approved):

src/main/kotlin/com/terraformation/backend/tracking/model/StratumModel.kt
sgrimm: I don't think it ended up being simpler, but it made the intent clearer, so it's a good change.

ice1080 (approved):

PR #4937, opened 2026-09-14
SW-8885 Normalize accession collector/site name whitespace

sgrimm:
The UI has a typeahead for accession collector and collection site names
that queries the list of existing values. Right now, in some
organizations, that list contains entries both with and without leading
and trailing whitespace, and sometimes with multiple consecutive spaces
between words. These look like duplicates in the UI, which is confusing
to users.

Update the code to remove excess whitespace in these fields, and update
all the existing values to match the new behavior.


sgrimm:
@codex review

sgrimm:
@codex review

src/main/kotlin/com/terraformation/backend/util/Extensions.kt
```diff
@@ -97,6 +97,24 @@ fun <T> Sequence<T>.onChunk(chunkSize: Int, func: (List<T>) -> Unit): Sequence<T
   return chunked(chunkSize).onEach { func(it) }.flatten()
 }
 
+private val multipleWhitespaceRegex = Regex("\\p{IsWhite_Space}+")
```
ice1080, line 100: Interesting. TIL about `\\p{IsWhite_Space}` as a better alternative to `\\s`

ice1080 (approved):

PR #4939, opened 2026-09-15
Use full clone of git repo on Buildkite

sgrimm:
Now that we're reusing the git checkout in all the pipeline steps,
there's much less benefit to doing a shallow clone, and using a
full clone means we no longer need to fetch additional revisions
on the fly for things like generating release notes.

This change requires removing the `git-shallow-clone` plugin invocation
from the pipeline's initial steps in the Buildkite configuration.


sgrimm:
The initial clone is slower than I expected; this isn't worth doing.

PR #4940, opened 2026-09-15
SW-8885 Normalize accession collector names

sgrimm:
To reduce confusing duplicate names in the accession collector
typeahead, we want to use consistent capitalization and accents for a
particular collector's name across their organization.

With this change, newly-created accessions will use the capitalization
and accents of the collector names on existing accessions if there are
any matching ones. When an accession is updated, any capitalization or
accent changes to its collector names are applied to that name
organization-wide.


src/main/resources/db/migration/2026/09/V20260914.1402__AccessionCollectorNames.sql
```diff
@@ -0,0 +1,55 @@
+CREATE COLLATION case_accent_insensitive (
+    provider = icu,
+    locale = 'und-u-ks-level1',
+    deterministic = false
+);
```
ice1080, line 5: I see we dropped the existing `accession_collectors_name_check` which disallows whitespace-only names. Does this new collation cover that or do we need a similar one added back in?

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -1111,19 +1137,40 @@ class AccessionStore(
...
+                        // row with different capitalization than the name we're searching for.
+                        DSL.select(NAME)
+                            .from(ACCESSION_COLLECTORS)
+                            .where(ORGANIZATION_ID.eq(organizationId))
+                            .and(NAME.eq(name))
+                            .limit(1)
+                    ),
+                    DSL.value(name),
+                ),
+            )
```
ice1080, line 1170: Should we add something to strip leading/trailing whitespaces?

sgrimm: https://github.com/terraware/terraware-server/pull/4937 should have taken care of that, unless I missed a code path in that PR.

ice1080: Ah I see, the name is already normalized by this point.

src/main/resources/db/migration/2026/09/V20260914.1402__AccessionCollectorNames.sql
ice1080: I'm guessing not needed since we're normalizing before insert/update now

sgrimm: The new constraints from https://github.com/terraware/terraware-server/pull/4937 should handle that.

ice1080 (approved):

PR #4941, opened 2026-09-15
SW-8885 Normalize accession collection site names

sgrimm:
To reduce confusing duplicate collection site names in the UI, we
want to use consistent capitalization and accents for a particular
site name across the organization.

With this change, newly-created accessions will use the capitalization
and accents of the collection site names on existing accessions if
there are any matching ones. When an accession is updated, any
capitalization or accent changes to its site name are applied to that
site name organization-wide.


src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
```diff
@@ -514,6 +536,22 @@ class AccessionStore(
       )
       insertStateHistory(existing, accession, facilityId, organizationId)
 
+      if (
+          existing.collectionSiteName != accession.collectionSiteName &&
+              accession.collectionSiteName != null
+      ) {
```
ice1080, line 542: Does this need a case-insensitive equality check?

src/main/resources/db/migration/2026/09/V20260914.1517__AccessionSiteName.sql
ice1080, line 1: Do we need the same collation on site name as we're doing on collector name?

ice1080: Nm, I see it now in the first PR

src/main/kotlin/com/terraformation/backend/seedbank/db/AccessionStore.kt
sgrimm: Not here, because this code path is how you correct the capitalization on an existing site name.

ice1080 (approved):

PR #4959, opened 2026-09-16
Stop querying unused data for ad-hoc observations

sgrimm:
When the caller asked for ad-hoc observation results, the observation results
query code was fetching a bunch of data about assigned observations that it
then threw away. The fetching of irrelevant data took nearly 75% of the
execution time of the query for some planting sites in production.

Split the ad-hoc observation results query into a separate function that only
fetches the ad-hoc-relevant data.


ice1080:
Tommy is making a bunch of changes around observation queries in the FE; is there any chance that this will negatively impact his work?

sgrimm:
> Tommy is making a bunch of changes around observation queries in the FE; is there any chance that this will negatively impact his work?

We were pulling the data from the database but not including it in the API responses, so nothing in the FE should be able to tell the difference other than the requests taking less time.

ice1080 (approved):

PR #4965, opened 2026-09-16
SW-8909 Remove plaintext-email-specific strings

sgrimm:
Remove strings that were only used in plaintext email notifications, and
rename their HTML variants to stop referencing "html" since that's not
useful information any more.


ice1080 (approved):

ice1080:
Do we need to update `NotificationServiceEmailTest` to remove the boolean flags in `assertBodyContains`?

sgrimm:
> Do we need to update NotificationServiceEmailTest to remove the boolean flags in assertBodyContains?

That should be covered by the previous PR in the stack, https://github.com/terraware/terraware-server/pull/4964

PR #4975, opened 2026-09-17
Add endpoint for latest observation stats of a planting site

tommylau523:
Adds `ObservationResultsStoreV2.fetchSiteObservationStats` and
`GET /api/v1/tracking/observations/results/stats?plantingSiteId={id}`, which return the statistics
from the most recent observation of a planting site and of each of its strata and substrata.

Each area resolves independently, so the observation ID and completion time can differ between
areas, and areas that haven't been observed yet are included without statistics. Each node carries
its observation ID, completion time, survival rate, and planting density.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tommylau523:
* **#4975** <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4975?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/terraware/terraware-server/4975?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -779,4 +817,314 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
...
+
+  /** Total live and dead plants of all species, or null if no species were recorded. */
+  private fun totalPlantsField(
+      speciesTotals: Table<*>,
+      totalLive: Field<Int?>,
+      totalDead: Field<Int?>,
+      condition: Condition,
+  ): Field<Int?> =
+      DSL.field(
+          DSL.select(DSL.sum(totalLive.plus(totalDead)).cast(SQLDataType.INTEGER))
```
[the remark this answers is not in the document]
tommylau523: existing plants are not counted.

src/main/kotlin/com/terraformation/backend/tracking/api/ObservationStatsPayloads.kt
ice1080: It seems like all of the payloads in here are essentially subsets of the payloads in `MonitoringResultsPayloads`. To me this feels like a smell that the new code in the store is doing nearly the same thing as what the results fetching already does

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreStatsTest.kt
ice1080, line 1: This feels like we might not be testing all of the complex observation and site history cases that we cover in the other results fetching. I think we may need more tests to ensure the stats endpoints get the correct data in the various cases

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -104,6 +114,34 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
...
+  /**
+   * Statistics for a planting site and each of its strata and substrata. Each area resolves to its
+   * own observation: the most recent one that completed at least one monitoring plot in that area.
+   * Areas that have never been observed are included with no statistics.
+   */
+  fun fetchSiteObservationStats(plantingSiteId: PlantingSiteId): ObservationSiteStatsModel {
+    requirePermissions { readPlantingSite(plantingSiteId) }
+
+    return dslContext
+        .select(PLANTING_SITES.ID, latestSiteStatsMultiset, strataStatsMultiset)
```
ice1080: Fetching the site stats and the strata stats in the same query as separate multisets feels off to me. 
Looking at `latestSiteStatsMultiset`, it seems like it might could be extracted to a direct query that also gets planting_site_id instead of doing this external query and adding both multisets.
Same comment within the `strataStatsMultiset` and `substrataStatsMultiset`

ice1080 (commented):
Other discussion in slack

src/test/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreStatsTest.kt
ice1080: Actually an alternative could be to use all of the existing tests, and just have them also test the stats functions when they are testing the results functions.

tommylau523: For unit testing, this fetch function doesn't really need to know how the results table is written. I think it serves the purpose well enough. 

I do agree that it is better to add verification for stats in those test cases instead!

src/main/kotlin/com/terraformation/backend/tracking/db/ObservationResultsStoreV2.kt
```diff
@@ -779,4 +831,292 @@ class ObservationResultsStoreV2(private val dslContext: DSLContext) {
...
+              .and(OBSERVATION_PLOTS.COMPLETED_TIME.isNotNull)
+              .and(plotsInAreaCondition)
+      )
+
+  // Each level joins its own alias of OBSERVATIONS so the nested queries don't shadow each other.
+  private val substratumObservations = OBSERVATIONS.`as`("substratum_observations")
+  private val stratumObservations = OBSERVATIONS.`as`("stratum_observations")
+  private val siteObservations = OBSERVATIONS.`as`("site_observations")
+
+  private val latestSubstratumObservationId: Field<ObservationId?> = run {
```
ice1080: I think the `run` should be changed to a function instead of a `val` to be consistent with our stores in the codebase. Looks like the only other usage of `run` in a store is when it's adding a log statement before returning something

ice1080 (approved):

PR #4980, opened 2026-09-18
Add survival rates diagnostic page to admin

ice1080:
(260 lines are html)

Oftentimes when a survival rate is not calculated or when it is not what we think it should be, we look through various observation pages and t0 settings to determine what is going on. This can be time consuming and is often forgotten in between issues what exactly needs to be in place. 
Add a survival rates diagnostic page to admin, accessible via a link from the planting sites page, to help track down what is going on with the survival rates of a planting site.

Example:
image

Co-authored-by: Codex (GPT-6) <codex@openai.com>


ice1080:
I kept the unit test that codex wrote for the large model for this page, but am not tied to it, especially since we don't have any other admin tests.

sgrimm:
Let's ditch the model test. It isn't really adding much value considering this page is pretty much just for developer use.

src/main/kotlin/com/terraformation/backend/admin/SurvivalRatesPageModel.kt
```diff
@@ -0,0 +1,207 @@
...
+              },
+          survivalRateIncludesTempPlots = site.survivalRateIncludesTempPlots,
+      )
+    }
+  }
+}
+
+private val displayTimeFormatter = DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
+
+private fun Instant.toDisplayString(): String = displayTimeFormatter.format(this)
```
sgrimm: This seems unnecessary. We display timestamps using the default ISO-8601 formatting elsewhere in the admin UI.

src/main/resources/templates/admin/plantingSite.html
```diff
@@ -307,6 +307,14 @@ <h3>Upload New Shapefiles</h3>
...
 </th:block>
 
+<h3>Survival Rates</h3>
+
+<p>
+    <a th:href="|/admin/plantingSite/${site.id}/survivalRates|">
+        View survival rate diagnostics
+    </a>
+</p>
+
```
sgrimm: Wrap this in a conditional block so it doesn't show up for users who don't have permission to view the new page.

sgrimm (approved):

src/main/kotlin/com/terraformation/backend/admin/SurvivalRatesPageModel.kt
ice1080: I actually pulled this from the Test Clock page in admin, but I must have picked the wrong one since I see that in these two other areas we use the default formatting:
- Simplify Planting Site Histories
- Device Management - Advanced Settings

src/main/resources/templates/admin/plantingSite.html
ice1080: Ah wait, looks like we don't have conditionals right now on planting sites, and it only requires `readPlantingSite` to view this page. I added that conditional, but seems pointless if they're already reading the planting site page that this link exists on

sgrimm: The new controller has

```
@RequireGlobalRole([GlobalRole.SuperAdmin, GlobalRole.AcceleratorAdmin])
```

so if you're only a TF Expert and you click the link, you'll get an error, unless I'm missing something.

ice1080: But the planting site controller and the org controller both have that already. 
Actually I just tested it and TF Experts don't seem to have access to admin at all

sgrimm: Ah, yeah, that's a fair point!

PR #4988, opened 2026-09-18
Move test clock to DatabaseBackedTest

sgrimm:
To make it easier to write tests that write database records simulating
the passage of time, add an instance of `TestClock` to
`DatabaseBackedTest` and use it instead of hardwired `Instant.EPOCH`
values for the default timestamp value in insert functions.

Adding this clock means a lot of tests no longer have to create their
own clocks.


tommylau523 (approved):
I was able to grep changes that are not Instant.EPOCH -> clock.instant, and removal of import TestClock and initialization of clock. This makes the review pretty straightforward!

